PR #4, opened 2016-05-10
Efficiency fix

yebai:
This PR significantly improved efficiency of the SMC sampler - SMC now scales linearly with number of particles used.


adscib:
Can you add some comments in the definition of `ParticleContainer` to explain what `conditional` and `n_consume` are for?

yebai:
These fields are reserved for implementing rejuvenation steps (e.g. SMC2) - I've added some comments. These code will become more readable when the rejuvenation step is implemented.

PR #15, opened 2016-05-30
Enabled win64 support for appveyor.

yebai:


yebai:
@evsmithx I've added the missing win64 support for appveyor. Could you check whether the code is sensible to you and if so, merge this PR?

PR #16, opened 2016-06-03
Summarise `Chain` in `io.jl`

### LINKED ISSUE #13: Summarise `Chain` in `io.jl`.
The current `Chain` type dump all fields to the console by default. A potential better way is to Summarise the results, e.g. print the `mean` and `variance`. This can be done by using the [weighted](http://statsbasejl.readthedocs.io/en/latest/means.html#mean) `mean` function etc in the [StatsBase](http://statsbasejl.readthedocs.io/en/latest/weightvec.html) package.

evsmithx:
Added Base.show(io::IO, ch1::Chain) to resolve #13. The prints the weight of the chain and the weighted means of its symbols.


evsmithx:
I've messed up the commits here! I've put a clean pull request in #17

PR #45, opened 2016-11-14
Make compiler support Distributions wihtout parameters

xukai92:


xukai92:
For `has_ops`, it just check if a `Distribution()` has optional arguments. It gives `true` if we have something like `Normal(0, 1; :static=true)` and gives `false` if there is no optional arguments. It's my bad not giving comments. I'll add comment to it now.

xukai92:
Previously the compiler gave error to `@assume x ~ Binomial()` because `Binomial()` doesn't have any arguments. This PR is to fix this issues.

yebai:
That makes sense. Thanks!

PR #57, opened 2016-11-16
Move most notebooks to doc folder.

yebai:
I propose we keep a simple notebooks folder in the ROOT dir, and move examples to doc folder.   

The basic idea is that ROOT/notebooks only keep an introduction example to Turing, and the doc folder will provide many other examples (we can keep adding more).

I believe there is a way to produce readthedoc files from Jupyter notebook files. We probably should check it out.


yebai:
@evsmithx This is a mostly a restructuring PR. Can you have a look and accept it?

PR #59, opened 2016-11-16
make Makefile compatible with 0.4.7 (for JuliaBox)

xukai92:


xukai92:
@yebai But users could also meet the same problem: those who have both 0.4 and 0.5 installed, and uses Jupyter without launching Julia.

xukai92:
I guess this is enough: https://github.com/xukai92/Turing.jl/commit/d922d260c7f07ea0c859c74ed7842d1a5b14bfdb

Issue #55, opened 2016-11-16
New Interface Design 


yebai:
We have some undecided interface issues since the introduction of the Hamiltonian Monte Carlo (HMC) sampler. More specifically, we need an interface to tell the HMC sampler which variable it should handle. In addition, we also need a way to compose samplers (e.g. Gibbs sampling subsets of variables with different samplers). To solve these issues altogether, I propose the following new interface. 

```julia
##  Model definition
gaussdemo() =  begin  
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  for i=1:length(x)
    x[i] ~ Normal(m, sqrt(s))
  end
  return(s, m, x)
end

# Standard sampler interface
res = sample(gaussdemo, data = Dict(:x=>[1 2]), HMC(..., :s, :m))

# Compositional inference interface
res = sample(gaussdemo, data = Dict(:x=>[1 2]), HMC(..., :m),  PG(..., :s))

# Sampler's default behavior is to sample all random variables
res = sample(gaussdemo, data = Dict(:x=>[1 2]), HMC(...))

# Tagging parameters for samplers like PMC, SMC2
res = sample(gaussdemo, data = Dict(:x=>[1 2]), PMC(..., params=[:m]))

# Running program without a sampler produces a draw from the prior
s, m = sample(gaussdemo)
s, m = gaussdemo() 
```

In summary, first, this design removed macros `@assume`, `@observe` and `@predict`. Instead, it introduces additional parameters to each sampler to distinguish different types of variables. Second, this design introduces an interface to compose samplers (e.g. HMC and PG).

yebai:
Here is a summary of the naming scheme we've discussed. One simple solution is to use the following scheme:

```
varname_idx1_idx2
```

Here `varname` is the name of a variable, e.g.`x`, `m`, or `s`; moreover `idx1`, `idx2` etc are the indices for multiple dimensional variables. For scalar type variables, the name would simplify to `varname`.

There is one potential drawback with this naming scheme. To illustrate this, let's first start with an (arbitrary) example

```
gaussdemo() =  begin  
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  for i=1:length(x)
    x[i] ~ Normal(m, sqrt(s))
    w    ~ Normal(m, sqrt(s))  # definition 1 of var w
  end
    w  ~   InverseGamma(2, 3) # definition 2 of var w
  return(s, m, x)
end
```

[...]

PR #61, opened 2016-11-18
Modularize AD

xukai92:
This PR 
1. splits AD from HMC and makes it an individual function
2. add an new abstract for gradient-based sampler `GradientSampler`.


xukai92:
Following our discussion just now, there is no problem in the 2nd and 3rd bullet point. However, in AD, after consuming the model, we need to use the log-joint, which is bound with the sampler as well. 

I can come up with a (temporary) solution without changing a lot of things: we create a wrapper for logjoint and pass logjoint by reference to `get_gradient_dict`. Then everything inside it keeps the same.

However, this let me feel unhappy because we actually pass three attributes of `spl` to `get_gradient_dict` by reference. I kind of hope our model will return the logjoint by default. Maybe we also need to consider this in #55 .

xukai92:
@yebai I've done the requested changes.

yebai:
> However, this let me feel unhappy because we actually pass three attributes of spl to get_gradient_dict by reference. I kind of hope our model will return the logjoint by default. Maybe we also need to consider this in #55 .

Can we make `logjoint` a field of `GradientInfo`? 

Also, maybe merge `replay.jl` into `ad.jl`?

yebai:
PS. I just looked at `replay.jl`, it seems `replay.jl` too large to be merged into `ad.jl`. Maybe we can keep it separate for now.

xukai92:
> Can we make `logjoint` a field of `GradientInfo`?

No problem. I think it's just a decision problem. We can simply do this by amending each get/set of `spl.logjoint` to `priors.logjoint`. But I kind of fell `logjoint` itself, as an attribute, does not belong to `priors`. But I feel its better to change it as the way you proposed now & will do.

> Also, maybe merge `replay.jl` into `ad.jl`?
> PS. I just looked at `replay.jl`, it seems `replay.jl` too large to be merged into `ad.jl`. Maybe we can keep it separate for now.

I guess we can first rename it to `gradientinfo.jl` and put it to core, because I fell it should exist in the same place as `container.jl`.

yebai:
Cool - can you push the proposed changes before I merge this PR

xukai92:
Done!

PR #63, opened 2016-11-22
Matrix support

xukai92:
All the variables are now processed as follow

- In compiler, the type of variable is recorded and indicated by an integer: 1 for univariate 2 for multi and 3 for matrix, which is stored as an attribute of `Prior`.
- In HMC 
  - when generating the variable, all variables are converted to vectors
  - before calling the `logpdf()`, vectors are converted back according to its variable type


yebai:
It seems that the sole purpose of the `Prior` type is to identify random variables. If this was true, I think something like `VarName` or `VarID` is more meaningful.

xukai92:
Yes that's true. I agree with you and will change it.

src/samplers/hmc.jl
```diff
@@ -169,33 +172,43 @@ end
...
+  if prior.typ == 1
+    # Turn Array{Any} to Any if necessary (this is due to randn())
+    val = val[1]
+  elseif prior.typ == 2
+    # Turn Vector{Any} to Vector{T} if necessary (this is due to an update in Distributions.jl)
+    T = typeof(val[1])
+    val = Vector{T}(val)
+  elseif prior.typ == 3
+    T = typeof(val[1])
+    dim = Int(sqrt(length(val)))
```
yebai: Not all matrix-variates are square matrices, we might want to store the `dim` information in `Prior` as well.

xukai92: You're right. Let me make the change.

xukai92:
It's a bit tricky to pass `dim` information to `VarInfo` directly as we didn't know it until we run the real `assume` function. Alternatively, I add another container in `HMC` sampler to store `dim` information. Do you think it's fine?

And I guess I need to rename everything use the `prior` convention to something else.

PR #66, opened 2016-12-07
Constrained variable support

xukai92:
The following are now supported

1. Bounded uni
2. Simplex multi
3. Positive definite matrix

~Note:~

~1. I didn't do density transform (as in Stan's doc) but only do:~
    ~i. X -> R~
   ~ii. Move~
    ~iii. R -> X~
~2. Due to the fact that `Dual(f)` where `f` is a `Float64` will lose precision sometimes. When vectorise the variables, I force the last entry of the vector to be `1-sum[1:end-1]`.~

---

**Update**

Note 1 is incorrect. This PR finally do as follow:

1. For each prior do X->R
2. Do HMC in R with transformed density
3. Do R->X when output the variable

Note 2 doesn't exist in this way.

This PR also contains some refactors on HMC in the last few commits.


yebai:
The Mamba package actually has a implementation for transforming positive definite matrices [here](https://github.com/brian-j-smith/Mamba.jl/blob/master/src/distributions/pdmatdistribution.jl), might be worth to have a look!

yebai:
Also, see this [file](https://github.com/brian-j-smith/Mamba.jl/blob/master/src/distributions/distributionstruct.jl) for some useful 

- utility functions for flattening/vectorising multiple dimensional variables: `unlist`, `relistlength`
- fallback functions for computing log densities 

Sorry I didn't notice them earlier!

xukai92:
The test for AD is here in this commit https://github.com/yebai/Turing.jl/pull/66/commits/7d0af5f834b6911dc19e6792ea7085f29de5c432

xukai92:
The bug about volume preservation is fixed.

xukai92:
I made a mistake when refactoring my code, which is now fixed.

xukai92:
I found that the error is due to my wrongly placed code you pointed for fixing the precision problem before. And actually the precision problem doesn't appear when we do all HMC movement in R space. I believe it is previously due to the frequent transforms between X and R. That codes are removed and the result is correct.

xukai92:
@yebai I'm also doing some refactor for HMC, making the code easier to read and modularized. Do you want me to just push them to this PR or later creating a new one?

xukai92:
I guess I've finished the refactors I want to do so far :)

yebai:
Running the following code can cause the parameters go to `NaN` of `Inf`. Could this be a bug in the transform?

```julia
@model gaussdemo begin
         @assume s ~ InverseGamma(2,3);println("s=$s");
         @assume m ~ Normal(0,sqrt(s))
         @observe 1.5 ~ Normal(m, sqrt(s))
         @observe 2.0 ~ Normal(m, sqrt(s))
         @predict s m
end
chain = sample(gaussdemo, HMC(5000, 2, 5))
```

xukai92:
I think it is caused by the nature that we are moving model parameters in the log space and the step size 2 is too large.

According to this piece of codes:

```julia
# Half momentum step
function half_momentum_step(p, ϵ, val∇E)
  dprintln(3, "half_momentum_step...")
  for k in keys(p)
    p[k] -= ϵ * val∇E[k] / 2
  end
  p
end

# Leapfrog step
function leapfrog(values, val∇E, p, ϵ, model)
  dprintln(3, "leapfrog...")

  p = half_momentum_step(p, ϵ, val∇E) # half step for momentum
  for k in keys(values)               # full step for state
    values[k] = Array{Any}(values[k] + ϵ * p[k])
  end
  val∇E = get_gradient_dict(values, model)
  p = half_momentum_step(p, ϵ, val∇E) # half step for momentum

  # Return updated θ and momentum
  values, val∇E, p
end

# Transformas
link(d::PositiveDistribution, x::Real) = log(x)

invlink(d::PositiveDistribution, x::Real) = exp(x)
```

If we have a large stepsize and step number, we can easily move a long distance in the log space and when it is converted back to linear space it will become very large. 

I've created an issue here for this problem here: https://github.com/yebai/Turing.jl/issues/67

src/samplers/support/transform.jl
```diff
@@ -0,0 +1,190 @@
...
+  lp
+end
+
+################## Callback functions ##################
+
+link(d::Distribution, x) = x
+
+invlink(d::Distribution, x) = x
+
+function logpdf(d::Distribution, x, transform::Bool)
```
tkelman: similarly here, extending another package's method on another package's (or Base) types has global side effects and is not recommended

xukai92: Rename `logpdf` to another name in https://github.com/yebai/Turing.jl/pull/316

PR #70, opened 2016-12-22
New naming schema

xukai92:
## New naming schema is implemented as below

```julia
      x ~ D # => :(x)
   x[i] ~ D # => :(x[1])
 x[i,j] ~ D # => :(x[1,2])
x[i][j] ~ D # => :(x[1][2])
```

## Other update

1. Support users' setting of chunk size via: `sample(model, HMC(...); chunk_size=...)`
2. More tests (for both old and new codes), making coverage back to 90%

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -37,6 +37,7 @@ export dprintln
 const TURING = Dict{Symbol, Any}()
 global sampler = nothing
 global debug_level = 0
+global CHUNKSIZE
 
 ##########
 # Helper #
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -1,5 +1,3 @@
-const CHUNKSIZE = 1000
-
 doc"""
     get_gradient_dict(spl :: GradientSampler)
 
--- a/src/core/compiler.jl
+++ b/src/core/compiler.jl
@@ -83,7 +83,7 @@ macro assume(ex)
         )
       end
     )
-  elseif length(varExpr.args) == 2
+  elseif length(varExpr.args) == 2 && isa(varExpr.args[1], Symbol)
     esc(
       quote
         $(varExpr) = Turing.assume(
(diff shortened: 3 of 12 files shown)


xukai92:
I changed the naming a bit:

```julia
      x ~ D # => :(Tx)
   x[i] ~ D # => :(Tx[1])
 x[i,j] ~ D # => :(Tx[1,2])
x[i][j] ~ D # => :(T(x[1])[2])
```

This is due to the a problem occurring when:

1. We do inference on model with `:x` used
2. After that we assign a value to `x` globally
3. Try to do inference again

e.g.

```julia
@model test begin
  x ~ Normal(0, 1)
end
sample(test, HMC(10, 0.1, 5))
x = 1
sample(test, HMC(10, 0.1, 5))  # => gives error
```

I thinks its better to distinguish our internal symbol by starting with "T", and this problem is also solved by doing so.

yebai:
This looks very good. However the interface for `VarInfo` constructor seems complicated, can we simply it? For example, image we are given `x[i][j] ~ dist`, then we can try to 

- extract the left side altogether as one variable `varName` (ie. varName = :(x[i][j]) 
- then pass it to `VarInfo(varName, dist)`

Note here I also added `dist` to `VarInfo`. The purpose is to store all relevant variable information in one place. If this was implemented, we can remove variable `dists` in the HMC sampler (here)[https://github.com/yebai/Turing.jl/blob/master/src/samplers/hmc.jl#L34]

xukai92:
Hi Hong, I'm working on the new-interface (i.e. remove `@assume`, `@observe` and `@predict`), in which I've already done the two points you proposed. I also made some change to the `dists`

Can you accept this PR first and then I will create a new PR for the interface part (soon)?

yebai:
I see - will do now.

yebai:
Note there are also some limits in this PR, e.g. `x[i][j][k]`, or `x[i,j][k]` or `x[i,j,k]` can't be handed correctly.

xukai92:
I see - it's on my to-do list now.

PR #73, opened 2016-12-31
New interface and Gibbs sampling

xukai92:
I've made some progress for the new interface and Gibbs sampler. But it's current buggy and I guess it's better for us to look into the design and code together.

Here is a summary of what I've done:

- pass sampler, varInfo and data locally in stead of using global variables
  - global variables are still supported now because I haven't completed this change to PG
- step-ize HMC and PG
  - so that in the Gibbs sampler I can call them step by step
- redesign `VarInfo`
  - rename: `GradientInfo` -> `VarInfo` and `VarInfo` -> `Var`
  - rotated array is removed because with the new naming schema it's unnecessary
- replace `@assume` and `@observe` with `~`
- update `@model` to support new synatx
  - deal with default sampler, varInfo and data
  - locally assignment of data
  - `@predict` is removed by automatically convert the return value of model (e.g. `return(m,s)`) to `@predict` statement by `@model`.
- add a Gibbs sampler which can be used to compose samplers
  - I finally decided to implement a separate Gibbs sampler first as it's clear for both users and developers
  - PG and HMC now have a new filed named `space` to indicates which variables they need to deal with
  - two examples for Gibbs are in [test/gibbs.jl](https://github.com/yebai/Turing.jl/blob/new-interface/test/gibbs.jl) and [test/gibbs2.jl](https://github.com/yebai/Turing.jl/blob/new-interface/test/gibbs2.jl)

Here are points need to be discussed:

- we need a consistent output of model (i.e. return) and storage of predicts
- we need to discuss the consistent way of using `task.storage`

Here are knowing issues I haven't solved :

- the naming issue of `x[i,j,k]` and `x[i,j][k]`
  - this is easy to solve recursively if all the variables are global but need some thoughts in compiler level
- different steps for each sampler in Gibbs sampling
- how to deal with rejected HMC step
- allow user to step how many steps each sampler in Gibbs sampling is run
- the correctness of Gibbs sampler seems doubtable - I only did two tests, of which the mixture of gaussian one doesn't seems to work well (another one is Gaussian with conjugate prior which works fine)
- the current implementation has a bad performance (speed)


xukai92:
BTW the test for 32-bit windows machine is always broken.

xukai92:
### Bugs

- [ ] the correctness of Gibbs sampler seems doubtable
- [ ] correct way to deal with rejected HMC step

### Discussion

- [ ] we need a consistent output of model (i.e. return) and storage of predicts
- [ ] we need to discuss the consistent way of using task.storage

### Enhancement

- [ ] the naming issue of x[i,j,k] and x[i,j][k]
- [ ] allow user to step how many steps each sampler in Gibbs sampling is run
- [ ] the current implementation has a bad performance (speed)

src/samplers/sampler.jl
```diff
@@ -54,25 +54,31 @@ end
...
+  else  # if it isn't in space
+    if haskey(varInfo.values, var)
+      val = varInfo[var]
+      dist = varInfo.dists[var]
+      val = reconstruct(dist, val)
+      r = invlink(dist, val)
+      produce(logpdf(dist, r, true))
+    else
+      r = rand(current_trace(), dist)  # gen random
+      produce(logpdf(dist, r))
```
yebai: Here we are proposing `r` from its prior, so the importance weight is exactly `1`.
So I think this line should be `produce(log(1.0))`.

xukai92: Done it!

xukai92:
I found and fix the bug which caused the correctness error before. The bug is due to the fact that the way HMC and PG store result is different (HMC directly stores results in `VarInfo` but PG produces the predicts directly). Previously I didn't deal with it in the right manner but it seems to be OK now (e.g. for `test/gibbs.jl` and `test/beta_binomial.jl`).

yebai:
> I found and fix the bug which caused the correctness error before. The bug is due to the fact that the way HMC and PG store result is different (HMC directly stores results in VarInfo but PG produces the predicts directly). Previously I didn't deal with it in the right manner but it seems to be OK now (e.g. for test/gibbs.jl and test/beta_binomial.jl).

Nice work : )

PR #79, opened 2017-02-12
`Observe` statement now only use variable in the `data` dictionary.

yebai:


xukai92:
It seems that this PR causes some bugs which make the master branch fails the test.

yebai:
I've restored the master brach. Let's ignore this PR and stick to the macro design shown [here](https://github.com/yebai/Turing.jl/issues/55#issuecomment-274498257) and below.

```julia
chn   = @sample(gdemo(x), sampler)
```

Issue #82, opened 2017-02-26
Updated model interface: add parameters support.


yebai:
The current interface pass in variables through a dictionary. This design requires us implicitly copy all items in the data dictionary to model's scope, which is not the most consistent design with Julia's function interface. A possible better design would be:

```julia
@model gdemo(x) = begin   
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  for i=1:length(x)
    x[i] ~ Normal(m, sqrt(s)) 
  end
  return(s, m, x)
end

x = [2.0, 3.0]
sampler = Gibbs(HMC(..., :m), PG(..., :s))
chn     = @sample(gdemo(x), sampler)
```

The most important changes are:

1. Allow declaring parameters in model macro.
2. Support `=` in model declaration, e.g. `gdemo(x) = statement_block`. We might want to depreciate the syntax without `=` in the future (i.e. `gdemo(x) statement_block`) to stay in line with Julia's assignment form of [function](http://docs.julialang.org/en/release-0.4/manual/functions/) definitions.
3. The twiddle macro will determine the type of left-hand-side (LHS) variable through:
           -  data: LHS variable can be found in model's parameter list.
           -  parameter: otherwise.

EDIT (28 FEB 2017): removed obsolete comment.

xukai92:
@yebai What do you mean by "Note: we need to fetch x from gdemo(; data = Data(...))."?

yebai:
that comment is obsolete, sorry I forgot to delete it when copying the code.

xukai92:
No worries. 

There is a previous design decision which may affect the implementation of this update. Can you have a look at Gitter for discussion?

yebai:
Closed by PR #85

PR #85, opened 2017-03-02
Updated model interface

xukai92:
This PR implements the new interface described in #82.

Summary of PR:

1. Allow declaring parameters in model macro;
2. Support = in model declaration;
    - The old model declaration syntax is still support
3. Use `@sample` macro to run inference engine.
    - This macro construct the `data` dictionary in the previous design and run the `sample()` function.
    - Calling `@sample` is the same as constructing the `data` dictionary and call `sample(model, data=..., alg)`.


yebai:
@xukai92 It seems some temporary stan files (see tmp/) are also included in the recent commit. Could you exclude them?

xukai92:
@yebai Done.

src/samplers/support/hmc_helper.jl
```diff
@@ -25,7 +25,12 @@ end
 end
 
 Base.convert{N,T<:Real}(::Type{T}, d::Dual{N,T})  = d.value
-Base.convert(::Type{Float64}, d::Dual{0,Int}) = round(Int, d.value)
+Base.convert{N}(::Type{Int}, d::Dual{N,Float64}) = round(Int, d.value)
```
tkelman: Do you need to be extending the conversion methods of a different package's type here? That can change the behavior of unrelated code

xukai92: This is removed now.

PR #90, opened 2017-03-07
Mamba Chain

yebai:
This PR adds (initial) support for converting `Turing's` Chain type to [`Mamba's`](https://github.com/brian-j-smith/Mamba.jl/blob/master/src/output/chains.jl). This allows us to re-use Mamba's MCMC diagnostic and plotting functions, which is the also the default option for [`Stan.jl`](https://github.com/goedman/Stan.jl). 

-  Default chain structure is not changed.
- To convert to Mamba chain, run: `sim1 = MambaChains(turingChain); describe(sim1)`
- New dependent package: Mamba.


yebai:
@xukai92  Could you review this PR?

Issue #96, opened 2017-03-08
Refactoring code: Unify VarInfo, Trace, TaskStorage, Chain.


yebai:
At the moment, we have various data structures that stores intermediate (MCMC) sampling states. It probably makes sense to unify these data structures. 

(I don't have a complete idea how to do this yet, but let's start thinking about this issue and post ideas here.)

yebai:
Here is a (preliminary) plan that we have discussed:


- Update VarInfo:  Remove Var, instead store MCMC state in a flatten data structure (similar to Mamba’s Chain type)
- Update internal variable naming scheme: 
    - x[1,2,3] ==> x[i]   where i is linear index corresponding to [1,2,3] (e.g. sub2ind(dims, i, j, k...) -> index)
    - x[1][2][..]  ==> x[i][j][..]  (the nesting could continue N times)
    - Forbid variables with same name, with the exception 
        - loops, e.g. `for i=1:N,   x ~ Normal(0,1);  end`
        - recursions, e.g.  `f(x) = if x > 1, f(x-1); else x ~ Normal(0,1); end`
        - Note the following case is forbidden:   `f() = begin x ~ Normal(0,1); y ~ Normal(0,1) end`
- Remove predict, `~` will store MCMC state by default.
- Gradually merge Trace into VarInfo (maybe keep Trace as a field of VarInfo initially)

xukai92:
Note: We tend to store variables in their original space.

xukai92:
Note: I will next resolve the 2nd and 3rd points together by the following steps:
1. Change code to store untransformed variables in `VarInfo`
2. Remove `@predict` (with `@~` store variables into `VarInfo`)
3. Remove the use of `Sample` (as it's only copy things from `@predict`)

xukai92:
It seems that it's impractical to remove the use of `Task.storage` because IS-based algorithms will need variable information tied to task level.

xukai92:
@yebai Do you think it's better to move `predict()` into `@model` level instead of calling it in each `@~`?

yebai:
Ref #142;

xukai92:
I created an individual issue (https://github.com/yebai/Turing.jl/issues/146) for the naming stuff. Do you think we can close this one now?

yebai:
Fixed by #114, #108, #103.

Issue #77, opened 2017-01-24
Further work for the Gibbs PR


xukai92:
### Discussion

- ~we need a consistent output of model (i.e. return) and storage of predicts~
- ~we need to discuss the consistent way of using task.storage~

### Enhancement

- [x] the naming issue of x[i,j,k] and x[i,j][k]
- [x] allow user to step how many steps each sampler in Gibbs sampling is run
- ~the current implementation has a bad performance (speed)~
- ~unify the way of storing and producing variables~

yebai:
It seems the HMC sampler constructor is broken. Running the following code would throw an error

```julia
HMC(1000, 0.2, 5)
```

xukai92:
Which version (commit) do you meet this error?

xukai92:
Do you think it's better to keep both `VarInfo` and `Sample` and convert between them when necessary or use one data structure only?

yebai:
> Which version (commit) do you meet this error?

It is working now with the most recent `new-interface`. I might be using some intermediate commit.

> Do you think it's better to keep both VarInfo and Sample and convert between them when necessary or use one data structure only?

It makes sense to unify these data structures. However, let's focus on bug fixing and make it a TODO item for the next release.

PR #108, opened 2017-03-17
Remove var and support more general indexing

xukai92:
This PR removes the `Var` type (#96) and resolves the naming issue of `x[i,j,k]` and `x[i,j][k]` (#77).

Regarding the second improvement, it should allow the use any kind of nested indexing. Something like below is also supported.

```julia
using Distributions, Turing
@model test() = begin
    a = Array{Array{Float64, 2}}(2)
    for i=1:2 a[i] = Array{Float64, 2}(2, 2) end  # Init
    for i=1:2, j=1:2, k=2:3
        a[i][j, k-1] ~ Normal(0, 1)
    end
    return(a)
end
chain = @sample(test(), HMC(100, 0.2, 3));
```


yebai:
Looks very good. I suggest that we rename `parse_indexing` to `VarName` and move its definition to `varinfo.jl`.

src/Turing.jl
```diff
@@ -16,7 +16,7 @@ using Turing.Traces
 #################
 
 # Turing essentials - modelling macros and inference algorithms
-export @model, @sample, @predict, @~, @isdefined, InferenceAlgorithm, HMC, IS, SMC, PG, Gibbs, sample, Chain, Sample, Sampler, ImportanceSampler, HMCSampler, VarInfo, @predictall
+export @model, @sample, @predict, parse_indexing, @~, @isdefined, InferenceAlgorithm, HMC, IS, SMC, PG, Gibbs, sample, Chain, Sample, Sampler, ImportanceSampler, HMCSampler, VarInfo, @predictall
```
yebai: remove obsolete export: @isdefined

src/core/compiler.jl
```diff
@@ -2,67 +2,70 @@
...
+    else
+      # Evaluting the concrete value of the indexing variable
+      push!(uid_ex.args, quote push!(uid_list, string($evaling)) end)
+    end
+  end
+  push!(uid_ex.args, quote uid = reduce(*, uid_list) end)
+  uid_ex
+end
+
 function gen_assume_ex(left, right)
```
yebai: Maybe directly place this code into macro `~`, since it is much simpler now?

xukai92: Yes I agree to put them directly into `~`.

src/core/compiler.jl
```diff
@@ -72,6 +75,8 @@ end
 
 macro ~(left, right)
```
yebai: Can we store all variables in `~` and completely remove `predict` in the future?

xukai92:
Can we merge this PR first?

PR #109, opened 2017-03-20
Performance issues about `ad.jl`

### LINKED ISSUE #106: Performance issues about `ad.jl`
The following two lines in `ad.jl` could potentially cause substantial memory allocation and subsequently deteriorate performance of the HMC sampler

 - Large matrix allocation: https://github.com/yebai/Turing.jl/blob/master/src/core/ad.jl#L50
 - (potentially) unnecessary use of Dual type: https://github.com/yebai/Turing.jl/blob/master/src/core/ad.jl#L83

For the first case, we can use a pre-allocated vector to avoid repetitive memory allocation. In the second case, it seems using dual types for variables that don't require gradients is wasteful.

xukai92:
Fix #106


yebai:
Looks cool to me!

Issue #110, opened 2017-03-21
Generalise TArray


yebai:
The self-copying mechanism in `TArray` could be generalised, e.g. create a macro `@T`, that can automatically lift existing data structures to Turing-safe ones. 

On a side note, `TArray`'s copying detection mechanism can be documented and improved, ie. `taskcopy` can set a flag (e.g. `flag=gensym()`) in `task.localstorage` to inform that a task has been copied.

xukai92:
This sounds very useful!

yebai:
This could be done through introduce an abstract type (e.g. `TSafe`). All data structures that requires self-copying can be defined as subtypes of `TSafe`.

PR #112, opened 2017-03-22
New copy detection mechanism for TArray.

yebai:


yebai:
Ref #110

src/trace/tarray.jl
```diff
@@ -63,53 +56,62 @@ function Base.getindex(S::TArray, i::Real)
...
+  return c
+end
+
+
 Base.show(io::IO, S::TArray) = Base.show(io::IO, task_local_storage(S.ref)[2])
 Base.display(io::IO, S::TArray) = Base.display(io::IO, task_local_storage(S.ref)[2])
 Base.size(S::TArray) = Base.size(task_local_storage(S.ref)[2])
 Base.ndims(S::TArray) = Base.ndims(task_local_storage(S.ref)[2])
 
 Base.get(t::Task, S) = S
```
tkelman: can the type of S be constrained here? extending a Base method on Base types is discouraged

xukai92: This is actually not used. Commented out in https://github.com/yebai/Turing.jl/pull/316.

PR #115, opened 2017-03-23
Add new name/index to VarInfo for replaying

xukai92:


xukai92:
Up to this point I still keep both `randomness` and `val`, with `randomness` stores the original space variables and `val` stores the transformed variables. This is because in order to use only one (`randomness` ideally) I need to change a lot in HMC (make it do the transform internally instead of using the transformed variable generated in `assume()`). I think it's better to do that in another PR.

yebai:
That's OK.

src/core/varinfo.jl
```diff
@@ -1,31 +1,57 @@
...
+getdist(vi::VarInfo, uid::String) = vi.dists[getidx(vi, uid)]
+setdist!(vi::VarInfo, dist, uid::String) = vi.dists[getidx(vi, uid)] = dist
 
-export VarInfo, syms
+# The default getindex & setindex!() for get & set values
+Base.getindex(vi::VarInfo, uid::String) = getval(vi, uid)
+Base.setindex!(vi::VarInfo, val, uid::String) = setval!(vi, val, uid)
+
+addvar!(vi::VarInfo, uid::String, val, sym::Symbol, dist::Distribution) = begin
+  vi.idcs[uid] = length(vi.idcs) + 1
```
yebai: Sanity check: better to check whether `uid` already exists in `vi` here.

xukai92: I agree. I just added the sanity check there.

xukai92:
It seems the bug in HMC is that I didn't distinguish transformed variable and untransformed variable in the computing of log-joint c41c920662de901ba9fc821e465d8eee60a6713d.

I reconsider the logic of PG and understand the bug. As now we use the same replying method for PG and HMC, PG may cutoff the variables generated by HMC in resampling, but it seems that I cannot do this with the only one container. I guess I still need two independent container for PG and HMC?

yebai:
> I guess I still need two independent container for PG and HMC?

Let's keep both `VarInfo.randomness` and `VarInfo.vals` for now.

Issue #111, opened 2017-03-21
`unflatten` for Mamba Chain


yebai:
In order to support better access to Mamba Chain data structures, we can support the following interface

```julia

c # Mamba Chain
c[:x, sim=1] # Return all samples that has a name starting with `x`, e.g. `x[1][1]`.
c[i, sim=1] # Return the `i`'th sample containing all parameters, 
                     # i.e. in the format of `Dict{Symbol, Any}, where the key is like `:x`. 
```

EDIT (22 Mar 2017): add `c[i, sim=1]`.

xukai92:
Strictly saying, `c[:x]` should return all samples has a **heading symbol** equal to `x`. This distinguishs some variables who is a leading substring of others, e.g. if `x[*]` and `x1[*]` both exist, `c[:x]` only returns `x[*]`.

yebai:
This has now been supported through storing samples in both `Turing` and `Mamba` format (PR #116).

Issue #98, opened 2017-03-08
Code for dealing with non-return is buggy


xukai92:
Need to fix that.

xukai92:
Fixed by https://github.com/yebai/Turing.jl/pull/115/commits/40db08f3763a9dd3d6197a244d411efa3c674314

PR #116, opened 2017-03-24
Unify Mamba.Chain and Turing.Chain.

yebai:
Redefining `Turing.Chain` as a subtype of `Mamba.AbstractChains`. Now `Turing` can use all `Mamba`  ability functions, even without converting `Turing`'s chain type to `Mamba`'s.

Ref: #111


yebai:
@xukai92 It seems that we have a bug with dealing  programs without `return` statements.

Ref: #98

yebai:
Ref: #111

xukai92:
Yes there was. But the problem doesn't exist after we remove `@predict` (https://github.com/yebai/Turing.jl/pull/115/commits/40db08f3763a9dd3d6197a244d411efa3c674314). Sorry that I close the issue before merge the PR.

PR #118, opened 2017-03-24
Run tests in parallel

xukai92:
Make testing faster

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -89,25 +89,56 @@ testcases_excluded = [
   "predict"
 ]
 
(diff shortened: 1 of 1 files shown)


xukai92:
It seems that there is one machine (Linux) failing the parallel version of test running. Any idea why?

yebai:
Not sure, seems some Linux specific issue is causing Julia master worker not able to connect to slave workers.

xukai92:
### Update

Just tried this PR on my local Linux machine, which works fine.

Issue #124, opened 2017-03-27
Merge `randomness` dan `vals` by introducing a new filed `group`


xukai92:
Each inference algorithm should have a group id and `vi.groups[i]` is the variable `vi.vals[i]`'s group.

This should be done after #125

Issue #123, opened 2017-03-27
Unify the `rand()` interface


xukai92:
The interface of `rand()` using different replaying mechanism needs to be unified to 

```
rand(vi, uid, dist, spl)
```

xukai92:
The `rand()` interface should also have one more Boolean parameter which controls whether to 

- replay and update the variables **this** sampler is responsible for, or
- replay the variables **other** samplers are responsible for.

This flag (call `inside` for example) should be set as `true` by default.

Issue #122, opened 2017-03-27
The format of uid need to be enriched


xukai92:
Agreed format:

`"x{##720:1}[1][1,2]..."`,

where `#720` is generated by `gensym()`, `1` after `:` is the counter.

xukai92:
### Discussion

Do you think it's better to re-arrange the format to be `"{##720,x[1][1,2]...}:1"`? 

The reason I feel it's better is that it clearly shows that we are counting the unique set, {gensym(), flattened naming}.

yebai:
we can do that, but maybe have some utilities to easily parse it into a tuple (e.g. `(##720,1,x, [1][1,2]...)`).

xukai92:
Make sense. I think tuples are faster than strings. Is this right?

yebai:
might be, since tuples are stored on the stack

xukai92:
Closed by PR #128

PR #128, opened 2017-03-30
Enrich varname

xukai92:
Solve #122. 

This is the first step to address #123 and #124.


src/core/compiler.jl
```diff
@@ -92,33 +93,36 @@ macro ~(left, right)
       if isa(left, Symbol)
         # Symbol
         assume_ex = quote
-          sym = $(string(left))
+          sym = Symbol($(string(left)))
+          vn = nextvn(vi, Symbol($(string(gensym()))), sym, "")
```
yebai: `gensym()` has to be called during compile time

src/core/varinfo.jl
```diff
@@ -1,65 +1,116 @@
...
-  idcs        ::    Dict{String, Int}
+  idcs        ::    Dict{Tuple, Int}
   vals        ::    Vector{Any}
   syms        ::    Vector{Symbol}
   dists       ::    Vector{Distribution}
   logjoint    ::    Dual
 
-  names       ::    Vector{String}
+  names       ::    Vector{Tuple}
   tsyms       ::    Vector{Symbol}
```
yebai: what's `tsyms`?

Issue #129, opened 2017-03-30
Depreciate the use of `task.storage[:turing_predicts]`


yebai:
Since we have now got an enhanced `VarInfo` type to store all MCMC states, we should consider depreciate using `task.storage[:turing_predicts]`. For example, the following lines need to be updated

- https://github.com/yebai/Turing.jl/blob/master/src/samplers/hmc.jl#L113

xukai92:
Besides https://github.com/yebai/Turing.jl/blob/master/src/samplers/hmc.jl#L113, there are another two places we use `Task.storage`, which are

1. https://github.com/yebai/Turing.jl/blob/master/src/core/container.jl#L187L188
2. https://github.com/yebai/Turing.jl/blob/master/src/samplers/sampler.jl#L64L66

Although our new `VarInfo` store all MCMC states, it doesn't store the original Julia value, i.e. it stores `x[1]`, `x[2]`, ..., instead of the original `x`.

We tackled this problem in our Gibbs sampler by using `varInfo2samples()` (https://github.com/yebai/Turing.jl/blob/master/src/samplers/gibbs.jl#L81) to reconstruct the original Julia value. We can do the same thing for everywhere we uses `Task.storage`, but

1. Our `varInfo2samples()` is not generalized. It only deals with simple array reconstruction, i.e. combing `x[1]`, `x[2]`, ... to `x`.
2. Calling `varInfo2samples()` (i.e. reconstruct Julia type) in each `values(p::Particle)` seems to be expansive.

Do you think we should genealize `varInfo2samples()`? 

Actually I also remember you mentioned it's good to just use the flatten type (as `Mamba` needs it). If this approach is better, I guess we could just abandon the original Julia value and use the flatten version everywhere? If so, the removal of `Task.storage` will also be quite easy.

yebai:
Let's just use the flatten type and drop `varInfo2samples` from the code.

yebai:
Closed by PR #123 
Ref: #133

PR #130, opened 2017-04-01
Unify rand interface

xukai92:
Resolves https://github.com/yebai/Turing.jl/issues/123

### DIFF
--- a/benchmarks/benchmark.jl
+++ b/benchmarks/benchmark.jl
@@ -6,6 +6,7 @@ CONFIG = Dict(
     #"naive-bayes",
     #"normal-loc",
     #"normal-mixture",
+    "simplegauss",
     "gauss",
     "bernoulli",
     "gdemo-geweke"
--- a/benchmarks/gauss.run.jl
+++ b/benchmarks/gauss.run.jl
@@ -11,9 +11,5 @@ describe(sim1)
 sim2 = @sample(gaussmodel(gaussdata), HMC(2000, 0.25, 5))
 describe(sim2)
 
-sim3 = @sample(gaussmodel(gaussdata), HMC(2000, 0.25, 5))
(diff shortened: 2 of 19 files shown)


yebai:
@xukai92 It is probably better to have a few additional tests that can check `randrn` and `VarInfo`, for example

- try to override existing entries in `VarInfo` should return warning or error. (In the case of error, we can add an additional parameter `forceOverride` to `randrn`, and only allow overriding when `forceOverride` is set to true).

- Test whether values stored in `VarInfo` are identical with original copies.

xukai92:
- `randrn()` itself doesn't overwrite the content of `VarInfo`. It only either add a new entry or replay an existing one. 
  - The function HMC uses to change the values in `VarInfo` during its steps is `setval()`. I've added a `overwrite` parameter to `setval()`, with its default value being `false`. 
  - I think give warnings to each overwriting operation is not good as HMC overwrites values in `VarInfo` quite frequently. 
    - I tried adding warning to each overwriting operation and there are too many warnings. 
    - I ended up giving warning messages to `setval()` when users use it without setting `overwrite` to be true. 
- I don't get the second bullet point.
- I can fix #129 in this PR. Will do it soon.
- I believe I put the wrong test for the Gaussian model in benchmark. I've update it.

yebai:
> I don't get the second bullet point: "Test whether values stored in VarInfo are identical with original copies."

Can we add the following test

```julia

# Create a variable and store it in vi
vn  = ... # some variable name
x = randrc(vi, dist, vn, spl)

# Then, test whether vi[vn] == x;

```

Ideally, we should test univariate, vector variate, and matrix variate distributions.

xukai92:
Can we merge this PR?

yebai:
Seems that benchmarking is still broken: https://travis-ci.org/yebai/Turing.jl/jobs/217986302

xukai92:
I see. As we abandon the use original Julia type. We cannot get `:y` now but only `Symbol("y[1]")`  and `Symbol("y[2]")`. I will fix it now.

PR #140, opened 2017-04-07
Support recover Julia's original type

xukai92:
This PR resolves https://github.com/yebai/Turing.jl/issues/133 by converting back `Sample` to Julia's original type.

Some test cases for the supported format can be seen in the test file: https://github.com/yebai/Turing.jl/blob/chain-interface/test/chain_utility.jl


yebai:
This is cool. Does this PR handle general cases, like nested arrays?

xukai92:
Yes I think so. 

I tested a lot of cases in the `Sample` level (see https://github.com/yebai/Turing.jl/blob/chain-interface/test/chain_utility.jl). But only did a test in a model level using the Bayesian HMM one (https://github.com/yebai/Turing.jl/blob/chain-interface/notebooks/BayesHmm.ipynb now uses your code for plotting).

I guess that's fine because `Chain` just does the conversion for each `Sample`?

PR #145, opened 2017-04-10
Hmc space

xukai92:
Resolves https://github.com/yebai/Turing.jl/issues/142

By default HMC also stores untransformed variables in VarInfo but it will do X -> R and R -> X before and after each HMC step (but not each `leapfrog()` and `gradient()`)

Also I finally keep the `trans` field in `VarInfo` because within the HMC step we need to know which variables are temporarily in R, so that we can call `logpdf()` with the right parameter. This actually gives a very neat code in a higher level, e.g. the `assume()` of HMC is very clean now.

### DIFF
--- a/src/core/varinfo.jl
+++ b/src/core/varinfo.jl
@@ -37,6 +37,7 @@ type VarInfo
   vals        ::    Vector{Any}
   dists       ::    Vector{Distribution}
   gids        ::    Vector{Int}   # group ids
+  trans       ::    Vector{Bool}
   logjoint    ::    Dual
   index       ::    Int           # index of current randomness
   num_produce ::    Int           # num of produce calls from trace, each produce corresponds to an observe.
@@ -46,6 +47,7 @@ type VarInfo
     Vector{Any}(),
     Vector{Distribution}(),
     Vector{Int}(),
+    Vector{Bool}(),
     Dual(0.0),
     0,
     0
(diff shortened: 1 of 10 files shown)


src/core/varinfo.jl
```diff
@@ -169,56 +171,50 @@ end
 
 # Main behaviour control of rand() depending on sampler type and if sampler inside
 rand(vi::VarInfo, vn::VarName, dist::Distribution, spl::Sampler, inside=true) = begin
```
yebai: since `inside` needs to be tested for each `rand` call, maybe it makes sense to remove the `inside` parameter, and move the code testing the association of a variable and a sampler into `rand`?

src/samplers/support/helper.jl
```diff
@@ -48,18 +48,42 @@ end
...
-      dist = getdist(vi, uid)
-      val = reconstruct(dist, val)
-      val = invlink(dist, val)
-      val = Any[realpart(val)]
-      val = length(val) == 1 ? val[1] : val   # Remove un-necessary []'s
-    end
-    samples[sym(uid)] = val
+    dist = getdist(vi, uid)
+    r = reconstruct(dist, vi[uid])
+    r = gettrans(vi, uid) ? invlink(dist, r) : r
```
xukai92: Make sense!

yebai:
It is probably better to organise the `rand` function by samplers, ie 

`rand(vi::VarInfo, vn::VarName, dist::Distribution, spl::Sampler, inside=true)`

==>

`rand(vi, vn, dist, spl::{HMCSampler})`
`rand(vi, vn, dist, spl::{PGSampler})`
`rand(vi, vn, dist, spl::{NewSampler})`

src/core/varinfo.jl
xukai92: Yes we can. But PG still need to test the association of a variable and a sampler in order to run `produce(1.0)` in `assume()`.

yebai: I see. we can probably remove `produce(log(1.0))` since it doesn't change particle weights.


Ref: https://github.com/yebai/Turing.jl/blob/3ac05a4e19ead1e48a4e003fed68b39cecc46234/src/samplers/pgibbs.jl#L70

PR #150, opened 2017-04-11
Run Stan models (normal mixture)

xukai92:
Run the normal mixture model in Stan.

The script seems to work.

NOTE: I guess the build filed written by Stan.jl team is problematic itself. We don't necessary need to "build Stan.jl successfully" to use Stan.jl. From the log we can see the Stan model is successfully built and run, even we didn't build the Stan.jl successfully. This happens to my local Linux machine as well.


yebai:
Do we actually need to build Stan.jl?

xukai92:
We don't need to build it explicitly because when we call `Pkg.add("Stan")` it actually builds Stan.jl. I will remove `Pkg.build("Stan")`.

xukai92:
I think the script for bench works now. I'm gonna add Stan version of bernoulli and simplegauss now.

yebai:
It might be useful to compute and print the difference (e.g. mean, variance) between Stan and Turing

xukai92:
Yeah. That makes a lot of sense. Will do!

Issue #126, opened 2017-03-27
Make initialisation of Gibbs sampler standalone


xukai92:

xukai92:
We've agreed that we can have a specific sampler who's responsible for sampling from prior.

xukai92:
- [ ] fix this inside HMC step

PR #157, opened 2017-04-13
Support `data` in compiler

### LINKED ISSUE #135: Depreciate `Void` in `observe`
The following line uses a `Void` type to handle cases where sampler is `null`

https://github.com/yebai/Turing.jl/blob/master/src/samplers/hmc.jl#L141

It is better to have a more principled mechanism to handle such cases.

yebai:
This PR adds support for passing all observed variables using a `data` dictionary. For example

```julia
@model gdemo(x) = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  x[1] ~ Normal(m, sqrt(s))
  x[2] ~ Normal(m, sqrt(s))
  return s, m
end

# here `x` is observed, `s` and `m` are parameters
f1 = gdemo([1.0, 2.0]) # `x` is compulsory

# here data is a dictionary that contains all the observed variables (eg. x, m)
d = Dict(:x => [1.0, 2.0])
f3 = gdemo(data=d) # replace `x` with a `data` dictionary

```

Here if `m` is absent, it is treated as a parameter and inferred by the inference engine.

- [x] Merge two `model` macros into one
- [x] Remove `Turing.TURING` and introduce `compiler` as a keyword argument
- [x] Make initialisation of Gibbs sampler standalone #126 
- [x] Change `csym` generation mechanism #169 
- [x] Fix #135
- [x] Various bug fixes.


yebai:
- [ ] Gibbs `space` assertion is broken.
- [ ] `gibbs_constructor` test does not pass.

xukai92:
Looks great! I briefly reviewed the update but I still don't quite understand how the compiler works. Can you explain it to me tomorrow, and I can do one more careful review?

yebai:
> Looks great! I briefly reviewed the update but I still don't quite understand how the compiler works. Can you explain it to me tomorrow, and I can do one more careful review?

Yes let's meet tomorrow and talk about this.

yebai:
@xukai92  Seems that we need to accumulate `logjoint` to make autodiff work.

yebai:
- [ ] Fix comment
- [ ] Fix #176

yebai:
@xukai92 I've updated comments for the compiler.

xukai92:
> @xukai92 Seems that we need to accumulate logjoint to make autodiff work.

Yeah, I think that's because we previous decided to make calling `model()` directly can give the gradient information of variables.

> @xukai92 I've updated comments for the compiler.

Cool. Let me go over it again and merge the PR.

PR #158, opened 2017-04-13
Improve bench log

xukai92:
Resolve https://github.com/yebai/Turing.jl/issues/151

### DIFF
--- a/benchmarks/benchmark.jl
+++ b/benchmarks/benchmark.jl
@@ -65,7 +65,7 @@ CONFIG = Dict(
     "gauss",
     "bernoulli-stan",
     "bernoulli",
-    # "gdemo-geweke"
+    "gdemo-geweke"
     #"negative-binomial"
   ],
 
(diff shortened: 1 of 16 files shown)


yebai:
This looks good. 
BTW, is it possible to get the running time for Stan as well? This would be helpful for us to get a sense of the performance difference between Stan and Turing (say drawing equal numbers of HMC/NUTS samplers).

xukai92:
I looked into that yesterday but it seems that Stan.jl doesn't store time in the Chain. Alternatively I can call `@elapsed` on the call for Stan but it will include the compile time as well.

xukai92:
- [ ] Improve Simple Gaussian Mixture
- [ ] Get Stan time from .csv

xukai92:
@yebai I think this PR is ready to be merged. Can you please have a look?

xukai92:
> The job exceeded the maximum time limit for jobs, and has been terminated.
It seems that we can't run the benchmark script as long as we want.

yebai:
> There is no timeout for a build; a build will run as long as all the jobs do as long as each job does not timeout.

https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts

I think we might need to print something every 10 mins in order to keep the job alive.

xukai92:
Yes, I've encountered this problem and already done that, i.e. make PG prints progress.

I think our benchmark script did last for 50 mins and got stopped by Travis.

xukai92:
Just checked that the script was stopped after "Ran for 49 min 58 sec"

PR #164, opened 2017-04-14
Fix hmc bug

### LINKED ISSUE #160: ForwardDiff.Dual doesn't support operations between "ambigous" Dual numbers
E.g. blow will give error.

```julia
d1 = Dual(1.0, 0)
d2 = Dual(1.0, 0, 0)
d1 + d2
```
from
```
error("npartials($(typeof(a))) != npartials($(typeof(b)))")
```

NOTE: ForwardDiff write rules to forbid this in promotion and operation (e.g. + here) level.

xukai92:
This PR resolves #160.

The problem was between two HMC samplers, say one responsible for `a` and another for `b`, they `Daul` type value is not reset to `Real`. If `a` and `b` have different dimensions, when we do arithmatic operation on them, `ForwardDiff` will give error.

I added one more step inside HMC step to reset all `Dual` to `Real` to address this issue. The benchmark for Simple Normal Mixture now uses two HMC steps (for variables with different dimensions) as an example.

### DIFF
--- a/benchmarks/simple-normal-mixture.run.jl
+++ b/benchmarks/simple-normal-mixture.run.jl
@@ -2,7 +2,7 @@ include("simple-normal-mixture.data.jl")
 include("simple-normal-mixture.model.jl")
 
 # NOTE: I only run a sub-set of the data as running the whole is quite slow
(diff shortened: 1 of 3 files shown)


yebai:
Makes sense to reset vi after each `HMC` step. 

Ref: #160

PR #187, opened 2017-04-20
Add a thin flag for Gibbs and bugfix

xukai92:
Targeting https://github.com/yebai/Turing.jl/issues/159

Add a optional parameter to Gibbs sampler to control whether to thin the output samples or not.

```julia
s = Gibbs(1000, PG(10, 3, :s), HMC(2, 0.4, 8, :m); thin=false)
# => will output 1000 * (3 + 2) samples then
```

Also found a silly bug in the main loop of Gibbs which makes us actually only run each component for 1 iteration before.

### DIFF
--- a/src/samplers/gibbs.jl
+++ b/src/samplers/gibbs.jl
@@ -1,9 +1,10 @@
 immutable Gibbs <: InferenceAlgorithm
-  n_iters   ::  Int
-  algs      ::  Tuple
+  n_iters   ::  Int     # number of Gibbs iterations
+  algs      ::  Tuple   # component sampling algorithms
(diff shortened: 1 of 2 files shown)


src/samplers/gibbs.jl
```diff
@@ -64,7 +76,7 @@ function sample(model::Function, gibbs::Gibbs)
       # println(varInfo)
       if isa(local_spl, Sampler{HMC}) || isa(local_spl, Sampler{HMCDA})
 
-        for _ in local_spl.alg.n_samples
+        for _ = 1:local_spl.alg.n_samples
```
yebai: it's bit hard to notice this bug..

Issue #193, opened 2017-04-23
NUTS without Dual Averaging (Algorithm 3 of the NUTS paper)


xukai92:
I will implement this first and make sure this part of code is correct.

Issue #182, opened 2017-04-19
Dual Averaging (with default setting suggested by NUTS paper) doesn't work well in transformed space


xukai92:
The default setting for HMC with Dual Averaging method seems not work well with constraints. https://github.com/yebai/Turing.jl/pull/179/commits/49b90e9b417c7fb420b9e740ad68cd6787b1d2fe adds a file to reproduce the situation in `test/hmcda_nb.jl`.

To see the situation, it's simple way is to changed the Line 110 of `samplers/hmcda.jl` to a normal print function `println("leapfrog for $τ steps with step size $ϵ")` and run the test with `julia test/hmcda_nb.jl`. It will then show the step size before each leap frog loop, i.e. the one before the error. It can be seen that the first step size adapted is always around 2.7, which is controlled by the initial Dual Averaging parameters and the target accept rate 0.95 (which is already high to reduce the step size).

yebai:
@xukai92 Is it possible to isolate the numerical issue for transformed variables first? This would allow us to first fix the transform (ie. `simplex` transform here). I imagine this can be done by finding a value `r2` such that 

```julia
r = link(x, Dirichlet(...))  # apply transform

# perform some changes to r, e.g. simulating Hamiltonian dynamics
r2 = ...

x2 = invlink(r, Dirichlet(...)) # some elements of `x2` are `NaN`
```

xukai92:
Yes. I guess if we solve the numerical problem here then we are fine at the moment.

I further looked into the issue and found that the numerical issue only happens to the dual part of dual numbers. 

The below is the `invlink` for the simplex variables.

```julia
function invlink(d::SimplexDistribution, y::Vector)
  K = length(y)
  T = typeof(y[1])
  z = [invlogit(y[k] + log(1 / (K - k))) for k in 1:K-1]
  x = Vector{T}(K)
  for k in 1:K-1
    x[k] = (1 - sum(x[1:k-1])) * z[k]
  end
  x[K] = 1 - sum(x[1:K-1])
  x
end
```

My guess is that  `invlogit(y[k] + log(1 / (K - k)))` is not very safe and lose precision, when calculating the dual parts, the chain rule of derivatives may make the situation even worse. I'm looking into StatsFuns's implementation of some functions to see if I can use some to improve the code here.

xukai92:
I further locate the problem should origin from `gradient()`. It seems that `gradient()` returns `NaN` when numbers go extreme, which later used to update variables.

[...]

PR #194, opened 2017-04-23
NUTS without Dual Averaging (Algorithm 3 of the NUTS paper)

xukai92:
Address #193


xukai92:
Efficient NUTS also seems to suffer from the similar numerical problem in #182.

yebai:
> Efficient NUTS also seems to suffer from the similar numerical problem in #182.

I've finished the new numerical transform; need to cleanup the interface. Should be ready later today.

xukai92:
Cool! I guess I can finish NUTS if this efficient NUTS has no problem cause I have both No-U-Turn and Dual Averaging now.

xukai92:
The simple geweke test (a simpler Gaussian model with only unknown mean) for the sampler seems to be passed.

xukai92:
I will make NUTS PR directly.

Issue #196, opened 2017-04-24
Remove type promotion and conversion of Dual


xukai92:
We are suggested to remove type promotion and conversion of Dual from the ForwardDiff package.

xukai92:
Closed by #197

PR #197, opened 2017-04-24
Remove type promotion and conversion of Dual 

xukai92:
Address #196

### DIFF
--- a/src/samplers/support/helper.jl
+++ b/src/samplers/support/helper.jl
@@ -8,6 +8,8 @@ realpart(d::Array) = map(x -> realpart(x), d)
 dualpart(d::Dual)  = d.partials.values
 dualpart(d::Array) = map(x -> x.partials.values, d)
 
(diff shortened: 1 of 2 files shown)


src/samplers/support/helper.jl
```diff
@@ -8,28 +8,16 @@ realpart(d::Array) = map(x -> realpart(x), d)
...
-
 #####################################################
 # Helper functions for vectorize/reconstruct values #
 #####################################################
 
-vectorize(d::UnivariateDistribution, r)   = Vector{Dual}([r])
-vectorize(d::MultivariateDistribution, r) = Vector{Dual}(r)
-vectorize(d::MatrixDistribution, r)       = Vector{Dual}(vec(r))
+# QUES: will use Any below lead to worse performance?
+vectorize(d::UnivariateDistribution,   r) = Vector{Any}([r])
```
yebai: we can use `Real` here

xukai92: Updated!

PR #199, opened 2017-04-24
Numerically more stable `SimplexDist` transform

yebai:
Changes

- Move accumulation of `VarInfo.logjoint` into `randr` from `assume`.
- Better `Simplex` transform: logpdf, invlink (link might be still vulnerable, but I didn't break it so far).


xukai92:
Looks great! I guess we also need to do that for other two transforms. I experienced the numerical issue for `Beta` and `Gamma` as well.

yebai:
> Looks great! I guess we also need to do that for other two transforms. I experienced the numerical issue for Beta and Gamma as well.

I'll write code for these distributions too. Let's not merge this PR until those transforms are ready

yebai:
*UPDATE*

Running `test/hmcda_nb.jl` still produces `log(-0)` error. I looked into this error more closely, it appears that the error originates from `Dual` operations instead of `invlink`.

xukai92:
Yes. But my understanding for that error message is that the reason `Dual` throw `nan_dom_err` is because there is numerical under/overflow before that operation i.e. the some part of the `Dual` number becomes `NaN`, then later when the `Dual` number is operated, this error occurs.

yebai:
*UPDATE* 

`SimplexTransform` should be stable now. The `test/hmcda_nb.jl` runs without any errors.

xukai92:
Great work!

yebai:
* UPDATE *

The `simplex` transform is stable now.

xukai92:
Great!

yebai:
@xukai92 

Could you point me to some example models that face numerical problems on univariate distributions such as `Beta` and `Gamma`?

xukai92:
The `gauss-demo` will give error on `InverseGamma` (with HMCDA).
The `beta-binomial` will give error on `Beta` (with very large step-size of HMC).

yebai:
Cool - I'll take a look at these examples.

Issue #188, opened 2017-04-20
NUTS implementation


xukai92:
Before implementing the NUTS sampler I think it's better for us to clear up the following issues:

- [x] #174 Merge `ParticleSampler` and `HMCSampler` into `Sampler`?
- [x] #183 Initial step size before heuristically search
- [x] #182 Numerical problem for Dirichelet

Also it may be good to implement the Algorithm 3 in the NUTS paper first before (the one without Dual Averaging), just in purpose of make one part of NUTS work first.

PR #200, opened 2017-04-24
NUTS

xukai92:
Address #193 (eNUTS) #188 (NUTS)

**Still seems to be buggy**

I believe the No-U-Turn algorithm is buggy somehow (the geweke tests for eNUTS and NUTS show a line no strictly pass the origin; but the HMCDA works fine).


yebai:
Great work, will take a look at the geweke test for NUTS tomorrow morning.

PR #220, opened 2017-04-30
Refactor VarInfo

xukai92:
Addressing https://github.com/yebai/Turing.jl/issues/211

There is one point unfinished (make `vals` to a vector of vectors)

**Existing issue(s) to resolve**
- ~~matrix support seems to be unstable (`test_varname.jl`)~~
- [x] add a `Var` type
- [x] make functions that need both `spl` and `gid` clear


xukai92:
The code seems to be problematic for matrix support, of which I haven't figured out why yet.

src/core/varinfo.jl
```diff
@@ -27,25 +34,29 @@ cuid(vn::VarName) = (vn.csym, vn.sym, vn.indexing) # the uid which is only avail
 Base.getindex(graddict::Dict, vn::VarName) = graddict[uid(vn)]
```
yebai: perhaps we can remove `Base.getindex(graddict::Dict, vn::VarName)` and `Base.setindex!(graddict::Dict, val, vn::VarName)` now

src/core/varinfo.jl
```diff
@@ -2,21 +2,28 @@ import Base.string, Base.isequal, Base.==, Base.convert
...
+# VarName #
+###########
 immutable VarName
   csym      ::    Symbol        # symbol generated in compilation time
   sym       ::    Symbol        # variable symbol
   indexing  ::    String        # indexing
   counter   ::    Int           # counter of same {csym, uid}
 end
 
+typealias UID Tuple{Symbol,Symbol,String,Int}
```
yebai: Perhaps we can remove `UID`, if its existence is to enable us to use VarName as a key for dictionaries. See e.g. 

http://stackoverflow.com/questions/26458149/julia-dictionary-with-tuple-keys-containing-composite-types

src/core/varinfo.jl
xukai92: Yes!

src/core/varinfo.jl
xukai92: Thank you for pointing me to this! Will do it!

src/core/varinfo.jl
```diff
@@ -55,109 +66,158 @@ type VarInfo
...
-  # Rebuild index dictionary
-  for i = 1:length(vi.uids)
-    vi.idcs[vi.uids[i]] = i
-  end
   vi
 end
 
 # Add a new entry to VarInfo
-addvar!(vi::VarInfo, vn::VarName, val, dist::Distribution, gid=0) = begin
+addvar!(vi::VarInfo, vn::VarName, val, dist::Distribution) = addvar!(vi, vn, val, dist, 0)
```
yebai: Shall we introduce a `Var` type

```julia
type Var
         vn :: VarName
         d  :: Distributions
         gid :: Int
       end
```

then we can define `push!` to replace `addvar!`

```
push!(vi::VarInfo, v::Var)
```

yebai:
Remove 

- setdist!
    - Maybe we should remove `setdist!`, since we don't want to modify dist of a variable once the variable is created.
- settrans!
    - Perhaps we should remove `settrans!` and keep 
    -  vi2 = link(vi::VarInfo)
    -  vi2 = invlink(vi::VarInfo)
    - here `link` and `invlink` will produce a deep copy of `vi`, but add/remove the transformation
- `keeplast!`    ==> vi2 = last(vi::VarInfo)  
     - `last` produces a deep copy of `vi`, but only keeping the last row of `VarInfo.val`
- `duplicate!`  ==> `expand!`

src/core/varinfo.jl
xukai92: I meet further problem of this point. I found the problem that we cannot use `VarName` as a key origins from the way how Julia treats strings inside types. A minimum example is given below.

```julia
immutable B
    s   ::  String
end
b1 = B("1")
b2 = B("2")
hash(b1.s) # => 0x8f6e84dd949e9002
hash(b2.s) # => 0x2ba9e56d0c27e607
b1.s == b2.s # => false
```

This means no matter how I define hash it won't work because the strings inside are "different". Do you know why this happens?

xukai92: I guess I need further look into the probelm.

yebai: I don't quite understand the problem you are meeting. But will be in the lab this afternoon, let's discuss then.

xukai92: Yes. I created a issue to record that https://github.com/yebai/Turing.jl/issues/221

xukai92: I sorted this out!

src/core/varinfo.jl
xukai92: We can do this. Do we want it be a `type` or `immutable` then?

yebai: `immutable` makes sense. Note this `Var` is only a temporary data structure,  and we will keep the type of `VarInfo.val`, i.e.  `Vector{Vector{Real}}` .

xukai92: So type `Var` should also have a field `val` right?

yebai: Right

xukai92: It seems that the only place this `Var` can be used is in `initvar`, because `Var` also need `val` so it can only be used after the value is generated.

yebai: I see, maybe it doesn't help too much to have a `Var` type?

xukai92: I guess so. Or maybe we can enrich `Var` and add some method? Something like below

```julia
immutable Var
  vn    ::  VarName
  val   ::  Vector{Real}
  dist  ::  Distribution
  gid   ::  Int
end

macro dispatch(v)
    esc(
        quote
            vn = $(v).vn
            ...
        end
    )
end

@dispatch v
# then use vn
```

So we can use it when some fields are missing even

xukai92:
Along with removing `randr` from `VarInfo`, shall we also remove 

```julia
randr(t::Trace, vn::VarName, distr::Distribution) = randr(t.vi, vn, distr, true)
Distributions.rand(t::Trace, vn::VarName, dist::Distribution) = randr(t, vn, dist)
```

from `trace.jl`?

yebai:
Do we still have any dependency on these functions?

src/core/varinfo.jl
```diff
@@ -55,112 +57,167 @@ type VarInfo
...
+  dist = getdist(vi, vn)
+  r = reconstruct(dist, getval(vi, vn))
+  r = istransformed(vi, vn) ? invlink(dist, r) : r
+end
 
-uids(vi::VarInfo) = Set(keys(vi.idcs))            # get all uids
-syms(vi::VarInfo) = map(uid -> uid[2], uids(vi))  # get all symbols
+Base.setindex!(vi::VarInfo, r, vn::VarName) = begin
+  dist = getdist(vi, vn)
+  setval!(vi, vectorize(dist, r), vn)
```
yebai: do we need to check transformation here?

xukai92:
> Do we still have any dependency on these functions?
Only few places still use `rand(Trace, ...` but we can simply change them.

src/core/varinfo.jl
xukai92: Do you mean by transforming `r` to the corresponding format indicated in `VarInfo`? Actually this something not clear to me because we don't really know whether `r` is transformed or not. Shall we assume `r` is in the original space then?

xukai92: Or maybe we shall forbid `setindex!` by `VarName`?

xukai92:
appveyor not running?

src/trace/trace.jl
```diff
@@ -94,14 +81,10 @@ function forkc(trace :: Trace)
...
-    gid = trace.spl.alg.group_id
-  else
-    gid = 0
-  end
 
-  n_rand = min(trace.vi.index, length(groupvals(trace.vi, gid, trace.spl)))
-  newtrace.vi = retain(deepcopy(trace.vi), gid, n_rand, trace.spl)
+  n_rand = min(trace.vi.index, length(groupvals(trace.vi, trace.spl)))
+  newtrace.vi = deepcopy(trace.vi)
+  retain!(newtrace.vi, n_rand, trace.spl)
```
yebai: maybe merge `retain!` and `setindex!`, e.g.

```julia
 n_rand = min(trace.vi.index, length(groupvals(trace.vi, trace.spl)))
 newtrace.vi = retain(deepcopy(trace.vi), gid, n_rand, trace.spl)		 +  newtrace.vi = deepcopy(trace.vi)
 retain!(newtrace.vi, n_rand, trace.spl)
```

==>


```julia
vi[spl] = NaN
```

src/samplers/support/hmc_core.jl
```diff
@@ -53,83 +38,59 @@ function leapfrog(_vi, _p, τ, ϵ, model, spl)
...
-      for i = 1:length(val_vec)
-        val_vec[i] = val_vec[i] + ϵ * p[k][i]
-      end
-    end
+    p -= ϵ * grad / 2
+
+    expand!(vi)
+
+    ranges = getranges(vi, spl)
+    vi[ranges] += ϵ * p             # full step for state
```
yebai: maybe merge the following to lines

```julia
ranges = getranges(vi, spl)
vi[ranges] += ϵ * p             # full step for state
``` 

=>

```julia
vi[spl] = += ϵ * p         # full step for state
```

yebai:
Nice work, the `HMC` related code is much more readable now. One additional suggestion about `setindex!` and `getindex`

- Implement function `setindex!(vi::VarInfo, spl::Sampler)` and `getindex(vi::VarInfo, spl::Sampler)`

Then we can write code like 

```julia
vi[spl] = vi[spl] + ϵ * p 
```

src/trace/trace.jl
xukai92: For `retain!` we also need the `n_rand`, it would be weird to do something like `vi[spl, n_rand]` I think.

src/samplers/support/hmc_core.jl
xukai92: This looks cool! Let's do this.

xukai92:
I think we could get some speed-up by caching `getranges(vi, spl)`. But it seems that then branching statements life `if` will have issues?

src/trace/trace.jl
xukai92: Or the right hand side of `vi[spl]` should be a vector, something like

```julia
vals = vi[spl]
for i = 1:n_rand
  vals[end-i+1] = NaN
end
newtrace.vi[spl] = vals
```

yebai: perhaps we can write

```julia
val = newtrace.vi[spl]*NaN
newtrace.vi[spl] = val
```

yebai: Or?

```julia
newtrace.vi[spl, end-n_rand+1:end] = NaN
```

yebai:
> I think we could get some speed-up by caching getranges(vi, spl). But it seems that then branching statements life if will have issues?

I see, maybe leave this as is for now and come back later.

src/trace/trace.jl
xukai92: We cannot use `end` here because we don't have `ndims(VarInfo)`. However, we can't even implement this `ndims` because the total length is dependent on the sampler. So we have to write things like

```julia
l = length(getranges(newtrace.vi, spl))
newtrace.vi[spl, n_rand+1:l] = NaN
```

This extra line of code doesn't looks neat.

xukai92: Here is another problem here: `n_rand` is not essentially the length of `vals` to keep because all variables are flatten (but `newtrace.vi[spl]` is aimed to so if we make `n_rand+1:l` to be variable-level range then it is not consistent).

yebai: Maybe add a function 'getretain'

r = getretain(spl, vi)
vi[spl, r] = NULL

xukai92: Yeah we could do this

xukai92:
@yebai I've made the changes we discussed. Can you have a look at them?

xukai92:
@yebai I guess that there is something wrong/changed in the AppVeyor side - I re-run one commit (https://ci.appveyor.com/project/yebai/turing-jl/build/1.0.803/job/1qcn12mdpje12le4) which was successful on AppVeyor but it is also stuck before `ad.jl`.

PR #235, opened 2017-05-04
Nuts models

xukai92:
#229

The SV model is unfinished - not sure what does the last term of equation 23 mean.


yebai:
That's a student-t distribution, see e.g. https://en.wikipedia.org/wiki/Student%27s_t-distribution

yebai:
@xukai92 are we ready to merge this PR now?

xukai92:
Yes let's merge it

PR #243, opened 2017-05-05
Unify step

xukai92:
Address last two unsolved points in https://github.com/yebai/Turing.jl/issues/107

The `step` function for both PG and HMC only returns an object of `VarInfo` now:
- For HMC, I removed the return `true`/`false` value to indicate but store the accept history in `spl.info[:accept_his]` with in the `step` function
- For PG, there were two return variables: `ref_particle` and `samples`
  1. `samples` is easy to remove as it just the samples in `ref_particle`
  2. `ref_particle` is removed by re-created the particle (`fork2(TracR(...))`) using the corresponding `vi`
      - I noticed that in the original code `fork2` always clean up `vi` - does it means the reference particle is always a "new" one?

### DIFF
--- a/src/samplers/gibbs.jl
+++ b/src/samplers/gibbs.jl
@@ -79,12 +79,7 @@ function sample(model::Function, alg::Gibbs)
           dprintln(2, "recording old θ...")
           varInfo = step(model, local_spl, varInfo, i==1)
           if ~spl.alg.thin
-            if isa(local_spl.alg, Hamiltonian) && ~local_spl.info[:accept_his][end]
-              samples[i_thin] = samples[i_thin - 1]
-            else
-              samples[i_thin].value = Sample(varInfo).value
-            end
-            i_thin += 1
+            samples[i_thin].value = Sample(varInfo).value; i_thin += 1
           end
         end
       else
--- a/src/samplers/gibbs.jl
+++ b/src/samplers/gibbs.jl
@@ -62,7 +62,7 @@ function sample(model::Function, alg::Gibbs)
   end
 
   # Init parameters
-  varInfo = model(); ref_particle = nothing; n = spl.alg.n_iters; i_thin = 1
+  varInfo = model(); n = spl.alg.n_iters; i_thin = 1
 
   # Gibbs steps
   spl.info[:progress] = ProgressMeter.Progress(n, 1, "[Gibbs] Sampling...", 0)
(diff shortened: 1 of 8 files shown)


yebai:
Nice work!

yebai:
> I noticed that in the original code fork2 always clean up vi - does it means the reference particle is always a "new" one?

I'm not sure I understand the question. `fork2` re-creates a particle by replaying all random variables in `VarInfo`. The new particle should be identical with the replayed one. More specifically,

`fork2(s :: TraceR) = forkr(s, 0, true)` 

the last parameter `true` tells `forkr` NOT to clean contents of `VarInfo`.

xukai92:
I see. Thank you for pointing this. I mis-understand the flag when I read the code yesterday.

src/samplers/gibbs.jl
```diff
@@ -71,51 +71,37 @@ function sample(model::Function, alg::Gibbs)
...
-            #    which will add new elements to vi.idcs etc.
-            varInfo = old_vi
-          end
-          if ~spl.alg.thin
-            samples[i_thin].value = Sample(varInfo).value
-            i_thin += 1
-          end
+      if isa(local_spl.alg, GibbsComponent)
+        if isa(local_spl.alg, PG)
+          # Clean possible wrong counter from HMC
```
yebai: I think we can remove these code, because `fork2(vi::VarInfo)` will perform these operations anyway.

src/samplers/gibbs.jl
```diff
@@ -71,51 +71,37 @@ function sample(model::Function, alg::Gibbs)
...
-        varInfo = deepcopy(varInfo); varInfo[getretain(varInfo, 0, local_spl)] = NULL
-
-        # Local samples
         for _ = 1:local_spl.alg.n_iters
-          ref_particle, _ = step(model, local_spl, varInfo, ref_particle)
+          dprintln(2, "recording old θ...")
+          varInfo = step(model, local_spl, varInfo, i==1)
           if ~spl.alg.thin
-            samples[i_thin].value = Sample(ref_particle.vi).value
+            if isa(local_spl.alg, Hamiltonian) && ~local_spl.info[:accept_his][end]
```
yebai: can we merge these lines 

```julia
if isa(local_spl.alg, Hamiltonian) && ~local_spl.info[:accept_his][end]
              samples[i_thin] = samples[i_thin - 1]
            else
              samples[i_thin].value = Sample(varInfo).value
            end
```

==>

```julia
     samples[i_thin].value = Sample(varInfo).value
```

Since if a proposal is rejected, `step` will return the old `VarInfo` anyway.

src/samplers/gibbs.jl
xukai92: Yes let's do this.

src/samplers/gibbs.jl
xukai92: I did that in order to save some computation as the conversion from `VarInfo` to `Sample` still waste some time.

Regarding to this, I was thinking if we can just save `VarInfo` (somehow) and do the conversion from `VarInfo` to `Sample` when the inference is finished?

yebai: Yes, since `VarInfo.val`is a vector, it can store multiple MCMC samples.

xukai92:
@yebai I've done the changes. Let's merge this PR when the tests pass.

yebai:
Sure!

src/samplers/gibbs.jl
```diff
@@ -79,12 +79,7 @@ function sample(model::Function, alg::Gibbs)
           dprintln(2, "recording old θ...")
           varInfo = step(model, local_spl, varInfo, i==1)
           if ~spl.alg.thin
-            if isa(local_spl.alg, Hamiltonian) && ~local_spl.info[:accept_his][end]
-              samples[i_thin] = samples[i_thin - 1]
-            else
-              samples[i_thin].value = Sample(varInfo).value
-            end
-            i_thin += 1
```
yebai: Do we need to keep `i_thin += 1`?

xukai92: I put them in the same line.

Issue #259, opened 2017-05-17
Modularize DA and pre-cond


xukai92:
Duplicated code are in HMCDA and NUTS. Need to be modularized/merged.

xukai92:
Closed by #267

Issue #206, opened 2017-04-27
Save parameter adaption trajectory for HMCDA, NUTS


yebai:
It's helpful to save parameter adaption history in `Sampler.info` for later inspection, e.g. inspecting dual averaging divergence. 

For example
- epsilon, 
- simulated Hamiltonian length: L'*epsilon, where L' is the actual number of leapfrog steps before rejection.

xukai92:
I started to do this in #267 (e.g. commit 34819c30e94248ea797766b0a539a21070b23165). I finally add a more-riched interface to generate sampler, saying `Sample(vi, spl)`, which combines both info from `VarInfo` and `spl.info`. This is pretty neat for HMC(DA) and NUTS but seems Gibbs is still bit messy because for samplers like PG the `spl.info` doesn't have information like `epsilon` - so in order to make each sample from the Gibbs sample, I currently manually check and record the information from the previous HMC sampler and add them to the PG one so each sample have the same dimension of information (otherwise Chain will be broken).

xukai92:
Closed by PR #267 and in favor of issue https://github.com/yebai/Turing.jl/issues/269

PR #267, opened 2017-05-26
Modularize da precond

xukai92:
Addressing #259

### DIFF
--- a/src/samplers/support/adapt.jl
+++ b/src/samplers/support/adapt.jl
@@ -24,3 +24,13 @@ adapt_step_size(spl::Sampler, stats::Float64) = begin
     end
   end
 end
+
+macro init_da_parameters()
+  quote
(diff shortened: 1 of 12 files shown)


xukai92:
Cario build failed on AppVeyor

xukai92:
Pushed few new commits on addressing #206.

src/samplers/support/adapt.jl
```diff
@@ -0,0 +1,60 @@
+adapt_step_size(spl::Sampler, stats::Float64, δ::Float64) = begin
```
yebai: maybe use a more specific sampler type, e.g. `Sampler{Hamiltonian}`?

yebai: also for other adaption functions

xukai92: Sure, let me do this.

Issue #251, opened 2017-05-11
More organized visualization


xukai92:
Guess we need more more organized visualization scripts for each type of dataset.

TODOs
- [x] Clean and move scripts from https://github.com/xukai92/Turing-exps/tree/master/amazon-talk

xukai92:
Closed by PR #271.

PR #271, opened 2017-05-29
Add visualization script and notebook for topic models

xukai92:
#251


xukai92:
How to add things to the nbviewer you used?

yebai:
> How to add things to the nbviewer you used?

It will automatically pull from Github, so the notebook will be there when this PR is merged. 

BTW, let's stop adding the PDF file to the repo since nbviewer can correctly display all figures.

xukai92:
I see. Let me remove the .pdf file.

Issue #219, opened 2017-04-30
Better initialization


xukai92:
3 runs with the same code for `sample(gdemo(x), Gibbs(1000, PG(30, 3, :s), HMC(1, 0.2, 3, :m)))`

```julia
[Gibbs] Sampling...100% Time: 0:00:08
  1. s ≈ 49/24 (ϵ = 0.2) X
    s = 2.5293348244112015, diff = 0.487668157744535
  2. m ≈ 7/6 (ϵ = 0.2)   X
     m = 0.9544717098265469, diff = 0.2121949568401198
                  
[Gibbs] Sampling...100% Time: 0:00:11
  1. s ≈ 49/24 (ϵ = 0.2) ✓
    s = 2.223441561488678, diff = 0.18177489482201148
  2. m ≈ 7/6 (ϵ = 0.2)   ✓
    m = 0.9683037640417662, diff = 0.19836290262490053

[Gibbs] Sampling...100% Time: 0:00:13
  1. s ≈ 49/24 (ϵ = 0.2) ✓
    s = 1.9990257822648827, diff = 0.04264088440178382
  2. m ≈ 7/6 (ϵ = 0.2)   ✓
    m = 1.1000766092540895, diff = 0.06659005741257729
```

Why Gibbs is so sensitive to initialization?

PR #272, opened 2017-05-29
Better initialization #219

xukai92:
Addressing #219

### DIFF
--- a/src/samplers/sampler.jl
+++ b/src/samplers/sampler.jl
@@ -36,6 +36,8 @@ assume(spl::Void, dist::Distribution, vn::VarName, vi::VarInfo) = begin
     r = init(dist)
     push!(vi, vn, r, dist, 0)
   end
+  # NOTE: The importance weight is not correctly computed here because
+  #       r is genereated from some uniform distribution which is different from the prior
   acclogp!(vi, logpdf(dist, r, istrans(vi, vn)))
   r
 end
--- a/src/samplers/sampler.jl
+++ b/src/samplers/sampler.jl
@@ -7,6 +7,7 @@ include("support/resample.jl")
 end
 include("support/hmc_core.jl")
 include("support/adapt.jl")
+include("support/init.jl")
 include("hmcda.jl")
 include("nuts.jl")
 include("hmc.jl")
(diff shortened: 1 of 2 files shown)


src/samplers/sampler.jl
```diff
@@ -32,7 +33,7 @@ assume(spl::Void, dist::Distribution, vn::VarName, vi::VarInfo) = begin
   if haskey(vi, vn)
     r = vi[vn]
   else
-    r = rand(dist)
+    r = init(dist)
```
yebai: perhaps we should only do this for Hamiltonian samplers, since sometimes we need to sample from the prior and compute importance weights. 

```julia
isa(eltypes(spl), Hamiltonian) ? init(dist) : rand(dist)
```

```

Comment (xukai92) [inline src/samplers/sampler.jl]: Yes that makes sense, but the problem is this `assume()` is called with `spl = nothing` - we don't have the information of sampler. Maybe it's still OK because the the `assume()` with sampler given still uses `rand()` ?

Comment (yebai) [inline src/samplers/sampler.jl]: I see. Perhaps it's time to introduce a `Init` sampler type that takes care of initialisation. Then we can get rid of void samplers from our code, and 

```
type Init{T<:InferenceAlgorithm} end

# then for HMC
Init{HMC}

# for samplers using prior as a proposal
Init{SMC} # or Init{IS}, Init{PG}.
```

Comment (xukai92) [inline src/samplers/sampler.jl]: Yes! Good idea. Let's do this.

Comment (xukai92) [inline src/samplers/sampler.jl]: I'm a bit unsure if it is necessary when implementing this: the place this `Init` is need is the `model(sampler=Init(alg))` in `hmc.jl` and `gibbs.jl`. For both (even in `gibbs.jl`), the values initialized by `model()` seems to be never used by PG, because PG will always clear (i.e. set values to `NULL`) in the beginning of each step.

Comment (yebai) [inline src/samplers/sampler.jl]: That's true, where this matters is inside samplers like `IS`, which will repetitively call `model` to generate samples and compute importance weights.

Comment (xukai92) [inline src/samplers/sampler.jl]: But IS do have its own definition of `assume()` to generate samples (below). It always calls `model` with `sampler::Sampler{IS}`.
```julia
assume(spl::Sampler{IS}, dist::Distribution, vn::VarName, vi::VarInfo) = begin
  r = rand(dist)
  push!(vi, vn, r, dist, 0)
  r
end
```

Issue #152, opened 2017-04-11
support resume MCMC sampling, e.g. `resume(s::Chain, num_iter)`


yebai:
Migrated from #96.

TODOs:
- [x] Add `Sampler` as a field of `Chain`
- [x] Add model as a field of `Chain`
- [x] Add last vi as a field of `Chain`
- [x] Support resume for HMC
- [x] Support resume for PG
- [x] Support resume for Gibbs
- [x] Add a `resume(s::Chain, num_iter)` wrapper

xukai92:
After thinking possible implementations of this feature, I ended up believing the following interface is the most preferable

```julia
chain = sample(model, alg)
sample(model, alg; resume_from=chain)
```

The good point about this is that we can reuse most of the code in `sample()`. For the enriched-chain type, we just need to additionally store sampler (for adaptive sampler we need it because of the adaptive parameters), model and last vi; in `sample()`, we check if the `resume_from` is provided:

- if so, instead of initialising by calling `model()`, just fetch the vi from provided chain & instead of creating a new sampler, use the provided one;
- if not, do sample as before.

Of course, by introducing this interface, we can also create a wrapper called `resume()` suggested before.

One question here is that:

Should the returned chain containing the samples for 2 runs or just the new run, i.e. appending samples to the new chain or create a new one?

How do you think?

yebai:
Looks sensible to me!

> Should the returned chain containing the samples for 2 runs or just the new run, i.e. appending samples to the new chain or create a new one?

If we resume running one chain (e.g. continuing from previous `VarInfo`) then we should store all samples in one chain. Otherwise, we should return a new chain.

xukai92:
Cool. Make sense to me. Let me implement to this design

xukai92:
Closed by https://github.com/yebai/Turing.jl/pull/278

PR #278, opened 2017-06-04
Support resume MCMC sampling #152

xukai92:
Nearly finished; progress see checklist of https://github.com/yebai/Turing.jl/issues/152.

Key changes:

- Add a `info` dict to `Chain` to store last `vi`, `sampler` and `model`
- Add keyword arguments to `sample()` function
  - `save_state`: if set `true`, the state (vi, spl, model) will be save to Chain's `info`
  - `resume_from`: if provided with an existing chain, the last vi will be used to continue the chain
  - `reuse_adapt` (only for HMC): if set `true`, provided Chain's spl.info will be used.

Left to do:
- [x] Fix calculation of logevidence of resumed PG chain.


xukai92:
NOTE
a19f3cc57d0a91f1eca5bcdcd5bee4ea85f53021 fails due to: `Cairo had build errors.`

xukai92:
I'm thinking of removing the `reuse_adapt` but moving this feature of re-using sampler into the `resume(s::Chain, num_iter)` wrapper. Otherwise it seems a bit messy.

xukai92:
Also `PG` `HMC` cannot resume each other's chain now due to the problem mentioned in https://github.com/yebai/Turing.jl/issues/277

xukai92:
Tests are available at https://github.com/yebai/Turing.jl/blob/5b5068e16163583482d1167aa918d66be53c60f2/test/io.jl/save_resume_chain.jl

Issue #276, opened 2017-05-31
Vectorising `link` and `invlink`


yebai:
As part of vectorisation for IID distributions( #117, #255), we need to also vectorise the pair of transform functions (e.g. `link` and `invlink`) for bounded parameters.

xukai92:
Done in #281

PR #281, opened 2017-06-05
Vec assume

xukai92:
On working of https://github.com/yebai/Turing.jl/issues/255

I'm trying to vectorize the transformation as well (#276). I already done that for univariate and simplex.

TODOs
- [x] vec mat
- [x] vec mat transformation(s)

### DIFF
--- a/src/samplers/hmc.jl
+++ b/src/samplers/hmc.jl
@@ -163,7 +163,7 @@ assume{A<:Hamiltonian,D<:Distribution}(spl::Sampler{A}, dists::Vector{D}, vn::Va
 
   acclogp!(vi, sum(logpdf(dist, rs, istrans(vi, vns[1]))))
 
-  if isa(dist, UnivariateDistribution)
+  if isa(dist, UnivariateDistribution) || isa(dist, MatrixDistribution)
     var = rs
   elseif isa(dist, MultivariateDistribution)
     if isa(var, Vector)
@@ -175,8 +175,6 @@ assume{A<:Hamiltonian,D<:Distribution}(spl::Sampler{A}, dists::Vector{D}, vn::Va
     else
       error("[Turing] unsupported variable container")
     end
-  elseif isa(dist, MatrixDistribution)
-    # nothing now
   end
 
   var
--- a/src/samplers/hmc.jl
+++ b/src/samplers/hmc.jl
@@ -95,7 +95,10 @@ function sample{T<:Hamiltonian}(model::Function, alg::T;
   for i = 1:n
     samples[i] = Sample(weight, Dict{Symbol, Any}())
   end
-  vi = model()
+
+  vi = resume_from == nothing ?
+       model() :
+       deepcopy(resume_from.info[:vi])
 
   if spl.alg.gid == 0
     link!(vi, spl)
@@ -107,7 +110,7 @@ function sample{T<:Hamiltonian}(model::Function, alg::T;
   for i = 1:n
     dprintln(2, "$alg_str stepping...")
 
-    time_elapsed = @elapsed vi = step(model, spl, vi, i==1)
+    time_elapsed = @elapsed vi = step(model, spl, vi, i == 1)
     time_total += time_elapsed
 
     if spl.info[:accept_his][end]     # accepted => store the new predcits
@@ -138,6 +141,8 @@ function sample{T<:Hamiltonian}(model::Function, alg::T;
   end
   c = Chain(0, samples)       # wrap the result by Chain
   if save_state               # save state
+    # Convert vi back to X if vi is required to be saved
+    if spl.alg.gid == 0 invlink!(vi, spl) end
     save!(c, spl, model, vi)
   end
 
--- a/src/samplers/hmc.jl
+++ b/src/samplers/hmc.jl
@@ -169,13 +169,16 @@ assume{A<:Hamiltonian,D<:Distribution}(spl::Sampler{A}, dists::Vector{D}, vn::Va
   acclogp!(vi, sum(logpdf(dist, rs, istrans(vi, vns[1]))))
 
   if isa(dist, UnivariateDistribution) || isa(dist, MatrixDistribution)
+    @assert size(var) == size(rs) "[assume] variable and random number dimension unmatched"
     var = rs
   elseif isa(dist, MultivariateDistribution)
     if isa(var, Vector)
+      @assert length(var) == size(rs)[2] "[assume] variable and random number dimension unmatched"
       for i = 1:n
         var[i] = rs[:,i]
       end
     elseif isa(var, Matrix)
+      @assert size(var) == size(rs) "[assume] variable and random number dimension unmatched"
       var = rs
     else
       error("[Turing] unsupported variable container")
--- a/src/samplers/hmc.jl
+++ b/src/samplers/hmc.jl
@@ -152,18 +152,34 @@ assume{T<:Hamiltonian}(spl::Sampler{T}, dist::Distribution, vn::VarName, vi::Var
   r
 end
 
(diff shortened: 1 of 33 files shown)


xukai92:
Note about using array of array v.s. mat result from `vec_assume_mv.jl`:


```julia
# Test for vectorize UnivariateDistribution
@model vdemo() = begin
  phi = Vector{Vector{Real}}(N)
  phi ~ [Dirichlet(beta)]
end

t_vec = @elapsed res_vec = sample(vdemo(), alg)

@model vdemo() = begin
  phi = Matrix(2,N)
  phi ~ [Dirichlet(beta)]
end

t_vec_mat = @elapsed res_vec_mat = sample(vdemo(), alg)

@model vdemo() = begin
  phi = Vector{Vector{Real}}(N)
  for i = 1:N
    phi[i] ~ Dirichlet(beta)
  end
end

t_loop = @elapsed res = sample(vdemo(), alg)

println("Time for")
println("  Loop : $t_loop")
println("  Vec  : $t_vec")
println("  Vec2 : $t_vec_mat")
```

gives

```julia
Time for
  Loop : 3.33750112
  Vec  : 15.286217032
  Vec2 : 2.782706592
```

Mat seems much faster (should be also related to my implementation of converting mat back to nested array)

xukai92:
Note: `Distributions.jl` doesn't support vectorisation version of `logpdf` of matrixvariate variables.

xukai92:
Also there is a typo(bug) in their fake verctorization implementation. For this  reason I will not include the test file in runtest.jl

yebai:
@xukai92 Do you want me to review this PR? Seems tests are currently broken.

xukai92:
Seems there is a bug of the resuming MCMC feature which may cause numerical error very easily. I guess it might be the case that when a chain is resumed, I call the `find_good_eps` again, which should be disabled. Let me fix this first.

xukai92:
@yebai I fixed few bugs and improve the stability a bit. Could you have a review now plz?

yebai:
>Time for
>  Loop : 9.823776041
>  Vec  : 21.934951717
>  Vec2 : 6.140146129
> with N=20.
> Mat seems much faster (should be also related to my implementation of converting mat back to nested array)

 impressive speedup! I've started reviewing this PR and will add my comments soon.

src/samplers/hmc.jl
```diff
@@ -152,18 +157,32 @@ assume{T<:Hamiltonian}(spl::Sampler{T}, dist::Distribution, vn::VarName, vi::Var
...
   vns = map(i -> copybyindex(vn, "[$i]"), 1:n)
 
-  rs = vi[vns]
+  rs = vi[vns]  # NOTE: inside Turing the Julia conversion should be sticked to
 
   acclogp!(vi, sum(logpdf(dist, rs, istrans(vi, vns[1]))))
 
-  rs
+  if isa(dist, UnivariateDistribution) || isa(dist, MatrixDistribution)
+    var = rs
```
yebai: perhaps we should add a dimensionality assertion here:

```julia
@assert size(var) == size(rs)
```

src/samplers/hmc.jl
```diff
@@ -152,18 +157,32 @@ assume{T<:Hamiltonian}(spl::Sampler{T}, dist::Distribution, vn::VarName, vi::Var
...
+  elseif isa(dist, MultivariateDistribution)
+    if isa(var, Vector)
+      for i = 1:n
+        var[i] = rs[:,i]
+      end
+    elseif isa(var, Matrix)
+      var = rs
```
yebai: dimensionality assertion?

src/samplers/sampler.jl
```diff
@@ -42,25 +42,42 @@ assume(spl::Void, dist::Distribution, vn::VarName, vi::VarInfo) = begin
...
   else
     rs = rand(dist, n)
-    for i = 1:n
-      push!(vi, vns[i], rs[i], dist, 0)
+
+    if isa(dist, UnivariateDistribution) || isa(dist, MatrixDistribution)
+      for i = 1:n
+        push!(vi, vns[i], rs[i], dist, 0)
+      end
+      var = rs
```
yebai: dimensionality assertion?

src/samplers/sampler.jl
```diff
@@ -42,25 +42,42 @@ assume(spl::Void, dist::Distribution, vn::VarName, vi::VarInfo) = begin
...
+    elseif isa(dist, MultivariateDistribution)
+      for i = 1:n
+        push!(vi, vns[i], rs[:,i], dist, 0)
+      end
+      if isa(var, Vector)
+        for i = 1:n
+          var[i] = rs[:,i]
+        end
+      elseif isa(var, Matrix)
+        var = rs
```
yebai: dimensionality assertion?

src/samplers/support/transform.jl
```diff
@@ -167,6 +198,23 @@ Distributions.logpdf{T}(d::SimplexDistribution, x::Vector{T}, transform::Bool) =
...
+      Z[k,:] = X[k,:] ./ (one(T) - sum(X[1:k-1,:],1))'
+    end
+    for k = 1:K-1
+      lp += log(Z[k,:]) + log(one(T) - Z[k,:]) + log(one(T) - sum(X[1:k-1,:], 1))'
+    end
+  end
+  lp
+end
+
+# REVIEW: why do we put this piece of code here?
```
yebai: the distributions package does not check `d.p[x] > 0.0`, so I override their definition here. It's probably better to submit a PR to Distributions package to fix this.

src/samplers/hmc.jl
xukai92: Yeah the assertions makes sense to me!

src/samplers/support/transform.jl
xukai92: Yeah I created an issue (https://github.com/yebai/Turing.jl/issues/285) for this.

xukai92:
Vec LDA 215s v.s. loop LDA 296s (https://travis-ci.org/yebai/Turing.jl/jobs/242001041)

yebai:
> Vec LDA 215s v.s. loop LDA 296s (https://travis-ci.org/yebai/Turing.jl/jobs/242001041)
Nice speedup! 

We should probably cleanup this PR, fix the tests and merge it into the master branch now.

xukai92:
Yes. Some recent commits fail on Travis because I disable the rescale of std in adaptation (we divided std by the `min(std)` before). The tests still pass on my local but they fail on Travis because of running too long for NUTS.

The reason I disable the re-scaling is because I observed that it will lead to very small accept rate & numerical error in some circumstances for HMCDA, mentioned here https://github.com/yebai/Turing.jl/issues/284.

Do you think re-scaling by mean may be good for both HMCDA and NUTS and make sense? I tried that on my local which looks fine and I just pushed another commit to see if it works on Travis.

xukai92:
@yebai Tests on Travis look fine now. But tests on Apprvor got the same problem before - freeze before running actual tests. I'm playing around the settings/scripts to see if I can resolve it.

xukai92:
NOTE: I did a profiling on the vectorized version of LDA model and it shows >70% time is spent by `gradient2()`.

xukai92:
@yebai Can you give me permissions on appveyor so that I can terminate or restart tests? I have to wait for each one to be terminated by the sever to try to fix the issue, which is a bit slow-responsed.

yebai:
Done - I've added you to appveyor permissions list.

xukai92:
@yebai Appveyor actually freezes when executing `using Turing`. Any idea why this happens?

yebai:
> @yebai Appveyor actually freezes when executing using Turing. Any idea why this happens?

`Turing` imports other packages when loaded. It's possible some packages being imported have errors/conflicts, most likely due to pre-compilation. To see these warnings or errors, try temporarily disable `@supress_err` function in Turing.

PR #290, opened 2017-06-21
Fix pos trans

xukai92:
- Fix a bug in initialisation
  - `find_good_eps` changed the uniformly initialised variables - thus the variables used for the second iteration is always "quite bad"
  -  that's probably the reason why some of the model is fragile
- Hopefully https://github.com/yebai/Turing.jl/issues/261 is resolved by this PR
- Also reduce the dimension of the vectorisation tests, which is too slow and sometimes make Travis exceeds time limit.

### DIFF
--- a/src/samplers/hmcda.jl
+++ b/src/samplers/hmcda.jl
@@ -54,9 +54,11 @@ function step(model, spl::Sampler{HMCDA}, vi::VarInfo, is_first::Bool)
 
       init_warm_up_params(vi, spl)
 
+      oldθ = vi[spl]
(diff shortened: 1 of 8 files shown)


src/samplers/hmcda.jl
```diff
@@ -54,9 +54,11 @@ function step(model, spl::Sampler{HMCDA}, vi::VarInfo, is_first::Bool)
 
       init_warm_up_params(vi, spl)
 
+      oldθ = vi[spl]
```
yebai: nice catch : )

Issue #238, opened 2017-05-05
Old notebooks


xukai92:
Shall we also move old notebooks (https://github.com/yebai/Turing.jl/tree/master/doc/source/notebooks) to https://github.com/yebai/Turing.jl/tree/master/example-models/notebooks?

yebai:
That makes sense. We can also remove some obsolete models.

xukai92:
Closed by https://github.com/yebai/Turing.jl/pull/321

PR #292, opened 2017-06-22
Notebook used in JuliaCon

xukai92:


xukai92:
So we have BayesNN demo back now, the only notebooks we want to move back  (#238) are

- https://github.com/yebai/Turing.jl/blob/master/doc/source/notebooks/04-Bench-Scalability.jl.ipynb
  - Do we still want this, considering we have a HMM one already and have a lot of benchmark models on testing scability
- https://github.com/yebai/Turing.jl/blob/master/doc/source/notebooks/05-Ornstein-Uhlenbeck.ipynb
- https://github.com/yebai/Turing.jl/blob/master/doc/source/notebooks/06-Demo-CRP.ipynb

yebai:
Some tasks have been moved to #296

PR #295, opened 2017-06-24
Support indexing by Expr (e.g. chn[:(x[1])])

xukai92:


xukai92:
I removed two piece of codes in https://github.com/yebai/Turing.jl/pull/295/commits/e8e67477650ff79fde0d3154c6a1186b94be4725. I think they are not used any more. How do you think?

xukai92:
@yebai Do you think https://github.com/yebai/Turing.jl/commit/e8e67477650ff79fde0d3154c6a1186b94be4725 is fine? If so let's merge this PR.

src/core/util.jl
```diff
@@ -66,3 +66,8 @@ align(x,y) = begin
 
   (x,y)
 end
+
+macro sym_str(var)
```
yebai: Perhaps we can rename this to ‘sym2str’?

src/core/io.jl
```diff
@@ -175,6 +175,8 @@ function Base.getindex(c::Chain, v::Symbol)
   end
 end
 
+Base.getindex(c::Chain, expr::Expr) = c[Symbol(string(expr))]
```
yebai: We need some sanity check here, e.g. verify ‘expr’ is an indexing expression in the form of ‘x[1]…’.

src/core/util.jl
xukai92: This is a trick I just learnt. If we write the macro name as `xxx_str` we can directly call it by `xxx"something"`. In our case, we can create a symbol by `sym"x[1]"`.

src/core/io.jl
xukai92: I see it makes sense!

src/core/util.jl
yebai: Oh, I see.

PR #301, opened 2017-06-30
Remove functions exports from Mamba

xukai92:
Addressing https://github.com/yebai/Turing.jl/issues/155

### DIFF
--- a/benchmarks/gdemo-geweke.run.jl
+++ b/benchmarks/gdemo-geweke.run.jl
@@ -1,5 +1,6 @@
 using Stats, Distributions, Turing
 using Gadfly
+using Mamba: describe
 
 # include("ASCIIPlot.jl")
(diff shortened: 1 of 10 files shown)


xukai92:
@yebai The AppVeyor is back after `using Mamba` is removed!

yebai:
That’s cool. Perhaps some Mamba dependency is broken on Appveyor, but why using a different import mechanism would solve the problem?

xukai92:
I'm not sure. My guess is that Mamba itself exports some functions which causes conflicts to others. Instead of call `using Mamba` but only `import Mamba: Chains, Abstract Chains` is called, other exports from Mamba will not make influence?

xukai92:
BTW it seems that still I cannot restart or do things to Appveyor build?

PR #303, opened 2017-06-30
Add a change-point model

xukai92:
A simple change-point model taken from http://www.mit.edu/~ilkery/papers/GibbsSampling.pdf

change-point


yebai:
@xukai92 Seems testing is broken here, shall we fix it before merging?

xukai92:
@yebai The Travis test failed because of running our of time - which is then because the pre-conditioning sometimes not converges. I restarted the test and it passed now.

PR #310, opened 2017-07-05
Optimization

xukai92:
I add one more benchmark to log the speed of ~ :vs: low-level apis and pre-allocation. I found that Travis doesn't run all of the scripts in the same time but only 4.


xukai92:
Let's keep pushing commits about optimization to this branch

src/samplers/support/distributions.jl
```diff
@@ -169,3 +169,38 @@ Distributions.logccdf{T<:Real}(d::Distributions.Truncated, x::T) =
     x <= d.lower ? zero(T) :
     x >= d.upper ? -T(Inf) :
     log(d.ucdf - cdf(d.untruncated, x)) - d.logtp
+
+
+
+
+
+# No info
+immutable Flat <: ContinuousUnivariateDistribution
```
yebai: what does `Flat` mean here?

xukai92: It's a flat distribution - so that I can use it as a prior to do maximum likelihood; some Stan models don't specify prior distributions

yebai: I see, is it equivalent with `Uniform(-Infty, Infty)`?

src/samplers/support/transform.jl
```diff
@@ -191,8 +198,8 @@ Distributions.logpdf{T}(d::SimplexDistribution, x::Vector{T}, transform::Bool) =
   if transform
     K = length(x)
     z = Vector{T}(K-1)
-    for k in 1:K-1
-      z[k] = x[k] / (one(T) - sum(x[1:k-1]))
+    @simd for k in 1:K-1
```
yebai: I'm not sure whether we should include `@simd` in Turing, since `@simd` is experimental according to Julia documentation https://docs.julialang.org/en/release-0.4/manual/performance-tips/

xukai92: I'm happy to remove them. They don't seems have obvious improvement neither.

yebai:
Since NUTS and HMCDA may get struck in the adaption stage, perhaps we should change all samplers in the test folder to HMC, so that Travis will pass.

src/samplers/support/distributions.jl
xukai92: Yes - I didn't know that we can use `Uniform(-Inf, Inf)` before

xukai92:
We can change all and disable those specifically for NUTS?

Issue #132, opened 2017-04-02
Support explicit return statement


xukai92:
```julia
@model example() = begin
  x = 1
  return x
end
```

i.e. didn't use `@~`

should still be supported.

yebai:
Maybe we can 

- restrict the user to write only `return x::Symbol`, but not `return x+1`.
- construct a special `VarInfo` that also stores variables referenced in the `return` statement.
- later in `Sample(vi::VarInfo)`, only extract variables referenced in the `return` statement.

Issue #264, opened 2017-05-25
Fix warning for re-defining model


yebai:
The current `@model` implementation will generate warning messages overtime when a model function is called (e.g. `gdemo(y)`). This is a bit annoying since we frequently make use of the syntax

```julia
sample (gdemo(y), engine)
```

xukai92:
Is it possible to add cache for `gdemo()` so that it will cache the generated function by `hash(y) => modelf`

yebai:
Yes, it is feasible. I'll give it a try.

yebai:
Fixed by #322

PR #322, opened 2017-07-25
Improve benchmark

yebai:
Kai: this PR was "Support explicit return statement #132" but #132 was decided to postponed to later release due to some difficulties we meet.


yebai:
@xukai92 There is a small issue for supporting explicit return statements. I have documented the issue in #323. It should be OK to merge this PR now as we can probably fix #323 later.

xukai92:
Do you think it might be better to just store them in `spl.info`?

yebai:
> Do you think it might be better to just store them in spl.info?

We want to be able to run the program without any samplers, so it's better to keep model function independent of sampler specific data structures. But as an alternative, we could modify the compiler so that model functions return `(v::VarInfo, s::Union{Void, Sample}` instead of `v::VarInfo` if needed.

xukai92:
I would prefer to return the tuple instead of putting things into `VarInfo`. What I am concerned is that the putting return samples will be carried with `VarInfo`, which may slow unnecessarily down the whole sampling?

Maybe we can also do things differently depending on if `sampler` is `nothing` - do nothing for return statement if sampler is missing and update them in sampler if it exists. In this way we always return vi still.

yebai:
> I would prefer to return the tuple instead of putting things into VarInfo. What I am concerned is that the putting return samples will be carried with VarInfo, which may slow unnecessarily down the whole sampling?

Let's go for a tuple for #323

xukai92:
I haven't make it to update the value yet

yebai:
> I haven't make it to update the value yet

That's OK. I will merge this PR when you are done.

xukai92:
I realise that my implementation actually has a bug where it doesn't handle coroutines correctly.

xukai92:
Regarding the corountine issue, I ended up adding a new filed to VarInfo, for the following reason:
1. We need store "predictions" in corountine level. There are many ways to do, e.g. what we did before is using task.storage. We can also add fields to particles but I choose to bind that to VarInfo for consistence.
2. Adding a new filed to VarInfo makes it easy to drop the predictions after each vi -> sample conversion, so that we don't waste time to copy them in the future.
  - This currently gives me error and need to be fixed. I disable the empty now.

yebai:
>Adding a new filed to VarInfo makes it easy to drop the predictions after each vi -> sample conversion, so that we don't waste time to copy them in the future.

Adding a `pred` field to `VarInfo` sounds good!

xukai92:
The update of `pred` into sample is on L67 in `helper.jl`. I was trying to reset `pred` after each conversion, but it gives me error when wrapping samples into `Chain` (seems for PG only). I put comments from L73 to L79. Can you have a look at it?

xukai92:
Another bug of explicit return:

When HMC dealing with numerical overflow, it only recovers the theta but not the vi.pred used by last numerical OK run of model.

yebai:
Note: I merged the part fixing #264 of this PR into the master branch.

yebai:
@xukai92 Yes, let's do that and merge this PR.

xukai92:
@yebai I've disabled the explicit return. Can you review this PR and accept it if everything is fine?

PR #327, opened 2017-07-30
SGHMC Sampler with tests

emilemathieu:
I did not commit the SGLD sampler because with `ϵ_t = 0.25 / (t+1)` (same step size annealing schedule as implemented in Edward) the performance is quite poor. In the reference paper, they choose a different annealing schedule  for each dataset, and it's not clear to me which one to choose.


emilemathieu:
Eventually, I included the SGLD sampler with `ϵ_t = initial_lr / t^35`. The poor performance was partially due to the fact that samples should be weighted by step size so as to reduce variance and enfance fixing (cf paper).

yebai:
@emilemathieu Thanks for the PR. Could you include an illustrative example using SGLD in the `examples` folder?

emilemathieu:
@yebai : I added an example in the `example-models/sgld-paper` folder

PR #339, opened 2017-08-24
Particle marginal Metropolis-Hastings (PMMH)

emilemathieu:
Related to https://github.com/yebai/Turing.jl/issues/332.

For now the parameters can only be sampled from the prior. I'm currently working on the ability to give a proposal (gaussian for instance) and compute the right acceptance ratio.

### DIFF
--- a/test/pmmh.jl/pmmh.jl
+++ b/test/pmmh.jl/pmmh.jl
@@ -15,11 +15,11 @@ x = [1.5 2.0]
 end
 
 check_numerical(
-  sample(pmmhtest(x), PMMH(100, SMC(30, :m), (:s, (s) -> Normal(s, sqrt(1))))),
+  sample(pmmhtest(x), PMMH(100, SMC(30, :m), (:s, (s) -> Normal(s, sqrt(10))))),
   [:s, :m], [49/24, 7/6]
 )
 
 check_numerical(
-  sample(pmmhtest(x), PMMH(100, SMC(30, :s), :m)),
+  sample(pmmhtest(x), PMMH(100, SMC(30, :m), :s)),
   [:s, :m], [49/24, 7/6]
 )
--- a/test/pmmh.jl/pmmh.jl
+++ b/test/pmmh.jl/pmmh.jl
@@ -16,6 +16,11 @@ x = [1.5 2.0]
 end
 
 check_numerical(
-  sample(pmmhtest(x), PMMH(1500, SMC(30, :s), :m)),
+  sample(pmmhtest(x), PMMH(100, SMC(30, :s), (:m, (m) -> Normal(m, sqrt(1))))),
+  [:s, :m], [49/24, 7/6]
+)
+
+check_numerical(
+  sample(pmmhtest(x), PMMH(100, SMC(30, :s), :m)),
   [:s, :m], [49/24, 7/6]
 )
--- a/test/pmmh.jl/pmmh.jl
+++ b/test/pmmh.jl/pmmh.jl
@@ -1,3 +1,5 @@
+include("../utility.jl")
+
 using Distributions
 using Turing
 using Base.Test
--- a/test/pmmh.jl/pmmh.jl
+++ b/test/pmmh.jl/pmmh.jl
@@ -1,7 +1,6 @@
 using Distributions
 using Turing
 using Base.Test
-
 srand(125)
 
 x = [1.5 2.0]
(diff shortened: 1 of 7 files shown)


emilemathieu:
@yebai : How is your review process defined ?

I've pushed the PMMH versioned allowing proposal for the parameter sampling.
There is one test, maybe I should add more.
I have also added comments in this PR on the parts I'm not completely confident.

@xukai92 do you know whether a space transformation is needed (such as for HMC) in PMMH ? Could you give me a reference on this subject ?

yebai:
> How is your review process defined ?

It's very similar to other Github projects, someone open a PR, then one developer will review the code and request some fixes if necessary. 

> do you know whether a space transformation is needed (such as for HMC) in PMMH ? Could you give me a reference on this subject ?

This depends on how you implement your proposal. If your proposal `q(theta)` is defined on the un-transformed `theta` space, then there is no need for performing transformation. But the caveat is that, some proposals will get rejected because they fall outside of the constraints imposed by the prior. If you proposal `q(vtheta)` is defined on the un-constrained space, then you need to apply the [change-of-variable](https://en.wikipedia.org/wiki/Probability_density_function#Dependent_variables_and_change_of_variables) correction when computing the MH ratio.

emilemathieu:
Thanks @yebai !
I think that's what is explained in https://darrenjw.wordpress.com/2012/06/04/metropolis-hastings-mcmc-when-the-proposal-and-target-have-differing-support/.
I need to fix this, currently there is no check that the proposed value lies in the support of the prior... (which can obviously yield errors)

emilemathieu:
@yebai: I have updated my PR; if the proposal for of the parameters is outside the support, I simply reject (without going through the SMC). I believe that the time spent on the parameters proposal is negligible compared to the SMC proposal so that's not a big issue.

This could be easily improved for Normal proposals ([see this blog post](http://fsaad.scripts.mit.edu/randomseed/metropolis-hastings-sampling-with-gaussian-drift-proposal-on-bounded-support/)), by proposing until the proposed value lies in the support, and correct this in the acceptance ratio.

emilemathieu:
Eventually, I have implemented the rejection sampling for Gaussian drift proposal on bounded support ([described here](http://fsaad.scripts.mit.edu/randomseed/metropolis-hastings-sampling-with-gaussian-drift-proposal-on-bounded-support/)).

Otherwise, I'm still not sure what the sampler should do when in `assume`, `if ~haskey(vi, vn)` is `true`. I've put a comment on this above.

yebai:
@emilemathieu Sorry for the delay, I'll review this PR in the next 2 days.

test/pmmh.jl/pmmh.jl
```diff
@@ -0,0 +1,25 @@
...
+@model pmmhtest(x) = begin
+  s ~ InverseGamma(2,3)
+  m ~ Normal(0,sqrt(s))
+  for i in 1:length(x)
+    x[i] ~ Normal(m, sqrt(s))
+  end
+  s, m
+end
+
+check_numerical(
```
yebai: @emilemathieu Hello. Did you define this function `check_numerical` somewhere in Turing? It seems that I can't find it.

yebai: Never mind - I found it in file `test/utility.jl`.

yebai:
> Eventually, I have implemented the rejection sampling for Gaussian drift proposal on bounded support (described here).

Thanks - that makes a lot of sense. 

> Otherwise, I'm still not sure what the sampler should do when in assume, if `~haskey(vi, vn)` is true. I've put a comment on this above.

This happens when model dimensions change, or during the first run of the program where `vi` is empty. More specifically to PMMH, this can happen 

1. adding global parameters `theta`, e.g. adding a new component in DP mixture model, or 
2. adding local parameters `x`, e.g. component indicators etc. 

For 1), we can initialise the added parameter by sampling from its prior, which I believe is what you implement in your code. When initialising the added parameter, there is no need to adjust MH ratio; otherwise, if the added parameter is proposed from some proposal `q`, you need to correct the MH ratio for PMMH. Then, for 2), I believe SMC will produce unbiased estimator of the marginal (ie, p(theta | y) whenever the importance weight is correctly computed. In both cases, I think sampling from the prior would produce correct answers for PMMH. Does this answer you question?

emilemathieu:
Thanks @yebai ! I'm closing issue https://github.com/yebai/Turing.jl/issues/332.

PR #341, opened 2017-08-28
Fix 173

yebai:


xukai92:
`sample.jl` (i.e. Gibbs with HMC + PG on `gdemo`) works on Julia 0.6 with 2 changes:

1. Make model wrapper call and sample function call separate (ffb9407868667bc33a10694d76a41dce1d9f3f23);
2. Update to ForwardDiff.jl's new signature (2a9466d3da251f65c9f41bd6070a50fb1c876cd5);

`1.` is just a work around for the test. Need to fix it.

xukai92:
I figure out how to resolve the issue by using `Base.invokelatest()` - but we may need to make change to the signature of `model` because `Base.invokelatest()` doesn't support keyword arguments. (Note: out model signature is like `model(vi=vi, sampler=spl)` which is keyword argument based.)

xukai92:
- [ ] Fix warning messages as many as possible

xukai92:
Most of the test are passed now. But Julia 0.6 (or other changes?) may lead to some minor changes to some cases. E.g.

```julia
p[i,j] ~ dist
```

generates a VarName called `p[1, 2]` (with space) now, which was `p[1,2]` (without space). This means we need to changes many of the tests/benchmarks for results fetching.

Any idea why?

yebai:
This might be because we introduced a mechanism for translating `~` operator into a macro.

xukai92:
@yebai I see.

All the tests passed on my local Linux machine now. I changed some deprecated functions but now all. I will do that in the next few days.

yebai:
I pumped Julia version to 0.6 on Travis and Appveyor. The tests run on my Mac as well, but seems quite slow at the moment.

xukai92:
Yes this is a bug in the test. My last commit fixes the test.

This is becauseI didn't actually implement the callback functions (I tried that naively but the compiler doesn't work as I expect) but always use `modelf(vi::VarInfo, sampler::Sampler)`. In empty argument cases before I changed those codes to `modelf(VarInfo(), nothing)`. It's wired my Linux doesn't give the error.

- [ ] implement callback functions

xukai92:
Known unfixed warnings:
- [ ] `WARNING: consume is now deprecated. Use Channels for inter-task communication.`

xukai92:
I also noticed Turing.jl becomes very slow after migrated to 0.6. I was guessing that's because those deprecations inside model function (which will be called mutiple times during model evatluation), e.g. `exp` vs `exp.`

```julia
julia> @time for _ = 1:1000 exp(a) end
  0.496221 seconds (83.00 k allocations: 12.161 MiB, 1.33% gc time)

julia> @time for _ = 1:1000 exp.(a) end
  0.088814 seconds (49.79 k allocations: 3.175 MiB)
```

I've fixed most of them - seems become a bit faster.

xukai92:
@yebai I finally resolve the callback issue.

In terms of the deprecation, I guess the only one unsolved is the `Channel` one. I don't completely understand what's going on with it. Do you think we can simply replace all `Task` with `Channel`?

yebai:
> In terms of the deprecation, I guess the only one unsolved is the Channel one. I don't completely understand what's going on with it. Do you think we can simply replace all Task with Channel?

I'm not sure about this yet; will take a look at the new Task interface.

yebai:
Yes, we can use `Channel` in Julia 0.6, but we don't need to drop `Task`. Basically, the `consume`/`produce` API is replaced by `take!`/`put!`. 

```julia
f(_c::Channel{Any}) = begin
      for i =1:1000
      println(i);put!(_c,i);
      end
end

a = Channel(f)

take!(a)
```

The part related to task copying might be a bit tricky. I will look into this further later.

yebai:
@xukai92  Seems the tests are failing. Could you take a look at Travis?

yebai:
@xukai92 I still haven't figured how to use channels correctly when tasks are copied. In the meantime, I implemented a temporary hack so that we can continue to use `produce` before I have a better solution.

xukai92:
List of un-fixed warnings (taken from https://travis-ci.org/yebai/Turing.jl/jobs/301057755):

- [ ] floor
- [ ] tanh
- [ ] abs
- [ ] logpdf
- [ ] log

src/samplers/sampler.jl
```diff
@@ -83,7 +83,7 @@ assume{T<:Distribution}(spl::Void, dists::Vector{T}, vn::VarName, var::Any, vi::
     end
   end
 
-  acclogp!(vi, sum(logpdf(dist, rs, istrans(vi, vns[1]))))
+  acclogp!(vi, sum(logpdf_with_trans(dist, rs, istrans(vi, vns[1]))))
```
xukai92: @yebai Plz have a look at this change - do you agree what we had before was a bug?

xukai92: @yebai This is actually (mathematically) correct ... as there is another bug - we didn't use customised `init` here for vectorised assume at all. I will push another commit to fix it.

yebai: yes - it looks correct to me.

PR #343, opened 2017-08-29
Support of Stan interface

xukai92:
Targeting https://github.com/yebai/Turing.jl/issues/326

Work in progress
- [x] Initial support of Stan interface
  - support `sample{T<:Function}(mf::T, num_samples::Int, num_warmup::Int, save_warmup::Bool, thin::Int, adapt::Stan.Adapt, alg::Stan.Hmc)`
- [x] Make parameters of adaptation no more constant
- [x] Warn user when setting unsupported settings in Turing
- [x] Use Stan's adaptation settings

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,2 +1,3 @@
 UnicodePlots
 Stats
+Stan
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,3 +1,2 @@
 UnicodePlots
 Stats
-Stan
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
 notifications:
     email: false
 script:
-  - if [ "$(uname)" != "Darwin" ]; then ln -s `which gcc-5` gcc && export PATH=`pwd`:$PATH && echo $PATH; fi
+  - if [ "$(uname)" != "Darwin" ]; then ln -s `which gcc-5` gcc && export PATH=`pwd`:$PATH && echo $PATH; sudo apt-get install hdf5-tools; fi
   - gcc -v
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia --check-bounds=yes -e 'Pkg.update();
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,6 @@ script:
   - gcc -v
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia --check-bounds=yes -e 'Pkg.update();
-             Pkg.add("Stan");
              Pkg.clone(pwd(), "Turing");
              Pkg.build("Turing");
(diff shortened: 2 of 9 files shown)


xukai92:
@yebai I've added the initial interface support of using `Stan.jl`'s types, but currently Turing.jl internally uses most of the adaptation parameters as constants, thus we need to decide a way to pass in those parameters. We can either

1. Add a keyword parameter in `sample()` and store them in `spl.info`, or
2. Store them as global constant and update them when necessary.

I would prefer option one. How do you think?

yebai:
> Add a keyword parameter in sample() and store them in spl.info, or

Let's do this first.

xukai92:
I've made the change we agreed on. 

It works for single HMC but would be tricky for Gibbs - if we want to set different adaptation parameter for different HMC component then passing adaptation as keyword to `sample()` is not a good design (which is not a problem in Stan though because no Gibbs support is necessary). For this reason it might be better to pass Stan.Adapt to algorithms (e.g. `NUTS(2000,1000,0.65; adapt_conf=Stan.Adapt())`).

Another possible design is to have another level of abstraction `Turing.Inference` to integrate `Turing.InferenceAlgorithm` and `Sampler.info[:adapt_conf]` (i.e Stan.Adapt). This is same as the `Stan.Sample` (http://goedman.github.io/Stan.jl/latest/index.html#Stan.Sample) in Stan.

xukai92:
BTW tests are OK on my local but fails on Travis. This might be because we have a new Stan dependency on Turing in this PR. Can you have a look at the error, or try to run tests on your local to see if it's still there?

ChrisRackauckas:
You need to change the travis script to install Stan

xukai92:
@ChrisRackauckas Thx. I'm trying that. BTW why adding Stan to REQUIRE is not enough (e.g. unlike ForwardDiff which doesn't need to be manually installed in Travis)?

ChrisRackauckas:
Because Stan isn't a Julia package. It needs to install the CmdStan stuff and compile it. Stan.jl doesn't do that for the user.

xukai92:
But as I remembered, the interface of Stan.jl (i.e. things before calling CmdSatn) can be used without CmdStan installed. Is that true?

ChrisRackauckas:
That I am not sure about. Sounds like that should be true? But in practice maybe there's still something calling into Stan.

yebai:
@xukai92 Looks good to me! 

@ChrisRackauckas Could you also have a look at the code before I merge this PR, please?

test/REQUIRE
```diff
@@ -1,2 +1,3 @@
 UnicodePlots
 Stats
+Stan
```
ChrisRackauckas: Since this is in the REQUIRE, it doesn't need to be in test/REQUIRE

ChrisRackauckas:
LGTM other than the doubled requirements.

PR #355, opened 2017-10-15
ESS computation: remove useless numerator 

emilemathieu:
See issue https://github.com/yebai/Turing.jl/issues/350

Current implementation of `effectiveSampleSize ` is
```julia
function effectiveSampleSize(pc :: ParticleContainer)
  Ws, _ = weights(pc)
  ess = sum(Ws) ^ 2 / sum(Ws .^ 2)
end
```

Weights are already normalised consequently `sum(Ws) ^ 2 = 1.0` .

Best,
Emile


yebai:
Thanks!

PR #361, opened 2017-10-19
Fix non optional resampling in CSMC

emilemathieu:
Fix issue https://github.com/yebai/Turing.jl/issues/357.

How can I run tests in Travis for this PR ?

When I run `Pkg.test("Turing"; coverage=false)` locally I currently get `ArgumentError: Normal: the condition σ > zero(σ) is not satisfied.` comping from `HMCDA`s tests. I believe this issue comes from `test/hmcda.jl/hmcda_geweke.jl`.

Best,
Emile

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -87,7 +87,7 @@ using Turing.Traces
 
 # Turing essentials - modelling macros and inference algorithms
 export @model, @~, @VarName                   # modelling
-export HMC, SGLD, SGHMC, HMCDA, NUTS, IS, SMC, CSMC, PG, PMMH, Gibbs   # sampling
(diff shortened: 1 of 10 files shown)


yebai:
> How can I run tests in Travis for this PR ?

If you work on the main Turing repo (https://github.com/yebai/Turing.jl/),  then all your PRs will get tested on Travis automatically. I think you are already a collabrator, so you should be able create branches directly on the main repo.

emilemathieu:
>If you work on the main Turing repo (https://github.com/yebai/Turing.jl/), then all your PRs will get tested on Travis automatically. I think you are already a collaborator, so you should be able to create branches directly on the main repo.

All right, I'll do this in the future. Yet, shouldn't Travis also build pull requests ?

Issue #330, opened 2017-08-09
Upgrade to v0.6


ChrisRackauckas:
Julia v0.6 is the current release. It seems there's no tests on v0.6, and when I try `using Turing` on v0.6 it stalls.

yebai:
Yes, we plan to work on that. There is an issue prevent us from using Julia 0.6 at the moment, see

https://github.com/yebai/Turing.jl/issues/173

https://github.com/JuliaLang/METADATA.jl/pull/10646#issuecomment-321925262

yebai:
I made some progresses on this issue. Now Turing can load within Julia 0.6, however, some tests are currently broken (`Pkg.test("Turing")`) -- the issue seems related to a change on the Julia side (https://github.com/JuliaLang/julia/issues/21356).

kleinschmidt:
See also JuliaLang/julia#19774

xukai92:
I found that if we do the model evaluation and calling `sample()`, Julia doesn't throw the error. E.g.

```julia
modelf = gdemo(x)
sample(modelf, alg)
```

xukai92:
From https://docs.julialang.org/en/stable/manual/methods/#Redefining-Methods-1
> But neither you, nor any of your callers, nor the functions they call, or etc. can call this new method definition!

emilemathieu:
Any news on this issue ?
Do we have a list of the issues/tests yet solved ?

xukai92:
@emilemathieu Please see https://github.com/yebai/Turing.jl/pull/341

In short Turing can be used in Julia 0.6 in this PR, and we are adapting the new co-routine and fixing some minor issues.

yebai:
@xukai92 let's try to merge the PR this week - I will make an effort of fixing coroutine issue; Could you fix other known issues?

emilemathieu:
Duplicate: https://github.com/yebai/Turing.jl/issues/293

yebai:
As of now, all tests in #341 are passing on Julia 0.6.

ChrisRackauckas:
For the warnings you can just install Femtocleaner 

https://github.com/JuliaComputing/FemtoCleaner.jl

PR #369, opened 2017-11-12
Metropolis-Hasting + PMMH refactoring

emilemathieu:
Solves issues https://github.com/yebai/Turing.jl/issues/366 and https://github.com/yebai/Turing.jl/issues/332

I refactored `PMMH`'s code to create a separate `MH` sampler, and use it inside the new `PMMH`.
The syntax is now the following
```julia
GaussianKernel(var) = (x) -> Normal(x, sqrt(var))
s1 = MH(N, (:s, GaussianKernel(3.0)), (:m, GaussianKernel(3.0)))
s2 = MH(N, :s, :m)
s3 = MH(N)
s4 = Gibbs(N, MH(5, :m), MH(5, :s))
s5 = PMMH(N, SMC(10, :s), MH(1,(:m, GaussianKernel(3))))
s6 = PMMH(N, SMC(10, :s), MH(1,:m))
s7 = PIMH(N, SMC(10))
```

One can build a Metropolis within Gibbs sampler (`s4`) or a particle independent Metropolis Hasting sampler (`s7`, which then calls `PMMH` constructor).


yebai:
@emilemathieu Thanks - is this ready for code review?

emilemathieu:
@yebai yes

src/samplers/mh.jl
```diff
@@ -0,0 +1,243 @@
...
+  function MH(n_iters::Int, space...)
+    new_space = Set()
+    proposals = Dict{Symbol,Any}()
+
+    # parse random variables with their hypothetical proposal
+    for element in space
+        if isa(element, Symbol)
+          push!(new_space, element)
+        else
+          @assert isa(element[1], Symbol) "[MH] ($element[1]) should be a Symbol. For proposal, use the syntax MH(N, (:m, (x) -> Normal(x, 0.1)))"
```
ae-foster: Could you set up it up to do this `[MH]` automatically? Or at least make it a variable. E.g. if I subclassed MH I might want to change the text that appears in the log messages.

src/samplers/pmmh.jl
```diff
@@ -1,104 +1,101 @@
...
 
   dprintln(2, "computing accept rate α...")
-  α = smc_spl.info[:logevidence][end] - spl.info[:old_likelihood_estimator]
-  if !isempty(spl.alg.proposals)
-    α += spl.info[:new_prior_prob] - spl.info[:old_prior_prob] + spl.info[:proposal_prob]
+  α = new_likelihood_estimate - spl.info[:old_likelihood_estimate]
+  α += new_prior_prob - spl.info[:old_prior_prob] + proposal_ratio
   end
 
   dprintln(2, "decide wether to accept...")
```
ae-foster: whether

src/samplers/pmmh.jl
```diff
@@ -163,63 +160,3 @@ sample(model::Function, alg::PMMH;
...
-      r
-    else
-      vi[vn]
-    end
-end
-
-assume{D<:Distribution}(spl::Sampler{PMMH}, dists::Vector{D}, vn::VarName, var::Any, vi::VarInfo) =
-  error("[Turing] PMMH doesn't support vectorizing assume statement")
-
-observe(spl::Sampler{PMMH}, d::Distribution, value::Any, vi::VarInfo) =
```
ae-foster: Just for my understanding, what happens if you call `observe(spl::Sampler{PMMH}, ...)` now? We don't have `PMMH <: MH` so how would it know which method to call?

src/samplers/mh.jl
```diff
@@ -0,0 +1,243 @@
...
+          stdG = Normal()
+          r = rand_truncated(proposal, lb, ub)
+          # cf http://fsaad.scripts.mit.edu/randomseed/metropolis-hastings-sampling-with-gaussian-drift-proposal-on-bounded-support/
+          spl.info[:proposal_ratio] += log(cdf(stdG, (ub-old_val)/σ) - cdf(stdG,(lb-old_val)/σ))
+          spl.info[:proposal_ratio] -= log(cdf(stdG, (ub-r)/σ) - cdf(stdG,(lb-r)/σ))
+
+        else # Other than Gaussian proposal
+          r = rand(proposal)
+          if (r < support(dist).lb) | (r > support(dist).ub) # check if value lies in support
+            spl.info[:violating_support] = true
```
ae-foster: So you're rejecting out-of-bounds proposals. But don't you need to deal with the acceptance ratio as we did for the Gaussian? (I.e. why is the blog post only relevant to Gaussian proposals)

src/samplers/pmmh.jl
emilemathieu: `observe(spl::Sampler{PMMH}, ...)` would be called inside`model(vi=vi, sampler=spl::Sampler{PMMH})`, which is never done. `model` is only ran with `MH` and `SMC` not `PMMH` itself.

src/samplers/mh.jl
emilemathieu: The blog post is relevant for the Gaussian case since we can analytically account for the truncated proposal (which depends on the current value). Maybe this can also be done for other distributions.

src/samplers/mh.jl
emilemathieu: good idea

src/samplers/mh.jl
ae-foster: Yh wouldn't your acceptance ratio be invalid otherwise?

src/samplers/pmmh.jl
ae-foster: ok cheers

src/samplers/mh.jl
emilemathieu: The ratio would be valid in the generic proposal case with a Gaussian proposal. But you could potentially be wasting proposals. Let's say we have 10 random variables, if the 10th proposal does not lie on the associated prior's support, we would have to throw all these proposal. With the Gaussian, we can avoid that by sampling from a truncated Gaussian, and correct the acceptance ratio to take into account that this truncation depends on our current value.

ae-foster:
@emilemathieu the MH looks good to me. Would wait for a :+1: form @yebai before merge though

yebai:
@emilemathieu Kai fixed some warnings on the master branch - presumably these fixes introduce conflict with this PR. Could you manually merge the master branch into this PR's branch?

emilemathieu:
I have weird errors now:
>LoadError: MethodError: no method matching ##mhtest_model#687()
The applicable method may be too new: running in world age 23509, while current world is 23513.

This doesn't happen for every test on my machine. The error comes from when `model` is called within `sample` for `MH`.

@yebai Would you have any idea of where that could come from ?

yebai:
> @yebai Would you have any idea of where that could come from ?

You need to merge the master branch in order to use Julia 0.6, see https://github.com/yebai/Turing.jl/issues/330#issuecomment-332391128

emilemathieu:
> You need to merge the master branch in order to use Julia 0.6, see #330 (comment)

Yes of course I've done that.

emilemathieu:
@yebai Doing `model = mhtest(x)` fixes it.

yebai:
Cool.

xukai92:
It's weird that you still need to do `model = mhtest(x)` to avoid the closure error (#330) which is already solved in #341. In which commit did you merge the master which contains #341?

emilemathieu:
@xukai92 yes it's weird, this seems to be specific to my machine, I'll figure that out on Monday

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
+doc"""
+    HMC(n_iters::Int)
```
yebai: This needs to be `MH`

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
...
+      vi[spl] = old_θ         # reset Θ
+      setlogp!(vi, old_logp)  # reset logp
+    end
+
+    vi
+  end
+end
+
+function sample(model::Function, alg::MH;
+                chunk_size=CHUNKSIZE,     # set temporary chunk size
```
yebai: perhaps we should remove `chunk_size` here since it's only required by HMC samplers.

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
...
+                save_state=false,         # flag for state saving
+                resume_from=nothing,      # chain to continue
+                reuse_spl_n=0,            # flag for spl re-using
+                )
+
+  default_chunk_size = CHUNKSIZE  # record global chunk size
+  setchunksize(chunk_size)        # set temp chunk size
```
yebai: remove these lines

```julia
default_chunk_size = CHUNKSIZE  # record global chunk size
setchunksize(chunk_size)        # set temp chunk size
```

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
...
+
+    if PROGRESS ProgressMeter.next!(spl.info[:progress]) end
+  end
+
+  println("[$alg_str] Finished with")
+  println("  Running time        = $time_total;")
+  accept_rate = sum(spl.info[:accept_his]) / n  # calculate the accept rate
+  println("  Accept rate         = $accept_rate;")
+
+  setchunksize(default_chunk_size)      # revert global chunk size
```
yebai: pls remove this line

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
...
+  end
+  c = Chain(0, samples)       # wrap the result by Chain
+  if save_state               # save state
+    save!(c, spl, model, vi)
+  end
+
+  c
+end
+
+function rand_truncated(dist, lowerbound, upperbound)
```
yebai: Does this work for multivariable Gaussian as well?

src/samplers/mh.jl
```diff
@@ -0,0 +1,249 @@
...
+
+assume(spl::Sampler{MH}, dist::Distribution, vn::VarName, vi::VarInfo) = begin
+    if isempty(spl.alg.space) || vn.sym in spl.alg.space
+      vi.index += 1
+      old_val = getval(vi, vn)[1]
+
+      if ~haskey(vi, vn)
+        r = rand(dist)
+        push!(vi, vn, r, dist, spl.alg.gid)
+        spl.info[:proposal_ratio] += (logpdf(dist, old_val) - logpdf(dist, r))
```
yebai: Might be a silly question: if `haskey(vi,vn)` equails to `false`, wouldn't `logpdf(dist, old_val)` throw an error since `old_val` does not yet exist?

src/samplers/mh.jl
emilemathieu: right

src/samplers/mh.jl
emilemathieu: My rejection-free gaussian proposal implementation works only for univariate normal distributions (`Distributions.Normal{Float64}`). I used the derivations of this [blog post](http://fsaad.scripts.mit.edu/randomseed/metropolis-hastings-sampling-with-gaussian-drift-proposal-on-bounded-support/). I'm pretty sure that similar formulas can be derived for the multivariate case.

yebai:
do we need to reset `spl.info[:proposal_ratio]` after each `step` call? Imagine that if a user runs `MH` for multiple iterations within a Gibbs sampler, then the `step` function will not reset `spl.info[:proposal_ratio]` in the current code.

src/samplers/mh.jl
emilemathieu: I think you are right ! Not sure if we can handle this case. `MH` may be restricted to a deterministic number of random variables (similar to `HMC`).
@twgr would know whether Metropolis-Hasting can handle stochastic existence of random variables ? i.e. should the set of random variables that is updated be fixed (as for `HMC` ?).

twgr: MH can handle stochastic existence of variables if you're careful - it is more difficult than usually to ensure your proposal leads to an ergodic Markov chain though as you need to make sure you can get to everywhere from everywhere.  The easiest way to do this is to have your proposal be a mixture of sampling from the prior and something else. There is also all the LMH stuff that is a particular case of MH with random existence.  As an example of the stuff that can go wrong here, there were a lot of things wrong with the original Wingate paper, for which I give an explanation in my thesis.

src/samplers/mh.jl
```diff
@@ -0,0 +1,243 @@
...
+        x = rand(dist)
+        notvalid = ((x < lowerbound) | (x > upperbound))
+    end
+    return x
+end
+
+assume(spl::Sampler{MH}, dist::Distribution, vn::VarName, vi::VarInfo) = begin
+    if isempty(spl.alg.space) || vn.sym in spl.alg.space
+      vi.index += 1
+      old_val = getval(vi, vn)[1]
```
emilemathieu: `getval(vi, vn)` is returning an array of size 1 for a univariate random variable, therefore I have to do `[1]` to get the real value.
But that's ugly and I'm pretty sure that would fail for the multivariate case.
@yebai would you know what should I do ?

yebai: @xukai92 is there any reasons why `getval(vi, vn)` has to return an array?

src/samplers/mh.jl
yebai: I would vote for the option of using a mixture of prior and Gaussian (or other proposals) since this would only require a small modification here. For example, we could sample `old_val` from the prior if it does not exist yet; then for following steps, we could flip a coin to decide to propose from the prior or user-specified proposals.

src/samplers/mh.jl
yebai: It might be worth to extend this to multivariate Gaussian, or at least throw an error message saying not supported yet when `dist` is a multivariate Gaussian.

src/samplers/mh.jl
emilemathieu: `getval(vi, vn)` calls `view(vi.vals[end], getrange(vi, vn))` which returns `1-element SubArray`.
You can try `view([1,2], 1:1)` for instance.

emilemathieu: Should I use `vi[vn]` instead ?

src/samplers/mh.jl
emilemathieu: I'll look how to extend this for the multivariate case, but not sure whether this is obvious or not.

Yet, that does not mean that proposing from a multivariate gaussian is not supported, but that if a proposed value (from the MVN) does not lie on the rv's support, the MH step will be rejected, i.e. this is not "proposal rejection-free".

src/samplers/mh.jl
emilemathieu: therefore something like `old_val = haskey(vi, vn) ? vi[vn] : rand(dist)`

src/samplers/mh.jl
yebai: yes, I think calling `vi[vn]` is the prefered way in Turing. `getval` is supposed to be for internal use for `VarInfo` related functions only.

src/samplers/mh.jl
yebai: perhaps open a new issue for the multivariate Gaussian case and solve that in a separate PR?

src/samplers/mh.jl
yebai: that looks good to me.

src/samplers/mh.jl
emilemathieu: good idea

emilemathieu: good idea: https://github.com/yebai/Turing.jl/issues/387

src/samplers/mh.jl
twgr: You might want to include some part of the proposal that allows single elements to be added or dropped -- though it provides ergodicity, relying on a prior component in the proposal to change the dimension can lead to very slow mixing for some models.

src/samplers/mh.jl
xukai92: Is there any reason we have to write this customized function instead of using Distributions.jl's truncated Normal?

src/Turing.jl
```diff
@@ -87,7 +87,8 @@ using Turing.Traces
 
 # Turing essentials - modelling macros and inference algorithms
 export @model, @~, @VarName                   # modelling
-export HMC, SGLD, SGHMC, HMCDA, NUTS, IS, SMC, CSMC, PG, PMMH, IPMCMC, Gibbs # sampling
+export MH, HMC, SGLD, SGHMC, HMCDA, NUTS      # sampling
+export IS, SMC, CSMC, PG, PIMH, PMMH, IPMCMC, Gibbs # sampling
```
xukai92: Can we make the comments more specific in stead of just say `sampling`? And probably make them aligned.

src/samplers/mh.jl
```diff
@@ -0,0 +1,243 @@
...
+                reuse_spl_n=0,            # flag for spl re-using
+                )
+
+  spl = reuse_spl_n > 0 ?
+        resume_from.info[:spl] :
+        Sampler(alg)
+  alg_str = "MH"
+
+  # Initialization
+  time_total = zero(Float64)
```
xukai92: The main purpose of using `zero(T)` is to avoid type conversion and promotion (which cost time). It seems that `zero()` makes no difference from `0.0` here () BUT it's OK to keep it.

src/samplers/mh.jl
```diff
@@ -0,0 +1,243 @@
...
+
+  c
+end
+
+function rand_truncated(dist, lowerbound, upperbound)
+    notvalid = true
+    x = 0.0
+    while (notvalid)
+        x = rand(dist)
+        notvalid = ((x < lowerbound) | (x > upperbound))
```
xukai92: `|` is the bitwise operator. Maybe better to use the boolean or `||`.

emilemathieu:
@xukai92 I have taken into account your feedbacks.
@yebai for now I'm throwing an error in the case of stochastic existence of a random variable. I'll see with Tom later how to rightly compute the acceptance ratio in that case (seems non trivial).

emilemathieu:
This is weird, tests are passing on this branch but failing once merged into master.
Specifically `test/pmmh.jl/...` tests are failing.`PIMH` is passing, it must therefore be related to how `PMMH` uses `MH`).

@xukai92 would you have any idea of how this could be possible ?

yebai:
> I have an idea ! I'm looking into it.

Thanks - let's create a new PR when it's fixed.

PR #384, opened 2017-11-17
Bugfix for Particle Gibbs

emilemathieu:
Proposition to solve https://github.com/yebai/Turing.jl/issues/383 and https://github.com/yebai/Turing.jl/issues/373.

### DIFF
--- a/src/core/container.jl
+++ b/src/core/container.jl
@@ -101,6 +101,8 @@ function Base.consume(pc :: ParticleContainer)
     score = consume(p)
     score = isa(score, ForwardDiff.Dual) ? realpart(score) : score
     if isa(score, Real)
+      score += getlogp(p.vi)
+      resetlogp!(p.vi)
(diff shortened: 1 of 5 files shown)


emilemathieu:
I still have to fix something

yebai:
> I still have to fix something

Right - let me know when you're done.

emilemathieu:
>Right - let me know when you're done.

Good, but it'd be nice to have another test targeting what we fixed.

Otherwise what do you think of using `gid=-1` in `push!(vi, vn, r, dist, -1)`(at https://github.com/yebai/Turing.jl/blob/bugfix-383-pg/src/samplers/pgibbs.jl#L167) since we don't know what is the sampler associated with the random variables encountered. As proposed here https://github.com/yebai/Turing.jl/issues/373#issuecomment-344213583, we could get the right `gid` by looking at `sub_alg`'s spaces. What do you think @yebai ?

PR #395, opened 2017-11-27
Random variables which can be sampled in a random order

emilemathieu:
There is a new array `orders :: Vector{Int}` in `VarInfo`, which for each random variable (indexed bu integers via `idcs`) maps the the index of the previous `observe` statement (i.e. `vi.num_produce`).
This array is filled each time a new random variable is encountered (thus just after `push!(vi, vn, r, dist, spl.alg.gid)` is called, we call `addorder!(vi, vn, vi.num_produce)`).

This array is then called in `getretain` so that all/only random variables (belonging to the current sampler) that have been sampled after the current value of `vi.num_produce` are returned.

Furthermore, in non reference particles, when a random variable which needs to be sampled (having a `NaN` value) is encoutared, the `orders` array is updated by calling `setorder!(vi, vn, vi.num_produce)`.

### DIFF
--- a/src/core/varinfo.jl
+++ b/src/core/varinfo.jl
@@ -42,6 +42,7 @@ type VarInfo
   index       ::    Int           # index of current randomness
   num_produce ::    Int           # num of produce calls from trace, each produce corresponds to an observe.
   indices     ::    Dict{Int,Vector{VarName}}     # indices of current randomness
+  orders      ::    Vector{Int}     #
   VarInfo() = begin
     vals = Vector{Vector{Real}}(); push!(vals, Vector{Real}())
     trans = Vector{Vector{Real}}(); push!(trans, Vector{Real}())
@@ -59,7 +60,8 @@ type VarInfo
       pred,
       0,
       0,
-      Dict{Int,Vector{VarName}}()
+      Dict{Int,Vector{VarName}}(),
+      Vector{Int}()
     )
   end
 end
@@ -208,13 +210,15 @@ push!(vi::VarInfo, vn::VarName, r::Any, dist::Distributions.Distribution, gid::I
   vi
 end
 
-addindex!(vi::VarInfo, vn::VarName, index::Int) = begin
-  if haskey(vi.indices, index)
-    push!(vi.indices[index], vn)
-  else
-    vi.indices[index] = [vn]
-  end
+addindex!(vi::VarInfo, vn::VarName, num_produce::Int) = begin
+  push!(vi.orders, num_produce)
+  vi
+end
 
+setindex!(vi::VarInfo, vn::VarName, index::Int) = begin
+  if vi.orders[vi.idcs[vn]] != index
+    vi.orders[vi.idcs[vn]] = index
+  end
   vi
 end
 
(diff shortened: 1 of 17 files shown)


emilemathieu:
The test `test/trace.jl/trace.jl` related to `TraceR` fails, but I don't get why this test is naming random variables with a global variable `n`, therefore when the particle is replayed in `forkr` `CSMC` sees new random variables. Can someone help me on that please ?

yebai:
> The test test/trace.jl/trace.jl related to TraceR fails, but I don't get why this test is naming random variables with a global variable n, therefore when the particle is replayed in forkr CSMC sees new random variables. Can someone help me on that please ?

@xukai92 can you help Emile on this?

xukai92:
@yebai I will do the review tonight. But I believe this test file was originally written by you (?) So I'm actually not that sure why it was done in that way...

emilemathieu:
@xukai92 @yebai I've deleted that part of the test since `TraceR` is not used for that case anymore. Therefore the test is passing !

src/core/varinfo.jl
```diff
@@ -39,8 +39,8 @@ type VarInfo
   trans       ::    Vector{Vector{Bool}}
   logp        ::    Vector{Real}
   pred        ::    Dict{Symbol,Any}
-  index       ::    Int           # index of current randomness
   num_produce ::    Int           # num of produce calls from trace, each produce corresponds to an observe.
+  orders      ::    Vector{Int}   # observe statements number associated with random variables
```
ae-foster: `observe_orders`?

src/core/varinfo.jl
```diff
@@ -206,6 +206,18 @@ push!(vi::VarInfo, vn::VarName, r::Any, dist::Distributions.Distribution, gid::I
   vi
 end
 
+addorder!(vi::VarInfo, vn::VarName, num_produce::Int) = begin
+  push!(vi.orders, num_produce)
+  vi
+end
+
+setorder!(vi::VarInfo, vn::VarName, index::Int) = begin
+  if vi.orders[vi.idcs[vn]] != index
```
ae-foster: Why do you need to check this condition? Couldn't you just overwrite it

src/samplers/pgibbs.jl
```diff
@@ -52,12 +52,12 @@ step(model::Function, spl::Sampler{PG}, vi::VarInfo, _::Bool) = step(model, spl,
 step(model::Function, spl::Sampler{PG}, vi::VarInfo) = begin
   particles = ParticleContainer{TraceR}(model)
 
-  vi.index = 0; vi.num_produce = 0;  # We need this line cause fork2 deepcopy `vi`.
+  vi.num_produce = 0;  # We need this line cause fork deepcopy `vi`.
```
ae-foster: Could you expand on the comment "We need this line cause fork deepcopy `vi`."

src/core/varinfo.jl
emilemathieu: yes, do you think it's more efficient (in Julia) ?

emilemathieu:
@ae-foster indeed it is lacking a regression test, we could check whether some samples are `NaN` I guess.
it's also lacking basic unit tests.

src/core/varinfo.jl
ae-foster: I was more thinking for clarity. I'm sure any performance differences are negligible.

ae-foster:
Guess we have to strike a balance between writing too many tests and not enough. I would say, for a fiddly thing like this, one test would probably be a good idea. There are always the chance that later code introduces off-by-one errors (e.g. in some places you have `num_produce`, in others `num_produce+1`). A test to warn for that would be useful.

src/samplers/pgibbs.jl
emilemathieu: I'm not sure about that comment, but we need to make sure that `vi.num_produce = 0` before the `CSMC` sweep.

src/samplers/pgibbs.jl
```diff
@@ -143,19 +143,19 @@ end
 assume{T<:Union{PG,SMC}}(spl::Sampler{T}, dist::Distribution, vn::VarName, _::VarInfo) = begin
   vi = current_trace().vi
   if isempty(spl.alg.space) || vn.sym in spl.alg.space
-    vi.index += 1
     if ~haskey(vi, vn)
       r = rand(dist)
       push!(vi, vn, r, dist, spl.alg.gid)
+      addorder!(vi, vn, vi.num_produce)
```
xukai92: I think it's better to actually put the `addorder!()` inside `push!()`. You call `addorder!()` after each `push!()` anyway.

xukai92:
I think before we addressing the `Chain` problem. Can we have another integrated test on random order support for `VarInfo`?

emilemathieu:
@xukai92 @ae-foster I've pushed a regression test in `test/varinfo.jl/orders.jl`. I've mimicked something very similar to the infinite mixture model issue, and it's doing it correctly !
Can we merge this PR ?

PR #405, opened 2017-12-06
SMC needs to resample at the end

ae-foster:
We need to add a resample statement to the SMC algorithm so that resampling is done at the end of the program.

Without this, any likelihood values from the last `observe` statement are discarded. In extreme cases with one observe statement in the program, this means we are just sampling from the prior 😮 

I also added a test for SMC, like the one we had already for IPMCMC

### DIFF
--- a/src/samplers/smc.jl
+++ b/src/samplers/smc.jl
@@ -84,6 +84,7 @@ function sample(model::Function, alg::SMC)
       resample!(particles,spl.alg.resampler,use_replay=spl.alg.use_replay)
     end
   end
+  resample!(particles,spl.alg.resampler,use_replay=spl.alg.use_replay)
   w, samples = getsample(particles)
   res = Chain(w, samples)
 
--- a/test/smc.jl/smc.jl
+++ b/test/smc.jl/smc.jl
@@ -1,0 +1,19 @@
+using Turing
+using Base.Test
+srand(125)
+
+x = [1.5 2.0]
+
+@model smctest(x) = begin
+  s ~ InverseGamma(2,3)
+  m ~ Normal(0,sqrt(s))
+  for i in 1:length(x)
+    x[i] ~ Normal(m, sqrt(s))
+  end
+  s, m
+end
+
+check_numerical(
+  sample(pmmhtest(x), SMC(100)),
+  [:s, :m], [49/24, 7/6]
+)


ae-foster:
Any hints why the tests failed? The error message was not very enlightening

emilemathieu:
@ae-foster a `produce(Val(:done))` is added at the end of the model in `SMC`, which mean the one before the last `consume` will be the last `observe` statement and particles will be resample after that. Therefore I don't think that there is an issue, or maybe I didn't properly get what you said.

ae-foster:
Hmmm you gotta convince me because on my example it definitely seemed to be giving the wrong answer until I added this change.

If what you said is true, there should be no change in results by adding that extra resample, right? I definitely observed a (substantial) change. I'll share the example but it's on another branch.

ae-foster:
I'll come back to you tomorrow with a proper regression test (one that will fail with the old code)

ae-foster:
But just to say there's no disagreement between what you said: `produce(Val(:done))` is added at the end, and what I said (we need a new resample statement). Indeed, the block
```
while consume(particles) != Val{:done}
    resample!(...)
end
```
doesn't run at the end of the program: you receive `Val{:done}` and then skip to `end` without evaluating the block one last time.

ae-foster:
@emilemathieu please disregard my last comment. It's totally incorrect 😆 . What's really happening is that the check `if ess <= spl.alg.resampler_threshold * length(particles)` means that we sometimes never resample. I've got a regression test that I'll push for you now.

ae-foster:
Conflicts probably best to fix via rebase. Lmk when you're happy to merge and I'll fix it

ae-foster:
Oh @emilemathieu the CI doesn't actually seem to be running my new test. Do you know how to include it?

xukai92:
You need to include your test file in `test/runtests.jl`

emilemathieu:
I'm not sure to get why you want to enforce resapling at the end. By doing so you loose a bit of information (not so much if you have quite a lot of particles)

emilemathieu:
@xukai92 are samples weights retuned in the Chain wrapper ? If not why ?

emilemathieu:
If the weights are not included in the returned Chain, indeed resapling is required, so I agree with u Adam ;)

ae-foster:
Thanks for clarifying Emile. I'll look into it a bit more.

emilemathieu:
Since it's not obvious how to get the weights we should by default resample at the end, but add an option to not resample, what do you think ?

xukai92:
Well ... I agree this is confusing (and kind of related to https://github.com/yebai/Turing.jl/issues/277).

Actually before we use `Mamba.Chain`, we do store each sample with the corresponding weight using our own `Chain` type. But I believe when we adapt to `Mamba`'s type we didn't do everything correctly.

Let's create an issue for this and I will be responsible fix the bug of returned `Chain`'s weight fields.

ae-foster:
Thanks for taking that on @xukai92 - it would be to have a cleaner interface for sure!

Issue #391, opened 2017-11-23
[Core.VarInfo] More robust mechanism for deleting entries from `VarInfo`


yebai:
At the moment, we use the following code to mark `VarInfo` entries for deletion

```julia
  vi[getretain(vi, 0, spl)] = NULL # NULL is defined as NaN.
```

One possible caveat of this approach is that it could confuse downstream code between deleted entries or actual `NaN` values. Perhaps we could change this to missing value type (e.g. use the [`Missings.jl`](https://github.com/JuliaData/Missings.jl) package) - like what I suggested in #386. 

@xukai92  Are you aware of any caveats if we replace `NaN` with [`Missings.jl`](https://github.com/JuliaData/Missings.jl)?

Ref: https://github.com/yebai/Turing.jl/issues/367#issuecomment-346671887

ChrisRackauckas:
NaN is different than a missing. NaN is of type Float64, while missing is its own type. That's type unstable, but there's special handling of small unions on v0.7/v1.0 to make that still fast. So the caveat is that on v0.6 this would be very slow.

emilemathieu:
If we do what I proposed in https://github.com/yebai/Turing.jl/issues/367#issuecomment-344032331, there is no need to delete entries in VarInfo. I don't know how this could impact performance.

yebai:
@xukai92 Are you aware of any reasons why we are setting deleted entries in `VarInfo` to `NaN` instead of just deleting them? See Emile's proposal in https://github.com/yebai/Turing.jl/issues/367#issuecomment-344032331.

yebai:
@emilemathieu perhaps you could copy and paste your code from https://github.com/yebai/Turing.jl/issues/367#issuecomment-344032331 into a PR and see whether the tests can pass?

emilemathieu:
@yebai good idea. And how can we assess performances ?

yebai:
There are some benchmarking scripts on Travis, you could use those numbers as a rough estimate of performance. But Travis is not a reliable platform for performing benchmarking because it is based on shared hardware.

emilemathieu:
@yebai I've realised that my solution only works for Iterated CSMC, therefore not for CSMC used within Gibbs (i.e. particle Gibbs).

yebai:
@xukai92 @emilemathieu  Is this issue solved by PR #395?

emilemathieu:
@yebai No, `NaN` is still used.
[...]

Issue #389, opened 2017-11-20
[Core.VarInfo] Allow adding right-hand-side information to `VarName`


yebai:
The current `VarName` in `VarInfo` is only dependent on left-hand-side information, we might want to generalise this to include RHS information to distinguish variables with the same symbol but live in a different scope. A example design could be 

```julia
x ~ Normal(0,1; lhs_vn = unique_id)
```

xukai92:
I think a solution is to introduce another filed to `VarInfo` named `mark`, and introduce different codes to mark variables. This also solves https://github.com/yebai/Turing.jl/issues/391 as now we can mark then can deleted variables.

xukai92:
Closed by #407

PR #407, opened 2017-12-10
Add flags to VarInfo

xukai92:
Resolving #389 and #391 by adding another filed to `VarInfo` called `flags`

TODOs
- [x] get rid of the use of `NULL`
- [x] turn `flags` to `VarName`-wise
  - Sparse by dictionary or not?
- [x] genearlize the signature of flags by ~~macro~~ better function signature


xukai92:
@yebai Do we still want the feature of VarInfo which supports storing multiple chains in the same time, which we never use, i.e. make vi.spl a vector of vector.

xukai92:
@emilemathieu I introduced a function `set_retained_vns_del_by_spl!` in `varinfo.jl` to replace our previous `getretain` and set to `NULL` statement (the function is almost the same as `getretain`). Because of it we actually don't need `getretain`. Do think we can just remove it?

xukai92:
Another design issues I was thinking of is: should we make `flags` sparse or dense

i) sparse: if we assume only a small part of variables will be set a flag, a set/hash map to represent the flag is more memory-efficient
ii) dense: if most of the variables will be set flags, an array based implement is good

src/core/varinfo.jl
```diff
@@ -310,8 +318,23 @@ end
 # Rand & replaying method for VarInfo #
 #######################################
 
-# Check if a vn is set to NULL
-isnan(vi::VarInfo, vn::VarName) = any(isnan.(getval(vi, vn)))
+check_flag(vi::VarInfo, vn::VarName, flag::String) = vi.flags[flag][getidx(vi, vn)]
```
emilemathieu: `get_flag` instead ?

src/core/varinfo.jl
```diff
@@ -310,8 +318,23 @@ end
...
+
+set_retained_vns_del_by_spl!(vi::VarInfo, spl::Sampler) = begin
+  gidcs = getidcs(vi, spl)
+  if vi.num_produce == 0
+    for i = length(gidcs):-1:1
+      vi.flags["del"][gidcs[i]] = true
+    end
+  else
+    retained = [idx for idx in 1:length(vi.orders) if idx in gidcs && vi.orders[idx] > vi.num_produce]
+    for i = retained
```
emilemathieu: `i = 1:retained` ?

src/samplers/pgibbs.jl
```diff
@@ -56,8 +56,8 @@ step(model::Function, spl::Sampler{PG}, vi::VarInfo) = begin
   ref_particle = isempty(vi) ?
                  nothing :
                  forkr(Trace(model, spl, vi))
-
-  vi[getretain(vi, spl)] = NULL
+
```
emilemathieu: empty space

emilemathieu:
Maybe a test could be useful (or modifying existing test)s to check that `vi.flags` is built and updated as planned. Otherwise, this PR seems right !

src/core/varinfo.jl
xukai92: Or maybe `is_flagged`? I was trying to find a name which fits the use case of this function. As it always return a `Bool` type, we always use it like 
```julia
if is_flagged(vi, vn, "missing")
  ...
end
```

src/core/varinfo.jl
xukai92: `retained` here is a vector of indices, and `i = retained` is equivalent to `i in retained`. But probably `i in retained` is more readable.

xukai92:
Good suggestion. I will add some tests for `vi.flags`!

xukai92:
Left to discuss
- [ ] merge trans or not
- [ ] sparse or dense

yebai:
@xukai92 Perhaps we can solve the following 3 tasks and merge this PR first?

- [ ] merge trans ~~or not~~
- [x] ~~sparse or~~ dense
- [ ] remove nested Array

We can turn ·VarInfo` into a submodule in a seprate PR.

xukai92:
Sure. will do it before Wednesday

xukai92:
@yebai Can you review the PR plz? I tried to modualize VarInfo but meet some issues - will probably do that in another PR.

xukai92:
Oops. Seems my modualization works now. It's a bit ugly as I literature export all functions required by Turing.jl from VarRepaly - but I guess we can solve the namespace issue later (?)

src/samplers/pgibbs.jl
```diff
@@ -56,8 +56,8 @@ step(model::Function, spl::Sampler{PG}, vi::VarInfo) = begin
   ref_particle = isempty(vi) ?
                  nothing :
                  forkr(Trace(model, spl, vi))
-
-  vi[getretain(vi, spl)] = NULL
+  
+  set_retained_vns_del_by_spl!(vi, spl)
```
yebai: I think the previous API is more readable, I would prefer something like 

```
vi.flags[spl] = DELETE #similiarly vi.flags[spl] = TRANSFORMED
```

PR #429, opened 2018-03-30
VarName fix

willtebbutt:
Added an extra constructor to `VarName` to `gensym` the `csym` field if it's not provided.

### DIFF
--- a/src/core/varinfo.jl
+++ b/src/core/varinfo.jl
@@ -240,8 +240,12 @@ VarName(vi::VarInfo, csym::Symbol, sym::Symbol, indexing::String) = begin
 end
 VarName(vi::VarInfo, syms::Vector{Symbol}, indexing::String) = begin
   # TODO: update this method when implementing the sanity check
-    VarName(syms[1], syms[2], indexing, 1)
+  VarName(syms[1], syms[2], indexing, 1)
 end
+
+# Automatically generate the csym field if it's not provided.
+VarName(vi::VarInfo, sym::Symbol, indexing::String) = VarName(gensym(), sym, indexing)
+
 
 #################################
 # Utility functions for VarInfo #


yebai:
@xukai92 could you take a look and approve this PR?

xukai92:
Hi Will, what's the use case of this constructor? If we want to use it in `compiler.jl`, it seems to be that the constructor will call `gensym()` in run-time instead of compilation time.

willtebbutt:
This PR makes this example in the wiki work: https://github.com/yebai/Turing.jl/wiki/Advanced-usages#avoid-using-model-macro

The question is whether or not this is the correct fix; I don't know whether adding this makes it easier to introduce bugs in which inference silently fails for some reason. The alternative is to just add a couple of calls to `gensym` to that example specifically. What are your thoughts?

xukai92:
I think it's better to do something like below, in the wiki example

```julia
...
s = Turing.assume(sampler,
                  InverseGamma(2, 3),
                  Turing.VarName(vi, [:any_identifier_not_change_in_runtime, :s], ""),
                  vi)
...
```

The main reason we use `gensym()` originally was to take care of internally naming by considering variables may be overwritten within the `@model`. As there we actually take care of naming manually, we can just use anything we want.

They reason use `gensym()` can lead to problem is that the internal naming is changed each time the function got called.

willtebbutt:
Sounds good to me. I'll close this and edit the wiki directly.

xukai92:
Cool. Please let me know if you spot any out-of-date stuff in the wiki and I'm happy to help.

Also, since you have permission to this repo, it may be better to push your development branch to this repo directly instead of your fork, to make use of the integrated CI set up in this repo. (Or there is a way which we can allow PRs to trigger the CI?)

Issue #422, opened 2018-02-10
Turing on Windows requires a compiler


ChrisRackauckas:
```
=========================================[ ERROR: Turing ]==========================================

LoadError: could not spawn `mingw32-make`: no such file or directory (ENOENT)
while loading C:\Users\Chris\.julia\v0.6\Turing\deps\build.jl, inexpression starting on line 2

====================================================================================================
```

Can this instead be done with BinaryProvider.jl and BinaryBuilder.jl?

xukai92:
@yebai Can you have a look into this issue please?

yebai:
@ChrisRackauckas I believe this can be easily solved by packages such as `BinaryProvider.jl` and `BinaryBuilder.jl. However, I don't have direct experience with these packages.  Would someone with more experience create a pull request?

cc'ing @antoine-soubret  (#258) and @hessammehr (#435).

hessammehr:
Okay, thanks for the pointers. I have no experience with BinaryBuilder but will have a go at it in a minute.

hessammehr:
So the issue here is that the build script depends on Julia itself (task.c uses julia.h and links against Julia libs), which BinaryBuilder.jl doesn't provide in its RootFS. And none of the examples that I looked at depended on Julia, so I'm clueless. For the Linux target one could `wget` the Julia binary tarball and extract it on the fly; not too sure about Windows.

yebai:
>  For the Linux target one could `wget` the Julia binary tarball and extract it on the fly; not too sure about Windows.

@hessammehr You may find the Appveyor configuration file helpful. Basically, we download Julia and MinGW first, then Turing's build script is run:

https://github.com/yebai/Turing.jl/blob/master/appveyor.yml

I'm glad that someone gave it an attempt at this issue, I'll be happy to help along the process if this can be solved.

hessammehr:
Thanks @yebai, I'll definitely check out your Appveyor config. Just wondering, though - how indispensable is task.c?

hessammehr:
So I have BinaryBuilder working for all Linux targets [here](https://github.com/hessammehr/LibtaskBuilder) (still waiting on Travis). I'm still struggling a bit with Windows though since I haven't figured out how to extract the Julia installer .exe file. Anyone know of a working portable Linux version of 7zip?
[...]

PR #439, opened 2018-04-19
libtask using BinaryProvider

hessammehr:
Fix for #422.


yebai:
@hessammehr Many thanks for the fix!

hessammehr:
No problem, thanks for all the help!

PR #455, opened 2018-07-08
`add_sample!`: Use Dot Syntax to Reduce Allocations

### LINKED ISSUE #451: `add_sample!`: Correctness
If we're not mistaken, the updates performed in [`add_sample!`](https://github.com/TuringLang/Turing.jl/blob/6f144d41f815b0fb2c72b6e26228279931e42234/src/samplers/support/adapt.jl#L16) are

```
mu_{n+1} = n / (n + 1) mu_n + 1 / (n + 1) s,
M_{n+1}  = M_n + (s - mu_n) * (s - mu_{n+1})
         = M_n + n / (n + 1) (s - mu_n)^2.
```

Is mixing `mu_n` and `mu_{n+1}` in the update of `M_{n+1}` intended? The resulting sequence for `M` weights later samples more. In the limit, the estimate will indeed be unbiased:

```
Limit_{n -> Infinity} 1 / n Sum_{i = 1 ... n} i / (i + 1) = 1
```

However, for finite `n`, this seems to produce a biased estimate.

As a side note, we can reduce the number of allocations by performing in-place updates instead:

```julia
  δ = s .- ve.μ
  ve.μ .+= δ / ve.n
  ve.M .+= δ .* (s .- ve.μ)
```

wesselb:
Just a simple change that saves some allocations.

Resolves #451.

### DIFF
--- a/src/samplers/support/adapt.jl
+++ b/src/samplers/support/adapt.jl
@@ -15,9 +15,9 @@ end
 
 add_sample!{T<:Real}(ve::VarEstimator{T}, s::Vector{T}) = begin
   ve.n += 1
-  δ = s - ve.μ
-  ve.μ += δ / ve.n
-  ve.M += δ .* (s - ve.μ)
+  δ = s .- ve.μ
+  ve.μ .+= δ / ve.n
+  ve.M .+= δ .* (s .- ve.μ)
 end
 
 get_var(ve::VarEstimator) = begin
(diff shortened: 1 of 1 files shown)


src/samplers/support/adapt.jl
```diff
@@ -15,9 +15,9 @@ end
 
 add_sample!{T<:Real}(ve::VarEstimator{T}, s::Vector{T}) = begin
   ve.n += 1
-  δ = s - ve.μ
-  ve.μ += δ / ve.n
-  ve.M += δ .* (s - ve.μ)
+  δ = s .- ve.μ
+  ve.μ .+= δ / ve.n
```
willtebbutt: Should the division here not also be dotted?

wesselb: Yes, it should. Good catch!

Issue #447, opened 2018-05-28
Remove Stan dependency on Turing


xukai92:
Proposed in https://github.com/TuringLang/Turing.jl/issues/445

> Use Require.jl for the part of the code which only for the support of Stan's construction

This needs a bit of work because the current code use `Stan.adapt()` to initialize out adaptaion parameters. This could be removed by simply hand-writting the same default values as Stan.

PR #459, opened 2018-08-06
 Use Requires.jl to remove the dependency of Stan

xukai92:
Addressing #447 

1. define default values: `STAN_DEFAULT_ADAPT_CONF = nothing` and `DEFAULT_ADAPT_CONF_TYPE = Void`
2. update them to `STAN_DEFAULT_ADAPT_CONF = Stan.Adapt()` and `DEFAULT_ADAPT_CONF_TYPE = Union{Void,Stan.Adapt}` if Stan is available, using Requires.jl
3. define corresponding functions using these two constants 
4. make sure the initialisation function in `adapt.jl` can deal with both situations


xukai92:
AppVeyor failed because Cairo had build errors. I meet the same thing on my macOS today, which is fixed by what suggested here https://github.com/JuliaGraphics/Cairo.jl/issues/230#issuecomment-408576133. But I guess there is no need to do this in the AppVeyor script as this bug will disappear when Cairo or Homebrew fix that (or when we migrate to Julia 0.7).

yebai:
Looks good - thanks, Kai!

PR #487, opened 2018-09-03
Refactor transform.jl

willtebbutt:
This PR contains additional modifications to #485 , so please don't even think about merging this until that branch is merged.

This is a substantial refactoring of transform.jl, including:
- Refactoring of vectorised code to more cleanly separate `link` / `invlink` / `logpdf_with_trans` calls from one another.
- A lot more testing. There were some bugs that had been missed in the 1.0 transition due to poor coverage.
- Proper fallbacks: there are now proper fallbacks for all distributions, which vectorise properly.
- Looser type constraints: all vectors are now `AbstractVector`s and matrices are `AbstractMatrices`.

Questions:
- the global TRANS_CACHE: did anyone do any benchmarking to determine whether or not having this cache provides a substantial reduction in complexity in actual use cases? I've removed the code that depends upon on for now, but will of course reinstate it if necessary.

### DIFF
--- a/src/transform.jl
+++ b/src/transform.jl
@@ -206,38 +206,30 @@ const PDMatDistribution = Union{InverseWishart, Wishart}
 
 function link(d::PDMatDistribution, X::AbstractMatrix{T}) where {T<:Real}
   Y = cholesky(X).L
-  dim = size(Y)
-  for m in 1:dim[1]
+  for m in 1:size(Y, 1)
     Y[m, m] = log(Y[m, m])
   end
-  for m in 1:dim[1], n in m+1:dim[2]
-    Y[m, n] = zero(T)
-  end
-  Matrix{T}(Y)
+  return Y
 end
 
-###### FIX THIS. IT'S BROKEN!
-function invlink(d::PDMatDistribution, Z::AbstractMatrix{<:Real}) where {T<:Real}
-  dim = size(z)
-  for m in 1:dim[1]
-    z[m, m] = exp.(z[m, m])
+function invlink(d::PDMatDistribution, Y::LowerTriangular{T}) where {T<:Real}
+  X, dim = copy(Y), size(Y)
+  for m in 1:size(X, 1)
+    X[m, m] = exp(X[m, m])
   end
-  for m in 1:dim[1], n in m+1:dim[2]
-    z[m, n] = zero(T)
-  end
-  Matrix{T}(z * z')
+  return X * X'
 end
 
-function logpdf_with_trans(d::PDMatDistribution, x::AbstractMatrix{<:Real}, transform::Bool)
-  lp = logpdf(d, x)
+function logpdf_with_trans(d::PDMatDistribution, X::AbstractMatrix{<:Real}, transform::Bool)
+  lp = logpdf(d, X)
   if transform && isfinite(lp)
-    U = cholesky(x).U
+    U = cholesky(X).U
     for i in 1:dim(d)
       lp += (dim(d) - i + 2.0) * log(U[i, i])
     end
     lp += dim(d) * log(2.0)
   end
-  lp
+  return lp
 end
 
 
--- a/src/transform.jl
+++ b/src/transform.jl
@@ -26,9 +26,6 @@
   > SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 =#
 
-const AV{T} = AbstractVector{T}
-const AM{T} = AbstractMatrix{T}
-
 #############
 # a ≦ x ≦ b #
 #############
@@ -87,11 +84,11 @@ end
 const RealDistribution = Union{Cauchy, Gumbel, Laplace, Logistic,
                                NoncentralT, Normal, NormalCanon, TDist}
 
-@inline link(d::RealDistribution, x::Real) = x
+link(d::RealDistribution, x::Real) = x
 
-@inline invlink(d::RealDistribution, x::Real) = x
+invlink(d::RealDistribution, x::Real) = x
 
-@inline logpdf_with_trans(d::RealDistribution, x::Real, transform::Bool) = logpdf(d, x)
+logpdf_with_trans(d::RealDistribution, x::Real, transform::Bool) = logpdf(d, x)
 
 
 #########
@@ -102,11 +99,11 @@ const PositiveDistribution = Union{BetaPrime, Chi, Chisq, Erlang, Exponential, F
                                    Gamma, InverseGamma, InverseGaussian, Kolmogorov, LogNormal,
                                    NoncentralChisq, NoncentralF, Rayleigh, Weibull}
 
-@inline link(d::PositiveDistribution, x::Real) = log(x)
+link(d::PositiveDistribution, x::Real) = log(x)
 
-@inline invlink(d::PositiveDistribution, x::Real) = exp(x)
+invlink(d::PositiveDistribution, x::Real) = exp(x)
 
-@inline function logpdf_with_trans(d::PositiveDistribution, x::Real, transform::Bool)
+function logpdf_with_trans(d::PositiveDistribution, x::Real, transform::Bool)
   return logpdf(d, x) + transform * log(x)
 end
 
@@ -117,11 +114,11 @@ end
 
 const UnitDistribution = Union{Beta, KSOneSided, NoncentralBeta}
 
-@inline link(d::UnitDistribution, x::Real) = logit(x)
+link(d::UnitDistribution, x::Real) = logit(x)
 
-@inline invlink(d::UnitDistribution, x::Real) = invlogit(x)
+invlink(d::UnitDistribution, x::Real) = invlogit(x)
 
-@inline function logpdf_with_trans(d::UnitDistribution, x::Real, transform::Bool)
+function logpdf_with_trans(d::UnitDistribution, x::Real, transform::Bool)
   return logpdf(d, x) + transform * log(x * (one(x) - x))
 end
 
@@ -132,9 +129,13 @@ end
 
 const SimplexDistribution = Union{Dirichlet}
 
-link(d::SimplexDistribution, x::AV{<:Real}) = link!(similar(x), d, x)
-function link!(y::AV{<:Real}, d::SimplexDistribution, x::AV{T}) where {T<:Real}
-  K = length(x)
+link(d::SimplexDistribution, x::AbstractVector{<:Real}) = link!(similar(x), d, x)
+function link!(
+  y::AbstractVector{<:Real},
+  d::SimplexDistribution,
+  x::AbstractVector{<:Real},
+)
+  K, T = length(x), eltype(x)
 
   sum_tmp = zero(T)
   z = x[1]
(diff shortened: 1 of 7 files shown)


trappmartin:
I really like the changes as it makes the code much cleaner. Just two questions:

1. Do we need the constants you defined? I understand the convenience but I feel the codebase becomes inconsistent when not used everywhere.

2. Do we need to explicitly inline the code? From what I understood at JuliaCon this might not be necessary anymore as the compiler improved quite a bit. I would prefer to benchmark this and see if the explicit inline hurts the compiler in optimising the code base or if it improves or doesn’t matter.

willtebbutt:
> Do we need the constants you defined?

Probably not. I always have mixed feeling about using these constants for exactly the reason you point out, to the extent that I chop and change between using them and not using them quite regularly. I'll refactor to use `AbstractVector` and `AbstractMatrix` explicitly, and take the code-bloat hit.

> Do we need to explicitly inline the code?

Ah, yeah, they just shouldn't be there. You're correct that they're almost certainly unnecessary; I'm very much in favour of letting the compiler do whatever it thinks it best with these functions.

willtebbutt:
Okay @trappmartin I've made the changes as suggested, but now I'm on the fence again regarding the `AbstractVector` / `AbstractMatrix` abbreviations :P The code is quite a bit more difficult to read IMHO. What are your thoughts?

@xukai92 @yebai I could use a hand understanding some weird behaviour: I've changed the `link` and `invlink` implementations for `PDMatDistribution`s (see the `transform.jl` file). Previously, the `invlink` implementation was silently modifying it's second argument, which is clearly bad. Upon doing so, I get a number of bugs appearing throughout the package to do with `cholesky` factorisations failing, despite the fact that my new tests for `link` and `invlink` pass nicely. Do either of you know if the mutating behaviour was intentional for some reason? I would appreciate it if you could run the test and let me know if anything obvious jumps out at you. If not, I'll dig a little deeper.

src/transform.jl
```diff
@@ -206,38 +206,30 @@ const PDMatDistribution = Union{InverseWishart, Wishart}
...
-###### FIX THIS. IT'S BROKEN!
-function invlink(d::PDMatDistribution, Z::AbstractMatrix{<:Real}) where {T<:Real}
-  dim = size(z)
-  for m in 1:dim[1]
-    z[m, m] = exp.(z[m, m])
-  end
-  for m in 1:dim[1], n in m+1:dim[2]
-    z[m, n] = zero(T)
+function invlink(d::PDMatDistribution, Y::LowerTriangular{T}) where {T<:Real}
+  X, dim = copy(Y), size(Y)
```
xukai92: Do you intend to write `zero(Y)`

xukai92:
> Do either of you know if the mutating behaviour was intentional for some reason?
This is just intended to save memory allocations. But yes this is bad.

For the failing test I added a comment to the code.

xukai92:
Also we shall merge this into `flux-support` as a lot of commits are from that incomplete PR?

willtebbutt:
> how about merging flux-support into master first?

My intent was definitely that `flux-support` be merged into master before this does.

src/transform.jl
willtebbutt: I don't believe so. Previously we were just mutating `Y` directly, so I think that `copy`ing is the appropriate behaviour.

xukai92: It seems that we don't have the L225-L227 of the original code.

xukai92: Thanks for the clarification. I think here should be no problem.

willtebbutt: Okay. I'll dig into the code a little deeper and try to figure out what's going wrong.

willtebbutt: Okay, I think that I've figured out what the root cause of the problem is here. The story is as follows:
1. With `PDMatDistribution`s, I want their unconstrained form to be `LowerTriangular` type, and the constrained type to be `Matrix` (really I want it to be `PDMat`, but that's another discussion). 
2. [link](https://github.com/TuringLang/Turing.jl/blob/e6782521706f8343ea7e364cb87537461b6cc13b/src/core/varinfo.jl#L125) and [invlink](https://github.com/TuringLang/Turing.jl/blob/e6782521706f8343ea7e364cb87537461b6cc13b/src/core/varinfo.jl#L139) require that the manner in which `reconstruct` and `vectorize` function depend only upon the distribution `dist` and a vector of `Real`s pulled from a `var_info` object. Crucially, this means that the particular method of `reconstruct` used _cannot_ depend on whether the thing being reconstructed was previously a `LowerTriangular` or a `Matrix`. This also explains why the previous implementation worked nicely: the matrix ways always represented as a `Matrix`, never a `LowerTriangular`, meaning that this simply wasn't an issue.

I'll find a hack for now, but we should really address this - moving forward it will become problematic if we have to keep hacking our way around this.

willtebbutt:
@xukai92 my recent commit resolves the issue discussed above, but also refactors some loosely-related code that I had to dig into to figure out what was going wrong. Please let me know if you're happy with the refactoring; if not I'll just revert it.

xukai92:
You mean the type changes in `src/helper.jl`?

willtebbutt:
> You mean the type changes in src/helper.jl?

Yeah, that and the general re-organisation of the code; it was just a bit messy before.

willtebbutt:
The only outstanding issue relates to `TRANS_CACHE`, and whether or not we need it. As I understand it, it's just a cache which keeps track of matrices which have already been allocated for use in the `link!` and `invlink!` functions.

What was the original motivation for its inclusion? My opinion is that we probably shouldn't be doing stuff like this, unless there's a strong performance case, since it
1. is itself a potentially serious performance bottleneck: it's not hard to imagine a case in which holding onto references to matrices unnecessarily and preventing the GC from reallocating their memory causes a programme to use significantly more memory than required. Moreover one could imagine this causing programmes to crash in degenerate cases.
2. bloats the code and makes it less readable.
3. It adds more global mutable state that we have to keep track of.

willtebbutt:
Found some more bugs, please don't merge.

willtebbutt:
The latest commits remove `link!` and `invlink!` methods in favour of `link` and `invlink` methods for `SimplexDistribution`s as these weren't actually getting used anywhere, and were essentially just code bloat.

@xukai92 @yebai I'll remove the old `TRANS_CACHE` code if there are no objections, then we'll be good to go.

yebai:
>  @xukai92 @yebai I'll remove the old TRANS_CACHE code if there are no objections, then we'll be good to go.

Hi @willtebbutt, the main use of `TRANS_CACHE` is in Gibbs sampling, where during the PG step, all parameters assigned to HMC does not change. Since the PG uses many particles and may run several iterations, it is wasteful to perform `link/invlink` for parameters that do not change. I agree that `TRANS_CACHE` is not ideal, but we do need some mechanism for avoiding wasteful transformation here.

yebai:
Perhaps we can resolve this issue in a separate PR?

xukai92:
> Hi @willtebbutt, the main use of TRANS_CACHE is in Gibbs sampling, where during the PG step, all parameters assigned to HMC does not change. Since the PG uses many particles and may run several iterations, it is wasteful to perform link/invlink for parameters that do not change. I agree that TRANS_CACHE is not ideal, but we do need some mechanism for avoiding wasteful transformation here.

I think currently in Gibbs we do `link`/`invlink` after each HMC `step` so this cache might not be used at all. But I realise this design also have unnecessary `link`/`invlink` when each HMC component has multiple steps. 

Note this does not happen in pure HMC because of the condition in `if spl.alg.gid != 0 link!(vi, spl); cleandual!(vi) end` and `if spl.alg.gid != 0 invlink!(vi, spl); cleandual!(vi) end`.

Seems like we should redesign this interface such that there is another wrapper over single `step` which performs multiple steps without doing transformation in between, or make this `step()` function having an optional parameter for how many steps it. Note we didn't do so because it introduces another problem of how to consistently fetch intermediate samples in each HMC component.

willtebbutt:
(Apologies in advance for the essay). I've run some timings for various implementations of `link` with the Dirichlet distribution, for dimensionality `K ∈ (3, 1000)` (ie. 3- and 1000-dimensional Dirichlet distributions) using BenchmarkTools. Results are compressed to keep this as brief as possible.
```julia

dist = Dirichlet(K, 3);
X = rand(dist, 100);
Y = link(dist,  X);

# Fallback: iterate over single-sample version.
@benchmark link($dist, $X)

# TRANS_CACHE-based implementations.
@benchmark link_simplex($dist, $X)

# New implementation.
@benchmark link_newimpl($dist, $X)
```
In the above snippet, `link` uses the current implementation on this branch, `link_simplex` uses the `TRANS_CACHE`-based implementation, and `link_newimpl` is a manual loop-based implementation of `link` which doesn't use all of the `view`s that `link` uses. For `K = 3`, timings are:
```julia
# link:
  memory estimate:  15.83 KiB
  allocs estimate:  201
  median time:      6.840 μs (0.00% GC)

# link_simplex:
  memory estimate:  7.36 KiB
  allocs estimate:  19
  median time:      4.574 μs (0.00% GC)

# link_newimpl:
  memory estimate:  1.77 KiB
  allocs estimate:  1
  median time:      1.357 μs (0.00% GC)
```
So for a problem with small dimensionality, the fallback implementation isn't great, but isn't horrific. The 201 allocations in `link` arise because two views have to be constructed for each of the 100 samples, plus one to allocate `Y`. The manually-unrolled implementation is frankly really good. It exploits the fact that we actually really don't need to allocate a large matrix temporary at all -- everything can be done in place.
For `K = 1000`:
```julia
# link:
  memory estimate:  1.54 MiB
  allocs estimate:  202
  median time:      2.518 ms (0.00% GC)

# link_simplex:
  memory estimate:  386.51 MiB
  allocs estimate:  19384
  median time:      84.977 ms (21.41% GC)

# link_newimpl:
  memory estimate:  781.33 KiB
  allocs estimate:  2
  median time:      2.282 ms (0.00% GC)
```
The picture is quite different for problems with high dimensionality. The `link` and `link_newimpl` implementations scale really nicely; the extra allocation comes from allocating a much larger output array apparently -- I'm not sure why Julia does this, but it's not something that we have control over anyway.
The `link_simplex` implementation on the other hand has scaled poorly. It looks to me like this is due to some type instabilities.

Here's the `link_newimpl` code:
```julia

function link_newimpl(d::SimplexDistribution, X::AbstractMatrix{T}) where T<:Real
  Y, K, N = similar(X), size(X, 1), size(X, 2)

  for n in 1:size(X, 2)
    sum_tmp, z = zero(T), X[1, n]
    Y[1, n] = logit(z) - log(one(T) / (K-1))
    for k in 2:K-1
      sum_tmp += X[k - 1, n]
      z = X[k, n] / (one(T) - sum_tmp)
      Y[k, n] = logit(z) - log(one(T) / (K - k))
    end
    Y[K, n] = zero(T)
  end

  return Y
end
```
In short, I'm reasonably sure that it's strictly better than the existing implementations on either this branch or master. Also, the results will look similar for `invlink`. My recommendation is that we
1. use the hand-optimised version of `link` (i.e. rename `link_newimpl`) for `SimplexDistribution`s,
2. keep the fallback implementation of `link` which uses views for other transformations, and
3. discard the `TRANS_CACHE`-based implementation, as we have a better one in the hand-optimised implementation.

willtebbutt:
I'm happy for this to be merged once the tests have passed if everyone else is.

yebai:
Looks good to me!

PR #496, opened 2018-09-05
Small clean up in compiler.

yebai:


src/core/compiler.jl
```diff
@@ -10,7 +10,7 @@ macro VarName(ex::Union{Expr, Symbol})
   s = string(gensym())
   if isa(ex, Symbol)
     ex_str = string(ex)
-    return :(Symbol($ex_str), Symbol($s))
+    return :(Symbol($ex_str), nothing, Symbol($s))
```
trappmartin: Why not return an empty tuple? This might be a more consistent behaviour?

Issue #495, opened 2018-09-05
AD Implementations


willtebbutt:
I was skimming through the AD implementations and am a bit confused by the ForwardDiff implementation, so could use some help understanding it.

If I was to implement a function to compute the gradient of the log joint of a model using ForwardDiff in the current version of Turing, the first thing I would think to do would be along the lines of the following:
```julia
"""
gradient_forward(
    θ::AbstractVector{<:Real},
    vi::VarInfo,
    spl::Sampler,
    model::Function,
)

Computes the gradient of the log joint of `θ` for the model specified by
`(vi, spl, model)`.
"""
function gradient_forward(
    θ::AbstractVector{<:Real},
    vi::VarInfo,
    spl::Union{Nothing, Sampler},
    model::Function,
)
    # Record old parameters.
    θ_old = vi[spl]

    # Define function to compute log joint.
    function f(θ)
        vi[spl] = θ
        return runmodel(model, vi, spl).logp
    end

    # Set chunk size and do ForwardMode.
    config = GradientConfig(f, θ, Chunk{min(CHUNKSIZE, length(θ))}())
    ∂l∂θ = gradient!(similar(θ), f, θ, config)

    # Replace old parameters to ensure this function doesn't mutate `vi`.
    vi[spl] = θ_old

    return ∂l∂θ
end
```
A minimal working example might be something like this:
```julia
using Turing, ForwardDiff, Distributions
using Turing: VarInfo, Sampler, runmodel, CHUNKSIZE

# Define a model.
@model ad_test() = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  1.5 ~ Normal(m, sqrt(s))
  2.0  ~ Normal(m, sqrt(s))
  return s, m
end
model = ad_test()
vi = model(Turing.VarInfo(), nothing)
spl = nothing

# Get parameters + compute gradient.
θ = map(Float64, vi[spl])
∂l∂θ = gradient_forward(θ, vi, spl, model)
```

The `gradient_forward` function here is similar to the `gradient_r` function as currently implemented. So my first question is, other than the caching stuff at the top of the function, what is all of the rest of the code for? It looks to me like it's just emulating what `ForwardDiff`'s `GradientConfig` would already do for us.

yebai:
[...]

PR #499, opened 2018-09-06
WIP: Refactoring AD

willtebbutt:
What it says on the tin. Relevant issue is #495 . I decided to refactor the reverse-mode implementation a bit as well to make it consistent with the new forward-mode implementation. I've also renamed `gradient`->`gradient_forward` and `gradient_r`->`gradient_reverse` for the sake of clarity.

Currently `gradient_forward` is reconstructing the `Chunk` on each application, which has a non-trivial effect on the performance of smaller models. It's not clear a priori whether this has a larger effect than the overhead associated with the retrieval of global mutable state, and therefore whether it's worth caching `Chunk` objects for various chunk sizes.

TODO

- [x] Refactor slightly to make `gradient_forward` and `gradient_reverse` have completely consistent interfaces.
- [x] Ascertain whether or not chunk-caching is generally worthwhile by doing some benchmarking.

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -32,7 +32,7 @@ function gradient_forward(
   # Replace old parameters to ensure this function doesn't mutate `vi`.
   vi[spl] = θ_old
 
-  return ∂l∂θ
+  return getlogp(vi), ∂l∂θ
 end
 
 """
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -47,27 +47,27 @@ Computes the gradient of the log joint of `θ` for the model specified by `(vi, 
 using reverse-mode AD from Flux.jl.
 """
 function gradient_reverse(
-  x::AbstractVector{<:Real},
+  θ::AbstractVector{<:Real},
   vi::Turing.VarInfo,
   model::Function,
   spl::Union{Nothing, Sampler}=nothing,
 )
   # Specify objective function.
-  function f(x)
-    vi[spl] = x
+  function f(θ)
+    vi[spl] = θ
     return -runmodel(model, vi, spl).logp
   end
 
   # Compute forward and reverse passes.
-  l, ȳ = Tracker.forward(f, x)
-  ∂l∂x = ȳ(1)[1]
+  l, ȳ = Tracker.forward(f, θ)
+  ∂l∂θ = ȳ(1)[1]
 
   # Remove tracking info from variables in model (because mutable state).
   vi.logp = Tracker.data(vi.logp)
   vi[spl] .= Tracker.data.(vi[spl])
 
   # Return non-tracked gradient value
-  return Tracker.data(l), Tracker.data(∂l∂x)
+  return Tracker.data(l), Tracker.data(∂l∂θ)
 end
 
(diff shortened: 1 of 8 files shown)


src/core/ad.jl
```diff
@@ -1,127 +1,81 @@
...
+  ∂l∂θ = ForwardDiff.gradient!(similar(θ), f, θ, config)
 
-  # Split keys(vi) into chunks,
-  @debug "making chunks..."
-  vn_chunk = Set{VarName}(); vn_chunks = []; chunk_dim = 0;
+  # Replace old parameters to ensure this function doesn't mutate `vi`.
+  vi[spl] = θ_old
 
-  vns = getvns(vi, spl); vn_num = length(vns)
+  return ∂l∂θ
```
xukai92: I think we should make the return between forward and reverse mode consistent.

xukai92: By which I mean `return getlogp(vi), ∂l∂θ` here

willtebbutt: I entirely agree -- see TODOs above.

willtebbutt:
I'm going to propose that we just go with my most recent commit for now. In some basic benchmarking I've done on the following model, the new version that constructs a new config each time takes around 2 / 3 of the time that the version on master does to compute `gradient_forward` / `gradient` respectively:
```julia
@model ad_test() = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  1.5 ~ Normal(m, sqrt(s))
  2.0  ~ Normal(m, sqrt(s))
  return s, m
end
```
On my laptop I get the following benchmark results on master:
```julia
julia> @benchmark Turing.gradient($vi, $model_, $spl)
BenchmarkTools.Trial:
  memory estimate:  20.98 KiB
  allocs estimate:  122
  --------------
  minimum time:     16.921 μs (0.00% GC)
  median time:      18.327 μs (0.00% GC)
  mean time:        20.449 μs (8.15% GC)
  maximum time:     2.248 ms (97.05% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
whereas the pushed branch gets:
```julia
julia> @benchmark Turing.gradient_forward($θ, $vi, $model_, $spl)
BenchmarkTools.Trial:
  memory estimate:  2.16 KiB
  allocs estimate:  53
  --------------
  minimum time:     10.700 μs (0.00% GC)
  median time:      11.632 μs (0.00% GC)
  mean time:        12.257 μs (3.07% GC)
  maximum time:     3.806 ms (98.88% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
and time taken to compute `θ = realpart(vi[spl])` is < 200ns.

yebai:
Thanks, @willtebbutt. I did some quick tests, here are my findings

```julia
julia> @time ForwardDiff.Chunk{52}()
  0.000345 seconds (577 allocations: 34.820 KiB)
Chunk{52}()

julia> @time ForwardDiff.Chunk{52}()
  0.000031 seconds (4 allocations: 160 bytes)
Chunk{52}()
```

Apparently the `ForwardDiff` package has already implemented some caching mechanism for `Chunk`..

Issue #486, opened 2018-09-03
HMC is type unstable


mohamed82008:
https://github.com/TuringLang/Turing.jl/blob/e6782521706f8343ea7e364cb87537461b6cc13b/src/samplers/hmc.jl#L31

Is there any reasonable default choice for the `space` set element? And can it change during the life of the object? If it can't change, then only the second constructor is problematic since the type of `space` is not known.

mohamed82008:
A search through the package shows 0 instances of modifying the `space` after object creation, it is a read-only. I assume this means we can just use an empty set `Set{Nothing}` as the default.

mohamed82008:
This is a common problem for other algorithms as well causing a performance hit.

PR #503, opened 2018-09-07
[WIP] Make structs parametric and type stable and take constructors out

mohamed82008:
Closes https://github.com/TuringLang/Turing.jl/issues/486. I took the constructors out because that's easier when dealing with parameters since the non-parametric constructor is defined by default and I can call it from other external constructors.

### DIFF
--- a/src/core/io.jl
+++ b/src/core/io.jl
@@ -86,18 +86,18 @@ mean(chain[:mu])      # find the mean of :mu
 mean(chain[:sigma])   # find the mean of :sigma
 ```
 """
-mutable struct Chain <: AbstractChains
+mutable struct Chain{R<:AbstractRange{Int}} <: AbstractChains
   weight  ::  Float64                 # log model evidence
   value2  ::  Array{Sample}
   value   ::  Array{Float64, 3}
-  range   ::  AbstractRange{Int} # TODO: Perhaps change to UnitRange?
-  names   ::  Vector{AbstractString}
+  range   ::  R # TODO: Perhaps change to UnitRange?
+  names   ::  Vector{String}
   chains  ::  Vector{Int}
   info    ::  Dict{Symbol,Any}
 end
 
-Chain() = Chain(0, Vector{Sample}(), Array{Float64, 3}(undef, 0,0,0), 0:0,
-                Vector{AbstractString}(), Vector{Int}(), Dict{Symbol,Any}())
+Chain() = Chain{AbstractRange{Int}}(0.0, Vector{Sample}(), Array{Float64, 3}(undef, 0, 0, 0), 0:0,
+                Vector{String}(), Vector{Int}(), Dict{Symbol,Any}())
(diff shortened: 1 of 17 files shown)


src/core/io.jl
```diff
@@ -86,18 +86,18 @@ mean(chain[:mu])      # find the mean of :mu
 mean(chain[:sigma])   # find the mean of :sigma
 ```
 """
-mutable struct Chain <: AbstractChains
+mutable struct Chain{R<:AbstractRange{Int}} <: AbstractChains
```
willtebbutt: Consider `AbstractRange{<:Integer}`, or perhaps simple `AbstractRange`.

src/samplers/pmmh.jl
```diff
@@ -11,17 +11,17 @@ alg = PMMH(100, SMC(20, :v1), MH(1,:v2))
 alg = PMMH(100, SMC(20, :v1), MH(1,(:v2, (x) -> Normal(x, 1))))
 ```
 """
-mutable struct PMMH <: InferenceAlgorithm
+mutable struct PMMH{T, A} <: InferenceAlgorithm
```
willtebbutt: Consider `A <: Tuple`, to ensure the same type constraints are in place as before. I'm not sure if this is entirely necessary though.

willtebbutt:
How far are you planning on pushing these changes? There's still quite a lot of concretely typed fields (eg. in `Chain`) that it might be better to make parametric. I appreciate that this is a WIP, so maybe you're already thinking about this...

src/core/io.jl
mohamed82008: Any reason?

src/samplers/pmmh.jl
mohamed82008: Sure.

mohamed82008:
I tried using a specific range type in `Chain` but apparently it changes somewhere in the code so that gave a conversion error. I will look into it again.

mohamed82008:
I guess the rule of thumb is when you need more parameters, add them. Parameters come at the cost of convenience sometimes. This PR just deals with type instability issues, not arbitrary types of fields, e.g. handling `Float32`s. This can be done in a different PR after the design of the package has stabilized.

yebai:
> I tried using a specific range type in Chain but apparently it changes somewhere in the code so that gave a conversion error. I will look into it again.

Martin is working towards a rewrite of `io.jl` -- let's leave `Chain` as is for now.

src/core/io.jl
willtebbutt: It's just less concrete, and I don't think we really lose anything. A completely generic `AbstractRange` might be a bit loose though actually.

mohamed82008:
So I rebased this on master. If tests pass and no further comments are there, I believe this is ready to go in.

yebai:
Thanks @mohamed82008!

Issue #509, opened 2018-09-07
Refactoring: remove `cleandual!` and `realpart!`


yebai:
Shall we remove the function `cleandual!` since our AD is using standard `ForwardDiff` and `Flux` interface now? We can always do

```julia
vi[spl] = theta::Vector{Float64}
```
after each AD iteration to ensure `vi` has the right types.

willtebbutt:
I agree we should remove it, but this actually highlights a bug that I hadn't realised would exist in the AD code. We're not ensuring that the `logp` field of `VarInfo` isn't a `Dual` or `Tracked`. I'll tack that onto my soon-to-be-opened AD-related PR, and remove `cleandual!`.

willtebbutt:
We might also be able to get rid of `realpart` if we're not contaminating things with `Dual`s anymore. Will look into it.

willtebbutt:
Closed by #515

Issue #450, opened 2018-06-30
Minor Comment Correction


wesselb:
```
  stats = stats > 1 ? 1 : stats # stats = δ - H_t, where δ is target accept prob
```
from [here](https://github.com/TuringLang/Turing.jl/blob/6f144d41f815b0fb2c72b6e26228279931e42234/src/samplers/support/adapt.jl#L95): Is `stats = δ - H_t` in the comment correct? I think `stats` is the average acceptance MH probability.

willtebbutt:
Closed by #515

PR #515, opened 2018-09-10
Tidy up AD interface + remove cleandual! and realpart!. Do not merge.

willtebbutt:
This PR does quite a few things. Firstly, it moves `sghmc` and `sgld` over to using the new `gradient` interface, and resolves an interesting bug that that threw up.

The second important thing is that it improves `gradient` by ensuring that it has no side-effects, and cleans up after itself. In particular it ensures that all tracking information is removed from `vi.vals`. This enables the satisfactory resolution of #509 and further enables the complete removal of `realpart`, as we're no longer carrying around which appears to concerned precisely the removal of tracking information.

There are also a few formatting fixes that I undertook whenever significant modifications were made to a file.

There remain a couple of loose ends to tie up, and I would advise against merging this until #513 is merged anyway to avoid clashes.


xukai92:
The changes look good to me in general. I will take a review on it when it's ready.

willtebbutt:
I've taken the liberty of addressing #450 since it's a one-line fix.

src/core/ad.jl
```diff
@@ -1,83 +1,120 @@
...
+    l, ∂l∂θ = Tracker.data(l_tracked), Tracker.data(ȳ(1)[1])
+
+    # Remove tracking info from variables in model (because mutable state).
+    vi.vals, vi.logp = vals_old, logp_old
+
+    # Strip tracking info from θ to avoid mutating it.
+    θ .= Tracker.data.(θ)
+
+    # Return non-tracked gradient value
+    return l, ∂l∂θ
```
xukai92: Do you think it might be good for us to wrapper the result with `DiffResults.DiffResult` (http://www.juliadiff.org/DiffResults.jl/stable/#Constructing-a-DiffResult-1). And retrieve them later using `DiffResults.value` and `DiffResults.gradient`.

willtebbutt: Good idea, but let's do this in a separate PR when incorporating DynamicHMC.

Issue #507, opened 2018-09-07
Assume in loops and array type


hessammehr:
The following code seems to work fine:

```julia
alg = HMC(1000, 0.2, 4)

@model vdemo() = begin
  x = Vector{Real}(undef, N)
  for i = 1:N
    x[i] ~ Normal(0, sqrt.(4))
  end
end

t_loop = @elapsed res = sample(vdemo(), alg)
```

However, changing `Real` to `Float64` seems to make it break.

```julia
alg = HMC(1000, 0.2, 4)

@model vdemo() = begin
  x = Vector{Float64}(undef, N)
  for i = 1:N
    x[i] ~ Normal(0, sqrt.(4))
  end
end

t_loop = @elapsed res = sample(vdemo(), alg)
```

`ERROR: MethodError: no method matching Float64(::Flux.Tracker.TrackedReal{Float64})`

There seems to be an easy fix:
```julia
Float64(f::Flux.Tracker.TrackedReal{Float64}) = f.data
```

A few thoughts/questions:
- Is there a reason we use an abstract type `Real` instead of, e.g., `Float64`?
- Are we interested in adding the above shim?
- Why the broadcast in the `sqrt` call?

yebai:
> Is there a reason we use an abstract type Real instead of, e.g., Float64?

This is required to make AD work. `Flux.Tracked` is a subtype of `Real`.

> Are we interested in adding the above shim?

I'm not sure cause it might cause incorrect gradients.

yebai:
@xukai92 might be able to elaborate on this.

xukai92:
I agree with @yebai on the first two answers. + I think doing `Float64(f::Flux.Tracker.TrackedReal{Float64}) = f.data` is against type piracy (https://docs.julialang.org/en/v1.0.0/manual/style-guide/#Avoid-type-piracy-1)?

> Why the broadcast in the sqrt call?

I don't think it's necessary to use broadcast here.

willtebbutt:
Adding
```julia
Float64(f::Flux.Tracker.TrackedReal{Float64}) = f.data
```
will almost certainly cause AD to fail silently, so we should _definitely_ not do this.

willtebbutt:
If the above `sqrt` broadcasting is something that is in one of our examples, then @hessammehr it would be good if you could make a PR to fix it. Otherwise we should close this issue.

hessammehr:
@willtebbutt Will do.

[...]

PR #519, opened 2018-09-11
Remove spurious broadcasts (see #507).

hessammehr:
Pretty straightforward change of `sqrt.(scalar)` to `sqrt(scalar)`,
mostly in test code.


hessammehr:
@yebai I see and agree, but do you think we could fix that in a different PR, please?

yebai:
@hessammehr sure, thanks!

PR #523, opened 2018-09-12
A little cleanup

### LINKED ISSUE #500: logsum -> logsumexp
https://github.com/TuringLang/Turing.jl/blob/70640d35805bb46a54c04d9ff25e5a90f22eee2b/src/core/util.jl#L23

I believe this is the famous log-sum-exp function.

mohamed82008:
Does a few things and closes #500.


src/core/util.jl
```diff
@@ -61,25 +62,21 @@ end
 
 auto_tune_chunk_size!(mf::Function, rep_num=10) = begin
   dim = length(mf().vals)
+  chunk_size = 8
```
yebai: Perhaps we should remove this function since it's no longer used after recent AD refactoring?

src/core/varinfo.jl
```diff
@@ -306,7 +306,7 @@ end
 
 # Get all indices of variables belonging to gid or 0
 getidcs(vi::VarInfo) = getidcs(vi, nothing)
-getidcs(vi::VarInfo, spl::Nothing) = filter(i -> vi.gids[i] == 0 || vi.gids[i] == 0, 1:length(vi.gids))
+getidcs(vi::VarInfo, spl::Nothing) = filter(i -> vi.gids[i] == 0, 1:length(vi.gids))
```
yebai: Nice catch!

willtebbutt:
Also looks like `kl` isn't used. If #500 is resolved and @yebai's suggestions implemented, then the `core/util.jl` could probably be removed completely if new homes can be found for `randcat` and `sym_str`.

yebai:
> Also looks like kl isn't used. 

`kl` might be used in some downstream code (e.g. tests, benchmarking).

mohamed82008:
> `kl` might be used in some downstream code (e.g. tests, benchmarking).

https://github.com/TuringLang/Turing.jl/blob/cba33402b3d71e95aa5cee9c043db5698b9f3b8c/test/util.jl/util.jl#L16

willtebbutt:
Yes, but that test just tests the `kl` implementation is correct. It's not used for anything else.

test/is.jl/importance_sampling.jl
```diff
@@ -13,7 +14,7 @@ function reference(n :: Int)
     samples[i] = reference()
     logweights[i] = samples[i][:logweight]
   end
-  logevidence = Turing.logsumexp(logweights) - log(n)
+  logevidence = logsumexp(logweights) - log(n)
```
yebai: ==> StatsFuns.logsumexp?

mohamed82008: Well, there is only one `logsumexp`.

mohamed82008:
This PR seems ready. If no further comments are there, it can be merged.

src/utilities/util.jl
```diff
@@ -0,0 +1,18 @@
+########
+# Math #
+########
+
+@inline invlogit(x::T) where T<:Real = one(T) / (one(T) + exp(-x))
```
willtebbutt: These can probably be removed as `StatsFuns` provides them.

test/util.jl/util.jl
```diff
@@ -1,4 +1,4 @@
-using Turing: @VarName, invlogit, logit, randcat, kl, align
+using Turing: @VarName, invlogit, logit, randcat
```
trappmartin: The logit related exports can probably be removed.

src/utilities/io.jl
```diff
@@ -200,7 +200,7 @@ function Base.vcat(c1::Chain, args::Chain...)
   Chain(0, value2)
 end
 
-save!(c::Chain, spl::Sampler, model::Function, vi::VarInfo) = begin
+save!(c::Chain, spl::Sampler, model::Function, vi) = begin
```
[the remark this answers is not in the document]
mohamed82008: The reason is that `VarInfo` is defined in a file that's `include`d after the `utilities` files. This would error after moving the `io.jl` to the `utilities` folder, unless I include this file before and the rest of the utilities files after, which looked weird. Removing the unnecessary `VarInfo` annotation was more elegant.

Issue #542, opened 2018-09-15
Add numerical tests to transformations


yebai:
The current test file for `transform.jl` only test the functionality of `logpdf_with_trans`

https://github.com/TuringLang/Turing.jl/blob/ce780b87fa7ac64a4a516c8bfa1e32fe8152f55d/test/transform.jl/transform.jl#L11

We need to add correctness tests for `logpdf`, e.g.

```julia
Distributions.logpdf(d, x) == Turing.logpdf_with_trans(d, link(x), true)
```

willtebbutt:
Good catch. I completely forgot to sort that out when I was refactoring. Will sort it out now.

willtebbutt:
#543 now resolves most of these issues, but there remain a couple of outstanding issues:

1. A few distributions (`Dirichlet`, `Wishart`, `InverseWishart`) don't have tests to ensure that their link functions are consistent with their `logpdf_with_trans` implementation. Due to the way that we transform between the constrained and unconstrained representations the Jacobian of the transformation is singular. This isn't a fundamental issue, but I'm not completely sure what the best way to resolve it is beyond directly hacking away at stuff.
2. The logpdf of the Kolmogorov distribution returns `-Inf` for `x` close to zero. I assume this is a numerical stability issue, so I've disabled testing for that for now as it messes everything up, and is probably on the `Distributions.jl` end.

yebai:
> A few distributions (Dirichlet, Wishart, InverseWishart) don't have tests to ensure that their link functions are consistent with their logpdf_with_trans implementation. Due to the way that we transform between the constrained and unconstrained representations the Jacobian of the transformation is singular. This isn't a fundamental issue, but I'm not completely sure what the best way to resolve it is beyond directly hacking away at stuff.

I did some eye-ball tests a while ago and didn't find any obvious errors.

willtebbutt:
> I did some eye-ball tests a while ago and didn't find any obvious errors.

Sure, I agree that it might be fine, but we really should be testing this stuff _heavily_ as, as you point out, the numerical stability of all of our HMC-related code depends on it quite heavily.

xukai92:
> Due to the way that we transform between the constrained and unconstrained representations the Jacobian of the transformation is singular. This isn't a fundamental issue, but I'm not completely sure what the best way to resolve it is beyond directly hacking away at stuff.

Will a numerical approximation of Jacobian work here?

willtebbutt:
[...]

PR #543, opened 2018-09-15
Better testing for transforms + code which conforms to the style guide

willtebbutt:
Partially addresses #542. Doesn't currently do vector- or matrix-variate distributions for technical reasons.

Also sorts out indentation in `src/utilities/transform.jl`.

### DIFF
--- a/src/utilities/transform.jl
+++ b/src/utilities/transform.jl
@@ -81,8 +81,9 @@ end
 # -∞ < x < -∞ #
 ###############
 
-const RealDistribution = Union{Cauchy, Gumbel, Laplace, Logistic,
-                               NoncentralT, Normal, NormalCanon, TDist}
+const RealDistribution = Union{
+    Cauchy, Gumbel, Laplace, Logistic, NoncentralT, Normal, NormalCanon, TDist,
+}
 
 link(d::RealDistribution, x::Real) = x
 invlink(d::RealDistribution, y::Real) = y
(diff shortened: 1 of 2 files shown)


willtebbutt:
Current failures appear to be the usual statistical ones. I'll try to sort out tests for vector- and matrix-variate distributions this afternoon, then we should be good to go.

mohamed82008:
Could a missing `Symmetric` wrapper be causing https://travis-ci.org/TuringLang/Turing.jl/jobs/428954293#L764? I cannot reproduce it on my machine so it could be floating point error causing some asymmetry in a matrix on which `cholesky` is called. If the matrix is not perfectly symmetric, it will trigger a positive definite error, unless wrapped by `Symmetric`.

willtebbutt:
I've added a number of tests using the `@test_broken` macro where I couldn't figure out the best way to check for consistency between the link function and `logpdf_with_trans` implementation using AD. These won't prevent CI from passing, but they will flag up in the test summary. This way, it will remind us that we have things that we need to fix, but doesn't halt development, and allows us to merge the code that I've already added in this PR (which I think is strictly better than what we had before).

willtebbutt:
Turns out that `LogMvNormal` wasn't being properly transformed, and our tests didn't cover this. I've now added an appropriate transformation for `LogMvNormal`, and added tests to ensure that if we take arbitrary points in the unconstrained domain and map them to the constrained domain using the appropriate inverse link function the `logpdf` is finite. These tests are statistical in nature (obviously you can't cover the entire space) but do a much better job than we were doing previously (which was nothing at all).

yebai:
Thanks, @willtebbutt. The code looks good to me!

PR #549, opened 2018-09-18
[WIP] DynamicHMC support

xukai92:
TODOs
-  ~figure out how to pass target accept rate and number of adaptations to DynamicHMC~
- [x] clean up style
- [x] figure our how to handle unregistered dependencies
- [x] use [Requires.jl](https://github.com/MikeInnes/Requires.jl) to resolve the unregistered dependencies
- [x] change the keyword argument way to a new type for DynamicHMC

Summary
The way I'm current supporting DynamicHMC is through a keyword argument called `implementation` in `sample`, e.g. `sample(model_func, DynamicNUTS(...))`. See the test file for a complete example.


yebai:
> figure out how to pass target accept rate and number of adaptations to DynamicHMC

I think passing in number of adaptions is probably enough here. We can leave target accept rate as default.

xukai92:
This PR is almost finished. One question on testing: for functionality which using Requires.jl, what's the best way to put them in our test system? Do we actually want to manually set up travis to install, say, DynamicHMC.jl related packages and do the tests?

yebai:
> This PR is almost finished. One question on testing: for functionality which using Requires.jl, what's the best way to put them in our test system? Do we actually want to manually set up travis to install, say, DynamicHMC.jl related packages and do the tests?

Perhaps we can install them from `travis` script for now?

xukai92:
I wonder if the new environment thing can be used? If not we will do that manually.

yebai:
> I wonder if the new environment thing can be used? If not we will do that manually.

Let’s do this manually for now.

trappmartin:
I’m using runtest.jl to install packages only required for tests in MCMCChain. Adding Pkg.add to Travis didn’t work for me.

yebai:
> Adding Pkg.add to Travis didn’t work for me.

Hmm, strange. Any idea why?

yebai:
Seems the issue is not on the Turing side -- something weird is going on with `Stats.jl`.

mohamed82008:
> I’m using runtest.jl to install packages only required for tests in MCMCChain. Adding Pkg.add to Travis didn’t work for me.

The proper way to do this is using a `test/REQUIRE` file.

trappmartin:
> > I’m using runtest.jl to install packages only required for tests in MCMCChain. Adding Pkg.add to Travis didn’t work for me.
> 
> The proper way to do this is using a `test/REQUIRE` file.

Can you refer to unregistered packages in the REQUIRE file?

mohamed82008:
If the unregistered package has a `Project.toml`, you can use add its UUID to the `Manifest.toml` file. The best example is probably in this link https://discourse.julialang.org/t/how-to-install-non-registered-dependencies-in-julia-0-7-1-0/13971/27

trappmartin:
Cool, thanks @mohamed82008!

mohamed82008:
Oh but you would still have to install it manually actually, since that's what Fredrick did there. So maybe not worth the effort. If it is for testing, I think you can just use your approach in `runtests.jl`. The link above is to stop Pkg3 from complaining about unregistered real deps, not test deps AFAICT. So I guess your approach should be fine.

mohamed82008:
> Seems the issue is not on the Turing side -- something weird is going on with Stats.jl.

Seems related to the recent automatic capping https://discourse.julialang.org/t/package-compatibility-caps/15301. I think `Stats` was in the blind spot of the heuristic used: updated before July, does not use `Compat`, and has Julia 0.6 lower limit.

mohamed82008:
Closing and reopening again to trigger CI.

xukai92:
The vec_assume_mat.jl test fails (https://travis-ci.org/TuringLang/Turing.jl/jobs/431989506#L897-L949) on both the Linux Travis and AppVeyor (but is OK for macOS Travis).

Issue #552, opened 2018-09-19
Replace `Stan.Adapt + WarmUpManager + adaption code` with `DynamicHMC.AbstractTuner` 


yebai:
The `AbstractTuner` type in DynamicHMC looks like a sensible replacement for 
`Stan.Adapt + WarmUpManager + adaption code` 

https://github.com/tpapp/DynamicHMC.jl/blob/master/src/sampler.jl#L148

perhaps we can adapt this bit of code for our HMC sampler family? This would fix the following related issues:

- Better adaptation configuration interface https://github.com/TuringLang/Turing.jl/issues/491 
- Support pre-cond with full covariance https://github.com/TuringLang/Turing.jl/issues/279
- Robust adaption for NUTS https://github.com/TuringLang/Turing.jl/issues/324
- Remove Stan dependency on Turing https://github.com/TuringLang/Turing.jl/issues/447

yebai:
It appears that DynamicHMC is taking an opaque approach here: although there exists DualAveragingParameters type, there is no NUTS constructor that accept and pass through dual averaging parameters. NUTS is always using default adaption parameters for leapfrog stepsize. 

The default parameters should be good enough here. Why do you want to change adaption parameters? We almost always use the default values in Turing too..

xukai92:
Current targets:
- Move adaptation related parameters from `Sampler` to a separate type `Tuner`
- Change the interface of `sample` so that it takes both sampler and tuner to do the sampling
- Adapt the design of `Tuner` from DynamicHMC

xukai92:
Closed by https://github.com/TuringLang/Turing.jl/pull/597

PR #560, opened 2018-09-22
Reexport Stan and automatically download cmdstan when building Turing

mohamed82008:
This PR automatically downloads and uncompresses `cmdstan` when Turing is built if `cmdstan`'s path does not exist already in `ENV["CMDSTAN_HOME"]`. It also reexports `Stan`. The function `Turing.cmdstan_home()` returns the path of the internal `cmdstan`. I tried defining it in Stan, but I think it must be defined in Main using `set_cmdstan_home!(Turing.cmdstan_home())`.

We can control which version of `cmdstan` to automatically download by setting the url to the zip file in https://github.com/mohamed82008/Turing.jl/blob/cmdstan/deps/cmdstan_url.txt. On the downside, this will slow down Travis and AppVeyor since `cmdstan` will be downloaded every time. To work around this, I can add `ENV["CMDSTAN_HOME"] = "_"` in the AppVeyor and Travis scripts to avoid setting up `cmdstan` every time.

This script was going to go in `TuringBenchmarks` initially but then I thought `Turing.jl` might be a more suitable place for it. I tried to think of a good reason not to setup cmdstan automatically with fresh installs, but I couldn't find any hence this PR. Please let me know if you think otherwise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ docs/build/
 docs/site/
 docs/build/
 .DS_Store
+
+src/cmdstan_home.jl
+cmdstan/
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,7 +3,7 @@ using Pkg
 if !haskey(ENV, "CMDSTAN_HOME") || ENV["CMDSTAN_HOME"] == ""
     # Make the cmdstan home directory
 
-    CMDSTAN = joinpath(@__DIR__, "..", "cmdstan", )
+    CMDSTAN = abspath(joinpath(@__DIR__, "..", "cmdstan", ))
     if !ispath(CMDSTAN)
         mkdir(CMDSTAN)
     end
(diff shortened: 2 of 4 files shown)


mohamed82008:
So the build passes on Travis, but the tests don't use Stan directly. I will look into this for some more to make sure it is properly tested, so please don't merge.

yebai:
Hi @mohamed82008, thanks for this PR.

The reason that we are depending on Stan is for benchmarking purpose. There are some places in Turing where we use Stan.jl for adaption data types. But we have discussed to remove these code refactoring our adaption interface (see #552). So for the next release, we'd like actually get rid of Stan dependencies from the main Turing repo. Perhaps we can install Stan from TuringBenchmarks instead of Turing.jl?

mohamed82008:
Fair enough, I will add it there then.

yebai:
Thanks!

PR #561, opened 2018-09-22
Use StatsBase instead of Stats

nalimilan:
Stats is a meta-package which in its current form simply reexports StatsBase, 
and which does not support Julia 1.0 for now.

See https://github.com/JuliaStats/Stats.jl/pull/7.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,3 +1,3 @@
 UnicodePlots
-Stats
+StatsBase
 StatsFuns 0.7.0
--- a/test/geweke.jl
+++ b/test/geweke.jl
@@ -1,4 +1,4 @@
-using Stats, Turing
+using StatsBase, Turing
 using Gadfly
 
 # include("ASCIIPlot.jl")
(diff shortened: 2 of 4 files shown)


trappmartin:
Thanks! I think it would be good to add a version number of StatsBase to the REQUIRE file before merging. This will ensure that we depend on a StatsBase version that supports Julia 0.7 / 1.0.

nalimilan:
> Thanks! I think it would be good to add a version number of StatsBase to the REQUIRE file before merging. This will ensure that we depend on a StatsBase version that supports Julia 0.7 / 1.0.

I don't really see the point of doing that: the package manager won't allow installing a version of StatsBase which doesn't support the current Julia.

nalimilan:
> Specifying version ranges for dependencies is the officially recommended procedure.

Yes, but the version should reflect specific requirements for package features, not those on the Julia version, which are redundant. What would be interesting is to know which StatsBase version is really needed, but that would probably require testing on Julia 0.6, which is a bit pointless.

Issue #540, opened 2018-09-14
Refactoring: move `transform.jl` into a separate package: `Bijectors.jl`


yebai:
Perhaps we can separate out `utilities/transform.jl` and related tests into a standalone package, for several reasons:

- it can be shared by other packages (e.g. Mamba, DynamicHMC) 
- its tests are not very stable which causes Turing tests to fail

Related files

- https://github.com/TuringLang/Turing.jl/tree/master/test/transform.jl
- https://github.com/TuringLang/Turing.jl/blob/master/src/utilities/transform.jl

yebai:
Related: #542 #522 #368

PR #565, opened 2018-09-28
Depend on Bijectors

mohamed82008:
Deletes utilities/transform.jl and depends on Bijectors.jl. Requires https://github.com/TuringLang/Bijectors.jl/pull/2 to be merged first, then the Bijectors.jl package should be registered. Closes https://github.com/TuringLang/Turing.jl/issues/540.


yebai:
Registration PR for `Bijectors`: 

https://github.com/JuliaLang/METADATA.jl/pull/18473

We can merge this PR once the registration is accepted.

PR #568, opened 2018-09-30
Site Beautification

cpfiffer:
In order to unify the site's theme across the landing page and the documentation, I've migrated the entire site from a Frankensteined-together Jekyll/Documenter stack to a far more cohesive mostly-Jekyll stack. 

There is a running example of this site being served from my fork, found [here](http://cameron.pfiffer.org/Turing.jl/). Please take a moment to click around and let me know what works and what doesn't, though I would like to try to get this up before Tuesday morning (UK time). 

The changes are essentially cosmetic. Content changes have been exceptionally minimal, mostly additions of YAML front matter to documentation pages and some link updating here and there. A lot of the non-visual changes are the additions of some textual processing code to the `update-homepage.jl` and `make.jl` scripts designed to improve LaTeX handling and make all the markdown files play nicely with Documenter. Everything should be automatically buildable by Travis the same way it is now.

The biggest changes that everyone needs to know about are that the documentation links have changed to `turing.ml/docs/` and not `turing.ml/latest/` as in the previous. The tutorials now have their own section on the site under `turing.ml/tutorials/`. We can change that pretty easily, as I don't really have a strong opinion on it.

There's also now support for a blog or news feed if we'd like to start doing that, with a bit of effort.


yebai:
Hi Cameron, many thanks for the PR. it looks great!

Just wondering is it possible to:

1) tweak the `navigation` style (font size etc), perhaps in a way similar to the current Turing.ml/latest page; 
2) tweak the bottom `previous`/`next` button; perhaps in a way similar to the current Turing.ml/latest page; 
3) add a link "Edit on Github"

cc'ing @xukai92 @willtebbutt @wesselb @trappmartin Feedbacks are welcome!

xukai92:
Nice work! I really like the new design! I generally agree with Hong's comments. Additionally, I feel the right side bar is kind of "unfinished" because of the (probably) default icon on the top. Also there is no hovering highlighting on the left side bar - maybe we should have it?

cpfiffer:
I've added the Lato font (what the turing.ml/latest uses) and made the documentation navigator font bigger, as well as added link underlining/slight highlight on hover.

The next/previous buttons are now roughly equivalent to how they appeared previously, though there's an issue with what they're pointing to -- it doesn't seem to be doing it in quite the order I want. I'll have to look at how it's paginating things a bit later.

I've changed the default icon to something a bit better to indicate a table of contents.

These changes won't show up yet in the demo site, since I've changed most of the configuration settings to work for `turing.ml` and not my personal site, but we can tweak all that stuff over time to get it up to snuff.

PR #580, opened 2018-10-09
Remove a load of code and refactor what is still here

willtebbutt:
There appears to have been quite a bit of redundant code in [src/utilities/resample.jl](https://github.com/TuringLang/Turing.jl/blob/master/src/utilities/resample.jl). This PR removes a lot of that stuff. @yebai @xukai92 I could really use your input here to make sure that I've not removed something that really shouldn't have been removed.

### DIFF
--- a/src/utilities/resample.jl
+++ b/src/utilities/resample.jl
@@ -20,7 +20,7 @@ end
 function resampleSystematic(w::AbstractVector{<:Real}, N::Int)
 
     Q = cumsum(w)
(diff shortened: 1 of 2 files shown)


yebai:
@willtebbutt  The original reason for adding them is to support multiple `resample` methods for particle filters. These `resample` functions have different influences on the performance of particle filters. See [Douc](https://arxiv.org/pdf/cs/0507025.pdf) for example.

willtebbutt:
Aha, thanks. I'll change this PR significantly.

Issue #575, opened 2018-10-06
Add assertions to compiler


trappmartin:
We should add assertions to the compiler code to be able to throw useful errors in case of errors in the model definition.

- Add assertion if right-hand side of ~ is not a distribution or vector of distributions.
- ~Add assertion if right-hand side of = is a distribution.~ <- Note sure this would be a problem.

mohamed82008:
Typo: first `left-hand` -> `right-hand`.

trappmartin:
Closing because of PR #585

PR #585, opened 2018-10-12
Compiler assertions according to issue #575

trappmartin:
This PR implements some simple assertions to allow for meaningful erroring of the compiler if the right-hand side of a tilde operant is not a Distribution or a Vector of Distributions.

### DIFF
--- a/src/core/compiler.jl
+++ b/src/core/compiler.jl
@@ -27,6 +27,9 @@ function var_tuple(sym::Symbol, inds::Expr=:(()))
     return esc(:($(QuoteNode(sym)), $inds, $(QuoteNode(gensym()))))
 end
 
+
+wrong_dist_errormsg = "Right-hand side of a ~ must be subtype of Distribution or a vector of Distributions."
+
 """
     generate_observe(observation, distribution)
 
@@ -35,6 +38,14 @@ Generate an observe expression.
 function generate_observe(observation, distribution)
     return esc(
         quote
+            # Sanity check.
+            isdist = if isa($(distribution), Vector)
+                all(d -> isa(d, Distribution), $(distribution))
+            else
+                isa($(distribution), Distribution)
+            end
+            @assert isdist @error($(wrong_dist_errormsg))
+
             vi.logp += Turing.observe(
                 sampler,
                 $(distribution),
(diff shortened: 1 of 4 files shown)


src/core/compiler.jl
```diff
@@ -35,6 +38,14 @@ Generate an observe expression.
 function generate_observe(observation, distribution)
     return esc(
         quote
+            # Sanity check.
+            isdist = if isa($(distribution), Vector)
```
willtebbutt: It's not immediately obvious from the way that the `distribution` variable is named that it is allowed to be either a `Distribution` or `Vector` thereof. Could we please make this clearer in the docstring for each method that has this variable as an argument?

willtebbutt: Also, `Vector` or `AbstractVector`?

src/core/compiler.jl
```diff
@@ -83,6 +103,14 @@ function generate_assume(variable::Expr, distribution)
             indexing = mapfoldl(string, *, idcs, init = "")
             varname = Turing.VarName(vi, Symbol(csym_str), sym, indexing)
 
+            # Sanity check.
+            isdist = if isa($(distribution), Vector)
+                all(d -> isa(d, Distribution), $(distribution))
+            else
+                isa($(distribution), Distribution)
+            end
+            @assert isdist @error($(wrong_dist_errormsg))
```
willtebbutt: This same block of code is repeated several times. Could we please put it in a function?

test/compiler.jl/model_macro.jl
```diff
@@ -73,3 +73,17 @@ f01_mm = testmodel01()
...
+# Test for assertions.
+@model brokentestmodel(x1, x2) = begin
+    s ~ InverseGamma(2,3)
+    m ~ Normal(0,sqrt(s))
+
+    x1 ~ Normal(m, sqrt(s))
+    x2 ~ x1 + 2
+
+    return x1, x2
+end
```
willtebbutt: This test is nice. Could we please implement one that will check both of the `generate_assume` methods? I believe that this one will only trigger the code in `generate_observe`.

src/core/compiler.jl
trappmartin: Good point!

test/compiler.jl/model_macro.jl
trappmartin: Correct, I'll add another test.

yebai:
@trappmartin is this ready for merging now?

trappmartin:
I would say yes but there seems to be something breaking on the Documenter part. Not sure where this comes from.

cc: @cpfiffer

cpfiffer:
I think I know what's going on here, I'll fix it when I'm back at a computer.

cpfiffer:
All good. Should be ready to merge this over now.

trappmartin:
Thanks!

PR #587, opened 2018-10-14
Switch to nbviewer and fix broken link

devmotion:
I think it might be preferable to link to nbviewer.jupyter.org instead of Github (e.g. on my mobile browser Github displays only raw text files of notebooks). Moreover, I discovered a broken link in the documentation.


cpfiffer:
Looks good to me, thanks for catching the link!

PR #590, opened 2018-10-21
Some fixes for benchmarks and other stuff

mohamed82008:
This PR fixes some problems with Turing to make TuringBenchmarks work properly, and organizes other parts, e.g. use of Requires.

### DIFF
--- a/src/samplers/hmcda.jl
+++ b/src/samplers/hmcda.jl
@@ -52,9 +52,6 @@ function HMCDA(n_iters::Int, n_adapt::Int, delta::Float64, lambda::Float64, spac
     _space = isa(space, Symbol) ? Set([space]) : Set(space)
     return HMCDA(n_iters, n_adapt, delta, lambda, _space, 0)
 end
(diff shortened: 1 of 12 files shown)


mohamed82008:
If no objections merging tomorrow.

src/samplers/hmcda.jl
```diff
@@ -52,9 +52,6 @@ function HMCDA(n_iters::Int, n_adapt::Int, delta::Float64, lambda::Float64, spac
     _space = isa(space, Symbol) ? Set([space]) : Set(space)
     return HMCDA(n_iters, n_adapt, delta, lambda, _space, 0)
 end
-function HMCDA(n_iters::Int, n_adapt::Int, δ::Float64, λ::Float64, space::Set, gid::Int)
```
trappmartin: Why was this dropped?

mohamed82008: This constructor is available automatically by the type definition, so this was overwriting the definition of the automatic inner constructors.

Issue #374, opened 2017-11-13
[BNP] Q-class


emilemathieu:
Sample from Q-class: `rand` and `logpdf` for total mass and size biased sampling.

 - [x] Implement DP, PYP via stick breaking: [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/pyp.jl)
  - [x] Q-class total mass: `rand(ExpTiltedSigma(sigma, tau))`: [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/ngip.jl) 
  - [ ] Q-class total mass: `logpdf(ExpTiltedSigma(sigma, tau))`
  - [x] Q-class new weight with sigma=0.5: `rand` : [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/ngip.jl) 
  - [ ] Q-class new weight with sigma=0.5: `logpdf`
  - [ ] Q-class new weight with sigma<0.5: `rand` and `logpdf`

Draft here https://github.com/yebai/Turing.jl/tree/project-bnp/src/samplers/support/rpm

ae-foster:
For efficient sampling of exponentially tilted stable random variables, Devroye double rejection sampling sounds efficient: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.157.9416&rep=rep1&type=pdf (see page 21 for the actual algorithm)

mlomeli1:
The current code is based on Hoffert (2011)
http://delivery.acm.org/10.1145/2050000/2043638/a3-hofert.pdf?ip=131.111.5.148&id=2043638&acc=ACTIVE%20SERVICE&key=BF07A2EE685417C5%2E6CDC43D2A5950A53%2E4D4702B0C3E38B35%2E4D4702B0C3E38B35&CFID=1005798912&CFTOKEN=54131981&__acm__=1510664595_6a074af8f4a60fe0f0f69844e6a1389b
He refers to Devroye (2009) and compares against it. If you would like to know when Devroye is more suitable, please read the paper.

ae-foster:
Oh tx @mlomeli1 - I guess we're good to leave as is until we notice a bottleneck in the total mass sampler :)

mlomeli1:
[...]

Issue #370, opened 2017-11-12
[BNP] Bayesian Nonparametrics project


emilemathieu:
This is a macro issue to follow the development of Turing being able to handle nonparametric models.

- [x] Defining the class of models we aim to handle: Q-class cf https://github.com/yebai/Turing.jl/issues/374
  - [x] Implement DP, PYP via stick breaking: [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/pyp.jl)
  - [x] Q-class total mass: `rand(ExpTiltedSigma(sigma, tau))`: [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/ngip.jl) 
  - [ ] Q-class total mass: `logpdf(ExpTiltedSigma(sigma, tau))`
  - [x] Q-class new weight with sigma=0.5: `rand` : [Here](https://github.com/yebai/Turing.jl/blob/project-bnp/src/samplers/support/rpm/ngip.jl) 
  - [ ] Q-class new weight with sigma=0.5: `logpdf`
  - [ ] Q-class new weight with sigma<0.5: `rand` and `logpdf`

- [x] Choose & implement model / metrics for the tests cf https://github.com/yebai/Turing.jl/issues/371

- [ ] Infinite array data structure for weights and locations of the RPM
  - [x] Can use `tzeros(Float64, N)` with `N` being the number of datapoints for both locations and weights.
  - [ ] Can use `Dict{Int, Any}()` for locations ?
  - [ ] Need a specific data structure `InfiniteArray` ?
 
- [x] `SMC` working on those models which implies handling array of random length cf https://github.com/yebai/Turing.jl/issues/367#issue-271525324
  - [ ] `Chain` wrapper issue https://github.com/yebai/Turing.jl/issues/386
  - [ ] [cache issue for `getidcs`](https://github.com/TuringLang/Turing.jl/blob/project-bnp/src/core/varinfo.jl#L260)

- [x] `CSMC` working
  - [x] Bug: missing prior of conditioned random variables in particles-weight https://github.com/yebai/Turing.jl/issues/383
  - [x] Handling random variables sampled in random order cf https://github.com/yebai/Turing.jl/issues/367#issue-271525324.  Solutions proposed https://github.com/yebai/Turing.jl/issues/367#issuecomment-343270076
[...]

PR #591, opened 2018-10-22
BNP priors for random partitions

trappmartin:
This PR is a work in progress PR, integrating the existing codes of #370 and #374 for random partitions.

TODO:
- [x] Add tests for distributions.
- [x] Add SB, SBS, CRP codes for DP, PYP related models.
- [x] Modify `Turing.Chain` to work with missing values.
- [x] Add CRP example & test.
- [x] Add Stick-Breaking example & test.
- [x] Add SBS example & test.
- [x] Check if we sample from correct posterior, see: https://github.com/TuringLang/Turing.jl/blob/project-bnp/test/rpm.jl/pym_posterior.jl

Changes to code base:
* Added BNP priors

cc: @mlomeli1 , @emilemathieu


mohamed82008:
`@TuringBot benchmark`

mohamed82008:
No action taken (submission was from an edit or delete).

mohamed82008:
`@TuringBot benchmark`

mlomeli1:
> This PR is a work in progress PR, integrating the existing codes of #370 and #374 for random partitions.
> 
> **Do not merge!**
> 
> TODO:
> 
> * Add tests for distributions.
> * Add missing codes.
> * Modify `Turing.Chain` to work with missing values.
> * Add DP-MM example.
> 
> cc: @mlomeli1 , @emilemathieu

That's great @trappmartin . If you want to have a look at https://github.com/mlomeli1/SMC-MPhilproject/tree/master/MFM_for_Turing , there is some code for SMC for DPM as well. I believe you have access to this private repo. I also have my Phd Matlab code for the Q-class, let me know if you would like access to that repo if it is useful :)

trappmartin:
Thanks, @emilemathieu and @mlomeli1!
There are still some things breaking but I keep you up to date. 

I changed the implementation of BNP priors by separating the representation from the stochastic process. For example, a Pitman-Yor process can now be constructed as follows:

```julia
a = 0.5
θ = 0.1
t = 2

# stick-breaking representation
d = StickBreakingProcess(PitmanYorProcess(a, θ, t))

# size-biased sampling representation
surplus = 2.0
d = SizeBiasedSamplingProcess(PitmanYorProcess(a, θ, t), surplus)

# CRP representation
cluster_counts = [2, 1]
d = ChineseRestaurantProcess(PitmanYorProcess(a, θ, t), cluster_counts)
```

Let me know what you think of the new interface. I hope it's easier to use and allows us to have a more flexible interface for BNP priors.

Cheers,
Martin

emilemathieu:
I believe such an interface is way better !
The only question is how such processes are represented internally ? Thus how inference can be performed (eg SMC)?

trappmartin:
Thanks, I think it should not have much of an influence on the sampling process. I should see soon. :D

trappmartin:
Chinese Restaurant Process Example using current implementation:

```julia
@model infiniteMM(y; H = Normal(mean(y), std(y) * 2), rpm = DirichletProcess(0.1) ) = begin
    
    # Latent assignments.
    N = length(y)
    z = tzeros(Int, N)

    # Cluster counts.
    cluster_counts = tzeros(Int, N)

    # Cluster locations.
    x = tzeros(Float64, N)

    for i in 1:N

        # Draw assignments using a CRP.
        z[i] ~ ChineseRestaurantProcess(rpm, cluster_counts)
        if cluster_counts[z[i]] == 0
            # Cluster is new, therefore, draw new location.
            x[z[i]] ~ H
        end
        cluster_counts[z[i]] += 1

        # Draw observation.
        y[i] ~ Normal(x[z[i]], 0.5)
    end
    return z
end
```

emilemathieu:
Nice ! Do you think that interface allows to easily extends to NIGP and such ?

trappmartin:
Yes, I'm pretty certain this should be possible. For this PR the focus is only on DP and PYP but it totally makes sense to extend the code after merging this PR.

trappmartin:
Ups, I think I broke something. o.O

emilemathieu:
changing `data = vcat(rand(Normal(0, 0.5), 10), rand(Normal(8, 0.5), 10))` to `data = vcat(rand(Normal(0, 0.5), 10), rand(Normal(1, 0.5), 10))` yields an error.
I was pushing the cluster closer to try to get different value for the particles since they are all equals...

trappmartin:
The issue was related to the wrong handling of missing values. This is fixed now and I'll now add test to check if we sample from correct posterior.

trappmartin:
I added the test for the stick-breaking representation. However, I'm a bit unsure if my implementation is correct / necessary or if we can use the one by @emilemathieu. 

```julia
@model sbimm(y, rpm, trunc) = begin
    # Base distribution.
    H = Normal(mu_0, sigma_0)

    # Latent assignments.
    N = length(y)
    z = tzeros(Int, N)

    # Infinite collection of stick pieces and weights.
    v = tzeros(Float64, trunc)
    w = tzeros(Float64, trunc)
    K = 0

    # Cluster locations.
    x = tzeros(Float64, trunc)

    for i in 1:N

        # Draw a slice ∈ [0,1]
        u ~ Beta(1, 1)

        while (sum(w) < u) && (K < trunc)
            K += 1
            v[K] ~ StickBreakingProcess(rpm)
            x[K] ~ H
            w[K] = v[K] * prod(1 .- v[1:(K-1)])
        end

        w_ = w[1:K] / sum(w[1:K])
        z[i] ~ Categorical(w_)

        # Draw observation.
        y[i] ~ Normal(x[z[i]], sigma_1)
    end
end
```

@emilemathieu and @yebai what are your thoughts?

emilemathieu:
Hi Martin! This seems to be a valid implementation of a stick breaking process :)
PS: yet it is inefficient as we argue in our workshop paper

trappmartin:
@cpfiffer the tests of this PR seem to be broken due to some bug in displaying MCMCChains. Can you have a look?

yebai:
> @cpfiffer the tests of this PR seem to be broken due to some bug in displaying MCMCChains. Can you have a look?

It's solved on the master branch; you need to rebase master into this PR.

trappmartin:
@emilemathieu I did some minor adjustment on your code for the SBS. Could you let me know if this is correct for simulation based sampling. I'm still not too familiar with the SBS and probably should read the paper again. :) 

Thanks!

```Julia
@model sbsimm(y,rpm) = begin
    # Base distribution.
    H = Normal(mu_0, sigma_0)

    # Latent assignments.
    N = length(y)
    z = tzeros(Int, N)

    x = tzeros(Float64, N)
    J = tzeros(Float64, N)
    z = tzeros(Int, N)

    k = 0
    surplus = 1

    for i in 1:N
        ps = vcat(J[1:k], surplus)
        z[i] ~ Categorical(ps)
        if z[i] > k
            k = k + 1
            J[k] ~ SizeBiasedSamplingProcess(rpm, surplus)
            x[k] ~ H
            surplus -= J[k]
        end
        y[i] ~ Normal(x[z[i]], sigma_1)
    end
end
```

trappmartin:
@yebai once the SBS sampling example is correct, this PR is ready for review and merging.

trappmartin:
This PR is ready to be merged from my side.

test/rpm.jl/sb.jl
```diff
@@ -0,0 +1,100 @@
...
+        # Instantiate new cluster.
+        while (sum(w) < u[i]) && (K < trunc)
+            K += 1
+            v[K] ~ StickBreakingProcess(rpm)
+            x[K] ~ H
+            w[K] = v[K] * prod(1 .- v[1:(K-1)])
+        end
+
+        # Find truncation point
+        K_ = findfirst(u[i] .< cumsum(w))
```
yebai: This seems like a non-typical slice sampler for DPs. Do you have a reference for this `slice` sampling representation?

trappmartin: You are right. After looking at it again, it seems rather odd and is probably not quite correct. I can change it to the retrospective sampler by Papaspiliopoulos and Roberts which seems straight forward in Turing to me. Or do you have a preference for another one, e.g. Walker et al.?

yebai: I don't have a preference for this; the method in Papaspiliopoulos and Roberts sounds good to me.

yebai:
It's probably good to have a dedicated folder for customised distributions in Turing, e.g. `src/distrs/`. If so, we can consider placing the main BNP module in `src/distrs/RandomMeasures.jl`.

src/randomMeasures/RandomMeasures.jl
```diff
@@ -0,0 +1,14 @@
...
+using ..Core, ..Core.VarReplay, ..Utilities
+using Distributions
+using LinearAlgebra
+using StatsFuns: logsumexp
+
+import Distributions: sample, logpdf
+import Base: maximum, minimum, rand
+
+# include concrete implementations
+include("rpm.jl")
```
yebai: it's ok to directly include the contents of `rpm.jl` here since it's not a huge amount of code. It also makes future maintenance easier since it will only involve one file.

trappmartin: Hm, maybe we can move the generic bits here and move the implementation related to random partitions into a file `randomPartitions.jl` or so. This way it would be easier to extend the code later on.

test/rpm.jl/sb.jl
trappmartin: Sounds good to me. I already started a BNP tutorial and will put the Papaspiliopoulos and Roberts code in there. With basic recursive stick breaking you mean a truncated implementation with fixed truncation point, right?

src/randomMeasures/RandomMeasures.jl
yebai: I don't have a strong preference here - but have seen many Julia basic libraries contained in a single file separated by sections. I think that makes sense cause have several small files doesn't add any flexibility, right?

trappmartin: Fair enough.

test/rpm.jl/sb.jl
yebai: I mean Alg 1 from the following paper: 

http://www.robots.ox.ac.uk/~twgr/assets/pdf/bloemreddy2017rpm.pdf

This version doesn't involve any truncation through the use of a random coin-flip based termination criterion.  I still need to read the original paper to have a better understanding of why this is equivalent to the standard stick-breaking, but my guess is that the expectation of the process converges to the standard stick-breaking process.

trappmartin: I see. Yes, it looks like it does. I'll read the paper more carefully again as I forgot about the recursive coin-flipping.

trappmartin: Algo 1 (coin-flipping based) is complicated to implement in Turing because of the variable name issue we have, i.e. the coin will not be resampled recursively because of: https://github.com/TuringLang/Turing.jl/blob/6c0f4b8e9cecc42d098b788f9dd7063da8ebcd3d/src/inference/Inference.jl#L114-L116  I'll keep the truncated stick-breaking test for now.

src/Turing.jl
```diff
@@ -128,6 +128,9 @@ end
     include("inference/dynamichmc.jl")
 end
 
+# Random probability measures.
+include("distributions/randomMeasures.jl")
```
yebai: Perhaps a capital `RandomMeasures.jl`, since it's a module?

yebai:
Excellent work - Ready to merge except one minor filename issue (see above).

test/rpm.jl/sb.jl
yebai: I see, perhaps move this into a separate issue?

This should be fixed soon - see related discussion https://github.com/TuringLang/Turing.jl/pull/720#pullrequestreview-216110025.

PR #595, opened 2018-11-11
Stan -> CmdStan

mohamed82008:
Stan.jl has been deprecated in favour of CmdStan.jl. This PR uses CmdStan.jl when available, much like how we used Stan.jl.


mohamed82008:
Changes are minimal, and tests pass so merging.

PR #610, opened 2018-11-28
Update REQUIRE

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/19695 failed because it still thinks it should be running on v0.7. This updates the REQUIRE to be v1.0+. The tag should be deleted and re-done with this change.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.7
+julia 1.0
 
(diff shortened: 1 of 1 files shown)


yebai:
@ChrisRackauckas thanks, I've retagged v0.5.2.

PR #611, opened 2018-11-28
Change default AD backend to ForwardDiff #608

yebai:


mohamed82008:
One error somewhat reduced to:
```julia
using ForwardDiff, Distributions, Turing

D = 2
@model vdemo2(x) = begin
  μ ~ MvNormal(zeros(D), ones(D))
  x ~ [MvNormal(μ, ones(D))]
end
vdemo2_model = vdemo2(randn(D,1000))

m = ForwardDiff.Dual{ForwardDiff.Tag{getfield(Turing, Symbol("#f#92")){Turing.VarReplay.VarInfo,typeof(vdemo2_model),Turing.Sampler{HMC{Any}}},Real},Real,2}.(rand(3));
s = deepcopy(m)

MvNormal(m, s)
```
It seems that `Distributions` can't handle all `Real` types, or somewhere along the line a fall back to `Float64` is happening. Investigating further...

mohamed82008:
Further reduced to:
```julia
using PDMats
PDMats.PDiagMat(s)
```
The fallback to `Float64` for the second parameter in `Dual` is happening because `typeof(inv(s::Dual{T, Real, 1})) == Dual{T, Float64, 1}`, when `s.value isa Float64`. This is intentional in `ForwardDiff` to handle cases where the input type is different from the output, e.g. `inv(::Int)::Float64`. The `inv` function is called in https://github.com/JuliaStats/PDMats.jl/blob/eeec1c1bfdddb8f522952f81b3f6a3d5ec8aa7bd/src/pdiagmat.jl#L18. 

It seems that the easiest fix is to change `AbstractVector{T}` to `AbstractVector` in the arguments to the `PDiagMat` constructor in https://github.com/JuliaStats/PDMats.jl/blob/master/src/pdiagmat.jl#L9.  This fix will promote the inputs to the constructor thus getting back `Real` in the second argument in `Dual`. But the problem is that it will allocate a new vector every time we call the constructor. Perhaps a better fix is to see if `Real` can be replaced entirely or not in the second argument of `Dual` in `Turing`. I will make a PR in `PDMats` for now.

mohamed82008:
https://github.com/JuliaStats/PDMats.jl/pull/86

mohamed82008:
Second error reduced to:
```julia
using Turing, ForwardDiff, LinearAlgebra
using Turing.VarReplay: VarInfo

@model hmcmatrixsup() = begin
  v ~ Wishart(7, [1 0.5; 0.5 1])
  v
end
model_f = hmcmatrixsup()

T = ForwardDiff.Tag{getfield(Turing, Symbol("#f#92")){VarInfo,typeof(hmcmatrixsup_model),Turing.Sampler{HMC{Any}}},Real}
TD = ForwardDiff.Dual{T}
TM = Array{ForwardDiff.Dual{T, Real, 4}, 2}

X11 = TD(2.0874877441826407,1,0,0,0)
X12 = TD(0.0,0,0,1,0)     
X21 = TD(0.42908063526362944,0,1,0,0)
X22 = TD(0.6440131884557554,0,0,0,1)
X = [X11 X12;
     X21 X22]
X = TM(X) # works without this line

LowerTriangular(X) * LowerTriangular(X)'
```

May turn out to be a bug in `LinearAlgebra`. The cause of the error here is `Real` as the second parameter of `Dual`. When it is `Float64`, it works fine.

mohamed82008:
The root cause of the third error I am getting is from `eps(Real)`. This is used in the new Bijectors.jl for handling numerical edge cases. The same error happens when doing `eps(ForwardDiff.Dual{Any, Real, 1})`. The fastest fix for this, is to define `_eps` in `Bijectors.jl` which falls back on `Float64` for `Real` and `ForwardDiff.Dual{<:Any, Real, N} where N` using Requires.jl for the latter.

mohamed82008:
So I believe this PR should be passing tests after the `PDMats` PR gets merged.

PR #612, opened 2018-12-02
Load Flux and ForwardDiff before Bijectors

mohamed82008:
This PR prepares for Bijectors using Requires.jl to define `_eps` function as explained in  https://github.com/TuringLang/Turing.jl/pull/611.


mohamed82008:
https://github.com/TuringLang/Bijectors.jl/pull/11

PR #613, opened 2018-12-02
Compiler refactor 2.0

### LINKED ISSUE #544: Running model without data returns error
```julia
using Turing

@model gdemo(x) = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  x[1] ~ Normal(m, sqrt(s))
  x[2] ~ Normal(m, sqrt(s))
  return s, m
end

g = gdemo([1.5, 2]) # This works
g()           # returns: (0.8660317849113424, 0.820409281237518)

g = gdemo() # Returns an error.
g()
```

Here is the error message:
```julia
julia> g = gdemo()
┌ Warning: Data `x` not provided, treating as parameter instead.
└ @ Main ~/.julia/dev/Turing/src/core/compiler.jl:326
[ Info:  Assume - `s` is a parameter
[ Info:  Assume - `m` is a parameter
[ Info:  Assume - `x` is a parameter
gdemo_model (generic function with 4 methods)

julia> g()
ERROR: UndefVarError: x not defined
Stacktrace:
 [1] macro expansion at /Users/hg344/.julia/dev/Turing/src/core/compiler.jl:78 [inlined]
 [2] gdemo_model(::Turing.VarReplay.VarInfo, ::Turing.SampleFromPrior) at /Users/hg344/.julia/packages/MacroTools/4AjBS/src/utils.jl:297 (repeats 2 times)
 [3] top-level scope at none:0
```

@trappmartin

mohamed82008:
# Design summary and remaining tasks

Here is a summary of the new design of the compiler. Please ask questions if something is not clear. As of now, all tests pass.

## Model definition syntax
```julia
@model model_function(x = default_x, y)
	...	
end
```
## Model variables

`args`: `x`, `y`
`dvars`: variables in `args`, not equal to `nothing`, and used in LHS of `~` 
`pvars`: variables used in LHS of `~` and not in `dvars`

When no value is given for `x`, it is treated as a parameter and then initialized to `default_x`.

## Expanded model definition

```julia
model_function(; x = nothing, y = nothing) = model_function(x, y)
function model_function(x = nothing, y = nothing)
    # Get the pvars and dvars from types of x and y in the form of Tuple{:x, :y}.
    # If either is `nothing`, it is not part of the tuple type. This information is available at compile time.
    pvars_tupletype, dvars_tupletype = Turing.get_vars($(Tuple{pvars...}), (x = x, y = y))

    # Get the `data` NamedTuple including only the vars in `dvars_tupletype`, i.e. not equal to `nothing`.
    # This information is available at compile time
    data = Turing.get_data(dvars_tupletype, (x = x, y = y))

    # Define closure signatures.
    closure(sampler::Turing.AnySampler, model) = closure(model)
    closure(model) = closure(Turing.VarInfo(), Turing.SampleFromPrior(), model)
    closure(vi::Turing.VarInfo, model) = closure(vi, Turing.SampleFromPrior(), model)
    function closure(vi::Turing.VarInfo, sampler::Turing.AnySampler, model)
        # Unwrap the `data` NamedTuple to avoid closures. 
        # These if statements are compiled away by the Julia compiler so this will be type stable.
        local x
        if isdefined(model.data, :x)
            x = model.data.x
        else
            x = default_x
        end
        local y
        if isdefined(model.data, :y)
            y = model.data.y
        else
            y = nothing
        end
        vi.logp = zero(Real)
        ...
    end
    
    # Define callable model    
    model = Turing.CallableModel{pvars_tupletype, dvars_tupletype}(closure, data)

    return model
end 
```

## Model

```julia
struct Model{pvars, dvars, F, TD}
    f::F
    data::TD
end
function Model{pvars, dvars}(f::F, data::TD) where {pvars, dvars, F, TD}
    return Model{pvars, dvars, F, TD}(f, data)
end
(m::Model)(args...; kwargs...) = m.f(args..., m; kwargs...)
```

When `m::Model` is called, it simply passes all its `args` and `kwargs` to `m.f` which is the closure defined above, in addition to the model `m` itself which gets passed as an argument. The model `m` is passed as an argument to give `m.f` access to `pvars`, `dvars` and `data`. Variables in `dvars` are [observed](https://github.com/TuringLang/Turing.jl/blob/mt/compiler2.0/src/core/compiler.jl#L76) and variables in `pvars` are [assumed](https://github.com/TuringLang/Turing.jl/blob/mt/compiler2.0/src/core/compiler.jl#L100). This is checked at [compile time](https://github.com/TuringLang/Turing.jl/blob/mt/compiler2.0/src/core/compiler.jl#L176) so the checking statement is optimized away by the Julia compiler. Note that `args` is known when parsing the model, but `pvars` and `dvars` can only be known once we know which of `x` and `y` are defined and which are defaulted to `nothing`. That is only when we know the input types to `model_function`, that we can deduce `pvars` and `dvars` and they never change within the scope of `model_function` or during the lifetime of `closure`.

Finally, any of `x` and `y` that is not `nothing` is put in a `data::NamedTuple` which is a field in the `Model` struct. This then gets unwrapped inside the `closure` body. This avoids the need for closing over `x` and `y`.

This PR also lets us generate the model using keyword arguments as follows: `model_function(y = 1)`, where the unassigned variables are treated as parameters. This lets us define variables in any order or leave them undefined thereof.

- [x] Make `pvars` and `dvars` type information
- [x] Make `data` a field in `CallableModel` and pass it to the closure.
- [x] Fix #544
- [x] Make use of keyword arguments or named tuples in `outer_function` as opposed to simple positional arguments.
- [x] Clean up the compiler, make the variable names more descriptive, and add comments
- [x] Update design above
- [x] Rename `CallableModel` to `Model`
- [x] Use `gensym` for `vi`, `sampler` and `logp`

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -30,7 +30,7 @@ getADtype(::Type{<:Hamiltonian{AD}}) where {AD} = AD
 gradient(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
 
@@ -40,7 +40,7 @@ Computes the gradient of the log joint of `θ` for the model specified by
 @generated function gradient(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
     sampler::TS=nothing,
 ) where {TS <: Union{Nothing, Sampler}}
     if TS == Nothing
@@ -68,7 +68,7 @@ end
 gradient_forward(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
     spl::Union{Nothing, Sampler}=nothing,
     chunk_size::Int=CHUNKSIZE[],
 )
@@ -111,7 +111,7 @@ end
 gradient_reverse(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
 
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -79,7 +79,7 @@ using forwards-mode AD from ForwardDiff.jl.
 function gradient_forward(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
     sampler::Union{Nothing, Sampler}=nothing,
     ::Val{chunk_size}=Val(CHUNKSIZE[]),
 ) where chunk_size
@@ -121,7 +121,7 @@ Computes the gradient of the log joint of `θ` for the model specified by
 function gradient_reverse(
     θ::AbstractVector{<:Real},
     vi::Turing.VarInfo,
-    model::Function,
+    model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
     vals_old, logp_old = copy(vi.vals), copy(vi.logp)
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -30,7 +30,7 @@ getADtype(::Type{<:Hamiltonian{AD}}) where {AD} = AD
 gradient(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model,
+    model::Model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
 
@@ -40,7 +40,7 @@ Computes the gradient of the log joint of `θ` for the model specified by
 @generated function gradient(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model,
+    model::Model,
     sampler::TS=nothing,
 ) where {TS <: Union{Nothing, Sampler}}
     if TS == Nothing
@@ -68,7 +68,7 @@ end
 gradient_forward(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model,
+    model::Model,
     spl::Union{Nothing, Sampler}=nothing,
     chunk_size::Int=CHUNKSIZE[],
 )
@@ -79,7 +79,7 @@ using forwards-mode AD from ForwardDiff.jl.
 function gradient_forward(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model,
+    model::Model,
     sampler::Union{Nothing, Sampler}=nothing,
     ::Val{chunk_size}=Val(CHUNKSIZE[]),
 ) where chunk_size
@@ -111,7 +111,7 @@ end
 gradient_reverse(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model,
+    model::Model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
 
@@ -121,7 +121,7 @@ Computes the gradient of the log joint of `θ` for the model specified by
 function gradient_reverse(
     θ::AbstractVector{<:Real},
     vi::Turing.VarInfo,
-    model,
+    model::Model,
     sampler::Union{Nothing, Sampler}=nothing,
 )
     vals_old, logp_old = copy(vi.vals), copy(vi.logp)
--- a/src/core/compiler.jl
+++ b/src/core/compiler.jl
@@ -4,11 +4,31 @@ using Base.Meta: parse
 # Overload of ~ #
 #################
 
-struct CallableModel{F}
+struct CallableModel{pvars, dvars, F}
     f::F
-    pvars::Set{Symbol}
-    dvars::Set{Symbol}
 end
+function CallableModel{pvars, dvars}(f::F) where {pvars, dvars, F}
+    return CallableModel{pvars, dvars, F}(f)
+end
+pvars(m::CallableModel{params}) where {params} = Tuple(params.types)
+function dvars(m::CallableModel{params, data}) where {params, data}
+    return Tuple(data.types)
+end
+@generated function inpvars(::Val{sym}, ::CallableModel{params}) where {sym, params}
+    if sym in params.types
+        return :(true)
+    else
+        return :(false)
+    end
+end
+@generated function indvars(::Val{sym}, ::CallableModel{params, data}) where {sym, params, data}
+    if sym in data.types
+        return :(true)
+    else
+        return :(false)
+    end
+end
+
 (m::CallableModel)(args...; kwargs...) = m.f(args..., m; kwargs...)
 
 # TODO: Replace this macro, see issue #514
@@ -122,7 +142,7 @@ function _tilde(vsym, left, dist, model_info)
         end
 
         return quote 
-            if $(QuoteNode(vsym)) ∈ model.dvars
+            if Turing.indvars($(Val(vsym)), model)
                 $(generate_observe(left, dist))
             else
                 $(generate_assume(left, dist, model_info))
@@ -195,8 +215,8 @@ macro model(fexpr)
         :closure_name => gensym(),
         :args => modeldef[:args],
         :kwargs => modeldef[:kwargs],
-        :dvars => Set{Symbol}(),
-        :pvars => Set{Symbol}()
+        :dvars => Symbol[],
+        :pvars => Symbol[]
     )
     # Unwrap ~ expressions and extract dvars and pvars into `model_info`
     fexpr = translate(fexpr, model_info)
@@ -220,6 +240,7 @@ macro model(fexpr)
     dvars_nt = :($([:($var = $var) for var in dvars]...),)
     return esc(quote
         function $(outer_function_name)($(fargs...))
+            pvars, dvars = Turing.get_vars($(Tuple{pvars...}), $dvars_nt)
             $closure_name(sampler::Turing.AnySampler, model) = $closure_name(model)
             $closure_name(model) = $closure_name(Turing.VarInfo(), Turing.SampleFromPrior(), model)
             $closure_name(vi::Turing.VarInfo, model) = $closure_name(vi, Turing.SampleFromPrior(), model)
@@ -227,20 +248,24 @@ macro model(fexpr)
                 vi.logp = zero(Real)
                 $closure_main_body
             end
-            model = Turing.CallableModel($closure_name, $pvars, $dvars)
-            Turing.update_vars!(model, $dvars_nt)
+            model = Turing.CallableModel{pvars, dvars}($closure_name)
             return model
         end
     end)
 end
-function update_vars!(model, dvars_nt)
-    for (sym, val) in pairs(dvars_nt)
-        if val == nothing
-            pop!(model.dvars, sym)
-            push!(model.pvars, sym)
-        end
-    end
-    return
+
+@generated function get_vars(pvars::Type{<:Tuple}, dvars::Tuple{})
+    return :(pvars, $(dvars))
+end
+@generated function get_vars(pvars::Type{Tpvars}, dvars_nt::NamedTuple) where {Tpvars <: Tuple}
+    pvar_syms = [Tpvars.types...]
+    dvar_syms = [dvars_nt.names...]
+    dvar_types = [dvars_nt.types...]
+    append!(pvar_syms, [dvar_syms[i] for i in 1:length(dvar_syms) if dvar_types[i] == Nothing])
+    setdiff!(dvar_syms, pvar_syms)    
+    pvars = Tuple{pvar_syms...}
+    dvars = Tuple{dvar_syms...}
+    return :($pvars, $dvars)
 end
 
 function warn_empty(body)
--- a/src/core/compiler.jl
+++ b/src/core/compiler.jl
@@ -4,6 +4,13 @@ using Base.Meta: parse
 # Overload of ~ #
 #################
 
+struct CallableModel{F}
+    f::F
+    pvars::Set{Symbol}
+    dvars::Set{Symbol}
+end
+(m::CallableModel)(args...; kwargs...) = m.f(args...; kwargs...)
+
 # TODO: Replace this macro, see issue #514
 """
 Usage: @VarName x[1,2][1+5][45][3]
@@ -37,8 +44,7 @@ Generate an observe expression for observation `observation` drawn from
 a distribution or a vector of distributions (`distribution`).
 """
 function generate_observe(observation, distribution)
-    return esc(
-        quote
+    return quote
             isdist = if isa($(distribution), AbstractVector)
                 # Check if the right-hand side is a vector of distributions.
                 all(d -> isa(d, Distribution), $(distribution))
@@ -55,19 +61,20 @@ function generate_observe(observation, distribution)
                 vi
             )
         end
-    )
 end
 
 """
-    generate_assume(variable, distribution, syms)
+    generate_assume(variable, distribution, compiler)
 
 Generate an assume expression for parameters `variable` drawn from 
 a distribution or a vector of distributions (`distribution`).
 
 """
-function generate_assume(variable, distribution, syms)
-    return esc(
-        quote
+function generate_assume(variable, distribution, compiler)
+    sym, idcs, csym = @VarName(variable)
+    csym = Symbol(compiler[:name], csym)
+    syms = Symbol[csym, variable]
+    return quote
             varname = Turing.VarName(vi, $syms, "")
 
             isdist = if isa($(distribution), AbstractVector)
@@ -97,17 +104,14 @@ function generate_assume(variable, distribution, syms)
             end
             vi.logp += _lp
         end
-    )
 end
 
-function generate_assume(variable::Expr, distribution)
-    return esc(
-        quote
-            sym, idcs, csym = @VarName $variable
-            csym_str = string(Turing._compiler_[:name])*string(csym)
-            indexing = mapfoldl(string, *, idcs, init = "")
-            varname = Turing.VarName(vi, Symbol(csym_str), sym, indexing)
-
+function generate_assume(variable::Expr, distribution, compiler)
+    sym, idcs, csym = @VarName variable
+    csym_str = string(compiler[:name])*string(csym)
+    indexing = mapfoldl(string, *, idcs, init = "")
+    return quote
+            varname = Turing.VarName(vi, Symbol($csym_str), $sym, $indexing)
             # Sanity check.
             isdist = if isa($(distribution), Vector)
                 all(d -> isa(d, Distribution), $(distribution))
@@ -124,7 +128,6 @@ function generate_assume(variable::Expr, distribution)
             )
             vi.logp += _lp
         end
-    )
 end
 
 """
@@ -140,67 +143,85 @@ Example:
 @~ x Normal()
 ```
 """
-macro ~(left, right)
-    return tilde(left, right)
+#macro ~(left, right)
+#    return tilde(left, right)
+#end
+
+function tilde(left::Number, right, compiler)
+    generate_observe(left, right)
 end
 
-function tilde(left, right)
-    return generate_observe(left, right)
+function tilde(left, right, compiler)
+    return quote
+        println($left)
+        if $left isa Nothing
+            $(generate_assume(left, right, compiler))
+        else
+            $(generate_observe(left, right))
+        end
+    end
 end
 
-function tilde(left::Symbol, right)
-
+function tilde(left::Symbol, right, compiler)
     # Check if left-hand side is a observation.
-    if left in Turing._compiler_[:args]
-        if !(left in Turing._compiler_[:dvars])
+    if left in compiler[:args]
+        if !(left in compiler[:dvars])
             @debug " Observe - `$(left)` is an observation"
-            push!(Turing._compiler_[:dvars], left)
+            push!(compiler[:dvars], left)
         end
 
-        return generate_observe(left, right)
+        return quote 
+            if $left isa Nothing
+                $(generate_assume(left, right, compiler))
+            else
+                $(generate_observe(left, right))
+            end
+        end
     else
         # Assume it is a parameter.
-        if !(left in Turing._compiler_[:pvars])
+        if !(left in compiler[:pvars])
             msg = " Assume - `$(left)` is a parameter"
             if isdefined(Main, left)
                 msg  *= " (ignoring `$(left)` found in global scope)"
             end
 
             @debug msg
-            push!(Turing._compiler_[:pvars], left)
+            push!(compiler[:pvars], left)
         end
 
-        sym, idcs, csym = @VarName(left)
-        csym = Symbol(Turing._compiler_[:name], csym)
-        syms = Symbol[csym, left]
-
-        return generate_assume(left, right, syms)
+        return generate_assume(left, right, compiler)
     end
 end
 
-function tilde(left::Expr, right)
+function tilde(left::Expr, right, compiler)
     vsym = getvsym(left)
     @assert isa(vsym, Symbol)
 
-    if vsym in Turing._compiler_[:args]
-        if !(vsym in Turing._compiler_[:dvars])
+    if vsym in compiler[:args]
+        if !(vsym in compiler[:dvars])
             @debug " Observe - `$(vsym)` is an observation"
-            push!(Turing._compiler_[:dvars], vsym)
+            push!(compiler[:dvars], vsym)
         end
 
-        return generate_observe(left, right)
+        return quote 
+            if $vsym isa Nothing
+                $(generate_assume(left, right, compiler))
+            else
+                $(generate_observe(left, right))
+            end
+        end
     else
-        if !(vsym in Turing._compiler_[:pvars])
+        if !(vsym in compiler[:pvars])
             msg = " Assume - `$(vsym)` is a parameter"
             if isdefined(Main, vsym)
                 msg  *= " (ignoring `$(vsym)` found in global scope)"
             end
 
             @debug msg
-            push!(Turing._compiler_[:pvars], vsym)
+            push!(compiler[:pvars], vsym)
         end
 
-        return generate_assume(left, right)
+        return generate_assume(left, right, compiler)
     end
 end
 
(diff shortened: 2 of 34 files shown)


src/samplers/hmc.jl
```diff
@@ -47,7 +47,7 @@ function HMC(alg::HMC, new_gid::Int)
 end
 HMC{T}(alg::HMC, new_gid::Int) where {T} = HMC(alg, new_gid)
 
-function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::HMC, momentum_sampler::Function;
+function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::HMC, momentum_sampler;
```
trappmartin: Why is `momentum_sampler` not a function anymore? Changes on the compiler should not effect this, or?

src/samplers/hmc.jl
```diff
@@ -78,7 +78,7 @@ Sampler(alg::Hamiltonian, adapt_conf::DEFAULT_ADAPT_CONF_TYPE) = begin
     Sampler(alg, info)
 end
 
-function sample(model::Function, alg::Hamiltonian;
+function sample(model, alg::Hamiltonian;
```
trappmartin: `model::CallableModel`?

src/samplers/hmc.jl
```diff
@@ -115,7 +115,7 @@ function sample(model::Function, alg::Hamiltonian;
 
     vi = if resume_from == nothing
         vi_ = VarInfo()
-        Base.invokelatest(model, vi_, HamiltonianRobustInit())
+        model(vi_, HamiltonianRobustInit())
```
trappmartin: 👍

src/samplers/hmcda.jl
```diff
@@ -53,7 +53,7 @@ function HMCDA(n_iters::Int, n_adapts::Int, delta::Float64, lambda::Float64, spa
     return HMCDA(n_iters, n_adapts, delta, lambda, _space, 0)
 end
 
-function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::HMCDA, momentum_sampler::Function;
+function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::HMCDA, momentum_sampler;
```
trappmartin: Why does changes on the compiler code change the `momentum_sampler`? I guess this should not be the case.

src/samplers/ipmcmc.jl
```diff
@@ -64,7 +64,7 @@ function Sampler(alg::IPMCMC)
   Sampler(alg, info)
 end
 
-step(model::Function, spl::Sampler{<:IPMCMC}, VarInfos::Array{VarInfo}, is_first::Bool) = begin
+step(model, spl::Sampler{<:IPMCMC}, VarInfos::Array{VarInfo}, is_first::Bool) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/ipmcmc.jl
```diff
@@ -94,7 +94,7 @@ step(model::Function, spl::Sampler{<:IPMCMC}, VarInfos::Array{VarInfo}, is_first
   VarInfos[nodes_permutation]
 end
 
-sample(model::Function, alg::IPMCMC) = begin
+sample(model, alg::IPMCMC) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/is.jl
```diff
@@ -35,7 +35,7 @@ Sampler(alg::IS) = begin
     Sampler(alg, info)
 end
 
-sample(model::Function, alg::IS) = begin
+sample(model, alg::IS) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/mh.jl
```diff
@@ -68,18 +68,18 @@ Sampler(alg::MH) = begin
   Sampler(alg, info)
 end
 
-propose(model::Function, spl::Sampler{<:MH}, vi::VarInfo) = begin
+propose(model, spl::Sampler{<:MH}, vi::VarInfo) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/mh.jl
```diff
@@ -68,18 +68,18 @@ Sampler(alg::MH) = begin
...
   spl.info[:proposal_ratio] = 0.0
   spl.info[:prior_prob] = 0.0
   spl.info[:violating_support] = false
   runmodel!(model, vi ,spl)
 end
 
-function step(model::Function, spl::Sampler{<:MH}, vi::VarInfo, is_first::Val{true})
+function step(model, spl::Sampler{<:MH}, vi::VarInfo, is_first::Val{true})
```
trappmartin: `model::CallableModel`?

src/samplers/mh.jl
```diff
@@ -68,18 +68,18 @@ Sampler(alg::MH) = begin
...
   return vi, true
 end
 
-function step(model::Function, spl::Sampler{<:MH}, vi::VarInfo, is_first::Val{false})
+function step(model, spl::Sampler{<:MH}, vi::VarInfo, is_first::Val{false})
```
trappmartin: `model::CallableModel`?

src/samplers/mh.jl
```diff
@@ -104,7 +104,7 @@ function step(model::Function, spl::Sampler{<:MH}, vi::VarInfo, is_first::Val{fa
   return vi, is_accept
 end
 
-function sample(model::Function, alg::MH;
+function sample(model, alg::MH;
```
trappmartin: `model::CallableModel`?

src/samplers/nuts.jl
```diff
@@ -43,7 +43,7 @@ function NUTS(n_iters::Int, delta::Float64)
 end
 NUTS(alg::NUTS, new_gid::Int) = NUTS(alg.n_iters, alg.n_adapts, alg.delta, alg.space, new_gid)
 
-function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::NUTS, momentum_sampler::Function;
+function hmc_step(θ, lj, lj_func, grad_func, H_func, ϵ, alg::NUTS, momentum_sampler;
```
trappmartin: Why does this change?

src/samplers/pgibbs.jl
```diff
@@ -47,9 +47,9 @@ Sampler(alg::PG) = begin
   Sampler(alg, info)
 end
 
-step(model::Function, spl::Sampler{<:PG}, vi::VarInfo, _) = step(model, spl, vi)
+step(model, spl::Sampler{<:PG}, vi::VarInfo, _) = step(model, spl, vi)
```
trappmartin: `model::CallableModel`?

src/samplers/pgibbs.jl
```diff
@@ -47,9 +47,9 @@ Sampler(alg::PG) = begin
...
 
-step(model::Function, spl::Sampler{<:PG}, vi::VarInfo) = begin
+step(model, spl::Sampler{<:PG}, vi::VarInfo) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/pgibbs.jl
```diff
@@ -79,7 +79,7 @@ step(model::Function, spl::Sampler{<:PG}, vi::VarInfo) = begin
   return particles[indx].vi, true
 end
 
-sample(model::Function, alg::PG;
+sample(model, alg::PG;
```
trappmartin: `model::CallableModel`?

src/samplers/pmmh.jl
```diff
@@ -61,7 +61,7 @@ function Sampler(alg::PMMH)
   Sampler(alg, info)
 end
 
-step(model::Function, spl::Sampler{<:PMMH}, vi::VarInfo, is_first::Bool) = begin
+step(model, spl::Sampler{<:PMMH}, vi::VarInfo, is_first::Bool) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/smc.jl
```diff
@@ -44,7 +44,7 @@ Sampler(alg::SMC) = begin
   Sampler(alg, info)
 end
 
-step(model::Function, spl::Sampler{<:SMC}, vi::VarInfo) = begin
+step(model, spl::Sampler{<:SMC}, vi::VarInfo) = begin
```
trappmartin: `model::CallableModel`?

src/samplers/smc.jl
```diff
@@ -68,7 +68,7 @@ step(model::Function, spl::Sampler{<:SMC}, vi::VarInfo) = begin
 end
 
 ## wrapper for smc: run the sampler, collect results.
-function sample(model::Function, alg::SMC)
+function sample(model, alg::SMC)
```
trappmartin: `model::CallableModel`?

src/samplers/support/hmc_core.jl
```diff
@@ -122,7 +122,7 @@ end
 
 ###
 
-function runmodel!(model::Function, vi::VarInfo, spl::Union{Nothing,Sampler})
+function runmodel!(model, vi::VarInfo, spl::Union{Nothing,Sampler})
```
trappmartin: `model::CallableModel`?

src/samplers/support/hmc_core.jl
```diff
@@ -135,7 +135,7 @@ function leapfrog(θ::AbstractVector{<:Real},
                   p::AbstractVector{<:Real},
                   τ::Int,
                   ϵ::Real,
-                  model::Function,
+                  model,
```
trappmartin: `model::CallableModel`?

src/samplers/support/hmc_core.jl
```diff
@@ -185,12 +185,12 @@ end
 
 function _hmc_step(θ::AbstractVector{<:Real},
                    lj::Real,
-                   lj_func::Function,
-                   grad_func::Function,
-                   H_func::Function,
+                   lj_func,
```
trappmartin: Why did those change?

src/samplers/support/hmc_core.jl
```diff
@@ -185,12 +185,12 @@ end
...
+                   grad_func,
+                   H_func,
                    τ::Int,
                    ϵ::Real,
-                   momentum_sampler::Function;
+                   momentum_sampler;
```
trappmartin: Same here. I guess the compiler changes should not affect the `momentum_sampler`. Or what is the rational behind that?

src/samplers/support/hmc_core.jl
```diff
@@ -238,7 +238,7 @@ end
 # TODO: remove used Turing-wrapper functions
 
 # Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/base_hmc.hpp
-function find_good_eps(model::Function, spl::Sampler{T}, vi::VarInfo) where T
+function find_good_eps(model, spl::Sampler{T}, vi::VarInfo) where T
```
trappmartin: `model::CallableModel`?

test/ad.jl/ad.jl
```diff
@@ -4,6 +4,7 @@ using Turing: VarInfo, gradient_forward, gradient_reverse
 @model foo_ad() = begin
     x ~ Normal(3, 1)
     y ~ Normal(x, 1)
+    return x, y
```
trappmartin: A return statement is not mandatory in Turing. Or is it now?

test/ad.jl/adr.jl
```diff
@@ -23,8 +23,8 @@ mvn = collect(Iterators.filter(vn -> vn.sym == :m, keys(vi)))[1]
 _s = getval(vi, svn)[1]
 _m = getval(vi, mvn)[1]
 
-x = map(_->Float64(_), vi[nothing])
-∇E = gradient_reverse(x, vi, ad_test_f)
+x = map(_x->Float64(_x), vi[nothing])
+∇E = gradient_reverse(x, vi, ad_test_f)[2]
```
trappmartin: I would prefer if `gradient_reverse` would still behave as before. Is there a reason to change it's behavior?

test/compiler.jl/model_macro.jl
```diff
@@ -16,45 +15,17 @@ expr = Turing.generate_observe(:x, :y)
...
-            :kwargs => [],
-            :body => :(return testmodel_comp_model(Turing.VarInfo(), Turing.SampleFromPrior()))
-           )
-@test c[:alias3] == MacroTools.combinedef(alias3)
-@test length(c[:closure].args[2].args[2].args) == 6
-@test mapreduce(line -> line.head == :macrocall, +, c[:closure].args[2].args[2].args) == 4
 
 # check if drawing from the prior works
-@model testmodel0(x) = begin
+@model testmodel0() = begin
```
trappmartin: Why did you remove the argument?

test/compiler.jl/model_macro.jl
```diff
@@ -16,45 +15,17 @@ expr = Turing.generate_observe(:x, :y)
...
     x ~ Normal()
     return x
 end
 f0_mm = testmodel0()
 @test mean(f0_mm() for _ in 1:1000) ≈ 0. atol=0.1
 
-@model testmodel01(x) = begin
+@model testmodel01() = begin
```
trappmartin: Why did you remove the argument?

test/gibbs.jl/gibbs_constructor.jl
```diff
@@ -24,14 +24,14 @@ c4 = sample(gdemo(), s4)
 c5 = sample(gdemo(), s5)
 
 # Very loose bound, only for testing constructor.
-for c in [c1, c2, c3 ,c4, c5]
-  check_numerical(c, [:s, :m], [49/24, 7/6], eps=1.0)
-end
+#for c in [c1, c2, c3 ,c4, c5]
```
trappmartin: I suppose this is work in progress. Maybe add something to ensure that this changes will not move into the master.

test/hmc_core.jl/bayes_lr.jl
```diff
@@ -81,9 +81,10 @@ lj = lj_func(θ)
 chn = []
 accept_num = 1
 
-
+#=
```
trappmartin: Maybe add a note as a reminder that this should be reverted?

test/hmc_core.jl/gdemo_nuts.jl
```diff
@@ -44,7 +44,7 @@ for test_id = 1:2
 
     totla_num = 10000
     for iter = 1:totla_num
-      
+      global logϵ, lj_func, grad_func, M_adapt, δ, μ
```
trappmartin: ?

test/hmcda.jl/hmcda_geweke.jl
```diff
@@ -43,7 +43,8 @@ x = [s[:y][1]...]
 s_bk = Array{Turing.Chain}(undef, N)
 
 simple_logger = Base.CoreLogging.SimpleLogger(stderr, Base.CoreLogging.Debug)
-with_logger(simple_logger) do
+Base.CoreLogging.with_logger(simple_logger) do
```
trappmartin: WIP ?

src/core/ad.jl
```diff
@@ -12,7 +12,7 @@ Computes the gradient of the log joint of `θ` for the model specified by
 function gradient(
     θ::AbstractVector{<:Real},
     vi::VarInfo,
-    model::Function,
+    model,
```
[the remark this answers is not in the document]
trappmartin: I think it would be better if this stays to be a function and not `model::CallableModel`. Or not?

trappmartin:
Could you provide a description on the new compiler design on this PR? This would be really helpful. Besides of my comments this looks really nice. Really looking forward to this.

mohamed82008:
Thanks a lot for your comments. I will respond to your comments above, and write a design summary in the weekend. Although the design is still partly being figured out!

src/core/ad.jl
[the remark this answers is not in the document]
mohamed82008: A `Function` is too restrictive. Julia lets us use [callable structs](https://docs.julialang.org/en/v1/manual/methods/index.html#Function-like-objects-1) which can store interesting information. The only purposes of input type annotations that I know of in Julia are: 1) to guide dispatch, 2) to throw meaningful errors for user-facing functions, or 3) to signal that a certain interface needs to be defined for the input, e.g. `AbstractVector`. I don't think `Function` is used for dispatch here, and the only "interface" assumed is that it needs to be callable so using `::Function` means more code and less features ;) Think of a user defining their own callable model struct, which can have a `supertype` other than `Function`, and then they want to sample from it. Not restricting the input here means that they can pass their callable struct through and it will just work, at least it should!

src/core/compiler.jl
```diff
@@ -55,19 +61,20 @@ function generate_observe(observation, distribution)
...
 
 """
-function generate_assume(variable, distribution, syms)
-    return esc(
-        quote
+function generate_assume(variable, distribution, compiler)
+    sym, idcs, csym = @VarName(variable)
+    csym = Symbol(compiler[:name], csym)
+    syms = Symbol[csym, variable]
+    return quote
```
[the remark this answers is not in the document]
mohamed82008: This was needed because the `~` macro was expanded separately, so it needed its own `esc`. Now the `~` expressions are expanded in the `model` macro, so only the `model` macro's `esc` is needed.

test/compiler.jl/model_macro.jl
mohamed82008: This is a mistake. I was testing that it works without `x`, and forgot to add it again, then noticed it after pushing! I will fix it. Thanks for pointing it out.

test/gibbs.jl/gibbs_constructor.jl
mohamed82008: Yes I will double check all comments before making this camera-ready.

test/hmc_core.jl/bayes_lr.jl
mohamed82008: Sure thing.

test/hmc_core.jl/gdemo_nuts.jl
mohamed82008: This was not working when I copied and pasted the code into REPL. The `global` here is needed in Julia 1.0 because the global scoping rules changed. This is a probably skipped file in the tests iiuc, but I am not sure.

test/hmcda.jl/hmcda_geweke.jl
mohamed82008: Again was not working when I copied and pasted the code into REPL to check the new compiler works, so I made a quick fix! These should probably be separated out in a separate PR.

src/core/compiler.jl
```diff
@@ -140,67 +143,85 @@ Example:
...
-    return tilde(left, right)
-end
+#macro ~(left, right)
+#    return tilde(left, right)
+#end
 
-function tilde(left, right)
-    return generate_observe(left, right)
+function tilde(left::Number, right, compiler)
+    generate_observe(left, right)
```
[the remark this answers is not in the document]
mohamed82008: Sure, thanks.

src/core/compiler.jl
```diff
@@ -245,36 +266,29 @@ end
...
-        @warn("Model definition seems empty, still continue.")
-    end
-
+    warn_empty(modeldef[:body])
     # construct compiler dictionary
     compiler = Dict(
         :name => modeldef[:name],
         :closure_name => Symbol(modeldef[:name], :_model),
-        :args => [],
+        :args => modeldef[:args],
```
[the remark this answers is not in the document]
mohamed82008: Hmm, the whole compiler `Dict` will probably go away or get restructured. I wanted to start with something simple and I believe `compiler[:args]` was being used somewhere with the assumption that it is essentially `modeldef[:args]` so I did this. This may have been a source of errors, so I will double check it next weekend.

src/core/trace.jl
```diff
@@ -8,7 +8,7 @@ mutable struct Trace
 end
 
 # NOTE: this function is called by `forkr`
-function Trace(f::Function)
+function Trace(f)
```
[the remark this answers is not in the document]
mohamed82008: Since I am allowing general callable structs, and passing the `CallableModel` around as if it was a function, this needs to be made generic.

src/core/trace.jl
```diff
@@ -19,7 +19,7 @@ function Trace(f::Function)
   res
 end
 
-function Trace(f::Function, spl::Sampler, vi :: VarInfo)
+function Trace(f, spl::Sampler, vi :: VarInfo)
```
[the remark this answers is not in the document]
mohamed82008: Not necessary; no conflicting `Trace` function signatures, or ambiguity in dispatch, and the parameter is not needed inside the function. `f(vi, spl)` when `f::CallableModel` will just forward the function call to `f.f`.

src/samplers/dynamichmc.jl
```diff
@@ -13,7 +13,7 @@ function Sampler(alg::DynamicNUTS{T}) where T <: Hamiltonian
   return Sampler(alg, Dict{Symbol,Any}())
 end
 
-function sample(model::Function, alg::DynamicNUTS, chunk_size=CHUNKSIZE[]) where T <: Hamiltonian
+function sample(model, alg::DynamicNUTS, chunk_size=CHUNKSIZE[]) where T <: Hamiltonian
```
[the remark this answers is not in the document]
mohamed82008: Possibly, but nothing `CallableModel` specific is used inside, so it can be kept generic to handle all callables.

src/samplers/hmc.jl
mohamed82008: If I remember correctly, this was just me getting too `Function`-allergic. I will return it for now.

src/samplers/hmc.jl
mohamed82008: Nothing `CallableModel` specific going on inside iiuc.

src/samplers/hmcda.jl
mohamed82008: You are probably right.

src/samplers/ipmcmc.jl
mohamed82008: Nothing `CallableModel`-specific inside iiuc.

src/samplers/ipmcmc.jl
mohamed82008: Same comment.

src/samplers/is.jl
mohamed82008: Same comment.

src/samplers/mh.jl
mohamed82008: Same comment.

src/samplers/nuts.jl
mohamed82008: It probably shouldn't.

src/samplers/support/hmc_core.jl
mohamed82008: hmm not sure, I will revert them.

src/samplers/support/hmc_core.jl
mohamed82008: You are probably right.

test/ad.jl/ad.jl
mohamed82008: No I don't think so. I will revert this.

test/ad.jl/adr.jl
mohamed82008: This was not doing the expected thing, it was giving an error. I just fixed the error.

mohamed82008: Wait this may be a bug. I will look into it.

test/compiler.jl/model_macro.jl
mohamed82008: A bug in sampling fake data.

mohamed82008: I was testing that sampling from prior works, then forgot to add it back.

src/core/trace.jl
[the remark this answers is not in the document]
trappmartin: Thanks, now I got it. :)

mohamed82008:
I have added a design summary above. Please let me know if something is not clear or if you want to squeeze more work in this PR.

trappmartin:
Thanks for the effort! I'll have a close look at the PR next week.

mohamed82008:
So while looking further into the type stability of Turing, I thought it may be necessary to de-scope this PR a little to avoid bloating it too much when it doesn't really have to. So I will consider this PR camera-ready if it provides at least the same features we had before, with at least the same speed we had before but without the global scope issue, which means that the `@model` macro can be used in other modules and even inside of functions. Currently, the features are on par IIUC, but the speed has worsened on my local benchmark. So I will look into it. Other big tasks to be done in other PRs include:
1. Making AD mostly type stable (started working on that). This involves encoding the AD type and parameters in the algorithm's type to push the type instability up and allow the user to fully eliminate it. However, it gets a little tricky when deciding the chunk size in `ForwardDiff`. I am saying mostly type stable because `VarInfo`'s type instability infects the AD functions.
2. Making `VarInfo` type stable. This one is very critical as it touches most of the code in `Turing.jl` so I am taking my time thinking about and making thought experiments with potential solutions as I am improving the other bits. I don't really have a feasible solution yet, but a few half-baked proposals which all seem to fail at least one use case in my head. If you have ideas, you can comment in https://github.com/TuringLang/Turing.jl/issues. While I don't have a full-fledged solution, I have a few incremental improvements which are easier to do and should improve the performance IIUC.

trappmartin:
Sounds like a good idea to me. I'll check the PR soon and then we should be ready to go.

Regarding the `VarInfo` type instability. I agree that this is very crucial, we (Hong, Kai and me) had already some discussions on this and I'm happy to jump if I'm of any help. As you said, we should have a good plan for this first. `VarInfo` currently also contains lots of unnecessary fields that we should get rid of.

mohamed82008:
Design description updated above.

yebai:
```julia
sample(gdemo()) # gdemo()()
```

mohamed82008:
@cpfiffer may you please take care of updating the docs after this is merged? The main update I can think of is that there is no need for a section on avoiding the `@model` macro, since it will be perfectly usable in other modules or functions after this PR is merged. You can perhaps rename it to Turing internals or something. I will remove item from the checklist above for now.

mohamed82008:
This is ready for one more review.

cpfiffer:
Can do. Excellent work!

yebai:
> This is ready for one more review.

Thanks, @mohamed82008. I'll take a review this PR soon.

mohamed82008:
Thanks. Btw, this is for you @yebai and @willtebbutt. I hope that's what you meant by phase-based compiler.

https://github.com/TuringLang/Turing.jl/blob/0ecf959b1554cf6df970458d22807e4cf24bda2e/src/core/compiler.jl#L256

mohamed82008:
Thanks for the review @willtebbutt. I will fix it up.

willtebbutt:
> Thanks. Btw, this is for you @yebai and @willtebbutt. I hope that's what you meant by phase-based compiler.

This is the general idea. We'll need a mechanism to allow people to add passes that aren't hard-coded into the compiler though. I'm not sure what that would look like, and we can leave it for another PR.

willtebbutt:
Changes look good to me. Great work @mohamed82008

yebai:
Thanks, @mohamed82008. Could you add a test for https://github.com/TuringLang/Turing.jl/issues/544?

mohamed82008:
Sure, I can do that.

trappmartin:
Look good to me. The only comment I have is that the `sample` function parametrisation is currently inconsistent. Sometimes you changed it to expect a `model::Model` and sometimes a `model`. Maybe it would be good to unify this so that all the `sample` functions expect a `model::Model`?

mohamed82008:
> Maybe it would be good to unify this so that all the sample functions expect a model::Model?

Sure I can do that.

PR #622, opened 2018-12-10
Update "Compositional Sampling Using Gibbs" example in docs

mohamed82008:
The current way the compiler does things, we cannot iterate over `xs` using `x = xs`, the index and variable name have to be on the LHS of `~`. Also `Categorical` was giving an error, so I assumed we want `Bernoulli`. Correct me if I am wrong. @cpfiffer


cpfiffer:
Oh, good catch. I think that was a pretty dated example that I hadn't noticed.

mohamed82008:
`Categorical` works like this ` Categorical([0.5, 0.5])`.

trappmartin:
Looks good, can this be merged?

Issue #559, opened 2018-09-22
DynamicHMC is (~5 times) faster than Turing


yebai:
Collecting 2 million samples 
- Turing.NUTS: Running time  = 4933.499192934352 seconds
- Turing.DynamicNUTS: 848.741643 seconds (7.86 G allocations: 251.076 GiB, 32.13% gc time)

using the following 

```julia

using DynamicHMC, Turing, Test

@model gdemo(x, y) = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0,sqrt(s))
  x ~ Normal(m, sqrt(s))
  y ~ Normal(m, sqrt(s))
  return s, m
end

mf = gdemo(1.5, 2.0)

@time chn1 = sample(mf, DynamicNUTS(2000000));

chn2 = sample(mf, Turing.NUTS(2000000, 0.65));

```

skanskan:
What about Stan and Klara?

tpapp:
This is how I would code the above in DynamicHMC 1.0:
```julia
using DynamicHMC, TransformVariables, LogDensityProblems, Distributions

struct GaussianDemo{T,P}
    y::T
    x::T
    s_prior::P
end

function (p::GaussianDemo)((m, s))
    d = Normal(m, √s)
    logpdf(d, p.x) + logpdf(d, p.y) + logpdf(p.s_prior, s) + logpdf(Normal(0, √s), m)
end

p = GaussianDemo(1.5, 2.0, InverseGamma(2, 3))
P = TransformedLogDensity(as((m = asℝ, s = asℝ₊)), p)
∇P = ForwardDiffLogDensity(P)
@time chain, tuned = NUTS_init_tune_mcmc(∇P, 10_000)
```
which on my laptop yields
```julia
  0.430758 seconds (3.74 M allocations: 211.077 MiB, 18.41% gc time)
```
Note that I **changed the sample size** since
```julia
julia> using MCMCDiagnostics
julia> mapslices(effective_sample_size, get_position_matrix(chain); dims = 1)
1×2 Array{Float64,2}:
 3841.64  4080.44
```
so I don't think you need 2 million samples. Finally, for real-world problems with higher dimensions, I would have used
```julia
∇P = FluxGradientLogDensity(P)
```
since reverse-mode AD is faster as the dimension increases (YMMV).

PR #623, opened 2018-12-11
Add a DynamicHMC section to the docs

cpfiffer:
Given the speed increase you can get from `DynamicNUTS` (see #559), I figured we should let folks know about it a bit more obviously.

This adds a small section to the documentation site about how to use DynamicHMC. If there's any other tips, tricks, or pitfalls that should be added here, let me know and I'll toss them in there.

### DIFF
--- a/docs/src/dynamichmc.md
+++ b/docs/src/dynamichmc.md
@@ -1,0 +1,27 @@
+---
+title: Using DynamicHMC
+permalink: /docs/dynamichmc/
+---
+
+Turing supports the use of [DynamicHMC](https://github.com/tpapp/DynamicHMC.jl) as a sampler through the use of the `DynamicNUTS` function. This is a [faster](https://github.com/TuringLang/Turing.jl/issues/559) version of Turing's native `NUTS` implementation.
+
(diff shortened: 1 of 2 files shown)


docs/src/dynamichmc.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Using DynamicHMC
+permalink: /docs/dynamichmc/
+---
+
+Turing supports the use of [DynamicHMC](https://github.com/tpapp/DynamicHMC.jl) as a sampler through the use of the `DynamicNUTS` function. This is a [faster](https://github.com/TuringLang/Turing.jl/issues/559) version of Turing's native `NUTS` implementation.
```
yebai: Thanks, @cpfiffer. This is quite helpful. Perhaps we should give some (possible) explanations on why Turing.NUTS is slow compared to DynamicHMC:

- Turing.NUTS adapt pre-conditioning matrix

@xukai92 any other notable differences between Turing.NUTS and DynamicHMC?

and also say `Turing.NUTS` support compositional inference (i.e. can be used together within a Gibbs sampler) and its efficiency will slowly catch up with other implementations.

PR #624, opened 2018-12-13
poissonlogpdf Fix for ForwardDiff

cpfiffer:
This fixes `poislogpdf` to work with `Dual` numbers. Thanks to @mohamed82008 for doing near everything.


mohamed82008:
It would be nice to add some tests and maybe fix https://github.com/TuringLang/Turing.jl/blob/master/test/ad.jl/AD_compatibility_with_distributions.jl if possible.

mohamed82008:
I will merge now, then add tests and other stuff in another PR.

PR #625, opened 2018-12-13
Fix and add more AD tests

mohamed82008:
This PR extends #624 and adds another workaround for `binomlogpdf` and `ForwardDiff`. It also adds tests for these AD functions and fixes other broken AD tests.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -11,7 +11,6 @@ MacroTools
 StatsFuns 0.7.0
 SpecialFunctions
 Bijectors
-FDM
 
 ProgressMeter 0.6.0
 BinaryProvider 0.4.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -11,6 +11,7 @@ MacroTools
 StatsFuns 0.7.0
 SpecialFunctions
 Bijectors
+FDM
 
 ProgressMeter 0.6.0
 BinaryProvider 0.4.0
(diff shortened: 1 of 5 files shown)


REQUIRE
```diff
@@ -11,6 +11,7 @@ MacroTools
 StatsFuns 0.7.0
 SpecialFunctions
 Bijectors
+FDM
```
yebai: Move the dependency on `FDM` to `test/REQUIRE`?

yebai:
Thanks, @mohamed82008. This PR looks good to me. Just one minor comment about dependency, see code review.

REQUIRE
mohamed82008: Well, that's what I thought I did! Thanks for pointing this out.

test/ad.jl/adr.jl
```diff
@@ -23,8 +23,8 @@ mvn = collect(Iterators.filter(vn -> vn.sym == :m, keys(vi)))[1]
 _s = getval(vi, svn)[1]
 _m = getval(vi, mvn)[1]
 
-x = map(_->Float64(_), vi[nothing])
-∇E = gradient_reverse(x, vi, ad_test_f)
+x = map(x->Float64(x), vi[nothing])
+∇E = gradient_reverse(x, vi, ad_test_f)[2]
```
trappmartin: Is the `[2]` on purpose?

PR #626, opened 2018-12-14
Make autodiff mostly type stable

mohamed82008:
This PR makes autodiff gradient functions mostly type stable. The AD type and parameters, for example the ForwardDiff chunk size are encoded in the type of `Hamiltonian`. I am getting a significant speedup in local benchmarks. The results are shown in https://github.com/TuringLang/TuringBenchmarks/blob/58f35460657e609d3530ecd0cefe6e23155dae32/benchmark_results/873aaec_302e61a/report.md. The `chunk` variable in ForwardDiff is still type unstable since it depends on the length of `θ` which is runtime information. I can make a safe/unsafe parameter in the `ForwardDiffAD` struct that can turn this check on or off at compile time, which enables further speedup for the experts who know what they are doing.

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -1,7 +1,7 @@
 abstract type ADBackend end
 struct ForwardDiffAD{chunk} <: ADBackend end
 struct FluxTrackerAD <: ADBackend end
-struct NoAD <: ADBackend end
+
 ADBackend() = ADBackend(ADBACKEND[])
 ADBackend(T::Symbol) = ADBackend(Val(T))
 function ADBackend(::Val{T}) where {T}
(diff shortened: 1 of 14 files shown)


src/core/ad.jl
```diff
@@ -1,26 +1,96 @@
...
+        return FluxTrackerAD
+    end
+end
+getADtype(::Any) = NoAD
+getADtype(s::Sampler) = getADtype(typeof(s))
+getADtype(s::Type{<:Sampler{TAlg}}) where {TAlg} = getADtype(TAlg)
+getADtype(alg::Union{Hamiltonian, Gibbs, PMMH}) = getADtype(typeof(alg))
+getADtype(::Type{<:Hamiltonian{AD}}) where {AD} = AD
+
+@generated function getADtype(::Union{Type{<:Gibbs{A}}, Type{<:PMMH{T, A}}}) where {T, A <: Tuple}
```
xukai92: Can we add some comments to say what's the logic of this function. E.g. if no sampling alg uses AD, returns NoAD; if only one sampling alg uses AD, returns its type and if more than one uses AD, return only if they are consistent otherwise throw an error.

xukai92: Also we can probably support mixed AD if we tie the AD type to samplers?

xukai92:
Thanks for the PR. I think it's looks great!

PR #632, opened 2018-12-18
Clearer documentation structure

cpfiffer:
This commit fully separates the documentation files into two folders. One folder (`docs/site`) is specifically for the the appearance and structure of the site, whereas the other (`docs/src`) is where all site _content_ is located. 

In general, a simple heuristic is to put anything that is a markdown (`.md`) file into `docs/src` and anything else into `docs/site`. `docs/src/` also contains any site assets, which is things like logos and some other miscellaneous site-wide files.

Hopefully this will alleviate the confusion I've been having as to which file to edit. If all you are doing is writing text, it is probably best to not mess with the `docs/site` folder. The exception here is when you need to add a new document to the documentation section or the tutorials section. This needs to be changed in the file `docs/site/_data/navigation.yml`. I've left this here because it seems a more natural location, but I can change that too if it seems strange to anyone else.

I've also made some (very) minor fixes to the advanced usage guide.


yebai:
Thanks, @cpfiffer!

Issue #631, opened 2018-12-15
Move all numerical/correctness tests to dedicated travis build stage


yebai:
- `hmc.jl/matrix_support.jl `

yebai:
Closed by #633.

PR #633, opened 2018-12-18
Move numerical tests to separate Travis build stage

cpfiffer:
See #631.

The `runtests` function has changed a bit. The keyword `tests` has been changed to `test_folders`, and there are two new keywords:

- `exclude` is an array of specific test files to exclude. 
- `specific_tests` is an array of specific test file to include. This overwrites all the other arguments and only tests the given files.

Currently, the only test treated separately is `hmc.jl/matrix_support.jl`, but this can be quickly updated by changing the `numerical_tests` variable in `runtests.jl`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -11,7 +11,21 @@ cd(path); include("utility.jl")
 @debug("[runtests.jl] utility.jl loaded")
 @debug("[runtests.jl] testing starts")
 
-# run all tests
-runtests()
+if get(ENV, "TRAVIS", false)
+    # If Travis is testing, separate the tests.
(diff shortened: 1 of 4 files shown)


cpfiffer:
This should be good to merge once Appveyor passes.

yebai:
Thanks, @cpfiffer!

test/runtests.jl
```diff
@@ -11,7 +11,21 @@ cd(path); include("utility.jl")
 @debug("[runtests.jl] utility.jl loaded")
 @debug("[runtests.jl] testing starts")
 
-# run all tests
-runtests()
+if get(ENV, "TRAVIS", false)
```
yebai: nice!

PR #636, opened 2018-12-29
Documentation Update

cpfiffer:
Contains a couple of updates to the docs after #613 was merged over.

- Adds an autodiff section. It's somewhat brief for the moment, but I think we should keep the information consolidated somewhere.
- The advanced guide now covers how to generate a model without the macro, according to the most current compiler structure.
- The regular guide document has a more expanded section on generated quantities.

As always, please take a look at what I've added and let me know if there's anything incorrect, misleading, or improvable.


docs/src/_docs/advanced.md
```diff
@@ -54,61 +54,79 @@ Distributions.rand(d::Flat, n::Int) = Vector([rand() for _ = 1:n])
 Distributions.logpdf{T<:Real}(d::Flat, x::Vector{T}) = zero(x)
 ```
 
-## Avoid Using the `@model` Macro
+## Model Internals
 
-When integrating Turing.jl with other libraries, it can be necessary to avoid using the `@model` macro. To achieve this, one needs to understand the `@model` macro, which works as a closure and generates an amended function by
+The `@model` macro accepts a function definition and generates a `Turing.CallableModel` struct for use by the sampler. Models can be constructed by hand without the use of a macro. Taking the `gdemo` model as an example, the two code sections below (macro and macro-free) are equivalent.
```
yebai: `Turing.CallableModel` ==> `Turing.Model`.

docs/src/_docs/autodiff.md
```diff
@@ -0,0 +1,39 @@
...
+permalink: /docs/autodiff/
+toc: true
+toc_sticky: true
+---
+
+## Switching AD Modes
+
+Turing supports two types of automatic differentiation (AD) in the back end during sampling. The current default AD mode is [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl), but Turing also supports [Flux's](https://github.com/FluxML/Flux.jl) [Tracker](https://fluxml.ai/Flux.jl/stable/internals/tracker.html#Flux.Tracker-1)-based differentation.
+
+To switch between `ForwardDiff` and `Flux.Tracker`, one can call function `Turing.setadbackend(backend_sym)`, where `backend_sym` can be `:forward_diff` or `:reverse_diff`.
```
yebai: nice!

docs/src/_docs/autodiff.md
```diff
@@ -0,0 +1,39 @@
...
+    s ~ InverseGamma(2, 3)
+    m ~ Normal(0, sqrt(s))
+    x ~ Normal(m, sqrt(s))
+    y ~ Normal(m, sqrt(s))
+    return s, m
+end
+
+# Sample using Gibbs and varying autodiff backends.
+c = sample(gdemo(1.5, 2),
+  Gibbs(1000,
```
yebai: nice!

docs/src/_docs/guide.md
```diff
@@ -143,15 +144,84 @@ Assign the function without inputs to a variable, and Turing will produce a samp
...
-# Output: (0.685690547873451, -1.1972706455914328)
+```
+Output:
+```
+(0.685690547873451, -1.1972706455914328)
+```
+
+#### Generating Vectors of Quantities
+
+More broadly, Turing models can be treated as generative by providing default values in the model declaration, and then calling that model without parameters.
```
yebai: default values can only be `missing` at the moment, any other value will be treated as missing values. 

@mohamed82008 is the above summary accurate?

trappmartin:
This looks good to me. So many improvements on the code and the documentation happened, really cool to read! 👍 

Thanks!

docs/src/_docs/guide.md
mohamed82008: It is kind of accurate yes. Default values currently can be anything but only the size and type information of the default value will be used if it is a vector. The value itself is not used anywhere. When the input is a vector (or matrix, etc), it is important to have a default input for the case when it is treated as a parameter to know the eltype and size to pre-allocate it. When the input is a scalar, the default value is useless, as we don't need to pre-allocate. I would just add more explanation explaining why this feature exists. Also perhaps change "calling that model without parameters" -> "calling that model without arguments" to avoid confusion with parameters and data variables.

cpfiffer: Are these all valid defaults? 

```julia
zeros(10)
Vector{Real}(undef, 10)
Vector{Union{Missing, Real}}([missing, missing])
Vector{Union{Missing, Real}}([missing, 2])
[1,2,3,9,10,16]
```

My proposed addition to explain this behavior:

> #### What to Use as a Default Value
> 
> Currently, the actual _value_ of the default argument does not matter. Only the dimensions and type of a non-atomic value are relevant. Turing uses default values to pre-allocate vectors when they are treated as parameters, because if the value is not provided, the model will not know the size or type of a vector. Consider the following model:
> 
> ```julia
> @model generator(x) = begin
>   s ~ InverseGamma(2,3)
>   m ~ Normal(0,sqrt(s))
>   for i in 1:length(x)
>       x[i] ~ Normal(m, sqrt(s))
>   end
>   return s, m
> end
> ```
> 
> If we are trying to generate random random values from the `generator` model and we call `sample(generator(), HMC(1000, 0.01, 5))`, we will receive an error. This is because there is no way to determine `length(x)`, whether `x` is a vector, and the type of the values in `x`.
> 
> A sensible default value might be:
> 
> ```julia
> @model generator(x = zeros(10)) = begin
>   s ~ InverseGamma(2,3)
>   m ~ Normal(0,sqrt(s))
>   for i in 1:length(x)
>       x[i] ~ Normal(m, sqrt(s))
>   end
>   return s, m
> end
> ```
> 
> In this case, the model compiler can now determine that `x` is a `Vector{Float64,1}` of length 10, and the model will work as intended. It doesn't matter what the values in the vector are — at current, `x` will be treated as a parameter if it assumes its default value, i.e. no value was provided in the function call for that variable.

mohamed82008: Yes these are valid defaults with one restriction. Variables used in Hamiltonian samplers need to have `Real` `eltype`. The last case is integer, so it can only be used for integer variables, which are obviously not Hamiltonian-compatible.

mohamed82008: When sampling from prior, we don't need the `Real` restriction.

cpfiffer: > When sampling from prior, we don't need the `Real` restriction.

What are the restrictions on there if `Real` is not required?

mohamed82008: Its `eltype` needs to admit the `eltype` of the variable, e.g. `Int` or `Float64`.

cpfiffer: So would it be fair to say the following:

> The type of the vector should be either `Int` or `Float64`, depending on which sampler is being used. A Hamiltonian sampler like `HMC` or `NUTS` requires values to be of type `Float64`, whereas non-Hamiltonian samplers like `SMC` or `PG` permit the use of `Int` variables.

mohamed82008: I think the following is better: 

"The element type of the vector (or matrix) should match the type of the random variable, `<: Integer` for discrete random variables and `<: AbstractFloat` for continuous random variables. Moreover, if the continuous random variable is to be sampled using a Hamiltonian sampler, the vector's element type needs to be `Real` to enable auto-differentiation through the model which uses special number types that are sub-types of `Real`. Finally, when using a particle sampler, a `TArray` should be used."

These are general tips when pre-allocating a vector of parameters, regardless of whether or not, it is a "default value".

cpfiffer: Good one, threw that into the guide.

cpfiffer:
This is probably good to merge over, barring further comments.

yebai:
Thanks, @cpfiffer!

PR #637, opened 2018-12-29
[WIP] Typed VarInfo and Turing reorganisation

mohamed82008:
This is a gigantic WIP PR that's not fully tested yet, and that I should probably break down into a few PRs. The main idea is to make a concretely typed `VarInfo` which I call `TypedVarInfo` and use a function barrier to do the expensive sampling using the `TypedVarInfo`. Before the sampling starts however, I run a type unstable function that figures out the `TypedVarInfo` by actually running the model on an `UntypedVarInfo`, formerly just `VarInfo`, and the converting it to a typed one. I will write a more complete summary at some point. If you feel adventurous, you can skim through it and give me your feedback.


yebai:
Thanks, @mohamed82008. Since it's challenging to make `VarInfo` type stable for all scenarios, having a type stable version of `VarInfo` for HMC samplers sounds like a good solution!

mohamed82008:
It is a bit more general that this but yes this is the idea. While making `TypedVarInfo`, I also realized that `spl.info` and `sample.value` need to be made type stable, beside a few other struct fields so I am lumping them together. Let's make Turing fast!

cpfiffer:
I'm really pumped for this!

xukai92:
Thanks @mohamed82008 ! I wonder if it would be useful for the purpose of type stability to introduce an intermediate type (say `RandomVariable`) which is a single instance inside `VarInfo` that contains the value, distribution, and other information, and make `VarInfo` a vector of `RandomVariable` plus `logp` and some other fields.

mohamed82008:
@xukai92 this is pretty much what I have done! https://github.com/TuringLang/Turing.jl/blob/mt/type_stability/src/Core/VarReplay/typed_varinfo.jl#L3

This file is pretty much ready, so please read it if you have time. There is one performance bug there which I fixed on my machine but it doesn't change the semantics.

xukai92:
That's great! Yes I read the code (briefly) before my last comment and noticed the `SingleVarInfo` type, but the type signature confused me (because they are still vectors of xxx). Why do we make it like this?

src/Core/AD.jl
```diff
@@ -1,3 +1,46 @@
+module AD
+
+import ForwardDiff
+import Flux.Tracker
+using ..VarReplay
```
xukai92: Probably dumb question - what do `..` and `...` do?

mohamed82008: One dot searches for the module in the same global scope (as opposed to normal package directories). Two dots go up one level, three dots go up 2 levels, etc. For example:

```julia
module A
    module B
    end
    using .B

    module C
         using ..B        
         module D
             using ...B
         end
         using .D
    end
end
```

xukai92: Great thx!

mohamed82008:
> Why do we make it like this?

This was the easiest to accommodate multi-variate and single-variate variables. A named tuple is immutable, so the single variable's value has to be a `Ref` or a `Vector`. It is technically possible to modify an immutable that is a field of a mutable (`vi` is) using the more advanced approach of [StaticArrays.MArray](https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/MArray.jl#L130), so this may change in the future but for now this is the easiest implementation.

src/Core/VarReplay/varinfo.jl
```diff
@@ -0,0 +1,325 @@
...
+    idcs        ::    Dict{VarName,Int}
+    vns         ::    Vector{VarName}
+    ranges      ::    Vector{UnitRange{Int}}
+    vals        ::    Vector{Real}
+    dists       ::    Vector{Distributions.Distribution}
+    gids        ::    Vector{Int}
+    logp        ::    Real
+    num_produce ::    Int           # num of produce calls from trace, each produce corresponds to an observe.
+    orders      ::    Vector{Int}   # observe statements number associated with random variables
+    flags       ::    Dict{String, BitVector}
```
xukai92: If this `flags` filed leads to instability, we can also ``unpack'' it into several fields. The main purpose of it is to introduce an unified interface for flag-like fields.

mohamed82008: No this is fine, since the way we use it is type stable as far as I can tell.

mohamed82008:
Module structure:
```julia
module Turing
    module Utilities
    end

    module Core
         module Compiler
         end

         module VarReplay
         end

         module Container
         end      

         module AD
         end
   end

    module Inference
        module Adapt
        end
    end
end
```

src/Core/AD.jl
```diff
@@ -1,3 +1,46 @@
...
+
+const ADSAFE = Ref(false)
+function setadsafe(switch::Bool)
+    @info("[Turing]: global ADSAFE is set as $switch")
+    ADSAFE[] = switch
+end
+
+const CHUNKSIZE = Ref(40) # default chunksize used by AD
+
+function setchunksize(chunk_size::Int)
```
cpfiffer: I'm glad all this stuff is getting moved out of `Turing.jl`.

PR #642, opened 2019-01-06
Adds code example and sampler path image to splash page

cpfiffer:
A minor update to the splash page, but one I think gets the splash page closer to something I'm happy with. 

New:
- A code excerpt showing the `gdemo` model.
- A nicer 3d plot showing the `NUTS` sampler path on a weird posterior.


yebai:
Thanks, @cpfiffer!

PR #648, opened 2019-01-17
Fix DynamicHMC support and related issues

mohamed82008:
This fixes #647 and other related issues introduced in #626. A little bit of cleanup is also done.

PR #649, opened 2019-01-19
Reorganization - no functional change

mohamed82008:
This PR dissects part of #637 and is based on #648. When #648 goes in, this PR can be trivially re-based. 

Three modules are created in this PR: `Utilities`, `Core` and `Inference`. This is less fine-grained than the splitting in #637 but is a step in the right direction IMO. There is no functional change here. I used this chance to fix some indentation and function syntax issues, but some more indentation inconsistency is remaining. Few structs and functions are defined in the `Turing` module itself, like `Sampler` and `Model`. `Sampler` for example is used in the other modules, so it was important to define it first. It's also nice to have `Turing.Sampler` and `Turing.Model` directly as opposed to `Turing.Inference.Sampler` and `Turing.Core.Model`.


yebai:
Thanks, @mohamed82008.

Issue #600, opened 2018-11-17
Is Binomial supported/working?


goedman:
Hi, I'm trying to get below Turing program to run:

```
using Turing, MCMCChain

data = (n = 9, k = 6)
@model globe_toss(n, k) = begin
  #prior
  theta ~ Beta(1, 1)
  #model
  k ~ Binomial(n, theta)
end

chn2 = sample(globe_toss(data.n, data.k), HMC(2000, 0.75, 5))
describe(chn2)
```
which produces below error:

julia> include("/Users/rob/.julia/dev/StatisticalRethinking/chapters/02/turing_globe_toss.jl")
[ Info:  Assume - `theta` is a parameter
[ Info:  Observe - `k` is an observation
ERROR: LoadError: MethodError: no method matching Float64(::Flux.Tracker.TrackedReal{Float64})
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:185
  Float64(::T<:Number) where T<:Number at boot.jl:725
  Float64(::Int8) at float.jl:60
  ...

This is on Julia 1.0.2 and using Turing v0.5.1. These are my first steps in using Turing, so it is likely I'm doing something wrong. Thanks.

yebai:
This should be already supported. We have a similar model here:
https://github.com/TuringLang/TuringExamples/blob/master/stan-models/bernoulli.model.jl

cpfiffer:
@yebai I've noticed this come up a couple times while I've been playing with some of the models. It seems like the parameters are being converted into tracked values which don't shoehorn well into the `binomlogpdf` function. I suspect this might need more triage because I've seen this come up a good deal.

The way I've patched this is by calling `Flux.Tracker.data()` on the parameters. It's not a good long-term solution but it certainly will get things working. Here's the modified version of @goedman's example:

```julia
using Turing, MCMCChain, Flux.Tracker

data = (n = 9, k = 6)
@model globe_toss(n, k) = begin
  #prior
  theta ~ Beta(1, 1)
  #model
  k ~ Binomial(n, Tracker.data(theta))
end

chn2 = sample(globe_toss(data.n, data.k), HMC(2000, 0.75, 5))
describe(chn2)
```

yebai:
>   k ~ Binomial(n, Tracker.data(theta))

@cpfiffer 
[...]

PR #656, opened 2019-01-22
Remove duplicate method definition.

yebai:
This PR fixes the following warning:

```julia
WARNING: Method definition <=(Flux.Tracker.TrackedReal{T} where T<:Real, Flux.Tracker.TrackedReal{T} where T<:Real) in module Tracker at /Users/hg344/.julia/packages/Flux/U8AZD/src/tracker/lib/real.jl:45 overwritten in module Core at /Users/hg344/.julia/dev/Turing/src/core/ad.jl:167.
```

Related: 

https://github.com/FluxML/Flux.jl/pull/499 

https://github.com/TuringLang/Turing.jl/issues/600

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -163,9 +163,6 @@ function gradient_reverse(
     return l, ∂l∂θ
 end
 
-import Base: <=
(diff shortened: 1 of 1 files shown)


mohamed82008:
My bad. I was using a pinned Flux.

PR #658, opened 2019-01-25
Fixed symbol parsing issue

cpfiffer:
This fixes a very minor issue noted over in [TuringTutorials](https://github.com/TuringLang/TuringTutorials/issues/20#issuecomment-457509791). The `parse` function was throwing a fit because it was being given a `SubString` to convert into an expression, so I called `Meta.parse` instead. Seems to have done the trick. Anyone see any obvious problems with this fix?

I guess my editor also removed some blank spaces at line endings, so I suppose I've tidied up a bit too.

### DIFF
--- a/src/utilities/io.jl
+++ b/src/utilities/io.jl
@@ -45,7 +45,7 @@ function getjuliatype(s::Sample, v::Symbol, cached_syms=nothing)
     # Fill sample
     for i = 1:length(syms)
         # Get indexing
-        idx = Main.eval(parse(idx_comp[i][1]))
+        idx = Main.eval(Meta.parse(idx_comp[i][1]))
         # Determine if nesting
         nested_dim = length(idx_comp[1]) # how many nested layers?
         if nested_dim == 1
@@ -97,12 +97,12 @@ mutable struct Chain{R<:AbstractRange{Int}} <: AbstractChains
 end
 
 function Chain()
-    return Chain{AbstractRange{Int}}( 0.0, 
-                                      Vector{Sample}(), 
-                                      Array{Float64, 3}(undef, 0, 0, 0), 
+    return Chain{AbstractRange{Int}}( 0.0,
+                                      Vector{Sample}(),
+                                      Array{Float64, 3}(undef, 0, 0, 0),
                                       0:0,
-                                      Vector{String}(), 
-                                      Vector{Int}(), 
+                                      Vector{String}(),
+                                      Vector{Int}(),
                                       Dict{Symbol,Any}()
                                     )
 end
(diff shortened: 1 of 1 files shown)


src/utilities/io.jl
```diff
@@ -45,7 +45,7 @@ function getjuliatype(s::Sample, v::Symbol, cached_syms=nothing)
     # Fill sample
     for i = 1:length(syms)
         # Get indexing
-        idx = Main.eval(parse(idx_comp[i][1]))
+        idx = Main.eval(Meta.parse(idx_comp[i][1]))
```
mohamed82008: `parse(Int, idx_comp[i][1])` seems to be what's needed here. There is also another `eval(parse(..))` above which should be replaced the same way.

mohamed82008:
For the record, the offending example:
```julia
using Turing
y = [ 1.0 ];
N = length(y);  K = 3;
@model BayesHmm(y, K) = begin
    N = length(y)
    s = tzeros(Int, N)
    m = Vector{Real}(undef, K)
    T = Vector{Vector{Real}}(undef, K)
    for i = 1:K
        T[i] ~ Dirichlet(ones(K)/K)
        m[i] ~ Normal(i, 0.5)
    end
    s[1] ~ Categorical(K)
    y[1] ~ Normal(m[s[1]], 0.1)
    for i = 2:N
        s[i] ~ Categorical(vec(T[s[i-1]]))
        y[i] ~ Normal(m[s[i]], 0.1)
    end
end;
g = Gibbs(1000, HMC(2, 0.001, 7, :m, :T), PG(20, 1, :s))
c = sample(BayesHmm(y, 3), g);
m_set = c[:m]
```

src/utilities/io.jl
cpfiffer: Doesn't the `Main.eval(Meta.parse(. . . ))` functionality allow for non-integer indices? Don't we have cases where we index by nonintegers?

mohamed82008: You are correct. But note that because of the way `idx` is used later, it can only be a tuple, array or integer. The following is an `eval`-free way of doing the right thing:
```julia
function parse_inds(inds)
	p_inds = [parse(Int, m.captures[1]) for m in eachmatch(r"(\d+)", inds)]
	if length(p_inds) == 1
		return p_inds[1]
	else
		return Tuple(p_inds)
	end
end
parse_inds("(1,1)")
parse_inds("1")
```

mohamed82008:
Thanks for this @cpfiffer . If no tests fail and if you have no more comments, I will merge this by the end of the day.

PR #660, opened 2019-01-26
VarInfo and type stability refactor 

mohamed82008:
This PR dissects the VarInfo and performance changes from #637.

The main change in this PR is that it defines 2 different subtypes of `AbstractVarInfo`:
1. `UntypedVarInfo`
2. `TypedVarInfo` 

The `UntypedVarInfo` is exactly the same as the current `VarInfo`. This `VarInfo` is important because it lets us figure out the types of the parameters by actually running the model once. When the types have been figured out, we can now reduce the `UntypedVarInfo` to a more performant `TypedVarInfo` with all the `eltype`s as tight as possible. The `TypedVarInfo` is designed around the `NamedTuple` construct in Julia, which lets us have a mapping from the parameters' `Symbol`s to their differently typed value vectors and other meta-information. So this lets us decouple the `eltypes` of different parameters which lowers the need for `Real` inside `TypedVarInfo`. However, there is one problem remaining. When using a Hamiltonian inference algorithm, the `eltype` of a sampler's parameter is not actually known in advance, since it depends on the function being differentiated by the autodiff package, which is part of the first parameter of `ForwardDiff.Dual` for example. Additionally, we should still be able to just run the model and have the parameter's `eltype` be `Float64`, not any autodiff type. This makes it difficult to just fix the concrete type of the instance of `TypedVarInfo` throughout the sampling process. The workaround used for this here is to create a new `VarInfo` instance using https://github.com/TuringLang/Turing.jl/blob/mt/type_stability2/src/core/ad.jl#L111, which replaces the `Float64` vectors of the `Sampler`'s parameters with similar vectors of the autodiff type before passing it to `model`. This is less than ideal but it lets us have concretely typed `TypedVarInfo` instances throughout. Note that to do this in a type stable way, the space of the `Sampler` was encoded in its type as a parameter.

These are the remaining tasks to do here:
- [x] Make `sampler.info` into a type stable struct, one for each inference algorithm
- [x] Make `sample.value` into 1 or more type stable fields
- [ ] Make sure tests pass
- [ ] Add tests for `TypedVarInfo`
- [ ] Add `_` before the names of functions in `VarReplay` which are only used internally
- [ ] Make sure all of the `_sample` functions are fully type stable (this is run in the second iteration onwards) except for the `Real` in model definitions discussed in #665


yebai:
> I will see if I can make this any faster.

Can you post some runtime numbers here?

mohamed82008:
I added a description of the main changes and design decisions made in this PR. Please give me your feedback on the concept. The code still needs a rebase so it is not clean.

src/core/ad.jl
```diff
@@ -95,28 +99,32 @@ using forwards-mode AD from ForwardDiff.jl.
...
 )
     # Record old parameters.
-    vals_old, logp_old = copy(vi.vals), copy(vi.logp)
+    vals_old, logp_old = copy(vi[sampler]), copy(vi.logp)
 
     # Define function to compute log joint.
     function f(θ)
-        vi[sampler] = θ
-        return -runmodel!(model, vi, sampler).logp
+        new_vi = NewVarInfo(vi, sampler, eltype(θ))
```
xukai92: Does `NewVarInfo` reallocate memory here or it just re-type things?

mohamed82008: It re-allocates. This is a performance issue which can be improved. I tried memoizing the output of `NewVarInfo` using `Memoize.jl` but it was slower because it introduced a type instability, and allocating is pretty fast. Perhaps the best option here would be to define the function `f` as a functor, and try to infer the AD type in advance, then allocate the AD vector as part of `vi`. This is a different surgical operation though so let's not do heart and brain surgery at the same time. I would be happy if we can just get this through without any repercussions elsewhere in the package, since I am touching quite a bit of Turing here. I am yet to add proper tests for `TypedVarInfo` to match those of `UntypedVarInfo` and compare the 2 together.

mohamed82008:
Currently, I am trying to "fix" the samplers and make their infos type stable. The following list shows the currently working inference algorithms on this simple program:
```julia
@model gdemo(x, y) = begin
	s ~ InverseGamma(2,3)
	m ~ Normal(0,sqrt(s))
	x ~ Normal(m, sqrt(s))
	y ~ Normal(m, sqrt(s))
end
```

- [x] IS `sample(gdemo(2.0, 1.5), IS(1000))`
- [x] MH `sample(gdemo(2.0, 1.5), MH(1000))`
- [x] HMC `sample(gdemo(2.0, 1.5), HMC(1000, 0.1, 5))`
- [x] SGHMC `sample(gdemo(2.0, 1.5), SGHMC(1000, 0.001, 0.1))`
- [x] SGLD `sample(gdemo(2.0, 1.5), SGLD(100, 0.01))`
- [x] HMCDA `sample(gdemo(2.0, 1.5), HMCDA(1000, 0.15, 0.65))`
- [x] NUTS `sample(gdemo(2.0, 1.5), HMCDA(1000, 0.65))`
- [x] Gibbs `sample(gdemo(2.0, 1.5), Gibbs(1000, MH(1, :s), SGLD(100, 0.01, :m)))`
- [ ] SMC
- [ ] IPMCMC
- [ ] PMMH
- [ ] PG

mohamed82008:
The following tests are failing for various reasons:

- [ ] compiler.jl / assume.jl
- [ ] compiler.jl / explicit_ret.jl
- [ ] compiler.jl / observe.jl
- [ ] compiler.jl / opt_param_of_dist.jl
- [ ] dynamichmc-support.jl / dynamic_hmc.jl
- [ ] gibbs.jl / gibbs_constructor.jl
- [ ] gibbs.jl / gibbs2.jl
- [ ] hmc.jl  / constrained_simplex.jl
- [ ] hmc.jl  / matrix_support.jl
- [ ] hmc.jl  / multivariate_support.jl
- [ ] hmc.jl  / hmcda_cons.jl
- [ ] io.jl / chain_utility.jl
- [ ] io.jl / save_resume_chain.jl
- [ ] ipmcmc.jl / ipmcmc.jl
- [ ] ipmcmc.jl / ipmcmc2.jl
- [ ] is.jl / importance_sampling.jl
- [ ] mh.jl / mh_cons.jl
- [ ] models.jl / single_dist_correctness.jl
- [ ] nuts.jl / nuts_cons.jl
- [ ] pmmh.jl / pmmh_cons.jl
- [ ] pmmh.jl / pmmh.jl
- [ ] resample.jl / particlecontainer.jl
- [ ] sghmc.jl / sghmc_cons.jl
- [ ] sgld.jl / sgld_cons.jl
- [ ] sgld.jl / sgld.jl
- [ ] trace.jl / trace.jl
- [ ] varinfo.jl / orders.jl
- [ ] varinfo.jl / test_varname.jl
- [ ] varinfo.jl / varinfo.jl
- [ ] vectorization.jl / vec_assume.jl
- [ ] vectorization.jl / vectorize_observe.jl
 
Wish me luck!

PR #664, opened 2019-01-27
AD workaround for nbinomlogpdf

### LINKED ISSUE #661: is NegativeBinomial AD supported?
```julia
using Turing

@model mymod(y, r) = begin
    p ~ Beta(2, 1)
    y ~ NegativeBinomial(r, p)
end

chn = sample(mymod(2,5), NUTS(1000, 200, .75))
```
I get the following error

MethodError: no method matching unsafe_convert(::Type{Float64}, ::ForwardDiff.Dual{Nothing,Float64,1})

Is it expected to draw from Gamma and Poisson in order to produce a NegativeBinomial?

mohamed82008:
Fixes #661. Someone please double check this.

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -181,10 +181,26 @@ Tracker.@grad function binomlogpdf(n::Int, p::Tracker.TrackedReal, x::Int)
 end
 
 import StatsFuns: nbinomlogpdf
+# Note the definition of NegativeBinomial in Julia is not the same as Wikipedia's.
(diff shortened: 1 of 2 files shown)


willtebbutt:
@mohamed82008 this looks good, but we _definitely_ need some finite-differencing-based unit testing here. I would strongly recommend making [FDM.jl](https://github.com/invenia/FDM.jl/) a test-only dependency and using this. We've just implemented `jvp` (jacobian-vector product) for testing forwards-mode sensitivities, and `j′vp` for testing reverse-mode sensitivities.

PR #671, opened 2019-01-31
Fix a warning message and provide a generic API for find_good_eps

xukai92:


src/inference/support/hmc_core.jl
```diff
@@ -251,12 +253,13 @@ function find_good_eps(model, spl::Sampler{T}, vi::VarInfo) where T
 
     # Heuristically find optimal ϵ
     while (iter_num <= 12)
```
trappmartin: Maybe we could make the 12 to be a keyword argument of find_good_eps?

src/inference/support/hmc_core.jl
```diff
@@ -221,28 +221,30 @@ function _hmc_step(θ::AbstractVector{<:Real},
...
 # Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/base_hmc.hpp
 function find_good_eps(model, spl::Sampler{T}, vi::VarInfo) where T
-    logpdf_func_float = gen_lj_func(vi, spl, model)
+    lj_func = gen_lj_func(vi, spl, model)
     momentum_sampler = gen_momentum_sampler(vi, spl)
+    grad_func = gen_grad_func(vi, spl, model)
     H_func = gen_H_func()
+    θ = vi[spl]
+    ϵ = _find_good_eps(θ, lj_func, grad_func, H_func, momentum_sampler)
+    @info "\r[$T] found initial ϵ: $ϵ"
```
trappmartin: Maybe move this to line: 283?

src/inference/support/hmc_core.jl
xukai92: Sure.

src/inference/support/hmc_core.jl
xukai92: In fact, I did this initially. But later moved it for two reasons: 1) `T` is not defined for `_find_good_eps`; 2) maybe it's better to keep functions starting with `_` not printing any information.

mohamed82008:
This needs a rebase.

PR #672, opened 2019-02-03
Fix Julia version in AppVeyor to 1.0

mohamed82008:


mohamed82008:
I will merge this for now. When Libtask is ready, we can revert this PR.

Issue #687, opened 2019-02-19
`gradient` interface inconsistent!


trappmartin:
I just came across the following breaking bug in the code base.

We currently return different results depending on the AD backend used. In particular, using a forward mode AD results in the the log joint and the gradient of the negative log joint, see:
https://github.com/TuringLang/Turing.jl/blob/0a95e366543d441d2f2449b8fac5de58656a3102/src/core/ad.jl#L106-L116
while using a reverse mode AD results in the *negative* log joint and the gradient of the negative log joint, see:
https://github.com/TuringLang/Turing.jl/blob/0a95e366543d441d2f2449b8fac5de58656a3102/src/core/ad.jl#L147-L154

To ensure correctness, @willtebbutt proposed the following steps:

- [ ] agree to compute `logp`,  _not_ `-logp` and
- [ ] rename `gradient` -> `gradient_logp`, `gradient_forward`->`gradient_logp_forward`, and `gradient_reverse`->`gradient_logp_reverse` to remove ambiguities
- [ ] add tests based on our (growing) suite of toy models that verify that `logp` and its gradient (as estimate by finite differencing or something) is correct.

I propose we should also:

- [ ] provide tests that verify gradient computations with any AD backend provide the same output
- [ ] add doc string that clearly states what the `gradient_logp` function returns, i.e. `logp` and `gradient_logp`

cc: @xukai92, @yebai

xukai92:
Thanks for catching this. Just for a side note: Turing's HMC sampler is not affected by this bug because we are not using the return log-joint from `gradient` function. But our DynamicHMC wrap probably does suffer from it (https://github.com/TuringLang/Turing.jl/blob/master/src/inference/dynamichmc.jl#L62).

PR #688, opened 2019-02-19
Fix gradient_forward and add tests

mohamed82008:
This is a temporary hot fix for the main problem raised in #687. The other changes proposed require changes in more places in Turing, so another PR would be required.


willtebbutt:
This looks good to me.

mohamed82008:
No worries. Tests pass locally, but let's see Travis and AppVeyor.

PR #641, opened 2019-01-06
[WIP] Update HMC adaptation interface, remove spl.info, and more tests

xukai92:
TODOs
- [x] Log statistics during adaptation
  - E.g. step size, accept rate, etc
- [x] Run MH as a baseline for sanity check on ESS
- [x] Adaptation interface
- [x] Write scripts to test each of below
  - HMC + DA
  - HMC + PD
  - HMC + DA + PD
  - NUTS + DA + PD
- Test cases:
  - [x] gdemo
  - ~~m12.6~~
  - [x] m12.2
  - ~~LDA~~

Issue #406, opened 2017-12-08
The weight filed of `Chain` is buggy


xukai92:
See discussion in #405

xukai92:
NOTE:

In `io.jl`

```julia
function Base.getindex(c::Chain, v::Symbol)
  # This strange implementation is mostly to keep backward compatability.
  #  Needs some refactoring a better format for storing results is available.
  if v == :logevidence
    log(c.weight)
  elseif v==:samples
    c.value2
  elseif v==:logweights
    c[:lp]
  else
    map((s)->Base.getindex(s, v), c.value2)
  end
end
```

xukai92:
To summarize the discussion in #405,

For `Chain`, we store original `Sample` in `Chain.vals2` so we can get all the weights by `weights = [results.value2[i].weight for i in range(1, length(results))]`.

However in current `bnp-project` branch, I guess due to the limitation of `Chain` that it cannot store varying dimensional variables, the `Vector{Sample}` is directly returned. So `weights = [results[i].weight for i in range(1, length(results))]` is needed to extract weights.

@emilemathieu does it sound correct to you?

xukai92:
Forget to @ae-foster

emilemathieu:
@xukai92 I believe the issue is mostly that the unbiased estimator in importance sampling or SMC is the __weighted__ sum over samples. Therefore, either a resampling step is needed at the end so that the weights are equals (thus useless) or a simple access to the weights is needed.

xukai92:
Yes, I understand this. I'm talking about how to give a simple & consistent access to the weights. There is such an interface but it seems not consistent for `Chain` type, which is returned after inference in the master, and `Vector{Sample}` type, which is used in the bnp branch (?).

yebai:
This should be fixed by #793.

PR #695, opened 2019-02-24
[Do not merge] MCMCChain → MCMChains Changeover

cpfiffer:
The MCMCChain package is changing its name to MCMCChains (note the "s" at then end). Currently, the package is not registered, but there is a [pull request](https://github.com/JuliaLang/METADATA.jl/pull/21638) open to do so. 

I will remove the draft status from this PR when 

1. The MCMCChains package is actually registered; and
2. We unify whether we actually store the log evidence in the `Chains.logevidence` field. Currently, some samplers store `exp.(logevidence)`, which seems strange to me at the moment. The importance sampling tests are currently failing (on purpose) until we pick some convention there. **EDIT** By convention, actual log evidence should be stored in the `logevidence` field. I believe I have updated the samplers that store this information.

**EDIT** MCMChains is registered on METADATA and the log evidence issue has been addressed, this should be ready for review.

In addition to changing the `using` statements everywhere, this pull request also slightly adapts Turing's internals to suit the MCMCChains API. 

Turing currently uses a `Turing.Chain` object, with the following definition:

```julia
struct Chain{R<:AbstractRange{Int}} <: AbstractChains
    weight  ::  Float64                 # log model evidence
    value   ::  Array{Union{Missing, Float64}, 3}
    range   ::  R # TODO: Perhaps change to UnitRange?
    names   ::  Vector{String}
    chains  ::  Vector{Int}
    info    ::  Dict{Symbol,Any}
end
```

This struct has been removed in this PR and replaced with the MCMCChains native version, `Chains`. `Chains` objects have the following fields:

```julia
struct Chains{A, T} <: AbstractChains
    value::AxisArray{Union{Missing,A},3}
    logevidence::T
    name_map::Dict{Any, Vector}
    info::Dict{Symbol, Any}
end
```

The `range` field can be retrieved with `range(chain)`, `names` can be retrieved with `names(chain)`, and the `weight` field has been renamed to `logevidence`.

Important changes:
* When using `save!` on a chain, the sample array is stuck inside the `chain.info[:samples]` field. This is not a long term solution, and I would still like to add functionality to recreate sample from the `chain.value` field. I believe this would require storing particle weights in the `value` field, but I haven't looked into it too much at the moment.
* Any variable with a name found in `"elapsed", "epsilon", "eval_num", "lf_eps", "lf_num", "lp"` will be moved into a separate section (called `:internals`) in the `Chains` object by default. This won't impact anything Turing side, but it will make the `describe`, `plot`, and other diagnostic functions output significantly more readable by automatically hiding `:internal` variables unless the user specifically asks to see them.
* Previously, Turing had it's own function defined with the signature `getindex(c::Chain, s::Symbol)`, which had some behavior specific to our `Chain` objects. I have ported the bulk of this functionality over to `MCMCChains`, but the ability to call `chain[:logevidence]` (which used to return `log(chain.weight)`) and `chain[logweights]` (which used to return `chain[:lp]`) is now gone.

Other than that, most of the changes have happened over at [MCMCChains](https://github.com/TuringLang/MCMCChains.jl). A good summary of the changes can be found [here](https://github.com/TuringLang/MCMCChains.jl/pull/33).


mohamed82008:
This change will conflict with #660 mainly because of the following 2 fields:
```julia
name_map::Dict{Any, Vector}
info::Dict{Symbol, Any}
```
I recommend changing these types to type parameters and assuming they can be a `NamedTuple` or a struct. I haven't read MCMCChains.jl and I am not sure if this will play well with other packages like Mamba and CmdStan or not. Do you think this is easily doable?

goedman:
For CmdStan I would follow the change to using NamedTuples. This will only affect StanMCMCChain and I'll combine it with the renaming to StanMCMCChains. At some point I do plan to make MCMCChains 'standard' in CmdStan as well.

The same is true for how I plan to use [DynamicHMC and MCMCChains](https://statisticalrethinkingjulia.github.io/DynamicHMCModels.jl/latest/05/m5.1d1/) in DynamicHMCModels.jl.

I haven't looked too much at what would be required for MambaModels but once MCMCChains is registered I will set up some of the advanced models in Mamba and then we can maybe go back to the Mamba folks?

cpfiffer:
How does the following definition sound:

```julia
struct Chains{A, T, K<:NamedTuple, L<:NamedTuple} <: AbstractChains
    value::AxisArray{Union{Missing,A},3}
    logevidence::T
    name_map::K
    info::L
end
```

This way, pretty much everything is immutable, except for the `value` field which can be indexed directly. I've got this working on a test branch on MCMCChains without too much change. It should even tidy some things up on the Turing side.

cpfiffer:
This is ready for a formal review, anyone have any comments on this before it is merged over?

src/utilities/io.jl
```diff
@@ -175,81 +137,35 @@ function flatten(names, value :: Array{Float64}, k :: String, v)
...
-        info = merge(info, c.info)
-    end
-
-    chn = Chain(weight,
-                value,
-                range,
-                names,
-                chains,
-                info)
+function Base.vcat(c1::Chains, args::Chains...)
```
trappmartin: This should have not been changed back to the old version which is brocken. Seem like you are accidentally overwriting master with an old version here..

yebai:
> When using save! on a chain, the sample array is stuck inside the chain.info[:samples] field. This is not a long term solution, and I would still like to add functionality to recreate sample from the chain.value field. I believe this would require storing particle weights in the value field, but I haven't looked into it too much at the moment.

Thanks, @cpfiffer - Excellent work. I agree that a long term solution should probably try to recreate `sample` from `Chain.value`. But this is ok for now.

src/utilities/io.jl
cpfiffer: Good catch, thank you! Updated.

yebai:
> Thanks, @cpfiffer - Excellent work. I agree that a long term solution should probably try to recreate sample from Chain.value. But this is ok for now. A related issue: #406

A quick thought, can we store `weights` as an internal parameter, in a way similar to `_internal_vars`?

cpfiffer:
> A quick thought, can we store `weights` as an internal parameter, in a way similar to `_internal_vars`?

Yeah, no problem at all with that. I can work on that after this PR goes through, if you like. Then we can clean up the `info[:samples]` field.

yebai:
Let's see whether @xukai92 has any comments before merging.

xukai92:
Looks nice to me. Is prefixing with `_` implemented though? I believe it's only in my WIP PR(https://github.com/TuringLang/Turing.jl/pull/641).

@yebai Actually I don't know what to do with #641 ATM. It contains some useful code for new interface to set-up adaptation methods as well as passing and prefixing internal stats with `_`. Otherwise codes for fixing adaptation (as you know) are experimental.

cpfiffer:
For the moment, the underscore functionality hasn't been integrated, though it's on the list for now. We can just add whatever your internal parameters to the `_internal_vars` vector.

yebai:
@cpfiffer  Appveyor X86 tests are broken - it seems the tests are taking substantial more time than X64 tests and may exceed Appveyor time constraints.

src/inference/pgibbs.jl
```diff
@@ -128,24 +128,24 @@ function sample(  model::Model,
 
     loge = exp.(mean(spl.info[:logevidence]))
```
trappmartin: Remove `exp`

src/inference/pgibbs.jl
```diff
@@ -128,24 +128,24 @@ function sample(  model::Model,
...
     if resume_from != nothing   # concat samples
-        pushfirst!(samples, resume_from.value2...)
-        pre_loge = resume_from.weight
+        pushfirst!(samples, resume_from.info[:samples]...)
+        pre_loge = resume_from.logevidence
         # Calculate new log-evidence
-        pre_n = length(resume_from.value2)
+        pre_n = length(resume_from.info[:samples])
         loge = exp.((log(pre_loge) * pre_n + log(loge) * n) / (pre_n + n))
```
trappmartin: Remove `exp`

src/inference/pmmh.jl
```diff
@@ -170,12 +170,12 @@ function sample(  model::Model,
     println("  Accept rate         = $accept_rate;")
 
     if resume_from != nothing   # concat samples
-      pushfirst!(samples, resume_from.value2...)
+      pushfirst!(samples, resume_from.info[:samples]...)
     end
     c = Chain(0.0, samples)       # wrap the result by Chain
```
trappmartin: Change to `-Inf`?

test/io.jl/save_resume_chain.jl
```diff
@@ -19,33 +19,39 @@ alg3 = Gibbs(500, PG(30, 10, :s), HMCDA(1, 500, 0.65, 0.05, :m))
 end
 
 chn1 = sample(gdemo([1.5, 2.0]), alg1; save_state=true)
-
 check_numerical(chn1, [:s, :m], [49/24, 7/6])
 
-chn1_resumed = resume(chn1, 1000)
+# Test chains read/write.
+write("chn.jlds", chn1)
```
trappmartin: Maybe we should move this IO test to `MCMCChains?`

test/io.jl/save_resume_chain.jl
```diff
@@ -19,33 +19,39 @@ alg3 = Gibbs(500, PG(30, 10, :s), HMCDA(1, 500, 0.65, 0.05, :m))
...
-
 check_numerical(chn1_contd2, [:s, :m], [49/24, 7/6])
 
 chn2 = sample(gdemo([1.5, 2.0]), alg2; save_state=true)
-
 check_numerical(chn2, [:s, :m], [49/24, 7/6])
 
-chn2_contd = sample(gdemo([1.5, 2.0]), alg2; resume_from=chn2)
+# Test chains read/write.
+write("chn.jlds", chn2)
```
trappmartin: Same here.

test/io.jl/save_resume_chain.jl
```diff
@@ -19,33 +19,39 @@ alg3 = Gibbs(500, PG(30, 10, :s), HMCDA(1, 500, 0.65, 0.05, :m))
...
+chn2_contd = sample(gdemo([1.5, 2.0]), alg2; resume_from=chn2)
 check_numerical(chn2_contd, [:s, :m], [49/24, 7/6])
 
 chn3 = sample(gdemo([1.5, 2.0]), alg3; save_state=true)
-
 check_numerical(chn3, [:s, :m], [49/24, 7/6])
 
-chn3_contd = sample(gdemo([1.5, 2.0]), alg3; resume_from=chn3)
+# Test chains read/write.
+write("chn.jlds", chn3)
```
trappmartin: And here.

test/io.jl/save_resume_chain.jl
cpfiffer: Works for me, I'll pull it out.

src/inference/pmmh.jl
cpfiffer: I'm assuming here you mean change

```julia
c = Chain(0.0, samples)
```

to

```julia
c = Chain(-Inf, samples)
```

I pushed a commit to that effect, but I just want to make sure that was correct.

trappmartin: Yes, but honestly I'm not sure what the default value should be. `-Inf` would be consistent with the old behaviour, i.e. `log(0.0)`.

cpfiffer:
I'm not sure what's going on with the AppVeyor tests. There's three tests in the queue ahead of the most recent commit, so I'll try to root around after it's done.

yebai:
> I'm not sure what's going on with the AppVeyor tests. There's three tests in the queue ahead of the most recent commit, so I'll try to root around after it's done.

This seems fixed by the most recent commit.

PR #696, opened 2019-02-24
Fix #687

### LINKED ISSUE #687: `gradient` interface inconsistent!

mohamed82008:
Fixes #687. This PR does the following:
- [x] compute `logp`,  _not_ `-logp` and
- [x] rename `gradient` -> `gradient_logp`, `gradient_forward`->`gradient_logp_forward`, and `gradient_reverse`->`gradient_logp_reverse` to remove ambiguities
- [x] add doc string that clearly states what the `gradient_logp` function returns, i.e. `logp` and `gradient_logp`

Testing that the backends give the same output was done before in #688, and testing against finite difference is already done for the lower level `logpdf` of distributions. I don't think this is necessary for the whole `gradient_logp`.

CC: @trappmartin @willtebbutt @xukai92

### DIFF
--- a/src/core/Core.jl
+++ b/src/core/Core.jl
@@ -37,12 +37,12 @@ export  @model,
         ForwardDiffAD, 
         FluxTrackerAD,
         value,
-        gradient,
+        gradient_logp,
         CHUNKSIZE, 
         ADBACKEND,
         setchunksize,
         verifygrad,
-        gradient_forward,
-        gradient_reverse
+        gradient_logp_forward,
+        gradient_logp_reverse
 
 end # module
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -57,17 +57,18 @@ getADtype(s::Sampler) = getADtype(typeof(s))
 getADtype(s::Type{<:Sampler{TAlg}}) where {TAlg} = getADtype(TAlg)
 
 """
-gradient(
+gradient_logp(
     θ::AbstractVector{<:Real},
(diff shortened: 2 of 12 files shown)


trappmartin:
Thanks!

Issue #537, opened 2018-09-14
Remove spl = nothing


trappmartin:
We currently have `assume(nothing, ...)` calls in the tests, e.g. in `test/ad.jl/ad1.jl`. These calls should be adjusted such that we can remove `Nothing` from the type union in assume.

PR #706, opened 2019-03-08
Remove `Nothing` from the Sampler union type

KDr2:
Try to fix https://github.com/TuringLang/Turing.jl/issues/537

### DIFF
--- a/src/core/VarReplay.jl
+++ b/src/core/VarReplay.jl
@@ -111,6 +111,7 @@ function Turing.runmodel!(model::Model, vi::VarInfo, spl::AbstractSampler)
     model(vi, spl)
     return vi
 end
+Turing.runmodel!(model::Model, vi::VarInfo) = Turing.runmodel!(model, vi, SampleFromPrior())
 
 const VarView = Union{Int,UnitRange,Vector{Int},Vector{UnitRange}}
 
--- a/src/core/VarReplay.jl
+++ b/src/core/VarReplay.jl
@@ -1,6 +1,7 @@
 module VarReplay
 
-using ...Turing: Turing, CACHERESET, CACHEIDCS, CACHERANGES, Sampler, Model
+using ...Turing: Turing, CACHERESET, CACHEIDCS, CACHERANGES, Model,
+    AbstractSampler, Sampler, SampleFromPrior
 using ...Utilities: vectorize, reconstruct, reconstruct!
 using Bijectors: SimplexDistribution
 using Distributions
@@ -102,9 +103,9 @@ mutable struct VarInfo
     end
 end
 
-function Turing.runmodel!(model::Model, vi::VarInfo, spl::Union{Nothing,Sampler})
+function Turing.runmodel!(model::Model, vi::VarInfo, spl::AbstractSampler)
     setlogp!(vi, zero(Real))
-    if spl != nothing && :eval_num ∈ keys(spl.info)
+    if isa(spl, Sampler) && :eval_num ∈ keys(spl.info)
         spl.info[:eval_num] += 1
     end
     model(vi, spl)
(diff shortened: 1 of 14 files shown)


src/core/compiler.jl
```diff
@@ -26,7 +26,7 @@ end
 
 
 function wrong_dist_errormsg(l)
-    return "Right-hand side of a ~ must be subtype of Distribution or a vector of" * 
+    return "Right-hand side of a ~ must be subtype of Distribution or a vector of " *
```
yebai: Nice to see this fixed!

src/core/ad.jl
```diff
@@ -30,6 +30,7 @@ getchunksize(::T) where {T <: ForwardDiffAD} = getchunksize(T)
 getchunksize(::Type{ForwardDiffAD{chunk}}) where chunk = chunk
 getchunksize(::T) where {T <: Sampler} = getchunksize(T)
 getchunksize(::Type{<:Sampler{T}}) where {T} = getchunksize(T)
+getchunksize(::SampleFromPrior) = getchunksize(Nothing)
```
yebai: Remove `getchunksize(::Nothing) = getchunksize(Nothing)`?

src/inference/Inference.jl
```diff
@@ -185,10 +172,16 @@ function assume(spl::A,
 
 end
 
+
+observe(::Nothing,
```
yebai: Nice!

src/core/VarReplay.jl
```diff
@@ -102,14 +103,15 @@ mutable struct VarInfo
     end
 end
 
-function Turing.runmodel!(model::Model, vi::VarInfo, spl::Union{Nothing,Sampler})
+function Turing.runmodel!(model::Model, vi::VarInfo, spl::AbstractSampler)
```
trappmartin: Would it make sense to use a generated function here?

src/core/VarReplay.jl
```diff
@@ -294,7 +296,8 @@ end
 
 # Get all indices of variables belonging to gid or 0
 getidcs(vi::VarInfo) = getidcs(vi, nothing)
-getidcs(vi::VarInfo, spl::Nothing) = filter(i -> vi.gids[i] == 0, 1:length(vi.gids))
+getidcs(vi::VarInfo, ::SampleFromPrior) = filter(i -> vi.gids[i] == 0, 1:length(vi.gids))
+# getidcs(vi::VarInfo, spl::Nothing) = filter(i -> vi.gids[i] == 0, 1:length(vi.gids))
```
trappmartin: Remove comment?

src/core/container.jl
```diff
@@ -1,8 +1,8 @@
 mutable struct Trace
   task  ::  Task
   vi    ::  VarInfo
-  spl   ::  Union{Nothing, Sampler}
-  Trace() = (res = new(); res.vi = VarInfo(); res.spl = nothing; res)
+  spl   ::  AbstractSampler
```
trappmartin: Maybe we can make `Trace` a parametric type?

src/core/compiler.jl
KDr2: 😄

src/core/ad.jl
KDr2: Done.

src/core/VarReplay.jl
KDr2: Hmm... I think it's not a big impact: a generated function only change the `isa` type-check to a multi-dispatch. Anyway, I changed it to a generated function.

src/core/VarReplay.jl
KDr2: Done.

src/core/container.jl
KDr2: Done.

Issue #713, opened 2019-03-10
Fresh install Julia 1.1: test errors -> No method iterate(::Chain), etc.


McCoyBecker:
Plenty of errors throwed.

Examples:

1. No method iterate(::Chain...)
2. LoadError: ArgumentError: at least one finite value must be provided to formatter.

Also mentions that it can't find Turing on the current path, which is bizarre because I just added it in pkg.

Is this a Julia 1.1 thing?

I should motivate this post by saying that I completely wiped Julia because I couldn't get the Visualize Samplers code to work on the website, because of no method iterate errors.

Any help would be greatly obliged.

Here's more information on the 1. No method iterate issue from test.

beta_binomial.jl: Error During Test at /home/mccoy/.julia/packages/Turing/dsVcO/test/utility.jl:105
  Got exception outside of a @test
  LoadError: MethodError: no method matching iterate(::Chains{Union{Missing, Float64},Float64,NamedTuple{(:parameters,),Tuple{Array{String,1}}},NamedTuple{(:hashedsummary,),Tuple{Base.RefValue{Tuple{UInt64,MCMCChains.ChainSummaries}}}}})
  Closest candidates are:
    iterate(!Matched::Core.SimpleVector) at essentials.jl:568
    iterate(!Matched::Core.SimpleVector, !Matched::Any) at essentials.jl:568
    iterate(!Matched::ExponentialBackOff) at error.jl:199

cpfiffer:
We changed some indexing behavior over at [MCMCChains](https://github.com/TuringLang/MCMCChains.jl), specifically in [PR 57](https://github.com/TuringLang/MCMCChains.jl/pull/57). It's not a Julia 1.1 issue, it's an oversight on my part to adapt the tests.

I've got a fix ready to go after testing completes, but for the most part, Turing should work fine. The `plot_samplers` function in the visualization library is also out of date for the indexing functionality, and I'll follow up shortly with an updated version of that function that should work for you.

McCoyBecker:
Awesome! Thank you.

McCoyBecker:
Apologies for closing, will let you close when good.

cpfiffer:
As to the sampler visualization, try changing `plot_sampler` to 

```julia
function plot_sampler(chain)
    # Extract values from chain.
    val = get(chain, [:s, :m, :lp])
    ss = link.(Ref(InverseGamma(2,3)), val.s)
    ms = val.m
    lps = val.lp

[...]

PR #715, opened 2019-03-10
Indexing fixes

cpfiffer:
Updates all the tests to use the MCMCChains indexing behavior, since `chn[:param]` now returns a `Chains` object and not an array. I also updated the sampler page to use the correct indexing.


cpfiffer:
This fixes all broken tests as noted in #713.

Issue #567, opened 2018-09-28
Replacing `algorithm.gid` with `algorithm.s::Selector`


yebai:
At the moment, each sampler algorithm has a field `alg.gid` to match/select variables in `VarInfo`. This is used to implement the Gibbs sampler, i.e. assigning variables to different component samplers. The current solution creates an unnecessary dependency on `Sampler` since there are cases when we want to select variables from `VarInfo` without referring to a sampler. Perhaps we can create a `Selector` type, and move the `gid` field to this selector type. 

- create a `Selector` type and make it a field of `Sampler`: `varinfo[spl::Sampler]`===> `varinfo[spl.s::Selector]`
- randomly generate a unique `gid` instead of using a manually assigned integer
- each variable in `v::VarInfo` contain an array of `s::Selector`. 
   - Note: each variable can be mapped to multiple samplers

yebai:
Fixed by #720

PR #720, opened 2019-03-18
Using new Selector type instead of gid

KDr2:
#567

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -63,14 +63,14 @@ end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
 
-mutable struct Selector
-    uuid :: UUID
+struct Selector
+    gid :: UUID
 end
 Selector() = Selector(uuid1())
-const DEFAULT_SELECTOR = Selector()
+const DEFAULT_SELECTOR = Selector(UUID("00000000-0000-0000-0000-000000000000"))
+const INVALID_SELECTOR = Selector(UUID("00000000-0000-0000-0000-000000000001"))
 const DEFAULT_GID = DEFAULT_SELECTOR
-const INVALID_SELECTOR = Selector()
-==(s1::Selector, s2::Selector) = s1.uuid == s2.uuid
+==(s1::Selector, s2::Selector) = s1.gid == s2.gid
 
 
 abstract type AbstractSampler end
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -91,13 +91,13 @@ Turing translates models to chunks that call the modelling functions at specifie
 then include that file at the end of this one.
 """
 mutable struct Sampler{T} <: AbstractSampler
-    parent   ::  AbstractSampler
     alg      ::  T
     info     ::  Dict{Symbol, Any} # sampler infomation
     selector ::  Selector
+    parent   ::  AbstractSampler
 end
-Sampler(alg, model::Model) = Sampler(alg)
-Sampler(alg, info::Dict{Symbol, Any}) = Sampler(SampleFromPrior(), alg, info, Selector())
+Sampler(alg, model::Model, parent=SampleFromPrior()) = Sampler(alg, parent)
+Sampler(alg, info::Dict{Symbol, Any}, parent=SampleFromPrior()) = Sampler(alg, info, Selector(), parent)
 
 include("utilities/Utilities.jl")
 using .Utilities
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -8,6 +8,7 @@ module Turing
 #       to indicate that's not implemented inside Turing.jl            #
 ########################################################################
 
+using UUIDs
 using Requires, Reexport, ForwardDiff
 using Bijectors, StatsFuns, SpecialFunctions
 using Statistics, LinearAlgebra, ProgressMeter
@@ -15,7 +16,7 @@ using Markdown, Libtask, MacroTools
 @reexport using Distributions, MCMCChains, Libtask
 using Flux.Tracker: Tracker
 
-import Base: ~, convert, promote_rule, rand, getindex, setindex!
+import Base: ~, ==, convert, promote_rule, rand, getindex, setindex!
 import Distributions: sample
 import MCMCChains: AbstractChains, Chains
 
@@ -62,6 +63,16 @@ end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
 
+mutable struct Selector
+    uuid :: UUID
+end
+Selector() = Selector(uuid1())
+const DEFAULT_SELECTOR = Selector()
+const DEFAULT_GID = DEFAULT_SELECTOR
+const INVALID_SELECTOR = Selector()
+==(s1::Selector, s2::Selector) = s1.uuid == s2.uuid
+
+
 abstract type AbstractSampler end
 
 """
@@ -83,10 +94,13 @@ Turing translates models to chunks that call the modelling functions at specifie
 then include that file at the end of this one.
 """
 mutable struct Sampler{T} <: AbstractSampler
-    alg   ::  T
-    info  ::  Dict{Symbol, Any}         # sampler infomation
+    alg      ::  T
+    info     ::  Dict{Symbol, Any} # sampler infomation
+    selector ::  Selector
 end
-Sampler(alg, model) = Sampler(alg)
+Sampler(alg, model::Model, new_selector=false) = Sampler(alg, new_selector)
+Sampler(alg, info::Dict{Symbol, Any}, new_selector=false) =
+    Sampler(alg, info, new_selector ? Selector() : DEFAULT_SELECTOR)
 
 include("utilities/Utilities.jl")
 using .Utilities
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -63,11 +63,10 @@ end
 function runmodel! end
 
 struct Selector
-    gid :: Tuple{Float64, Int32}
+    gid :: UInt64
 end
-Selector() = Selector((time(), rand(Int32)))
-const DEFAULT_SELECTOR = Selector((0.0, 0))
-const INVALID_SELECTOR = Selector((0.0, 1))
+Selector() = Selector(time_ns())
+const INVALID_SELECTOR = Selector(UInt64(1))
 ==(s1::Selector, s2::Selector) = s1.gid == s2.gid
 
 
@@ -92,13 +91,13 @@ Turing translates models to chunks that call the modelling functions at specifie
 then include that file at the end of this one.
 """
 mutable struct Sampler{T} <: AbstractSampler
+    parent   ::  AbstractSampler
     alg      ::  T
     info     ::  Dict{Symbol, Any} # sampler infomation
     selector ::  Selector
 end
-Sampler(alg, model::Model, new_selector=false) = Sampler(alg, new_selector)
-Sampler(alg, info::Dict{Symbol, Any}, new_selector=false) =
-    Sampler(alg, info, new_selector ? Selector() : DEFAULT_SELECTOR)
+Sampler(alg, model::Model) = Sampler(alg)
+Sampler(alg, info::Dict{Symbol, Any}) = Sampler(SampleFromPrior(), alg, info, Selector())
 
 include("utilities/Utilities.jl")
 using .Utilities
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -8,7 +8,6 @@ module Turing
 #       to indicate that's not implemented inside Turing.jl            #
 ########################################################################
 
-using UUIDs
 using Requires, Reexport, ForwardDiff
 using Bijectors, StatsFuns, SpecialFunctions
 using Statistics, LinearAlgebra, ProgressMeter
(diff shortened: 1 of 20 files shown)


src/Turing.jl
```diff
@@ -62,6 +63,16 @@ end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
 
+struct Selector
+    gid :: UUID
+end
+Selector() = Selector(uuid1())
```
yebai: Consider `uuid1()` ==> `UUIDs.uuid1()`?

src/Turing.jl
```diff
@@ -62,6 +63,16 @@ end
...
+const DEFAULT_SELECTOR = Selector(UUID("00000000-0000-0000-0000-000000000000"))
+const INVALID_SELECTOR = Selector(UUID("00000000-0000-0000-0000-000000000001"))
+const DEFAULT_GID = DEFAULT_SELECTOR
```
yebai: Is it possible to remove `DEFAULT_GID`, since it's an alias to `DEFAULT_SELECTOR ` anyway?

xukai92:
Regarding the broken tests, it seems like a deterministic bug simply because we fixed the random seed and use a small number of iterations (500) and small tolerance (0.1).  Thus the bug is tied to that particular random seed we set, which makes the estimate slightly larger than the tolerance with 500 iterations. Either simply increasing the number of iterations (e.g. both to 2,000) can fix the tests.

Question is why and where this PR change the random number generator process? With a skim on the changes it doesn't seem direct to me. Or does the use of UUID have any influence on it?

KDr2:
> Question is why and where this PR change the random number generator process? With a skim on the changes it doesn't seem direct to me. Or does the use of UUID have any influence on it?

It must be the UUID, if I change it to a timestamp, all the test cases pass.

src/core/VarReplay.jl
```diff
@@ -139,8 +140,13 @@ getsym(vi::VarInfo, vn::VarName) = vi.vns[getidx(vi, vn)].sym
 getdist(vi::VarInfo, vn::VarName) = vi.dists[getidx(vi, vn)]
 
 getgid(vi::VarInfo, vn::VarName) = vi.gids[getidx(vi, vn)]
-
-setgid!(vi::VarInfo, gid::Int, vn::VarName) = vi.gids[getidx(vi, vn)] = gid
+function setgid!(vi::VarInfo, gid::Selector, vn::VarName, shared=false)
```
yebai: Similar to `VarInfo.gids`, we should consider renaming this function.

src/Turing.jl
```diff
@@ -83,10 +92,13 @@ Turing translates models to chunks that call the modelling functions at specifie
...
 """
 mutable struct Sampler{T} <: AbstractSampler
-    alg   ::  T
-    info  ::  Dict{Symbol, Any}         # sampler infomation
+    alg      ::  T
+    info     ::  Dict{Symbol, Any} # sampler infomation
+    selector ::  Selector
 end
-Sampler(alg, model) = Sampler(alg)
+Sampler(alg, model::Model, new_selector=false) = Sampler(alg, new_selector)
```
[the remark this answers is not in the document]
KDr2: These two arguments constructors are in the algorithm implementation files, e.g.: ` Sampler(alg::PG, new_selector=false)` in `pgibbs.jl`.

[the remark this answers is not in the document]
yebai: I see - thanks!

yebai:
Ready to merge from my side.

KDr2:
> * Do we really need a `DEFAULT_SELECTOR`, since we can always generate a `selector` for new samplers on-the-fly?

Currently, the `DEFAULT_SELECTOR` is used for
1. `SampleFromPrior` and `SampleFromUniform`, in this case, we push var into VI with the default selector because these samplers has no selector. This can be easily fixed.
2. In components of Gibbs, their many `spl.selector == DEFAULT_SELECTOR` and `spl.selector != DEFAULT_SELECTOR`, I will look into this.

KDr2:
> Ready to merge from my side.

OK, if the PR is merged, I will file new PR for the renaming and `DEFAULT_SELECTOR` deletion if I can fix them  :)

KDr2:
> 2\. In components of Gibbs, their many `spl.selector == DEFAULT_SELECTOR` and `spl.selector != DEFAULT_SELECTOR`, I will look into this.

We use `spl.selector ==/!= DEFAULT_SELECTOR` to test if a Sampler is a top-level one or a component of another sampler, so if we remove `DEFAULT_SELECTOR`, we should add a new field to indicate this relation, e.g., a `parent` sampler for each sampler. Is it right?

yebai:
> We use spl.selector ==/!= DEFAULT_SELECTOR to test if a Sampler is a top-level one or a component of another sampler, so if we remove DEFAULT_SELECTOR, we should add a new field to indicate this relation, e.g., a parent sampler for each sampler. Is it right?

yes, maybe add a field to `Selector` instead of `Sampler` for this purpose? Given that each sampler only has one unique `selector` field, this might be the most straightforward approach.

KDr2:
> yes, maybe add a field to `Selector` instead of `Sampler` for this purpose? Given that each sampler only has one unique `selector` field, this might be the most straightforward approach.

I think that adding a `parent` field to the `Sampler` type may be better:
1. The hierarchy is of the samplers, not the selectors. `parent` in Sampler type shows this hierarchy more straightforwardly.
2. We can use the singleton `SampleFromPrior()` as the parent of the top-level samplers.
3. A selector can be reused in different Samplers or be used standalone without any sampler, and selectors have no hierarchies, adding a hierarchy field in it brings much confusion.
4. For now, the argument `new_selector` in the Sampler constructor is also very confusing. When we use `new_selector=true` to create a Sampler, we indeed mean to create a Sampler with a parent. So this argument should be changed to `parent`.

I add some new commits about this, please have a look. @yebai

src/core/VarReplay.jl
```diff
@@ -305,8 +301,8 @@ end
 #   vi.logp = vi.logp[end:end]
 # end
 
-# Get all indices of variables belonging to spl.selector or DEFAULT_SELECTOR
-getidcs(vi::VarInfo, ::SampleFromPrior) = filter(i -> DEFAULT_SELECTOR in vi.gids[i] , 1:length(vi.gids))
+# Get all indices of variables belonging to SampleFromPrior
```
yebai: consider adding a comment here to indicate that: if `VarInfo.gid` is empty for a variable, then it's assumed to be assigned to a `SampleFromPrior` sampler.

src/inference/dynamichmc.jl
```diff
@@ -52,7 +52,7 @@ function sample(model::Model,
     vi = VarInfo()
     model(vi, SampleFromUniform())
 
-    if spl.selector == DEFAULT_SELECTOR
+    if spl.parent == SampleFromPrior()
```
yebai: @xukai92 I'm confused by why we need `spl.selector == DEFAULT_SELECTOR` here. Can you clarify?

src/core/VarReplay.jl
KDr2: done.

src/inference/dynamichmc.jl
yebai: Perhaps we can introduce a field to `Selector` to indicate sampling algorithm types, e.g.

```julia
struct Selector
    gid :: UInt64
    alg :: Symbol # :default, :invalid, :Gibbs, :HMC
end
Selector() = Selector(time_ns(), :default)
Selector(a::Symbol) = Selector(time_ns(), a)
==(s1::Selector, s2::Selector) = s1.gid == s2.gid
```

Then we can 

- `spl.selector == DEFAULT_SELECTOR` ==> `spl.selector.alg==:default`
- Remove `INVALID_SELECTOR ` and change ` push!(vi, vn, r, dist, INVALID_SELECTOR)` ==> 
`push!(vi, vn, r, dist, Selector(:invalid))`
https://github.com/TuringLang/Turing.jl/blob/b7ceac22310175f0651f70ac150aaf3e9fad5f78/src/inference/pgibbs.jl#L173

KDr2: Good solution, I will have a try.

yebai:
Excellent work - thanks @KDr2!

src/core/VarReplay.jl
```diff
@@ -70,7 +71,7 @@ mutable struct VarInfo
     vals        ::    Vector{Real}
     rvs         ::    Dict{Union{VarName,Vector{VarName}},Any}
     dists       ::    Vector{Distributions.Distribution}
-    gids        ::    Vector{Int}
+    gids        ::    Vector{Set{Selector}}
```
mohamed82008: Why is it a set? IIUC, every variable can only belong to one sampler, hence one gid is enough, no?

src/core/VarReplay.jl
```diff
@@ -207,6 +207,9 @@ end
 Base.getindex(vi::VarInfo, vview::VarView) = copy(getval(vi, vview))
 Base.setindex!(vi::VarInfo, val::Any, vview::VarView) = setval!(vi, val, vview)
 
+Base.getindex(vi::VarInfo, s::Selector) = copy(getval(vi, getranges(vi, s)))
+Base.setindex!(vi::VarInfo, val::Any, s::Selector) = setval!(vi, val, getranges(vi, s))
```
mohamed82008: Looks identical to the next one, so use `Union{Selector, Sampler}` instead to avoid duplication.

src/core/VarReplay.jl
```diff
@@ -346,6 +356,10 @@ function getranges(vi::VarInfo, spl::Sampler)
     end
 end
 
+function getranges(vi::VarInfo, s::Selector)
```
mohamed82008: Call this function in the `getranges` with `::Sampler` as a second argument to avoid duplication.

src/inference/gibbs.jl
```diff
@@ -28,23 +28,24 @@ mutable struct Gibbs{A} <: InferenceAlgorithm
...
     samplers = Array{Sampler}(undef, n_samplers)
-
     space = Set{Symbol}()
 
     for i in 1:n_samplers
         sub_alg = alg.algs[i]
         if isa(sub_alg, GibbsComponent)
-            samplers[i] = Sampler(typeof(sub_alg)(sub_alg, i), model)
+            samplers[i] = Sampler(sub_alg, model)
+            samplers[i].selector.tag[] = Symbol(typeof(sub_alg))
```
mohamed82008: This and the equivalent in PMMH seem to be the only 2 places where `tag` is modified after construction of the sampler. Can we not pass the tag to the constructor, and avoid the use of `Ref` entirely for performance?

src/inference/mh.jl
```diff
@@ -45,15 +45,15 @@ function MH(n_iters::Int, space...)
...
 end
-MH{T}(alg::MH, new_gid::Int) where T = MH{T}(alg.n_iters, alg.proposals, alg.space, new_gid)
 
 function Sampler(alg::MH, model::Model)
     alg_str = "MH"
 
     # Sanity check for space
-    if alg.gid == 0 && !isempty(alg.space)
+    # TODO: if (this_sampler.selector.tag[] == :default) && !isempty(alg.space)
+    if false && !isempty(alg.space)
```
mohamed82008: ??

src/core/VarReplay.jl
yebai: This relaxed the assumption the one-to-one mapping between variables and samplers. In general, variables can be mapped to multiple component samplers in a Gibbs sampler. For example, a variable can be updated by: `Gibbs(PG(:z), HMC(:z))`. The result will still be correct and can lead to performance gains for some elaborate models.

mohamed82008: I see, thanks for the explanation.

src/inference/gibbs.jl
yebai: I agree that the use of `Ref` might be unnecessary if we avoiding mutating `selector`s after they are created. For this case, we can directly pass in the `tag` when constructing `samper`s, e.g.

```julia
samplers[i] = Sampler(sub_alg, model, Selector(Symbol(typeof(sub_alg))))
```

yebai:
> @KDr2 Sorry, I am late to the party! Thanks for the great PR. I left some comments. If you agree with me, please address them in another PR. Cheers!

I left some comments above. Other comments look reasonable to me and can be addressed by a new PR.

src/inference/mh.jl
KDr2: Fixed by #726.

src/inference/gibbs.jl
KDr2: Fixed by #726.

src/core/VarReplay.jl
KDr2: Fixed by #726.

src/core/VarReplay.jl
KDr2: Fixed by #726.

src/Turing.jl
```diff
@@ -62,6 +62,15 @@ end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
 
+struct Selector
+    gid :: UInt64
+    tag :: Ref{Symbol} # :default, :invalid, :Gibbs, :HMC, etc.
```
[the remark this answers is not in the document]
KDr2: Fixed by #726.

Issue #725, opened 2019-03-21
Run `Windows` tests on Travis CI 


yebai:
Travis CI now has [support](https://blog.travis-ci.com/2018-10-11-windows-early-release) for `Windows`. Given that Appveyor is painfully slow at the moment, we should probably try to replace Appveyor with Travis CI at some point. This allows us to use a unified test setup for all environments (Windows, Linux, OSX). 

Related: https://github.com/TuringLang/Turing.jl/issues/530

KDr2:
@TuringBenchBot bm("issue-725")
// just for test.

TuringBenchBot:
Hi Sir, a new benchmark job will be dispatched soon at your command,
you can track it here: https://github.com/TuringLang/TuringBenchmarks/issues/25.

TuringBenchBot:
Hi Sir, a new benchmark job will be dispatched soon at your command,
you can track it here: https://github.com/TuringLang/TuringBenchmarks/issues/26.

yebai:
Closed in favour of #823

PR #728, opened 2019-03-25
[WIP] Run test on windows@travis

KDr2:
#725

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,12 @@ jobs:
         - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
         - julia --check-bounds=yes -e 'using Pkg;
                    Pkg.test("Turing"; coverage=true)'
+    - stage: test
+      os: windows
+      language: c
+      env: STAGE=test
+      script:
+        - ./test/windows-travis.sh
     - stage: numerical
       allow_failures: true
       julia: 1.0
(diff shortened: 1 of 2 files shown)


KDr2:
After spending several hours on this task, I think the Windows platform of the Travis-CI is not available now:

1. The booting time of the virtual machine is not stable, sometimes it takes more than 5 mins to boot a virtual machine, sometimes it takes 10 secs to boot.
2. After the virtual machine boots, it hangs for a long time, usually longer the 5 mins, before it starts to the tasks.
3. It has a misery performance, many times the test fails with a "no output captured for more than 10 mins ...." message.
4. The most common situation, the job is marked as failed while it is still running, (at least looks like still running), no error message from the test cases, no error message from the CI platform, it just fails.

So I think it is not generally available so far.

I wrote a bash script to run the CI tasks in this PR, we can keep it for later use(when Travis's Windows is production ready) if possible.

If the reason we want to run test cases on Travis' Windows is that Appveyor is too slow, I think we can comment out the x86 platform and only keep x64 on Appveyor.

@yebai

KDr2:
See https://travis-ci.org/TuringLang/Turing.jl/jobs/510941663, the test job hangs and fails with no error.

yebai:
Closed in favour of https://github.com/TuringLang/Turing.jl/pull/823

PR #739, opened 2019-03-30
Replace HMC related codes using AHMC

xukai92:
Integrated the new AdvancedHMC.jl into Turing so we can start to test it.

Usage:
```julia
chn = sample(gdemo_default, ANUTS(10_000, 0.8), 1_000)
```

where `0.8` is the target accept ratio and `1_000` is the number of adaptation steps (happy to change the interface).

I feel we might want to play it with Turing for a while before replacing the current HMC codes completely.

TODOs
- ~~disable SGLD and SGHMC for now~~
  - Create a PR in AdvancedHMC.jl to move them there
- [x] move codes in `hmc_core.jl` to `hmc.jl`
- ~~remove `spl.info`~~
- [x] clean up codes
- [x] update doc strings


xukai92:
NOTE: ANUTS doesn't support compositional interface yet. 
Question: do we actually want dynamic HMC support that interface? I think @yebai and me had some discussion on the validation of the adaption phase of NUTS during Gibbs.

yebai:
> Question: do we actually want dynamic HMC support that interface? I think @yebai and me had some discussion on the validation of the adaption phase of NUTS during Gibbs.

Dual averaging should work fine within Gibbs, although might require more interactions before converging to a stable value. The no-U-turn criteria will be correct but might be suboptimal depending on the model structure.  I think we can try to integrate dual averaging to HMC within Gibbs as a first step (in a separate PR).

xukai92:
AHMC.jl is registered now so hopefully the tests shall pass. Also please don't merge this for now. I'm working with Jonas on his example of using DiffEq solvers together with Turing, which, works with Turing's NUTS but not AHMC's. Need to figure out why this happens.

xukai92:
I started an experimental interface; see https://github.com/TuringLang/Turing.jl/pull/739/commits/a32cec09d3acf5341d43a08c4c17e60ee62d8248#diff-4b1655bda3ebca1ef0e5a5a587d67152R42

xukai92:
CI should not pass until we tag the AHMC after https://github.com/TuringLang/AdvancedHMC.jl/pull/47 is merged.

yebai:
Is this ready for merging? @xukai92

xukai92:
TODOs
- [ ] tag a new release for AHMC
- [ ] sort out dependency files on Turing for AHMC

xukai92:
We are almost here (see the TODOs in https://github.com/TuringLang/Turing.jl/pull/739#issuecomment-484893504).

xukai92:
@yebai @mohamed82008 This is ready for a review.

src/inference/hmc.jl
```diff
@@ -80,25 +75,25 @@ function Sampler(alg::Hamiltonian, adapt_conf::Nothing, s::Selector)
...
 
     Sampler(alg, info, s)
 end
 
 function sample(model::Model, alg::Hamiltonian;
-                                save_state=false,                   # flag for state saving
-                                resume_from=nothing,                # chain to continue
-                                reuse_spl_n=0,                      # flag for spl re-using
-                                adapt_conf=STAN_DEFAULT_ADAPT_CONF, # adapt configuration
+                save_state=false,                                   # flag for state saving
```
yebai: Consider improving code style following Turing's style guide

src/inference/hmc.jl
```diff
@@ -114,107 +109,146 @@ function sample(model::Model, alg::Hamiltonian;
...
-    for i = 1:n
-        Turing.DEBUG && @debug "$alg_str stepping..."
+    # Init h, prop and adaptor
+    step(model, spl, vi, Val(true))
 
-        time_elapsed = @elapsed vi, is_accept = step(model, spl, vi, Val(i == 1))
-        time_total += time_elapsed
-
-        if is_accept # accepted => store the new predcits
+    # Sampling using AHMC
```
yebai: Consider a bit factoring here to replace the branching statements with a dispatching based design, e.g.

```julia
step(a::AdaptiveHamiltonian) # or step(a::NUTS)
step(a:: HMC)
step(a:: HMCDA)
```

or

```julia
s::VarInfo, a::Alg, d::Adaptor = step(a::HMC, d::AbstractAdaptor) 
```

yebai:
Thanks, @xukai92. This PR looks good. I left some comments above. Here are some further refactoring comments that I would consider:

- merge `support/hmc_core.jl`, `hmc.jl`, `hmcda.jl` and `nuts.jl` into a single file, i.e. `inference/hmc.jl`
- modify `stan-interface.jl` such that it returns a  `StanNUTSAdaptor` type. This would avoid passing around a `CmdStan` type in our code. https://github.com/TuringLang/Turing.jl/blob/kx/int-ahmc/src/utilities/stan-interface.jl 
- perhaps move `NUTS` `HMCDA` and `HMC` sampler type definition to AHMC? https://github.com/TuringLang/Turing.jl/blob/154069e41c6389a559b85de289f4605414a63e69/src/inference/nuts.jl#L36
   - After each adaption step, return a new sampler containing the full sampler configuration. This would create a fully immutable API design and enable us to remove `Sampler.info` safely.

src/inference/Inference.jl
```diff
@@ -230,14 +240,8 @@ end
...
-    if haskey(spl.info, :wum)
-        s.value[:epsilon] = getss(spl.info[:wum])
-    end
-    if haskey(spl.info, :lf_num)
-        s.value[:lf_num] = spl.info[:lf_num]
-    end
-    if haskey(spl.info, :eval_num)
-        s.value[:eval_num] = spl.info[:eval_num]
+    if haskey(spl.info, :adaptor)
+        s.value[:lf_eps] = AHMC.getϵ(spl.info[:adaptor])
```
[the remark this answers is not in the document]
xukai92: We have to add `h` and `prop` fields to either sampler or algorithm because of the Gibbs-compatible version still needs it.

src/inference/hmc.jl
xukai92: Could do `step(model, spl::Sampler{H}, vi::VarInfo) where {H}` version

xukai92:
TODO
- [ ] disable SGLD and SGHMC for now
  - Create a PR in AdvancedHMC.jl to move them there

xukai92:
@yebai It's ready for another look. I don't think I can remove `spl.info` easily, without making huge RFC oh either `Sampler` or other stuff. I'd do this in another PR. I also discussed this with Mohamed  and he also has some plan on this. I will create an issue on this and let discuss over there.

src/inference/support/hmc_core.jl
```diff
@@ -66,42 +65,11 @@ function _hmc_step(θ::AbstractVector{<:Real},
...
-    θ_new, r_new, is_valid = AHMC.step(prop.integrator, h, θ, r, prop.n_steps)
-    H_new = AHMC.hamiltonian_energy(h, θ_new, r_new)
-    # Accept via MH criteria
-    is_accept, α = AHMC.mh_accept(H, H_new)
-    if is_accept
-        θ, r = θ_new, -r_new
-    end
+    θ_new, r_new, α, H_new = AHMC.transition(prop, h, Vector{Float64}(θ), r)
+    # NOTE: as `transition` doesn't return `is_accept`, I use `H == H_new` as a check
+    is_accept = H != H_new  # If the new Hamiltonian enerygy is different
```
yebai: Would it be safer to test `θ == θ_new` here, instead of `H != H_new`?

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
...
+
+Arguments:
+
+- `n_iters::Int` : Number of samples to pull.
+- `n_adapts::Int` : Numbers of samples to use for adaptation.
+- `δ::Float64` : Target acceptance rate. 65% is often recommended.
+- `λ::Float64` : Target leapfrop length.
+- `init_ϵ::Float64=0.0` : Inital step size; 0 means automatically search by Turing.
+
+Example:
```
yebai: Consider removing the following doc section since they are generic for all samplers. Perhaps we can transfer them to the `sample` API once https://github.com/TuringLang/Turing.jl/issues/746#issuecomment-485222516 is fixed?

```julia

Example:
```julia
# Define a simple Normal model with unknown mean and variance.
@model gdemo(x) = begin
  s ~ InverseGamma(2,3)
  m ~ Normal(0, sqrt(s))
  x[1] ~ Normal(m, sqrt(s))
  x[2] ~ Normal(m, sqrt(s))
  return s, m
end
sample(gdemo([1.j_max, 2]), NUTS(1000, 200, 0.6j_max))
```

mohamed82008:
I will review this today.

src/inference/sghmc.jl
```diff
@@ -58,6 +54,7 @@ end
 function step(model, spl::Sampler{<:SGHMC}, vi::VarInfo, is_first::Val{false})
     # Set parameters
     η, α = spl.alg.learning_rate, spl.alg.momentum_decay
+    spl.info[:eval_num] = 0
```
mohamed82008: Why are we resetting `eval_num` here?

src/inference/sghmc.jl
```diff
@@ -32,10 +32,6 @@ mutable struct SGHMC{AD, T} <: StaticHamiltonian{AD}
     space::Set{T}    # sampling space, emtpy means all
 end
 SGHMC(args...) = SGHMC{ADBackend()}(args...)
-function SGHMC{AD}(learning_rate::Float64, momentum_decay::Float64, space...) where AD
```
mohamed82008: Why did we remove the limited `space` constructor? Isn't this used by Gibbs?

src/inference/sgld.jl
```diff
@@ -30,10 +30,6 @@ mutable struct SGLD{AD, T} <: StaticHamiltonian{AD}
     space   :: Set{T}    # sampling space, emtpy means all
 end
 SGLD(args...; kwargs...) = SGLD{ADBackend()}(args...; kwargs...)
-function SGLD{AD}(epsilon::Float64, space...) where AD
```
mohamed82008: Same `space` comment.

src/inference/sgld.jl
```diff
@@ -45,23 +41,22 @@ end
...
 
     spl.selector.tag != :default && invlink!(vi, spl)
     return vi, true
 end
 
 function step(model, spl::Sampler{<:SGLD}, vi::VarInfo, is_first::Val{false})
     # Update iteration counter
-    spl.info[:t] += 1
+    spl.info[:i] += 1
+    spl.info[:eval_num] = 0
```
mohamed82008: Same `eval_num` comment.

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
     epsilon   ::  Float64   # leapfrog step size
     tau       ::  Int       # leapfrog step number
     space     ::  Set{T}    # sampling space, emtpy means all
+    metricT
```
mohamed82008: Make a type parameter?

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
...
+Hoffman, Matthew D., and Andrew Gelman. "The No-U-turn sampler: adaptively setting path lengths in Hamiltonian Monte Carlo." Journal of Machine Learning Research 15, no. 1 (2014): 1593-1623.
+"""
+mutable struct HMCDA{AD, T} <: AdaptiveHamiltonian{AD}
+    n_iters   ::  Int       # number of samples
+    n_adapts  ::  Int       # number of samples with adaption for epsilon
+    δ         ::  Float64   # target accept rate
+    λ         ::  Float64   # target leapfrog length
+    space     ::  Set{T}    # sampling space, emtpy means all
+    init_ϵ    ::  Float64
+    metricT
```
mohamed82008: Please make this a type parameter.

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
...
+    gen_logπ(vi::VarInfo, spl::Sampler, model)
+
+Generate a function that takes `θ` and returns logpdf at `θ` for the model specified by
+`(vi, spl, model)`.
+"""
+function gen_logπ(vi::VarInfo, spl::Sampler, model)
+    function logπ(x)::Float64
+        x_old, lj_old = vi[spl], vi.logp
+        vi[spl] = x
+        runmodel!(model, vi, spl).logp
```
mohamed82008: `runmodel!(...).logp` -> `runmodel!(...)`

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
...
+        theta_mask = map(x -> !ismissing(x), init_theta_flat)
+        # Get all values
+        theta = vi[spl]
+        # Update those which are provided (i.e. not missing)
+        theta[theta_mask] .= init_theta_flat[theta_mask]
+        # Update in `vi`
+        vi[spl] = theta
+    end
+
+    # Convert to transformed sapce
```
mohamed82008: `sapce` -> `space`

src/inference/hmc.jl
```diff
@@ -48,190 +48,394 @@ mutable struct HMC{AD, T} <: StaticHamiltonian{AD}
...
+"""
+mutable struct NUTS{AD, T} <: AdaptiveHamiltonian{AD}
+    n_iters   ::  Int       # number of samples
+    n_adapts  ::  Int       # number of samples with adaption for epsilon
+    δ         ::  Float64   # target accept rate
+    space     ::  Set{T}    # sampling space, emtpy means all
+    max_depth ::  Int
+    Δ_max     ::  Float64
+    init_ϵ    ::  Float64
+    metricT
```
mohamed82008: Type parameter

src/inference/hmc.jl
```diff
@@ -262,6 +466,41 @@ function step(model, spl::Sampler{<:Hamiltonian}, vi::VarInfo, is_first::Val{fal
...
+        vi[spl] = ahmc_samples[i]
+        samples[i].value = Sample(vi, spl).value
+    end
+end
+
+function steps!(model, spl::Sampler{<:HMC}, vi, samples; rng::AbstractRNG=GLOBAL_RNG)
+    ahmc_samples =  AHMC.sample(rng, spl.info[:h], spl.info[:traj], Vector{Float64}(vi[spl]), spl.alg.n_iters)
+    for i = 1:length(samples)
+        vi[spl] = ahmc_samples[i]
+        samples[i].value = Sample(vi, spl).value
```
mohamed82008: `vi.logp` will be the same for all?

src/inference/hmc.jl
```diff
@@ -262,6 +466,41 @@ function step(model, spl::Sampler{<:Hamiltonian}, vi::VarInfo, is_first::Val{fal
...
+    end
+end
+
+function steps!(model, spl::Sampler{<:Hamiltonian}, vi, samples; rng::AbstractRNG=GLOBAL_RNG)
+    for i = 1:length(samples)
+        time_elapsed = @elapsed vi, is_accept = step(model, spl, vi, Val(i == 1))
```
mohamed82008: Unrolling the first iteration and using `Val(true)` and `Val(false)` will make this loop faster.

src/inference/sghmc.jl
mohamed82008: Why did we remove the partial `space` constructor?

src/inference/sghmc.jl
mohamed82008: Why is `eval_num` reset here?

src/inference/sgld.jl
mohamed82008: Why did we remove the partial space constructor?

src/inference/sgld.jl
mohamed82008: Why resetting `eval_num`?

src/inference/support/hmc_core.jl
xukai92: I guess it will be safer if we doing so. I just worried about the performance of comparing two vectors. In fact, the Hamiltonian energy should be a good enough approximation (before we actually return the `is_accept` from AHMC.jl) as not only the model parameter should be the same to make the Hamiltonian energy be the same but also a random momentum variable should by luck be the same, which I guess will not happen in practise.

src/inference/hmc.jl
xukai92: Good point. I was also thinking about removing the duplicated comments here. We should only leave examples for constructing algorithms here.

src/inference/sghmc.jl
xukai92: The original usage of `spl.info[:eval_num]` is to count how many times the model was executed for each `step()` call. But in fact these codes are removed now - I will add them back.

src/inference/sghmc.jl
xukai92: See motivation here: https://turingjl.slack.com/archives/CBUHPCAH1/p1555772916045000

src/inference/hmc.jl
xukai92: Yes - will do!

src/inference/hmc.jl
xukai92: Good catch

src/inference/hmc.jl
xukai92: When `Sample(vi, spl)` is called, it automatically copies the `logp` from `vi` to `value[:lp]`.

src/inference/hmc.jl
xukai92: I see. Nice point. I will change!

xukai92:
Thanks for the reviews. I will take a pass on fixing some of them tonight :)

src/inference/hmc.jl
mohamed82008: But unless the model is called on `vi`, this `logp` will be 0? Shouldn't the `logp` values be obtained from AHMC side?

xukai92: Oh sorry I see your point. I thought this comment was on the non-AHMC one. The interface there doesn't return `logp` so far. We may want to add it later.

yebai:
Ready for another review - tests pass on my local machine but Travis CI might fail before https://github.com/TuringLang/AdvancedHMC.jl/pull/61 is merged.

PR #740, opened 2019-03-30
TS 1: reorganize VarName and VarInfo and rename it to UntypedVarInfo

mohamed82008:
This is the first wave of the unwieldy type stability PR #660. The main point of this PR is to reorganize `VarInfo`, remove some unused fields and functions, and rename it to `UntypedVarInfo`. `VarName` is also modified to encode the symbol of the variable in the type as a type parameter. Feedback is welcome!

### DIFF
--- a/src/core/VarReplay.jl
+++ b/src/core/VarReplay.jl
@@ -11,7 +11,9 @@ import Base: string, isequal, ==, hash, getindex, setindex!, push!, show, isempt
 import Turing: link, invlink
 
 export  VarName, 
-        VarInfo, 
+        AbstractVarInfo,
+        VarInfo,
+        UntypedVarInfo,
         uid, 
         sym, 
         getlogp, 
@@ -36,11 +38,15 @@ export  VarName,
 ###########
 # VarName #
 ###########
-struct VarName
-  csym      ::    Symbol        # symbol generated in compilation time
-  sym       ::    Symbol        # variable symbol
-  indexing  ::    String        # indexing
-  counter   ::    Int           # counter of same {csym, uid}
+struct VarName{sym}
+    csym      ::    Symbol        # symbol generated in compilation time
+    indexing  ::    String        # indexing
+    counter   ::    Int           # counter of same {csym, uid}
+end
+VarName(csym, sym, indexing, counter) = VarName{sym}(csym, indexing, counter)
+
+function Base.getproperty(vn::VarName{sym}, f::Symbol) where {sym}
+    return f === :sym ? sym : getfield(vn, f)
 end
 
 # NOTE: VarName should only be constructed by VarInfo internally due to the nature of the counter field.
@@ -52,9 +58,10 @@ isequal(x::VarName, y::VarName) = hash(uid(x)) == hash(uid(y))
 ==(x::VarName, y::VarName)      = isequal(x, y)
 
 Base.string(vn::VarName) = "{$(vn.csym),$(vn.sym)$(vn.indexing)}:$(vn.counter)"
-Base.string(vns::Vector{VarName}) = replace(string(map(vn -> string(vn), vns)), "String" => "")
+Base.string(vns::Vector{<:VarName}) = replace(string(map(vn -> string(vn), vns)), "String" => "")
 
-sym(vn::VarName) = Symbol("$(vn.sym)$(vn.indexing)")  # simplified symbol
+sym_idx(vn::VarName) = Symbol("$(vn.sym)$(vn.indexing)")  # simplified symbol
+getsym(vn::VarName{sym}) where sym = sym
 
 cuid(vn::VarName) = (vn.csym, vn.sym, vn.indexing)    # the uid which is only available at compile time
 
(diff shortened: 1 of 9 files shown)


mohamed82008:
I believe this is ready for review.

test/core/VarReplay.jl
```diff
@@ -94,68 +94,20 @@ include("../test_utils/AllUtils.jl")
...
-        @test getretain(vi, spl1) == UnitRange[4:4,5:5,6:6]
         vi.num_produce += 1
         randr(vi, vn_z2, dists[1], spl1)
-        @test getretain(vi, spl1) == UnitRange[6:6]
         vi.num_produce += 1
         randr(vi, vn_z3, dists[1], spl1)
         randr(vi, vn_a2, dists[2], spl1)
         @test vi.orders == [1, 1, 2, 2, 3, 3]
         @test vi.num_produce == 3
-
```
yebai: why we are removing these tests?

src/core/VarReplay.jl
```diff
@@ -36,11 +38,15 @@ export  VarName,
 ###########
 # VarName #
 ###########
-struct VarName
-  csym      ::    Symbol        # symbol generated in compilation time
-  sym       ::    Symbol        # variable symbol
-  indexing  ::    String        # indexing
-  counter   ::    Int           # counter of same {csym, uid}
+struct VarName{sym}
+    csym      ::    Symbol        # symbol generated in compilation time
```
yebai: `sym` stores the original variable name, while `csym` stores a randomly generated name. If we remove `sym`, the original variable name is lost. Do you plan to add an alternative mechanism for storing original variable information?

src/core/VarReplay.jl
```diff
@@ -11,7 +11,9 @@ import Base: string, isequal, ==, hash, getindex, setindex!, push!, show, isempt
 import Turing: link, invlink
```
yebai: Consider renaming `VarReplay.jl` to `VarInfo.jl`, and renamaing submodule `VarReplay` to `VarInfo`.

src/core/VarReplay.jl
mohamed82008: `sym` is now part of the type of the `VarName`, it is not removed.

test/core/VarReplay.jl
mohamed82008: They only test `getretain` which is not used anywhere in Turing, so I removed it.

yebai: I think this was originally used by the Gibbs sampler/particle Gibbs sampler. 

@xukai92 do we have a replacement for `getretrain` now?

src/core/VarReplay.jl
mohamed82008: Can we call the module `Parameters`, `ParametersInfo`, `Variables`, `VariablesInfo` or `VarRecords` instead? Any other names are welcome too. `VarInfo` is used to construct `UntypedVarInfo` using `VarInfo()`, and will be used to construct `TypedVarInfo` using `VarInfo(m::Model)`. I like `VarInfo` for the constructor because it is short and we are used to it.

test/core/VarReplay.jl
xukai92: Just checked the code. It seems that the functionality is replaced by `is_flagged(vi, vn, "del")`.

At each PG step or when task is coped, `set_retained_vns_del_by_spl!(vi, spl)` is triggered to flag all variables for the sampler as `del`.

When the model is run, `assume` function behave depends on `is_flagged(vi, vn, "del")` and whether or not the variable belongs to the sampler.

The logic looks correct to me.

I'm not sure if it's me or @emilemathieu did this change.

test/core/VarReplay.jl
```diff
@@ -94,68 +94,20 @@ include("../test_utils/AllUtils.jl")
...
         @test vi.orders == [1, 1, 2, 2, 2, 3]
         @test vi.num_produce == 3
-        @test getretain(vi, spl1) == UnitRange[]
 
-        # Check getretain at different stages
-        # variables samples in different order: z1,a1,z2,z3,a2
         vi.num_produce = 0
-        @test getretain(vi, spl1) == UnitRange[6:6,5:5,4:4,2:2,1:1]
-        @test getretain(vi, spl2) == UnitRange[3:3]
         set_retained_vns_del_by_spl!(vi, spl1)
```
xukai92: Can we test if `is_flagged(vi, vn, "del")` is `true` for the corresponding `vns`?

src/core/VarReplay.jl
yebai: Shall we call it `RandomVariables` (or `RanVars` if you want something mysterious)? This makes the purpose of the submodule clearer than `VarInfo` actually.

xukai92: I'm up to `RandomVariables`!

test/core/VarReplay.jl
yebai: sounds good to adopt these tests for testing `is_flagged(vi, vn, "del")`, since it's important to ensure the functionality adding/deleting variables is correct.

mohamed82008: Done.

src/core/VarReplay.jl
mohamed82008: If no objections happen or better suggestions get made, I will make the change tomorrow.

PR #741, opened 2019-03-30
Update README.md

yebai:


cpfiffer:
I think we should probably only point people to the releases page since it has a better and more relevant history. The `RELEASES.md` file seems like it replicates the information in the release feed.

PR #742, opened 2019-03-31
TS 2: introduce TypedVarInfo and fix spl.info[:cache_updated]

mohamed82008:
This is the second wave of #660. In this PR:
1. The `TypedVarInfo` type is introduced with all its utility functions mapping that of `UntypedVarInfo`.
2. Unit tests mapping those of `UntypedVarInfo` are also defined.
3. One side change that was made in this PR is to make sure to invalidate the sampler's cache every time `push!` is called on `VarInfo`. This seemed like a latent bug that was "worked around" by invalidating the cache at the call site after calling `push!`. 
4. `getvals(vi, spl)` was removed as it seemed to be doing the same thing as `vi[spl]` and it was unused in the rest of Turing.

Your feedback is appreciated.


mohamed82008:
TODO: replace most of the generated functions with recursive functions. Example:
```julia
@generated function _getranges(vi::TypedVarInfo{Tvis}, idcs) where Tvis
    args = []
    for f in fieldnames(Tvis)
        push!(args, :($f = _map(vi, $(QuoteNode(f)), idcs.$f)))
    end
    if length(args) == 0
        nt = :(NamedTuple())
    else
        nt = :(($(args...),))
    end
    return nt
end
```
can be replaced with:
```julia
_getranges(vi::TypedVarInfo, idcs) = _getranges(vi.vis, vi, idcs)
@inline function _getranges(vis::NamedTuple{names}, vi::TypedVarInfo, idcs) where names
	length(names) === 0 && return NamedTuple()
	f = names[1]
	v = _map(vi, f, getfield(idcs, f))
	nt = NamedTuple{(f,), Tuple{typeof(v)}}(v)
	return merge(nt, _getranges(Base.tail(vis), vi, idcs))
end
```

yebai:
@mohamed82008 is this ready for a review?

mohamed82008:
Yes if you don't mind the generated functions. The semantics will not change by making them into normal functions.

src/core/RandomVariables.jl
```diff
@@ -118,32 +130,224 @@ mutable struct UntypedVarInfo <: AbstractVarInfo
...
+        sym_inds = collect(1:length(_inds))
+        sym_vns = getindex.((vi.vns,), _inds)
+        sym_idcs = Dict(a => i for (i, a) in enumerate(sym_vns))
+        sym_dists = getindex.((vi.dists,), _inds)
+        sym_gids = getindex.((vi.gids,), _inds)
+        sym_orders = getindex.((vi.orders,), _inds)
+        sym_flags = Dict(a => vi.flags[a][_inds] for a in keys(vi.flags))
+
+        _ranges = getindex.((vi.ranges,), _inds)
+        # `copy` is a workaround to reduce the eltype from Real to Int or Float64
```
xukai92: Do you mean `copy(::Vector{Real})` gives `Vector{Float64}` if each element is `Float64`?

test/core/RandomVariables.jl
```diff
@@ -190,41 +231,54 @@ include("../test_utils/AllUtils.jl")
         @test cuid(vn1) == cuid(vn2)
         @test vn11 == vn1
 
-        vi = VarInfo()
         dists = [Normal(0, 1), MvNormal([0; 0], [1.0 0; 0 1.0]), Wishart(7, [1 0.5; 0.5 1])]
+        function test_varinfo!(vi)
```
xukai92: Would it be nice to improve the tests here a bit, e.g. adding more comments on what's testing and expected, and improving the code (variable names, code reuse, etc).

test/core/RandomVariables.jl
```diff
@@ -1,9 +1,9 @@
 using Turing, Random
 using Turing: Selector, reconstruct, invlink, CACHERESET, SampleFromPrior
 using Turing.RandomVariables
-using Turing.RandomVariables: uid, cuid, getvals, getidcs,
+using Turing.RandomVariables: uid, cuid, getidcs,
```
xukai92: Are we aware of the test coverage of the `TypedVarInfo`?

src/core/RandomVariables.jl
```diff
@@ -67,9 +67,21 @@ cuid(vn::VarName) = (vn.csym, vn.sym, vn.indexing)    # the uid which is only av
 
 copybyindex(vn::VarName, indexing::String) = VarName(vn.csym, vn.sym, indexing, vn.counter)
 
-###########
-# VarInfo #
-###########
+function is_inside(vn::VarName, space::Set)::Bool
+    if vn.sym in space
+        return true
+    else
```
xukai92: Let's add some comments to this old function. E.g. here we can say "1. collect expressions from `space`. 2. convert them to strings (`strs`). 3. convert `vn` to string (`vn_str`). 4. check if `vn_str` is in the `strs`.

Also maybe rename `strs` to `expr_strs`?

src/core/RandomVariables.jl
```diff
@@ -118,32 +130,224 @@ mutable struct UntypedVarInfo <: AbstractVarInfo
 end
 VarInfo() = UntypedVarInfo()
 
+###########################
+# Single variable VarInfo #
```
xukai92: Does `SingleVarInfo` only contains a single random variable? If so why does `ranges` required for `SingleVarInfo` and fields like `dists`, `gids` are still vectors?

src/core/RandomVariables.jl
```diff
@@ -170,6 +380,27 @@ function link!(vi::UntypedVarInfo, spl::Sampler)
...
     end
 end
+function link!(vi::TypedVarInfo, spl::Sampler)
+    vns = getvns(vi, spl)
+    space = getspace(spl)
+    return _link!(vi.vis, vi, vns, space)
+end
+@inline function _link!(vis::NamedTuple{names}, vi, vns, space) where {names}
+    length(names) === 0 && return nothing
+    f = names[1]
```
xukai92: A bit confused by the variables here, what is `f`? Also we are calling `getfield(vns, f)` multiple times later, can we assign it to some local variable with a name?

src/core/RandomVariables.jl
```diff
@@ -184,9 +415,30 @@ function invlink!(vi::UntypedVarInfo, spl::Sampler)
...
     end
 end
+function invlink!(vi::TypedVarInfo, spl::Sampler)
+    vns = getvns(vi, spl)
+    space = getspace(spl)
+    return _invlink!(vi.vis, vi, vns, space)
+end
+@inline function _invlink!(vis::NamedTuple{names}, vi, vns, space) where {names}
+    length(names) === 0 && return nothing
+    f = names[1]
```
xukai92: Similar here.

src/inference/is.jl
```diff
@@ -58,7 +58,7 @@ end
 
 function assume(spl::Sampler{<:IS}, dist::Distribution, vn::VarName, vi::VarInfo)
     r = rand(dist)
-    push!(vi, vn, r, dist, spl.selector)
+    push!(vi, vn, r, dist, spl)
```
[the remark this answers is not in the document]
mohamed82008: No we are still pushing the same way. The reason why I am passing `spl` here is to reset the `:cache_updated` field inside the function.

src/core/RandomVariables.jl
mohamed82008: `copy.(::Vector{Real})` gives `Vector{Float64}` if each element is `Float64`. note the dot.

test/core/RandomVariables.jl
mohamed82008: I will add more comments to the src too.

test/core/RandomVariables.jl
mohamed82008: I made sure the test coverage of `TypedVarInfo` is no less than that of `UntypedVarInfo`. But I will see if I can improve it.

src/core/RandomVariables.jl
mohamed82008: Sure.

src/core/RandomVariables.jl
mohamed82008: In case we have a vector variable where each element has a different prior distribution, so I think `dists` still needs to be a vector. I think `gids` can be shared however. Not sure about `ranges`, in case we have a matrix variable whose columns are multivariate variables, we probably still need it.

src/core/RandomVariables.jl
mohamed82008: `f` is the first variable's symbol.

mohamed82008: But using recursion, this "first" variable will be a different variable in each level of the recursion.

mohamed82008:
I am working on docstrings for all the major functions and types in the module `RandomVariables`.

mohamed82008:
To do:
[ ] - More unit tests for `TypedVarInfo` and `UntypedVarInfo`
[ ] - More comments and docstrings
[ ] - Make `gids` for `TypedVarInfo` a resizeable `FillArray` type

src/core/RandomVariables.jl
xukai92: I see. Great thanks!

src/core/RandomVariables.jl
xukai92: Hmmm, I see your point, though I guess we should discourage users to do that. And I'm not sure whether it makes or not for a single multivariate random variable whose different dimension follows different distributions.  But I'm OK with it for now.

src/core/RandomVariables.jl
xukai92: But for the loop below we are calling `getfield(vns, f)` with same `vns` and `f` all the time do we?

mohamed82008: Yes, so we are iterating over all the `vn`s of the first symbol, then all the `vn`s of the second, and so on. Each symbol has a `vns` vector because it can consist of multiple random variables, e.g. multiple univariate variables arranged in vector form or multiple multivariate variables arranged in matrix form, etc. The last line is where recursion happens.

src/core/RandomVariables.jl
```diff
@@ -83,97 +168,514 @@ function Turing.runmodel!(model::Model, vi::AbstractVarInfo, spl::AbstractSample
...
+struct VarInfo{Tmeta, Tlogp} <: AbstractVarInfo
+    metadata::Tmeta
+    logp::Base.RefValue{Tlogp}
+    num_produce::Base.RefValue{Int}
+end
+```
 
+A light wrapper over one or more instances of `Metadata`. Let `vi` be an instance of `Metadata`. If `vi isa VarInfo{<:Metadata}`, then only `Metadata` instance is used for all the sybmols. `VarInfo{<:Metadata}` is aliased `UntypedVarInfo`. If `vi isa VarInfo{<:NamedTuple}`, then `vi.metadata` is a `NamedTuple` that maps each symbol used on the LHS of `~` in the model to its `Metadata` instance. The latter allows for the type specialization of `vi` after the first sampling iteration when all the symbols have been observed. `VarInfo{<:NamedTuple}` is aliased `TypedVarInfo`.
+
+Note: It is the user's responsibility to ensure that each symbol is visited at least once whenever the model is called, regardless of any stochastic branching.
```
yebai: presumably, it's ok if a variable is visited multiple times since they can have different `counter`? If so, can we say this explicitly in the comment to reduce potential confusion?

src/inference/is.jl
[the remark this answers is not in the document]
xukai92: In you opinion, do you think this cache thing is worth doing here at all. I feel it might be over-optimisation here...

src/core/RandomVariables.jl
```diff
@@ -208,19 +745,79 @@ function Base.getindex(vi::AbstractVarInfo, vns::Vector{<:VarName})
...
+"""
+`getindex(vi::VarInfo, spl::Union{SampleFromPrior, Sampler})`
+
+Returns the current value(s) of the random variables sampled by `spl` in `vi`. The value(s) may or may not be transformed to Eucledian space.
+"""
+Base.getindex(vi::AbstractVarInfo, spl::SampleFromPrior) = copy(getall(vi))
+Base.getindex(vi::UntypedVarInfo, spl::Sampler) = copy(getval(vi, _getranges(vi, spl)))
+function Base.getindex(vi::TypedVarInfo, spl::Sampler)
+    # Gets the ranges as a NamedTuple
+    # getfield(ranges, f) is all the indices in `vals` of the `vn`s with symbol `f` sampled by `spl` in `vi`
```
xukai92: There is no `getfield(ranges, f)` any more.

src/core/RandomVariables.jl
```diff
@@ -208,19 +745,79 @@ function Base.getindex(vi::AbstractVarInfo, vns::Vector{<:VarName})
...
+    ranges = _getranges(vi, spl)
+    return vcat(_get(vi.metadata, ranges)...)
+end
+# Recursively builds a tuple of the `vals` of all the symbols
+@inline function _get(metadata::NamedTuple{names}, ranges) where {names}
```
xukai92: Maybe name it as `_getindex` as it's only used by `Base.getindex`.

src/core/RandomVariables.jl
```diff
@@ -262,136 +880,265 @@ function push!(vi::UntypedVarInfo, vn::VarName, r::Any, dist::Distributions.Dist
...
+    push!(meta.dists, dist)
+    push!(meta.gids, gidset)
+    push!(meta.orders, vi.num_produce)
+    push!(meta.flags["del"], false)
+    push!(meta.flags["trans"], false)
+
+    return meta
+end
 
+"""
```
xukai92: Bookmark for Kai.

src/core/RandomVariables.jl
mohamed82008: Sure

src/core/RandomVariables.jl
mohamed82008: `ranges` refers to the output variable from `_getranges`. Calling `getfield(ranges, f)` is what this line refers to. I will try to make it clearer.

src/core/RandomVariables.jl
```diff
@@ -83,97 +168,514 @@ function Turing.runmodel!(model::Model, vi::AbstractVarInfo, spl::AbstractSample
...
 
 # X -> R for all variables associated with given sampler
+"""
+`link!(vi::VarInfo, spl::Sampler)`
+
+Transforms the values of the random variables sampled by `spl` in `vi` from the support of their distributions to the Eucledian space and sets their corresponding ``"trans"` flag values to `true`.
+"""
 function link!(vi::UntypedVarInfo, spl::Sampler)
-    vns = getvns(vi, spl)
+    # TODO: Change to a lazy iterator over `vns`
```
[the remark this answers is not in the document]
mohamed82008: As in we don't actually need to materialize `vns`, we can have an iterator over it.

src/core/RandomVariables.jl
```diff
@@ -83,97 +168,514 @@ function Turing.runmodel!(model::Model, vi::AbstractVarInfo, spl::AbstractSample
...
 
+"""
+`getlogp(vi::VarInfo)`
+
+Returns the log of the joint probability of the observed data and parameters sampled in `vi`.
+"""
 getlogp(vi::AbstractVarInfo) = vi.logp
+
+"""
+`setlogp!(vi::VarInfo, logp::Real)`
```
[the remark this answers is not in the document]
mohamed82008: We can overload `setproperty!` for `:log` and default that to an error, but that's probably overkill unless there is a good reason to stop people from doing this.

src/core/RandomVariables.jl
```diff
@@ -83,97 +168,514 @@ function Turing.runmodel!(model::Model, vi::AbstractVarInfo, spl::AbstractSample
...
+- `md.ranges[md.idcs[vn]]` is the index range of `vn` in `md.vals`.
+- `md.vals[md.ranges[md.idcs[vn]]]` is the vector of values of corresponding to `vn`.
+- `md.flags` is a dictionary of true/false flags. `md.flags[flag][md.idcs[vn]]` is the value of `flag` corresponding to `vn`. 
+
+To make `md::Metadata` type stable, all the `md.vns` must have the same symbol and distribution type. However, one can have a Julia variable, say `x`, that is a matrix or a hierarchical array sampled in partitions, e.g. `x[1][:] ~ MvNormal(zeros(2), 1.0); x[2][:] ~ MvNormal(ones(2), 1.0)` and is managed by a single `md::Metadata` so long as all the distributions on the RHS of `~` are of the same type. Type unstable `Metadata` will still work but will have inferior performance. When sampling, the first iteration uses a type unstable `Metadata` for all the variables then a specialized `Metadata` is used for each symbol along with a function barrier to make the rest of the sampling type stable.
+"""
+struct Metadata{TIdcs <: Dict{<:VarName,Int}, TDists <: AbstractVector{<:Distribution}, TVN <: AbstractVector{<:VarName}, TVal <: AbstractVector{<:Real}, TGIds <: AbstractVector{Set{Selector}}}
```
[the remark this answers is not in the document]
mohamed82008: Hmm, avoiding the flattening is an interesting idea. I am not sure what the best way to handle this is. Should probably open an issue to discuss it.

src/core/RandomVariables.jl
```diff
@@ -83,97 +168,514 @@ function Turing.runmodel!(model::Model, vi::AbstractVarInfo, spl::AbstractSample
...
+"""
+```
+struct VarInfo{Tmeta, Tlogp} <: AbstractVarInfo
+    metadata::Tmeta
+    logp::Base.RefValue{Tlogp}
+    num_produce::Base.RefValue{Int}
+end
+```
 
+A light wrapper over one or more instances of `Metadata`. Let `vi` be an instance of `Metadata`. If `vi isa VarInfo{<:Metadata}`, then only `Metadata` instance is used for all the sybmols. `VarInfo{<:Metadata}` is aliased `UntypedVarInfo`. If `vi isa VarInfo{<:NamedTuple}`, then `vi.metadata` is a `NamedTuple` that maps each symbol used on the LHS of `~` in the model to its `Metadata` instance. The latter allows for the type specialization of `vi` after the first sampling iteration when all the symbols have been observed. `VarInfo{<:NamedTuple}` is aliased `TypedVarInfo`.
```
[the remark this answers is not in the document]
mohamed82008: Thanks.

src/core/RandomVariables.jl
mohamed82008: Each random variable can only be visited once in a model call. But the symbol of the Julia variable can be visited more than once, e.g. `x[1] ~ ...` and `x[2] ~ ...`. This line refers to the symbol `x` not the random variables `x[1]` and `x[2]`. The requirement here is that each symbol, e.g. `x`, is visited at least once. I will try to make this clearer.

src/core/RandomVariables.jl
```diff
@@ -60,20 +108,57 @@ isequal(x::VarName, y::VarName) = hash(uid(x)) == hash(uid(y))
 Base.string(vn::VarName) = "{$(vn.csym),$(vn.sym)$(vn.indexing)}:$(vn.counter)"
 Base.string(vns::Vector{<:VarName}) = replace(string(map(vn -> string(vn), vns)), "String" => "")
 
+"""
+`sym_idx(vn::VarName)`
```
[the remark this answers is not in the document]
mohamed82008: It's the inverse of it.

src/core/RandomVariables.jl
```diff
@@ -38,20 +38,68 @@ export  VarName,
...
+"""
+```
+struct VarName{sym}
+    csym      ::    Symbol
+    indexing  ::    String
+    counter   ::    Int
+end
+```
+
+A variable identifier. Every variable has a symbol `sym`, indices `indexing`, and internal fields: `csym` and `counter`. The Julia variable in the model corresponding to `sym` can refer to a single value or to a hierarchical array structure of univariate, multivariate or matrix variables. `indexing` stores the indices that can access the random variable from the Julia variable.
```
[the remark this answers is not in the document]
mohamed82008: Will do.

src/core/RandomVariables.jl
```diff
@@ -38,20 +38,68 @@ export  VarName,
...
+    indexing  ::    String
+    counter   ::    Int
+end
+```
+
+A variable identifier. Every variable has a symbol `sym`, indices `indexing`, and internal fields: `csym` and `counter`. The Julia variable in the model corresponding to `sym` can refer to a single value or to a hierarchical array structure of univariate, multivariate or matrix variables. `indexing` stores the indices that can access the random variable from the Julia variable. 
+
+Examples:
+
+- `x[2] ~ Normal()` will generate a `VarName` with `sym == :x` and `indexing == "[1]"`
```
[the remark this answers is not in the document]
mohamed82008: Thanks.

src/inference/is.jl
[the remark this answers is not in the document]
mohamed82008: I think it is important in cases where the `VarInfo` is "filled" most of the time so the cache is valid most of the time. This is the case for most samplers except the particle ones I believe.

src/core/RandomVariables.jl
```diff
@@ -38,20 +38,74 @@ export  VarName,
...
+"""
+`VarName(csym, sym, indexing, counter)`
+`VarName{sym}(csym::Symbol, indexing::String)`
+
+Constructs a new instance of `VarName{sym}``
+"""
 VarName(csym, sym, indexing, counter) = VarName{sym}(csym, indexing, counter)
+function VarName(csym::Symbol, sym::Symbol, indexing::String)
+    # TODO: update this method when implementing the sanity check
+    VarName{sym}(csym, indexing, 1)
```
trappmartin: Is it correct that the counter is always `1` in the constructor?

src/core/RandomVariables.jl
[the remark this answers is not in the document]
trappmartin: I'm happy with removing those setter and getter methods if we don't need them. I don't see much benefit.

yebai:
Current exported API

https://github.com/TuringLang/Turing.jl/blob/mt/ts_wave2_untypedvarinfo/src/core/RandomVariables.jl#L13-#L36

mohamed82008:
I think we need to do the following in a separate PR:
1. Find which functions we need outside of `RandomVariables` to define the "necessary" API.
2. For internal functions, we can use any names we like.
3. For functions needed outside, we can try to find `Base` functions with similar enough semantics and overload those instead to minimize the exported names and make it easier to remember.

yebai:
> I think we need to do the following in a separate PR:

Sounds good to do the API redesign in a separate PR.

src/core/RandomVariables.jl
mohamed82008: I haven't changed this behavior. Maybe @xukai92 knows better.

mohamed82008:
I split the exported functions and internal functions into 2 separate files. In the future, we can see how to make the exported functions more "macroscopic" in the sense that they should serve high-level functions in the inference code. This means less fiddling with Core internals in the Inference module and therefore hopefully clearer and more readable code.

Once tests pass, I believe this PR will be ready.

yebai:
> I split the exported functions and internal functions into 2 separate files. In the future, we can see how to make the exported functions more "macroscopic" in the sense that they should serve high-level functions in the inference code. This means less fiddling with Core internals in the Inference module and therefore hopefully clearer and more readable code.

@mohamed82008 Great work - many thanks! The general idea of making submodules more "macroscopic" makes a lot of sense and we should reinforce this principle when moving forward. I'm ready to merge this PR once @xukai92 is happy.

mohamed82008:
Rebased on master

xukai92:
Looks great to me. 

The `@code_warntype` check doesn't pass on my local, althought it's not the issue from this PR. 

```julia
 @model gdemo_d() = begin
    s ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s))
    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))
    return s, m
end

mf = gdemo_d()

@code_warntype sample(mf, NUTS(2000, 0.8))
Body::Any
1 ─ %1 = invoke Turing.Inference.AHMCAdaptor(_3::NUTS{Turing.Core.ForwardDiffAD{40},Any})::AdvancedHMC.Adaptation.StanNUTSAdaptor
│   %2 = (Turing.Inference.:(#sample#19))(false, Turing.Inference.nothing, 0, %1, Turing.Inference.nothing, Turing.Inference.GLOBAL_RNG, $(QuoteNode(Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}())), #self#, model, alg)::Any
└──      return %2
```

xukai92:
OK it seems only the return type is not stable but not during sampling?

mohamed82008:
No, this PR doesn't hook into `sample` yet, this will be the next PR. And even if I hook them, `sample` won't be type stable; we need to address the `info` field of `Sampler` and the `value` field of `Sample` to get complete type stability from the second iteration onwards. Currently, `spl.info` is `Dict{Symbol, Any}` so when we retrieve values from it during sampling, the compiler can't infer its type and this muddies the rest of the sampling function. So 2 more PRs are needed to get complete type stability from the second iteration onwards for all algorithms: 
1. Hooking `TypedVarInfo` to `sample`, and 
2. Making `spl.info` and `sample.value` type stable.

mohamed82008:
Something weird happened when I merged, I will fix the error.

yebai:
> Making spl.info and sample.value type stable.

This issue should go away after #746 is fixed.

PR #743, opened 2019-03-31
[WIP] Compatiblilty of `MvNormal` and `Tracker`

### LINKED ISSUE #717: `MvNormal` and `Tracker`
We still can't compute the `logpdf` of a multivariate Normal distribution and backprop using `Tracker`, which is a bit of a shame. Here's a MWE:
```julia
]dev Tracker
using Tracker, PDMats, LinearAlgebra

A = randn(3, 3);
B = A * A' + I;
x, m = randn(3), randn(3), B;

y, back = Tracker.forward((x, m, B)->logpdf(MvNormal(m, PDMat(B)), x), x, m, B)
ERROR: MethodError: no method matching PDMat(::TrackedArray{…,Array{Float64,2}})
Closest candidates are:
  PDMat(::AbstractArray{T,2} where T, ::Cholesky) at /home/wct23/.julia/dev/PDMats/src/pdmat.jl:13
  PDMat(::Array{T,2} where T) at /home/wct23/.julia/dev/PDMats/src/pdmat.jl:19
  PDMat(::Symmetric) at /home/wct23/.julia/dev/PDMats/src/pdmat.jl:20
  ...
Stacktrace:
 [1] (::getfield(Main, Symbol("##7#8")))(::TrackedArray{…,Array{Float64,1}}, ::TrackedArray{…,Array{Float64,1}}, ::TrackedArray{…,Array{Float64,2}}) at ./REPL[13]:1
 [2] #20 at /home/wct23/.julia/dev/Tracker/src/back.jl:148 [inlined]
 [3] forward(::getfield(Tracker, Symbol("##20#22")){getfield(Main, Symbol("##7#8"))}, ::Params) at /home/wct23/.julia/dev/Tracker/src/back.jl:135
 [4] forward(::Function, ::Array{Float64,1}, ::Array{Float64,1}, ::Vararg{Any,N} where N) at /home/wct23/.julia/dev/Tracker/src/back.jl:148
 [5] top-level scope at none:0
```
I'm not entirely sure what the issue _actually_ is here if I'm honest.

(Note that `Tracker` has recently been split off from `Flux` in preparation for `Zygote`, presumably becoming the default AD, hence the lack of `using Flux.Tracker`.)

KDr2:
Try to fix #717 on the Turing side.

I will file a PR on the Distributions or Tracker side later, but I think there will be a fair time-window before the fix on their side.

### DIFF
--- a/src/core/ad_ext.jl
+++ b/src/core/ad_ext.jl
@@ -10,3 +10,22 @@ Distributions.logpdf(
     dist::Distribution{Multivariate,S} where S<:ValueSupport,
     ta::TrackedArray{T,N,A} where A<:AbstractArray{T, N} where {T, N}
 ) = logpdf(dist, [e for e in ta])
+
+
+mvnormlogpdf(u::Array, M::Array, x::Array) = logpdf(MvNormal(u, PDMat(M)), x)
+mvnormlogpdf(u::Tracker.TrackedArray, M::Tracker.TrackedArray, x::Tracker.TrackedArray) =
+    Tracker.track(mvnormlogpdf, u, M, x)
+
+Tracker.@grad function mvnormlogpdf(
+    u::Tracker.TrackedArray,
+    M::Tracker.TrackedArray,
+    x::Tracker.TrackedArray
+)
+    return mvnormlogpdf(Tracker.data(u), Tracker.data(M), Tracker.data(x)), function(Δ)
+        inv_M = inv(M)
+        z = u - x
+        du = - inv_M * z
+        dM = (inv_M * z * z' * inv_M - inv_M)/2
+        (du, dM, -du) .* Δ
+    end
+end
--- a/src/core/ad_ext.jl
+++ b/src/core/ad_ext.jl
@@ -1,0 +1,12 @@
+using Tracker, Distributions, PDMats
+
+# The code below introduces Tracker and Distributions to each other to
+# make them work well together. The code can be removed after that issue
+# is fixed on the Distributions side.
+
+PDMats.PDMat(ta::TrackedArray) = PDMat([e for e in ta])
+Distributions.MvNormal(ta::TrackedArray, mat::PDMat) = MvNormal([e for e in ta], mat)
+Distributions.logpdf(
+    dist::Distribution{Multivariate,S} where S<:ValueSupport,
+    ta::TrackedArray{T,N,A} where A<:AbstractArray{T, N} where {T, N}
+) = logpdf(dist, [e for e in ta])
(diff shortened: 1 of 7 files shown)


src/core/ad_ext.jl
```diff
@@ -0,0 +1,12 @@
+using Tracker, Distributions, PDMats
+
+# The code below introduces Tracker and Distributions to each other to
+# make them work well together. The code can be removed after that issue
+# is fixed on the Distributions side.
+
+PDMats.PDMat(ta::TrackedArray) = PDMat([e for e in ta])
```
trappmartin: We should probably add a note that this solution is inefficient.

KDr2: OK. 

Another question, we can use `ta.data` to replace `[e for e in ta]`, but `ta.data` is an ordinary (not tracked) array of ordinary (not tracked) elements. Is it OK for our case?

willtebbutt:
Thanks for the PR. I would recommend taking a look [here](https://github.com/FluxML/Zygote.jl/blob/6b1b80faa354d46a9b448f83ed42b62fdbc63a94/src/lib/array.jl#L192) for an efficient implementation of the `cholesky` that you should be able to adapt for our purposes here. You might have to extend it slightly, but it's definitely necessary for performance.

src/core/ad_ext.jl
mohamed82008: I don't think this is the correct fix. We should just teach `cholesky` how to operate on `TrackedArray` IMO. `ta.data` will drop the tracking which is important not to break the differentiation chain.

trappmartin: I agree.

mohamed82008:
Here is an implementation of the gradient of the lower triangle of the Cholesky, [ref](https://homepages.inf.ed.ac.uk/imurray2/pub/16choldiff/choldiff.pdf):

```julia
phi(x, i, j) = i < j ? zero(x) : (i > j ? x : x/2)  
Phi(M) = phi.(M, 1:size(M,1), (1:size(M,2))')
lowerchol(M::Matrix) = cholesky(M).L
lowerchol(M::TrackedMatrix) = Tracker.track(lowerchol, M)
Tracker.@grad function lowerchol(M::AbstractMatrix)
	L = cholesky(Tracker.data(M)).L
	invL = inv(L)
	grad = delta -> (L*Phi(invL*delta*invL'),)
	return L, grad
end

function f(x)
	@assert length(x) > 2
	n = length(x)
	A = x*x' + length(x)*I
	L = lowerchol(A)
	return norm(L, 2)
end

x = rand(3)
l_tracked, ȳ = Tracker.forward(f, x)
l, dl = Tracker.data(l_tracked), Tracker.data(ȳ(1)[1])
isapprox(dl, ForwardDiff.gradient(f, x))
```

yebai:
Closed in favour of https://github.com/TuringLang/Turing.jl/pull/749

PR #744, opened 2019-04-01
Fix a bug in `find_good_eps`

xukai92:
I guess we lost these lines of reverting theta and log-joint during some previous refactoring...


yebai:
@xukai92 seems some tests are broken with this PR.

xukai92:
Seems due to the change of stochastic processes and the random seed was a bit unlucky and lead to a mean exceeding the tolerance.

yebai:
@xukai92 The code looks good to me. However, does this actually influence anything except the initialization for HMC, since the `find_good_eps` is only called before HMC begins?

xukai92:
It only influences the initialisation of HMC.

PR #749, opened 2019-04-11
Fix MvNormal and Flux.Tracker support

mohamed82008:
This PR sets the grounds for fixing `MvNormal` and `Flux.Tracker` support. Requires https://github.com/JuliaStats/PDMats.jl/pull/89.

```julia
using Turing, Random, LinearAlgebra, Flux.Tracker, Distributions

Random.seed!(0);
A = randn(3, 3);
M = A * A' + I;
u, x = randn(3), randn(3);

f(u, M, x) = logpdf(MvNormal(u, M), x)
y, back = Tracker.forward(f, u, M, x)
back(1)
```

Some more work here is necessary:
- [ ] Cleanup the syntax for constructing `MvNormal`, by defining a method that dispatches on `Tracker` types, implementing the above workaround. 
- [ ] Add proper tests of `MvNormal` support with reverse AD against forward AD.
- [ ] Improve the performance of reverse AD on Cholesky by adapting the Zygote version.

Side note: All this type piracy we are doing in Turing needs to be properly resolved at some point by contributing the code to the relevant packages.


mohamed82008:
Supersedes #743

mohamed82008:
Reference for the Cholesky gradient: https://homepages.inf.ed.ac.uk/imurray2/pub/16choldiff/choldiff.pdf

yebai:
@mohamed82008  Nice work! I would suggest that we make this PR a priority and try to merge it ASAP.

willtebbutt:
I'm generally pro- this, but it's not clear to me why we need to have this `lowerchol` function? Is there a reason that we can't just at an `@grad` definition to `cholesky`? And is there a particular reason that you're working in terms of the `LowerTriangular` factorisation? (the default is usually `UpperTriangular`)

As a side-note, this probably isn't quite the optimal way of going about implementing the `cholesky` adjoint. Please see [here](https://github.com/FluxML/Zygote.jl/blob/81b9e2367ef1a837a39924ecd60fb16d7377aa67/src/lib/array.jl#L250) for a fairly optimal implementation, for which [this](https://arxiv.org/pdf/1710.08717.pdf) (section 4.2) is the appropriate reference.

mohamed82008:
> And is there a particular reason that you're working in terms of the LowerTriangular factorisation? (the default is usually UpperTriangular)

I just found it easier to work in terms of lower triangles than Cholesky when resolving the dispatch issues. Now that it is working, I can try to improve its performance by adapting the Zygote code, defining the grad for cholesky directly and using the upper triangular part. I will add that to the to-do.

mohamed82008:
@willtebbutt I tried implementing the Zygote version but the inability of `Flux.Tracker` to differentiate `setindex!` and inplace operations is a show stopper. I will just modify the current version to use `U` instead of `L`, and leave the same procedure which is `Tracker`-compatible.

yebai:
@mohamed82008 can we add some tests and merge this PR, then improve the performance in a separate PR?

mohamed82008:
Sure.

mohamed82008:
This PR also goes from Flux.Tracker to Tracker.jl, since Flux doesn't host Tracker anymore. If tests pass, you can merge this if you want. I have a slightly more efficient implementation that uses upper triangular matrices but it requires adding some methods to Tracker so I will do that in separate PRs.

mohamed82008:
This PR still needs a rebase.

mohamed82008:
Please don't merge.

mohamed82008:
@willtebbutt this test is failing. 
```julia
using ForwardDiff, Tracker, Distributions, Turing, Test

func = S->logpdf(MvNormal(zeros(2), S), [1.0, 1.0])
S = rand(2,2); S = S' * S
rtol = atol = 1e-8
reverse_grad = Tracker.gradient(func, S)[1]
forward_grad = ForwardDiff.gradient(func, S)
@test isapprox(reverse_grad, forward_grad, atol=atol, rtol=rtol)
```

Something funny seems to be going on here too, the `delta` matrix in the grad of `upperchol` is not symmetric, shouldn't it be?

yebai:
We should probably investigate why the test for this PR is failing and replace the trick in  https://github.com/TuringLang/Turing.jl/pull/757 with this PR.

yebai:
Following discussions from https://github.com/TuringLang/Turing.jl/issues/777#issuecomment-496278925, can we use `Zygote` to implement a special `Tracker` gradient function for  `MvNormal`?

yebai:
Closed in favour of https://github.com/TuringLang/Turing.jl/pull/798

Issue #634, opened 2018-12-20
Forbid `mf()` and implement explicit `rand` and `logpdf` functions for `mf`


xukai92:

yebai:
@willtebbutt @xukai92 
how about we always drop all **observe** statements in `gdemo()()`, this leads to a consistent behaviour and should be intuitive too.

xukai92:
I see. But what I mean by this issue is too actually have some concrete functions generated by evaluating `gdemo()` (I think @willtebbutt also proposed this some time ago).

Currently behaviour:
```julia
@model demo(y) = begin
   5 ~ Normal(0,1)   # rv1
   y ~ Normal(0,1) . # rv2
end
mf = gdemo()
mf() # => sampling from prior
```

But instead we could have
```julia
@model demo(y) = begin
   5 ~ Normal(0,1)   # rv1
   y ~ Normal(0,1) . # rv2
end
mf = gdemo() # this returns mf::UniqueRuntimeModelType and create some function definitions:
logpdf(::UniqueModelType, vi) = ... # (1) return logpdf by evaluating using vi
rand(::UniqueModelType) = ... # (2) gdemo()() in your example
sample(::UniqueModelType) = ... # (3) gdemo(2)() in your example
```

I guess it could also potentially give more space for performance because during sample we are explicitly using the specific version of `logpdf` and `rand`.

willtebbutt:
@yebai your proposal has the issue that the behaviour of the following programme becomes ambiguous:
```julia
@model demo() = begin
    5 ~ Normal(0, 1)
end
demo() # what does this do?
```
This makes me nervous. One the one hand, this programme is weird, on the other it's a valid Turing programme, so we should ensure that the behaviour is well-defined. My proposal below will mostly resolve this.

@xukai92 I completely agree regarding the format of `logpdf`. Regarding `rand` and `sample`: certainly we could define one to sample from the prior and the other the posterior, but this wouldn't be particularly intuitive (there's nothing in the names to suggest the proposed behaviour). Instead, I propose the following:

1. Don't define `sample`, just use `rand` and `logpdf` (or maybe `logp`, since `logpdf` isn't defined for all models supported by Turing?).
2. The current proposal regarding `gdemo(missing)` to indicate that the prior should be used for `y` seems sensible to me.
3. We implement `gdemo(Prior())` to indicate that _all_ random variables should be sampled from the prior. In this case, both `rv1` and `rv2` would be sample from the prior.
[...]

Issue #476, opened 2018-08-28
Change broadcasting for ~ macro


trappmartin:
Currently the `~` macro automatically uses broadcasting if the left-hand side is a vector and the right-hand side a univariate distribution. It would be more convenient for testing and for the user to change it to an explicit syntax, i.e.`.~`, which would also better match the standard Julia syntax.

@yebai What happens in the case of multivariate distributions?

yebai:
I think `~` will check whether the dimensionality of left-side is a matrix or a vector of vectors. If so, it will treat left-side as a vector of draws from multivariate distributions.

willtebbutt:
@trappmartin I agree with your point regarding Julia-ness, but we're quite closely tied in to the `Distributions.jl` conventions, and they make very specific assumptions regarding what multiple samples from a distribution look like depending upon whether it's a `UnivariateDistribution`, `MultivarateDistribution`, or a `MatrixDistribution` (see the `Distributions.jl` documentation for more info). Consequently, I think it actually makes sense _not_ to follows the standard Julia conventions here, but rather to vectorise in the traditional sense and follow the `Distributions.jl` conventions.

To provide a concrete example of where we would get in trouble using a `.~` notation, the Distributions convention for multiple samples from a `MultivariateDistribution` is to produce a matrix whose columns are observations. What, then, does the broadcasting `.~` notation mean? One the one hand our intent would probably be to broadcast observation-wise, but broadcasting is already defined for `AbstractMatrix` in an element-wise fashion, so it would necessarily involve introducing some non-standard behaviour.

trappmartin:
Thanks, @willtebbutt! The convention of the `Distributions` package is quite unfortunate. I understand that with this convention, we cannot generically broadcast. 

Let's keep this issue open and see if we can find a more transparent solution to this. I feel that the implicit broadcasting solution we currently have is sub-optimal.

willtebbutt:
Agreed regarding the convention. FWIW, I had the same problem in [Stheno.jl](https://github.com/willtebbutt/Stheno.jl) with regards to data sets. The inputs to a GP (or any other model for that matter) might be a vector of scalars, a matrix of for which each row or column should be interpreted as an observation, etc.

[...]

PR #750, opened 2019-04-13
[WIP] PR for issue #634

trappmartin:
This is a work in progress PR refactoring the assume and observe interface (#634). *Do not merge!* 

----
**Todo:**

- [x] Define abstract runner interface.
- [x] Define runner for particle based inference.
- [x] Define runner to evaluate `logjoint(m::Model, v::VarInfo)`.
- [ ] Define runner to evaluate `logpdf(m::Model, v::VarInfo)`.
- [x] Separate compiler and other core parts from `SampleFromPrior`.
- [ ] Use runners inside of samplers instead of overloading `assume` and `observe`.
- ~~[ ] Fix Vectorisation issues:  #760  #761 .~~
----


yebai:
> Why do we call `setgid!` only if assume is called for a single distribution during HMC sampling.
> Why do we increase `vi.num_produce` on `observe `only for a single distribution and not during vectorisation? Is this a bug?

These are likely bugs in the vectorisation implementation (related https://github.com/TuringLang/Turing.jl/issues/476). 

> We sometimes increase the logp of VarInfo using acclogp!(vi, ...) and sometimes not. Is there a rational behind this?

I'm not sure I understand the issue here. Can you clarify a bit?

> I'd prefer if we have a more flexible and consistent interface for manipulating the `lp_` variable. Currently, the user has no chance to manipulate this if needed. I feel something similar to `acclogp!(vi, ...)` but without the need of passing vi in the form of a macro would be a solution. Any objections?

Have an API for manipulating `logp` sounds good!

trappmartin:
> > We sometimes increase the logp of VarInfo using acclogp!(vi, ...) and sometimes not. Is there a rational behind this?
> 
> I'm not sure I understand the issue here. Can you clarify a bit?

We increase log p inside the compiler using: `$vi.logp += $lp` and additionally increase this values inside the particle Gibbs sampler:

```
./src/inference/pgibbs.jl:174:        acclogp!(vi, logpdf_with_trans(dist, r, istrans(vi, vn)))
```

I suspect this is a bug.

yebai:
Places where we modify `logp`:

Inside the compiler: 
- https://github.com/TuringLang/Turing.jl/blob/820051e71fc5f6035423d1b85847121abfb303c3/src/core/compiler.jl#L52

- https://github.com/TuringLang/Turing.jl/blob/820051e71fc5f6035423d1b85847121abfb303c3/src/core/compiler.jl#L104

Inside PG:
- https://github.com/TuringLang/Turing.jl/blob/e585eea53f05fcbfde006967da79b56d40685b04/src/inference/pgibbs.jl#L174

It does seem the line in PG is a duplicate of `L104` in the compiler. Since `PG` is not making use of `VarInfo.logp` in the resampling step, this is not hurting us yet...

yebai:
UPDATE:

Actually, the following line "cancels" the increment of `logp` in the compiler, since it returns `0`. 

https://github.com/TuringLang/Turing.jl/blob/e585eea53f05fcbfde006967da79b56d40685b04/src/inference/pgibbs.jl#L176

trappmartin:
It seems it's a very good idea that we refactor that bit of the code base.

src/inference/runners.jl
```diff
@@ -0,0 +1,218 @@
...
+end
+
+# NOTE: this is necessary as we cannot use broadcasting for MV dists.
+function observe(::Sampler{ComputeLogJointDensity}, dists::Vector{<:MultivariateDistribution}, values, ::VarInfo)
+    dist = _getdist(dists)
+    return sum(logpdf(dist, values))
+end
+
+#################################
+# Compute the log joint Runner. #
```
cpfiffer:
```suggestion
# Compute the particle filtering Runner. #
```

mohamed82008:
I suggest we keep the vectorization issues out of this, since that's a totally different beast. 

About this PR and its underlying issue, while I am not exactly against them, I just don't get the appeal of introducing a plethora of new types to do exactly the same thing we are doing now without these types at all. The abstractions we are introducing here don't add much functional value for now as far as I can see. If anything they are a bit confusing. For example, the relation between `SampleFromDistribution`, `ComputeLogJointDensity` and `Sampler` is a bit grey in my eyes since they all subtype `AbstractRunner` but are used in very different situations. For instance, why is `ComputeLogJointDensity` not just a function? Why do I have to do `Sampler(ComputeLogJointDensity(), selector)` which is unnecessarily complicated and obscure imo. What are we sampling? What are we selecting? Can any `AbstractRunner` be passed to `Sampler` as a first argument? Another question is when should we use `SampleFromPrior` vs `nothing` when trying to run the model initially. Why was `nothing` introduced back when we removed it earlier and replaced it with `SampleFromPrior` or `SampleFromUniform` everywhere? 

I think what we need to do is work from the use-cases backwards. So firstly, we specify the API functions that we want to work, e.g. `rand(::Sampler(::Model, ::InferenceAlgorithm))` then we write minimal code to define these functions. There is no need to introduce a new type if it doesn't add functional value via dispatch or data organization. There might be a philosophical appeal to abstracting everything and introducing a new type for everything, but that's just code that needs to be maintained later so simple is usually better.

I may be just failing to see the value of this re-structuring so please explain it to me if I got it all wrong. Sorry for the late night rant :)

yebai:
>About this PR and its underlying issue, while I am not exactly against them, I just don't get the appeal of introducing a plethora of new types to do exactly the same thing we are doing now without these types at all. 

This is meant to avoid direct dispatch `assume` and `observe` on `Sampler` types, and implement a standard set of APIs (e.g. `logpdf`, `rand`). The current code bases contain redundant `assume` and `observe` implementations, which are largely unnecessary. We can keep modifying the signature, e.g. 

```julia
assume(spl::{HMC, NUTS, SGLD}, ...)
```

But this looks a bit ugly and doesn't support plug-and-play inference. Introducing an intermediate type such as `Runner` allows samplers to share the same set of `assume` and `observe` implementations for many use cases. 

> For instance, why is ComputeLogJointDensity not just a function?

This is meant to reduce the number of functions the Turing compiler has to generate. By lowering the model into IRs and allow users to overload `assume` and `observe`, we save a lot of compiler transformations. I like to keep the compiler as simple as possible (though it's already quite complex) and do most stuff via standard Julia features such as multi-dispatching.

xukai92:
It seems that the issue is not introducing these new runner types, but the fact these types are currently used to construct `Sampler` are not very intuitive. Not sure how to change this though.

trappmartin:
I agree. I'll create a new PR in which I will try to find a solution that doesn't require to construct `Sampler`.

xukai92:
What's the plan for this issue?

trappmartin:
Once the inference changeover is merged I’ll work on a new PR that contains the main features of this PR.

yebai:
I suspect that a lot of features planned in this PR is now added in #965, e.g. the `Runner` type is replaced by a similar type `Context`. I propose we close this PR for now, and move useful code contained in this PR to a new PR if needed.

PR #757, opened 2019-04-20
Fix MvNormal and Tracker - Zhuo's trick

mohamed82008:
This PR brings back the trick from #743 which is the easiest way to support Tracker in MvNormal. This is a workaround for #749 until we figure out what's wrong there. We also move from Flux.Tracker to Tracker as Tracker is now removed from Flux.


mohamed82008:
https://github.com/JuliaStats/PDMats.jl/pull/90

mohamed82008:
This PR is just awaiting a new release for PDMats.

mohamed82008:
https://github.com/JuliaStats/PDMats.jl/issues/91

yebai:
Thanks, let's merge it from the Turing side then.

PR #765, opened 2019-04-25
Add Julia compatibility

cpfiffer:
[Fredrik suggested](https://github.com/JuliaRegistries/General/pull/304) adding a Julia compat section to the Project.toml file. Could someone look at the second I've added and see if it seems reasonable? Do we want to publish with `julia = ">= 1.0"`?


yebai:
@cpfiffer Setting a lower bound to `1.0` for julia looks good to me!

Issue #746, opened 2019-04-02
Design/improve/unifty `sample()` interface


xukai92:
Questions and my current thoughts
- How to pass initial state
  - Support `sample(..., init::Array{Union{Missing, Array{Float64,1}},1}=[missing])`
  - We flat `init`, and assign those variables which are not missed to `vi[spl]`
  - Constructing a correct `init` is the user's responsibility
- How to pass adaptation configuration
  - Take use of `AdvancedHMC.Adaptation`
  - `sample(..., adapt::Union{Nothing,AbstractAdaptor}=nothing)`
- How to pass continuing chain
  - `sample(..., resume::Union{Nothing,AbstractChain}=nothing)`
  - How do we resume an adapted sampler?
    - With the new interface of `AdvancedHMC.Adaptation`, the adapted sampler is just the types but not the adapter or any other thing. So if we can make the types in Turing to be 1:1 to those types, we can resume the adapted sampler.

Related issue:
- Provide a way to set initial state of HMC #571
- Better adaptation configuration interface #491
- Resume interface #566

trappmartin:
Does it make sense that we differentiate between initial value and resume sampling? Why not get rid of the resume interface and only provide a clean way to set an initial value?

xukai92:
Good idea! Then the only question leaves here is how to resume adapted sampler.

yebai:
I think we can use [`keyword`](https://docs.julialang.org/en/v1/manual/functions/index.html#Keyword-Arguments-1) arguments here to allow maximum flexibility and minimal `sample` design. For example:

```julia

function sample(rng::AbstractRNG, ℓ::Model, s::SamplerType, N::Integer; args...) where SamplerType
     for i=1:N
        step(rng::AbstractRNG, ℓ::Model, s::Sampler, N::Integer; args...)
    end
end

```

where `args` is a keyword argument, some example uses are

- `θ₀::AbstractVector = zeros(D)`
- `adaptor::Adaptor = StanNUTSAdaptor()`
- `n_chains::Int = 1`
- `resume::AbstractChain = chain`

yebai:
Related: https://github.com/TuringLang/Turing.jl/issues/723

yebai:
@cpfiffer Does it make sense moving this `sample` function to `MCMCChains`. This would allow packages such as `AdvancedHMC` can depend on `MCMCChains`, but not depend on `Turing`.

cpfiffer:
[...]

PR #773, opened 2019-05-06
Merge PMCMC samplers into one file - no functionality change. 

yebai:


yebai:
@trappmartin @cpfiffer @xukai92 @mohamed82008 This PR merges all particle samplers into one file. It uses sections instead of small files to keep functions organised. This approach is used in some Julia base libraries (e.g. `channel.jl`, `task.jl`) I would say this is easier to read and track changes than several smaller files. Any comments/thoughts on why this might be a good/bad practise?

mohamed82008:
I like it. The file is not that big, so I think it is reasonable.

cpfiffer:
I don't see an issue either. Plus, consolidating all the particle samplers might make it easier to have them share interface implementations where possible for #746.

trappmartin:
All the refactoring changes look really good to me too. Thanks for merging files together.

PR #783, opened 2019-05-12
Zygote AD backend

mohamed82008:
This is a WIP PR that introduces Zygote to our AD family. `MvNormal` is still not working and some inplace operations are tripping `Zygote`, so I need to figure out how to make them play well together.


xukai92:
By inplace operations do you mean the way we use `vi::VarInfo` requires the following?
```julia
vi[spl] = theta
model(vi, spl)
```

mohamed82008:
Seems so https://travis-ci.org/TuringLang/Turing.jl/jobs/531342906#L624, it is unclear to me what the solution here is.

cpfiffer:
Doesn't seem like much, so far: https://github.com/FluxML/Zygote.jl/pull/75. The PR to fix this hasn't made it through any may have some caveats for our case.

yebai:
@mohamed82008 @willtebbutt Just a quick thought, can we use `Tracker` to define an in-place mutation function, and then use it inside `Zygote`?

yebai:
@mohamed82008 Perhaps revisit this PR when refactoring `VarInfo`?

mohamed82008:
Sure, I will give it another look. Although the underlying issue in Zygote (https://github.com/FluxML/Zygote.jl/issues/265) was not resolved yet. But maybe I can work-around it.

mohamed82008:
https://github.com/FluxML/Zygote.jl/pull/459

mohamed82008:
https://github.com/TuringLang/Bijectors.jl/pull/73

mohamed82008:
https://github.com/TuringLang/DynamicPPL.jl/pull/31

src/core/ad.jl
```diff
@@ -189,3 +202,7 @@ for F in (:link, :invlink)
         end
     end
 end
+
+## Zygote support ##
+
+Zygote.@nograd Bijectors._debug
```
devmotion: Maybe move this to Bijectors?

mohamed82008: Hmm sure, I thought Bijectors would then have to depend on Zygote. But I forgot that we can use Requires.

Project.toml
```diff
@@ -26,6 +26,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
devmotion: It seems only ZygoteRules is needed (if `Zygote.@nograd Bijectors._debug` is moved to Bijectors)

src/core/ad.jl
devmotion: Yes, I thought one could just add a file `src/compat/zygote.jl` with the required definitions for Zygote.

Project.toml
mohamed82008: We may need Zygote because it has the definition of the `pullback` method that we are using here.

devmotion: No, `pullback` is actually defined in ZygoteRules.

devmotion: https://github.com/FluxML/ZygoteRules.jl/blob/b0dc38ed8b857f0beb79afe165cb853e16a1a261/src/adjoint.jl#L21

devmotion: BTW due to the same reason probably Zygote (currently) does not have to be an (optional) dependency of AdvancedHMC (https://github.com/TuringLang/AdvancedHMC.jl/pull/161/files#r366268742).

mohamed82008: Yes but I think calling it will error unless `Zygote` is loaded because the 2-arg method is defined in `Zygote`. Since `DistributionsAD` loads `Zygote` anyways, the method will be loaded. But I think it it's better to load `Zygote` here as well in case `DistributionsAD` stops loading `Zygote` in the future.

devmotion: Hmm I think I disagree. It seems the main purpose of ZygoteRules is for other packages to be able to define custom adjoints and refer to `pullback` without having to depend on Zygote and all its dependencies (I think ForwardDiff2 uses a similar approach with ChainRules). Especially if a package wants to support different AD backends, I think it is nice if users are not forced to install every single AD backend and all its dependencies.

It also seems that (currently) ForwardDiff is the default AD backend, and hence a user has to actively choose Zygote in which case, I guess, users should be expected to load Zygote explicitly. Nevertheless, I guess it might be useful to check that Zygote is actually loaded if the backend is switched to Zygote and provide a helpful error message in case it's not.

willtebbutt: @devmotion is exactly right regarding the motivation for `ZygoteRules`.

It would be a really great thing if the entire Turing stack depended only on `ZygoteRules`, rather than `Zygote`, precisely to ensure that users that don't want to install Zygote and its dependencies.

This will become even more acute once we've moved Zygote onto ChainRules (see [this issue](https://github.com/FluxML/Zygote.jl/pull/366)).

mohamed82008: Supporting Zygote optionally sounds good to me if that's the path we want to take. But note that we currently support Tracker by default too, and Zygote is currently used to define MvNormal adjoints for Tracker use. So if we want to optionally support Zygote, we will have to optionally support Tracker as well, or break Tracker and MvNormal. Both sound like bad options to me. Alternatively, we can support all these AD packages by default which currently won't increase our loading time at all because we load Zygote anyways in DistributionsAD to support Tracker. I think this is a better short-term solution.

devmotion: I'm not in favour of making all AD backends optional right away and breaking Tracker and MvNormal :stuck_out_tongue:

IMO it's fine to have the explicit dependency in DistributionsAD at the moment, even though I think it would be a nice long-term goal to get rid of such heavy dependencies. But I see that it might be difficult to achieve this (or even impossible if it's an indirect dependency from some non-Turing package).

Still, even if DistributionsAD depends on Turing, I think Turing shouldn't add Zygote as a new explicit dependency. Depending on ZygoteRules seems to be sufficient and the intended approach. Additionally, not depending on Zygote seems like moving into the right direction if one wants to pursue the long-term goal of slimming down dependencies.

mohamed82008: Ok that can be done. But when Zygote is no longer loaded in Turing by default, that will be a breaking change so we need to be mindful of that. I will replace Zygote with ZygoteRules.

devmotion: I'm not sure what's the best way to ensure that users can't use `ZygoteAD` if Zygote is not loaded, but maybe one could something similar to https://github.com/TuringLang/AdvancedHMC.jl/pull/161, i.e., use Requires and define optional backends such as `ZygoteAD` only if Zygote is loaded. More concretely, one could define
```julia
ADBackend(::Val) = error("The requested AD backend is not available. Make sure to load all required packages.")

function gradient_logp(
    θ::AbstractVector{<:Real},
    vi::VarInfo,
    model::Model,
    sampler::Sampler
)
    return gradient_logp(θ, vi, model, sampler, getADtype(sampler))
end

# optional dependencies
function __init__()
    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin 
         struct ZygoteAD <: ADBackend end

         ADBackend(::Val{:zygote}) = ZygoteAD

         function gradient_logp(
            θ::AbstractVector{<:Real},
            vi::VarInfo,
            model::Model,
            sampler::Sampler,
            ::ZygoteAD
        )
            T = typeof(getlogp(vi))

            # Specify objective function.
            function f(θ)
                new_vi = VarInfo(vi, sampler, θ)
                return getlogp(runmodel!(model, new_vi, sampler))
            end

            # Compute forward and reverse passes.
            l, ȳ = Zygote.pullback(f, θ)
            ∂l∂θ = ȳ(1)[1]
    
            return l, ∂l∂θ
        end
    end
end
```

mohamed82008: Yes that's reasonable.

devmotion: One could even update the default backend (I edited the comment above) and it does require neither Zygote nor ZygoteRules as dependencies :smiley:

docs/src/using-turing/compiler.md
```diff
@@ -207,8 +207,8 @@ function inner_function(vi::Turing.VarInfo, sampler::Turing.AbstractSampler, ctx
 
     ... # The code above is repeated for the other 2 variables, y and TV
 
-    # Reset `vi.logp`
-    vi.logp = 0
+    # Reset `vi.logp[]`
```
devmotion: Maybe use less code and just write something like
```suggestion
    # Reset the `logp` accumulator
```
It's not really relevant here that one can access it with `vi.logp[]`, it seems?

src/core/compat/zygote.jl
```diff
@@ -0,0 +1,39 @@
...
+    θ::AbstractVector{<:Real},
+    vi::VarInfo,
+    model::Model,
+    sampler::AbstractSampler=SampleFromPrior(),
+    backend::ADBackend,
+)
+
+Computes the value of the log joint of `θ` and its gradient for the model
+specified by `(vi, sampler, model)` using reverse-mode AD from the specified `backend`, e.g. `TrackerAD()` which uses `Tracker.jl` or `ZygoteAD()` which uses `Zygote.jl`.
+"""
```
devmotion: The function signature in the docstring is different from the one that's implemented. I'm wondering if it's actually useful/needed to copy-paste the docstring from the Tracker implementation here?

src/core/compat/zygote.jl
```diff
@@ -0,0 +1,39 @@
...
+function gradient_logp_reverse(
+    backend::ZygoteAD,
+    θ::AbstractVector{<:Real},
+    vi::VarInfo,
+    model::Model,
+    sampler::AbstractSampler = SampleFromPrior(),
+)
+    T = typeof(getlogp(vi))
```
devmotion:
```suggestion
```

src/core/ad.jl
```diff
@@ -102,43 +106,54 @@ function gradient_logp_forward(
...
 
     # Compute forward and reverse passes.
     l_tracked, ȳ = Tracker.forward(f, θ)
-    l::T, ∂l∂θ::typeof(θ) = Tracker.data(l_tracked), Tracker.data(ȳ(1)[1])
     # Remove tracking info from variables in model (because mutable state).
+    l::T, ∂l∂θ::typeof(θ) = Tracker.data(l_tracked), Tracker.data(ȳ(1)[1])
+
     return l, ∂l∂θ
 end
+function gradient_logp_reverse(
```
devmotion: Is this function (without explicit backend) actually needed? If only `gradient_logp` is called in the implemented algorithms, then the method with explicit backend is called anyways all the time.

mohamed82008: It's only called in tests I think.

mohamed82008: It's convenient when poking around AD bugs so let's keep it for now.

src/core/compat/zygote.jl
mohamed82008: You are right, I will remove it.

mohamed82008:
I added tests for Zygote, many of which are currently failing. I will try shooting them down one by one in the next few days.

mohamed82008:
Ran into https://github.com/FluxML/Zygote.jl/issues/468 for `Wishart` distribution.

mohamed82008:
Tests should pass after the following PRs are merged:
1. https://github.com/FluxML/Zygote.jl/pull/469
2.  https://github.com/TuringLang/DistributionsAD.jl/pull/25
3. https://github.com/TuringLang/Bijectors.jl/pull/74

in this order.

mohamed82008:
The only test failures here are because of DistributionsAD. I will work on them.

mohamed82008:
I think this is ready for a review.

Project.toml
```diff
@@ -35,17 +35,18 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
...
+DistributionsAD = "0.4.3"
 DocStringExtensions = "0.8"
 DynamicPPL = "0.4"
 EllipticalSliceSampling = "0.2"
+FiniteDifferences = "0.9"
 ForwardDiff = "0.10.3"
 Libtask = "0.3.1"
 LogDensityProblems = "^0.9, 0.10"
-MCMCChains = "3"
+MCMCChains = "3.0"
```
devmotion: This change does not restrict the compatible versions of MCMCChains, so maybe just keep
```suggestion
MCMCChains = "3"
```
to avoid any confusion (some people might assume incorrectly that it is compatible with only MCMCChains [3.0.0, 3.1.0) whereas actually it is still compatible with MCMCChains [3.0.0, 4.0.0)?

docs/src/using-turing/advanced.md
```diff
@@ -128,16 +128,18 @@ mf(vi, sampler, ctx, model) = begin
...
-        x, 
-        vi,
+    acclogp!(vi,
+        Turing.Inference.dot_tilde(
+            ctx, 
+            sampler, 
+            Normal(m, sqrt(s)), 
+            x, 
+            vi,
+        ),
```
devmotion: Maybe split this into two steps, i.e., `lp = ....; acclogp!(vi, lp)` as above? IMO that is easier to read.

docs/src/using-turing/compiler.md
```diff
@@ -91,9 +91,9 @@ if preprocessed isa Tuple
     vn, inds = preprocessed
     out = Turing.Inference.tilde(ctx, sampler, temp_right, vn, inds, vi)
     p[1] = out[1]
-    vi.logp += out[2]
+    acclogp!(vi, out[2])
 else
-    vi.logp += Turing.Inference.tilde(ctx, sampler, temp_right, preprocessed, vi)
+    acclogp!(vi, Turing.Inference.tilde(ctx, sampler, temp_right, preprocessed, vi))
```
devmotion: Same here, maybe two steps?

docs/src/using-turing/compiler.md
```diff
@@ -118,25 +118,25 @@ if preprocessed isa Tuple
     temp_left = x[1:2]
     out = Turing.Inference.dot_tilde(ctx, sampler, temp_right, temp_left, vn, inds, vi)
     left .= out[1]
-    vi.logp += out[2]
+    acclogp!(vi, out[2])
 else
     temp_left = preprocessed # x[1:2]
-    vi.logp += Turing.Inference.dot_tilde(ctx, sampler, temp_right, temp_left, vi)
+    acclogp!(vi, Turing.Inference.dot_tilde(ctx, sampler, temp_right, temp_left, vi))
```
devmotion: And here as well?

src/Turing.jl
```diff
@@ -70,6 +71,7 @@ export  @model,                 # modelling
         @logpdf,
         @sampler,
         DynamicPPL,
+        mapvcat,
```
devmotion: Do we need actually have to reexport `mapvcat`?

mohamed82008: No we don't. I will remove it.

docs/src/using-turing/advanced.md
mohamed82008: Fair enough.

src/core/ad.jl
```diff
@@ -69,8 +71,8 @@ function gradient_logp(
     ad_type = getADtype(sampler)
     if ad_type <: ForwardDiffAD
         return gradient_logp_forward(θ, vi, model, sampler)
-    else ad_type <: TrackerAD
-        return gradient_logp_reverse(θ, vi, model, sampler)
+    else
+        return gradient_logp_reverse(ad_type(), θ, vi, model, sampler)
```
devmotion: I guess a better design would be to just define `gradient_logp(backend, theta, vi, model, sampler)` instead of distinguishing between `gradient_logp_forward` and `gradient_logp_reverse`. Now that Turing supports > 2 backends IMO the branching should be handled by multiple dispatch.

BTW, I'm wondering if it would be better to replace `getADtype(sampler)` (that returns a type) with `getADbackend(sampler)` that would return an instance of an AD backend. It seems a bit strange to retrieve the type and then call `ad_type()`.

src/inference/Inference.jl
```diff
@@ -314,10 +316,13 @@ function AbstractMCMC.bundle_samples(
     N::Integer,
     ts::Vector,
     chain_type::Type{Chains};
+    raw_output::Bool=false,
```
devmotion: I don't think this is needed? That's the default in AbstractMCMC if you specify `chain_type = Any`:
https://github.com/TuringLang/AbstractMCMC.jl/blob/dc5a76f45acc3f3b1025ea5856808edc3f07c10a/src/AbstractMCMC.jl#L208-L218

devmotion:
```suggestion
```

src/inference/Inference.jl
```diff
@@ -314,10 +316,13 @@ function AbstractMCMC.bundle_samples(
...
     discard_adapt::Bool=true,
     save_state=false,
     kwargs...
 )
+    raw_output && return ts
+
```
devmotion:
```suggestion
```
(See my comment above)

src/inference/Inference.jl
```diff
@@ -865,8 +870,8 @@ function dot_observe(
     vi::VarInfo,
 )
     increment_num_produce!(vi)
-    Turing.DEBUG && @debug "dist = $dist"
-    Turing.DEBUG && @debug "value = $value"
+    Turing.DEBUG && _debug("dist = $dist")
```
devmotion: IIRC we need `_debug` due to some Zygote issues?

Project.toml
```diff
@@ -35,13 +35,14 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 AbstractMCMC = "0.5.2"
 AdvancedHMC = "0.2.20"
 AdvancedMH = "0.4"
-Bijectors = "0.4.0, 0.5, 0.6"
+Bijectors = "0.6.2"
 BinaryProvider = "0.5.6"
-Distributions = "0.21.11, 0.22"
-DistributionsAD = "0.4.1"
+Distributions = "0.22"
+DistributionsAD = "0.4.3"
```
willtebbutt: Consider
```suggestion
DistributionsAD = "~0.4.3"
```

src/inference/Inference.jl
mohamed82008: Yes

Project.toml
```diff
@@ -35,13 +35,14 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 AbstractMCMC = "0.5.2"
 AdvancedHMC = "0.2.20"
 AdvancedMH = "0.4"
-Bijectors = "0.4.0, 0.5, 0.6"
+Bijectors = "0.6.2"
```
willtebbutt: Consider 
```suggestion
Bijectors = "~0.6.2"
```

docs/src/using-turing/compiler.md
mohamed82008: Done

docs/src/using-turing/compiler.md
mohamed82008: Done

src/variational/VariationalInference.jl
```diff
@@ -23,6 +23,25 @@ function __init__()
...
+            model,
+            θ::AbstractVector{<:Real},
+            out::DiffResults.MutableDiffResult,
+            args...
+        )
+            f(θ) = if (q isa VariationalPosterior)
+                - vo(alg, update(q, θ), model, args...)
+            else
+                - vo(alg, q(θ), model, args...)
+            end
```
devmotion: What's the reason for not defining
```suggestion
            if q isa VariationalPosterior
                f = let alg=alg, q=q, model=model, args=args
                    θ -> - vo(alg, update(q, θ), model, args...)
                end
            else
                f = let alg=alg, q=q, model=model, args=args
                    θ -> - vo(alg, q(θ), model, args...)
                end
            end
```
or at least moving the `if` statement out of the function definition? Does it compile to the same thing?

src/core/ad.jl
mohamed82008: Sounds good.

Project.toml
```diff
@@ -35,13 +35,14 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
...
 BinaryProvider = "0.5.6"
-Distributions = "0.21.11, 0.22"
-DistributionsAD = "0.4.1"
+Distributions = "0.22"
+DistributionsAD = "0.4.3"
 DocStringExtensions = "0.8"
 DynamicPPL = "0.4"
 EllipticalSliceSampling = "0.2"
+FiniteDifferences = "0.9"
```
devmotion: Do we need the compatibility bound on the test dependency `FiniteDifferences`? IIRC, by default, Pkg will remove it if you run `] up` and CompatHelper won't open PRs for updating it.

src/variational/VariationalInference.jl
mohamed82008: The if statement should compile away yes. I just copied the functions from other backends' methods.

src/Turing.jl
```diff
@@ -9,11 +9,12 @@ module Turing
 ########################################################################
 
 using Requires, Reexport, ForwardDiff
-using Bijectors, StatsFuns, SpecialFunctions
+using DistributionsAD, Bijectors, StatsFuns, SpecialFunctions
 using Statistics, LinearAlgebra
 using Markdown, Libtask, MacroTools
 @reexport using Distributions, MCMCChains, Libtask
 using Tracker: Tracker
+using DistributionsAD: mapvcat
```
devmotion: I guess this not needed either?
```suggestion
```

src/core/ad.jl
```diff
@@ -4,12 +4,13 @@
 using Bijectors
```
devmotion: Can this be removed now that Bijectors is loaded in src/core/Core.jl?

mohamed82008: Sure

Project.toml
mohamed82008: Done

Project.toml
mohamed82008: Done

Project.toml
mohamed82008: Removed

src/core/ad.jl
mohamed82008: Done

Project.toml
```diff
@@ -35,14 +35,13 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 AbstractMCMC = "0.5.2"
 AdvancedHMC = "0.2.20"
 AdvancedMH = "0.4"
-Bijectors = "0.6.2"
+Bijectors = "~0.6.2"
```
devmotion: That's the same as
```suggestion
Bijectors = "0.6.2"
```
BTW, so the change is not needed.

Project.toml
```diff
@@ -35,14 +35,13 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
...
 BinaryProvider = "0.5.6"
 Distributions = "0.22"
-DistributionsAD = "0.4.3"
+DistributionsAD = "~0.4.3"
```
devmotion: Same here, the compatible versions are the same as the ones you get with the previous
```suggestion
DistributionsAD = "0.4.3"
```

willtebbutt: Well I learned something new today...

Project.toml
mohamed82008: I am fine with both. I added the ~ because of Will's comment above. I prefer the simpler version though so happy to bring it back.

src/core/ad.jl
```diff
@@ -39,14 +38,15 @@ ADBackend() = ADBackend(ADBACKEND[])
 ADBackend(T::Symbol) = ADBackend(Val(T))
 
 ADBackend(::Val{:forward_diff}) = ForwardDiffAD{CHUNKSIZE[]}
-ADBackend(::Val) = TrackerAD
+ADBackend(::Val{:reverse_diff}) = TrackerAD
+ADBackend(::Val) = error("The requested AD backend is not available. Make sure to load all required packages.")
 
 """
-getADtype(alg)
+getADbackend(alg)
```
devmotion: According to the [Julia documentation](https://docs.julialang.org/en/latest/manual/documentation/#Documentation-1), this should be indented:
```suggestion
    getADbackend(alg)
```

src/core/ad.jl
```diff
@@ -39,14 +38,15 @@ ADBackend() = ADBackend(ADBACKEND[])
...
 
-Finds the autodifferentiation type of the algorithm `alg`.
+Finds the autodifferentiation backend of the algorithm `alg`.
```
devmotion: According to the [Julia documentation](https://docs.julialang.org/en/latest/manual/documentation/#Documentation-1), one should use the imperative form:
```suggestion
Find the autodifferentiation backend of the algorithm `alg`.
```

src/core/ad.jl
```diff
@@ -66,26 +66,23 @@ function gradient_logp(
...
-        return gradient_logp_forward(θ, vi, model, sampler)
-    else ad_type <: TrackerAD
-        return gradient_logp_reverse(θ, vi, model, sampler)
-    end
+    return gradient_logp(getADbackend(sampler), θ, vi, model, sampler)
 end
 
 """
-gradient_logp_forward(
+gradient_logp(
```
devmotion: Same here, the function signature should be indented:
```suggestion
    gradient_logp(
```
(I guess you want to adjust the following lines as well).
BTW maybe one could make use of DocStringExtensions here.

src/core/ad.jl
```diff
@@ -66,26 +66,23 @@ function gradient_logp(
...
+    backend::ADBackend,
     θ::AbstractVector{<:Real},
     vi::VarInfo,
     model::Model,
-    spl::AbstractSampler=SampleFromPrior(),
+    sampler::AbstractSampler = SampleFromPrior(),
 )
 
 Computes the value of the log joint of `θ` and its gradient for the model
```
devmotion:
```suggestion
Compute the value of the log joint of `θ` and its gradient for the model
```

Project.toml
devmotion: See https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1.

Project.toml
devmotion: I realized too late that you discussed this change, but actually both versions are equivalent: https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1

src/core/ad.jl
mohamed82008: Maybe in another PR.

src/core/ad.jl
```diff
@@ -153,31 +140,7 @@ function verifygrad(grad::AbstractVector{<:Real})
...
-            y = $F(dist, x_data, Val{proj})
-            return  y, Δ -> begin
-                out = (ForwardDiff.jacobian(x -> $F(dist, x, Val{proj}), x_data)::Matrix{T})' * Δ
-                return (nothing, out, nothing)
-            end
-        end
-    end
-end
-
+# These still seem necessary
```
torfjelde: Should this maybe be moved into Bijectors.jl? (Though prob in a different PR)

src/variational/VariationalInference.jl
torfjelde: > The if statement should compile away yes. I just copied the functions from other backends' methods.

I was the one who did this, and I indeed did it because the `if` should compile away:)

src/core/ad.jl
mohamed82008: Actually Bijectors differentiates fine without these. There seems to be something at the Turing-Bijectors interface that's causing issues with Tracker. I didn't have time to dig deeper into the issue so I just brought back those for now.

yebai:
thanks @mohamed82008. Looks good to me, ready to merge from my side!

PR #784, opened 2019-05-12
Discard adaptation samples

cpfiffer:
This PR modifies the `hmc.jl` file to discard the adaptation samples from `AdaptiveHamiltonian` samplers. I've added a keyword to the `sample` function for `NUTS` and `HMCDA` which is temporary until the sampling API is reworked, but for now you can choose to keep the adaptation samples using 

```julia
c1 = sample(model(. . .), NUTS(1000, 200, 0.65), discard_adapt = true)
c2 = sample(model(. . .), NUTS(1000, 200, 0.65), discard_adapt = false)
```

In the above case, `c1` has length 800, while `c2` has length 1000. `discard_adapt` defaults to true, so you must call `discard_adapt = false` explicitly if you want to hang on to them for some reason.

I also threw in a couple tests just to make sure this was sane.

### DIFF
--- a/src/inference/hmc.jl
+++ b/src/inference/hmc.jl
@@ -220,6 +220,7 @@ function sample(
     adaptor = AHMCAdaptor(alg),
     init_theta::Union{Nothing,Array{<:Any,1}}=nothing,
     rng::AbstractRNG=GLOBAL_RNG,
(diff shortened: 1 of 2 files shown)


yebai:
Thanks, @cpfiffer!

PR #785, opened 2019-05-12
Interface Implementation Part 1

cpfiffer:
This is the first in a series of PRs intended to get Turing's internals working on a common interface. Currently there is **no functional change**, but it is a good idea to have the bones of the interface in the codebase for easier development later. 

Also note that the basic interface (`Interface.jl`) is ultimately intended to move over to MCMCChains, once all of Turing's relevant internals are updated.

In this PR:
1. The basic interface.
2. A document which I'll be updating over time, containing a general guide to interface implementation.
3. A basic hook of Turing into the interface, which basically only moves us from Turing's native `AbstractSampler` type to the `Interface` version.

**EDIT**

I've copied below the beginnings of the design document for the interface, so people don't have to click around for it:

# The Sampling Interface

Turing's sampling interface presents several structures and functions that one needs to overload in order to implement an interface-compatible sampler. Currently it is non-operational, but the intent is to have all of Turing's samplers using the API below.

1. A subtype of `AbstractSampler`, defined as a mutable struct containing state information
2. A subtype of `AbstractTransition`, which represents a single draw from the sampler
3. A function `transitions_init` which returns a preallocated vector of the `AbstractTransition` type used by an implementation of an `AbstractSampler`
4. A function `sample_init!` which performs any necessary set up
5. A function `step!` which returns an `AbstractTransition`
6. A function `sample_end!` which handles any sampler wrap-up
7. A function `MCMCChains.Chains` which accepts the post-sampling vector created by `transitions_init` returns an `MCMCChains` object

### DIFF
--- a/src/interface/Interface.jl
+++ b/src/interface/Interface.jl
@@ -1,0 +1,136 @@
+module Interface
+
+import Distributions: sample, Sampleable
+import Random: GLOBAL_RNG, AbstractRNG
+
+export AbstractSampler,
+       AbstractTransition,
+       sample_init!,
+       sample_end!,
+       sample,
+       step!
+
+"""
+    AbstractSampler
+
+The `AbstractSampler` type is intended to be inherited from when
+implementing a custom sampler. Any persistent state information should be
+saved in a subtype of `AbstractSampler`.
+
+When defining a new sampler, you should also overload the function
+`transition_type`, which tells the `sample` function what type of parameter
+it should expect to receive.
+"""
+abstract type AbstractSampler end
+
+transition_type(s::AbstractSampler) = AbstractTransition
+
+"""
+    AbstractTransition
+
+The `AbstractTransition` type describes the results of a single step
+of a given sampler. As an example, one implementation of an
+`AbstractTransition` might include be a vector of parameters sampled from
+a prior distribution.
+
+Transition types should store a single draw from any sampler, since the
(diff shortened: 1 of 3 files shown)


cpfiffer:
I've changed the part that instantiates the `AbstractTransition` vector to be an arbitrary function of the `sample` parameters, instead of requiring implementors to make some weird `transition_type` vector.

src/interface/Interface.jl
```diff
@@ -0,0 +1,146 @@
...
+) where {ModelType<:Sampleable, SamplerType<:AbstractSampler}
+    # Perform any necessary setup.
+    sample_init!(rng, ℓ, s, N; kwargs...)
+
+    # Preallocate the TransitionType vector.
+    t = transitions_init(rng, ℓ, s, N; kwargs...)
+
+    # Step through the sampler.
+    for i=1:N
+        t[i] = step!(rng, ℓ, s, N; kwargs...)
```
yebai: Perhaps `step!` should depend on the previous `t[i-1]::Transition` as an input?

cpfiffer: Good idea. I've added that.

yebai:
@cpfiffer Seems the tests are broken for Travis CI with recent commits.

cpfiffer:
That should do it. AppVeyor tests are failing on the `MvNormal` bug, but I think that's being fixed elsewhere.

PR #787, opened 2019-05-14
Created `stdlib` folder - no functionality change.

yebai:
This PR creates a `stdlib` folder for storing `distributions.jl` and `RandomMeasures.jl`.


src/utilities/Utilities.jl
```diff
@@ -12,16 +12,10 @@ export  vectorize,
         Sample,
         Chain,
         init,
-        vectorize,
-        logpdf_binomial_logit,
```
xukai92: Are these not exported any more?

yebai: It seems these are only for internal uses. Any reason for exporting them?

cpfiffer: No reason I can see.

yebai:
Ready to merge from my side.

PR #791, opened 2019-05-22
Change HMCDA default adaptation

xukai92:
Change default adaptor for HMCDA to a naive composite instead of Stan's .

This will fail the CI before https://github.com/TuringLang/AdvancedHMC.jl/pull/67 is merged in AHMC (and used by Turing).


yebai:
@kai This should be working after https://github.com/TuringLang/AdvancedHMC.jl/pull/72 is merged.

xukai92:
I just updated AHMC in `Manifest.toml`. Let's see if CI passes this time.

Issue #775, opened 2019-05-07
Variational Inference for TuringLang


yebai:
We can consider adding support for the following variational inference methods to Turing

1. Automatic derivatives variational inference [1,4]
2. Variational inference with normalizing flows [2]
3. Markov Chain Monte Carlo and variational inference: Bridging the gap [3]
4. Black-box α-divergence minimization [5]

Comment: Project 3 should be interesting and also relatively easy since we have HMC support already.  

Related projects:

- https://github.com/stan-dev/stan/tree/develop/src/stan/variational
- https://github.com/jmxpearson/VinDsl.jl (julia)
- https://github.com/biaslab/ForneyLab.jl (julia)

[1]: Kucukelbir, A., Tran, D., Ranganath, R., Gelman, A., & Blei, D. M. (2017). [Automatic differentiation variational inference](https://arxiv.org/abs/1603.00788). The Journal of Machine Learning Research, 18(1), 430-474. 
[2]: Rezende, D. J., & Mohamed, S. (2015). [Variational inference with normalizing flows](https://arxiv.org/abs/1505.05770). arXiv preprint arXiv:1505.05770.
[3]: Salimans, T., Kingma, D., & Welling, M. (2015, June). [Markov chain Monte Carlo and variational inference: Bridging the gap](https://arxiv.org/abs/1410.6460). In International Conference on Machine Learning (pp. 1218-1226).
[4]: Kucukelbir, A., Ranganath, R., Gelman, A., & Blei, D. (2015). [Automatic variational inference in Stan. In Advances in neural information processing systems](https://arxiv.org/abs/1506.03431) (pp. 568-576).
[5]: Hernández-Lobato, J. M., Li, Y., Rowland, M., Hernández-Lobato, D., Bui, T., & Turner, R. (2016). [Black-box α-divergence minimization](https://arxiv.org/abs/1511.03243).

cc @cpfiffer @xukai92 @willtebbutt @mohamed82008

willtebbutt:
Regarding point 2, @rpinsler gave a very interesting presentation on developments in this area recently, that we should maybe look at to decide what to target.

cpfiffer:
I suspect that mean-field VI is probably going to be the easiest to implement first, and it'll be easier to build up the infrastructure to support the more complex VI stuff like ADVI and stochastic VI. I think the infrastructure part will be very important to do in a reasonable way, because it seems like there's a fair amount of heterogeneity in VI methods. 

[...]

Issue #347, opened 2017-09-21
Approximate Bayesian computation?


juliohm:
Is it possible to perform ABC with Turing.jl?

yebai:
At the moment, ABC is not directly supported yet. However, it should be easy to add ABC give that we have other simulation-based inference algorithms including IS, SMC etc.  Please have a look at the importance sampling code below and see whether it can help you.

https://github.com/yebai/Turing.jl/blob/master/src/samplers/is.jl

ChrisRackauckas:
Yeah, this would be nice to have. Does it require anything different about the likelihood? I am quite unfamiliar with it but hear it's good for the parameter estimation stuff we've been doing with DiffEqBayes.jl.

yebai:
@juliohm Thanks for the pointer. It looks interesting. Would be good if someone could give it a try to use `ApproxBayes.jl` as a backend of `Turing.jl`. I'm happy to help with along the way. 

PS. The coroutine-based SMC implementation in Turing is the fastest SMC at the moment. I think coroutines (copying) from Turing might help ABC as well if the simulator involves complex stochastic control flows.

yebai:
ABC should be implemented as a standalone package supporting the AbstractMCMC interface.

fipelle:
I am actually working on that and I hope to publish a short proof of concept soon ([relevant discussion](https://github.com/JuliaApproxInference/LikelihoodfreeInference.jl/issues/5#issuecomment-1286338886)). It seems to be working fine with simple problems and ABC algorithms. However, I am still trying to figure out how to implement more advanced ABC algorithms on the basis of SMC ([relevant issue](https://github.com/TuringLang/AdvancedPS.jl/issues/68)).

PR #796, opened 2019-05-26
Variational Inference

torfjelde:
This PR is an attempt at implementing **variational inference (VI)** in Turing (#775).

# Interface
VI is implemented in a `Variational` submodule of `Turing` with the following example interface:

```julia
alg = ADVI()       # `<: VariationalInference`
q = vi(model, alg) # => `<: VariationalPosterior` with optimized params
rand(q, n)         # produces `n` samples from variational posterior
```

The types are:
- `abstract type VariationalInference`: similar to `Sampler` in Turing, this is meant to contain the parameters of a particular VI method.
- `abstract type VariationalObjective`: provides an abstract type for objectives used in variational inference, e.g. ELBO. The resulting object, e.g. `elbo = ELBO()`, should be callable with signature `elbo(q::VariationalPosterior, model::Model, args...)`.
  - This type is probably over-doing it. Initial design was instead to use `objective(vo::VariationalObjective, ...)`, but if we simply require the objective to be callable as we do now, then we can also accept arbitrary function calls with the correct signature. This type can still be useful for constructing certain objectives, but should probably not be type-enforced in the signature of the `optimize` call (see below).
- `VariationalPosterior`: simply an alias to `Distribution{Multivariate, Continuous}`, i.e. `VariationalPosterior =  = Distribution{Multivariate, Continuous}`. Variational posteriors should at the very least implement `_rand!(rng::AbstractRNG, q::VariationalPosterior, x::AbstractVector{T})` as required by Distributions.jl.
  - Maybe also implement `logpdf`?

With accompanying methods:
- `vi(model::Model, alg::VariationalInference, [q::VariationalPosterior])`: performs VI on `model` using `alg`, with optionally provided `VariationalPosterior` `q`. As of writing, there's no method supporting `q` as additional argument, since `ADVI` does not have this freedom. See notes below for discussion.
- `grad(vo::VariationalObjective, vi::VariationalInference, q::VariationalPosterior, model::Model)`: computes the gradient.
- `optimize(vo::VariationalObjective, vi::VariationalInference, q::VariationalPosterior, model::Model; kwargs...)`: simply wraps the call to `grad` and performs the updates.
  - For `ADVI` we currently have `optimize(vo, vi, q, model; optimizer = ADAGrad())`, where `ADAGrad` is from `Flux.Optimise`. Currently supports any of the optimizers from `Flux.Optimise`.
  - As mentioned above, should probably drop the type-enforcement on `vo`

## Supports
This interface seems to be a good starting point for quite general families of VI, supporting the following in a "nice" way:

- [x] Different gradient computations for different `(vo, vi)` combinations through `grad`. Necessary because:
  - not all VI families can compute gradient by using auto-grad on the objective call, e.g. _Unbiased Implicit Variational Inference (UIVI)_ [1]
  - closed form expressions available, e.g. _Mixed membership stochastic blockmodels_ [2]
- [x] Different optimization schemes for different `(vo, vi)` combinations. Necessary because:
  - some objectives requires a min-max optimization scheme, e.g. _Stein Variational Gradient Descent (SVGD)_ [3]
- [x] Control variates and other variance-reduction techniques are supported by customizing `grad`, again allowing different behavior depending on `(vo, vi, q)`, e.g. _Black-Box Variational Inference_ [4]
- [x] User-implemented VI methods `CustomVI <: VariationalInference` by implementing `vi(model, alg::CustomVI)`, `grad(alg::CustomVI, ...)`, `optimize` (_optionally_ if we implement a default `optimize` method, which seems reasonable to do).

# Notes / To-be discussed
1. Optimization is not _in-place_. This causes some potential weird behaviour when calling the optimisers repeatedly. As a hacky fix, I've just added a check to see if the `acc` field in the optimizer, which is a `IdDict`, already has a key with identical value as `vcat(q.mu, q.sigma)` I'll initialize the optimization using the corresponding this value. It works but most certainly can have weird effects if two _different_ keys in the `IdDict` has the same _value_. This is an additional pro of making the optimization in-place, the other argument simply being performance. Lose some composability though.
2. I believe there are a couple of good reasons to have VI in a different submodule than `Inference`. At some point we might add support for Approximate Bayesian Computing (ABC) (see issue #347). Such methods I believe fit well into the current `Inference` module, maybe as an `Approximate` submodule of `Inference`. This combined with the different interface `vi(model, alg) # => <: VariationalPosterior` and `sample(model, alg) # => Chain` I believe a separate submodule in Turing is appropriate. 
3. Instead of having callable objectives, we could instead implement `objective(vo, vi, q, model, ...)`. This might lead to a more uniform interface, but I believe the interface of `vo(vi, q, model, ...)` is more intuitive and nicer looking. Similarily, we _could_ implement a calling function for the `VariationalInference` objects as a replacement of the `optimization` function, e.g. `advi(elbo::ELBO, q, model::Model, ...)`. It makes for a "nicer" interface in total if the user wants to do more than the simple `vi(model, ADVI())` call, but could also make the interface less "uniform". In fact, this could then replace the `vi(model, alg)` with `alg = ADVI(); alg(model) # => <: VariationalPosterior`. 
4. ~~Currently using  `ForwardDiff` for `ADVI`; haven't thought much yet about how to allow different AD approaches. This is also something that needs to be addressed, especially since for amortized VI we're going to want to support _inference networks_ which can be constructed using Flux.jl [5].~~ Now support both `ForwardDiff` and `Flux.Tracker` (though `Tracker` is slow).
5. One requirement that is often imposed on variational posteriors is being transformed to have support on the real numbers. This means that we're heavily dependent on good and fast bijectors implemented. Currently using Bijectors.jl, but jacobians of these transforms and their inverses are not yet exposed, and so we've simply implemented these using `ForwardDiff` for the moment. This can definitively be improved.
6. More thought can go into the how to implement these `VariationalPosterior` subtypes. `MeanField` is currently the only one. Depending on the method, these need different levels of information about the original `Model`. `MeanField` only requires a list of the distributions, and currently obtains this by extracting `dists` and `ranges` from `VarInfo` when calling `vi(model, ADVI())`. Then there is _structural_ VI which basically requires a DAG representing the model [5]. Ideally we'd be able to separate this `Variational` submodule out from Turing at some point, at which point we need to have a standard approach on how to tackle this.
    1. Even the current `MeanField` approach could be improved. When sampling etc. from `MeanField` we iterate through `dists`, sample and inverse-transform using Bijectors.jl. If we had a slightly improved interface for transformed distributions, e.g. `Transformed{D <: Distribution}` or something, this would simply be a list of `rand` calls. See issue at https://github.com/TuringLang/Bijectors.jl/issues/6.
    2. Should these variational posteriors have their own abstract type, which is a subtype of `Distributions{Multivariate, Continuous}`? This would separate these models from standard `Distribution`, which might be suitable since `VariationalPosterior` also requires some information about the structure of the model; thus a `VariationalPosterior` is in some sense tied to the original `Model` it approximates, while a `Distribution` of course has no such "ties".
7. How to provide `q` probably needs some discussion, since for some methods (e.g. `ADVI`) the variational posterior `q` is really _fixed_ and not something to change. But for other methods where entire families of models are possible, user-specification of the `q` is necessary. We'd also like to the keep this specification separate from the actual inference algorithm, e.g. _not_ do something like `VariationalInference(..., q)`. Keyword arguments could be a bit ambiguous because of the reason before; some _require_ a choice, some don't even have a choice. Suggestions:
    1. `vi(m, alg, q=alg isa ADVI ? advi_q_init(m) : error("method require explicit variational distribution"))` by @xukai92.
8. There are some methods, e.g. Semi-Implicit VI, where they're not only using MC estimation of the ELBO, but instead estimating a _lower bound_ of the ELBO. It's not 100% clear to me if this should then be a separate `VariationalObjective` or we simply make `(::ELBO)(vi::SIVI, ...)` an implementation that actually computes this lower-bound of the ELBO rather than the ELBO itself. I'm partial to a different `VariationalObjective` so as to not confuse the user.
9. Currently using `Flux.Optimise` but could also use Knet.jl's optimizers, which does _not_ use an `IdDict` like `Flux` does.

# TODO
- [ ] Perform a convergence check for the optimization rather than a fixed `max_iter` (though this should also be specified). What metric to track? 
  - Running mean?
- [x] Support both `ForwardDiff` and `Tracker`

# References
[1] Titsias, M. K., & Ruiz, F. J. R., Unbiased Implicit Variational Inference, CoRR, (),  (2018). 
[2] Airoldi, E. M., Blei, D. M., Fienberg, S. E., & Xing, E. P., Mixed membership stochastic blockmodels, Journal of machine learning research, 9(Sep), 1981–2014 (2008). 
[3] Liu, Q., & Wang, D., Stein variational gradient descent: a general purpose bayesian inference algorithm, CoRR, (),  (2016). 
[4] Ranganath, R., Gerrish, S., & Blei, D. M., Black Box Variational Inference, CoRR, (),  (2013). 
[5] Zhang, C., Butepage, J., Kjellstrom, H., & Mandt, S., Advances in variational inference, CoRR, (),  (2017).


src/variational/advi.jl
```diff
@@ -0,0 +1,213 @@
...
+    end
+
+    @info time_elapsed
+
+    μ, ω = x[1:num_params], x[num_params + 1: end]
+
+    return μ, ω
+end
+
+function (elbo::ELBO)(q::MeanField, model::Model, μ::Vector{T}, ω::Vector{T}, num_samples) where T <: Real
```
trappmartin: I suppose the computation of the ELBO should be independent of the VI algo. If so, maybe you could move this code and use `link(rand(q))` instead of the hard-coded `center_diag_gaussian_inv`?

src/variational/advi.jl
```diff
@@ -0,0 +1,213 @@
...
+
+    # TODO: in (Blei et al, 2015) TRUNCATED ADAGrad is suggested; this is not available in Flux.Optimise
+    # Maybe consider contributed a truncated ADAGrad to Flux.Optimise
+
+    i = 0
+    prog = PROGRESS[] ? ProgressMeter.Progress(max_iters, 1, "[$alg_name] Optimizing...", 0) : 0
+
+    time_elapsed = @elapsed while (i < max_iters) # & converged # <= add criterion? A running mean maybe?
+        # TODO: separate into a `grad(...)` call; need to manually provide `diff_result` buffers
+        ForwardDiff.gradient!(diff_result, f, x)
```
trappmartin: Turing has a `grad` function, maybe this could be adjusted so that you can use it here instead?

src/variational/advi.jl
```diff
@@ -0,0 +1,213 @@
...
+            ζ = center_diag_gaussian_inv(η, μ_i, exp.(ω_i))
+            
+            # inverse-transform back to domain of original priro
+            θ = invlink(prior, ζ)
+
+            # update
+            var_info.vals[r] = θ
+
+            # add the log-det-jacobian of inverse transform
+            elbo_acc += log(abs(det(jac_inv_transform(prior, ζ)))) / num_samples
```
trappmartin: Maybe use `logabsdet` instead?

src/variational/advi.jl
```diff
@@ -0,0 +1,213 @@
...
+    # instead of creating a diagonal matrix, we just do elementwise multiplication
+    (σ .^(-1)) .* (x - μ)
+end
+
+function center_diag_gaussian_inv(η, μ, σ)
+    (η .* σ) + μ
+end
+
+# Mean-field approximation used by ADVI
+struct MeanField{T, TDists <: AbstractVector{<: Distribution}} <: VariationalPosterior where T <: Real
```
trappmartin: I'd say MeanField is not restricted to ADVI. Maybe move this some place else.

test/skipped/advi_demo.jl
```diff
@@ -0,0 +1,129 @@
+using Turing
```
trappmartin: I don't think it necessary to have the demo as a test case. Maybe replace this with unit-tests.

test/skipped/advi_demo.jl
```diff
@@ -0,0 +1,129 @@
...
+opt = ADAGrad()                  # optimizer taking value and gradient to new value
+advi = ADVI(10, 100)             # <: VariationalInference
+q = vi(m, advi; optimizer = opt) # => MeanField <: VariationalPosterior
+
+elbo = Variational.ELBO()        # <: VariationalObjective
+
+history = [elbo(q, m, 1000)]     # history of objective evaluations
+
+# setup for plotting
+using Plots, StatsPlots, LaTeXStrings
```
trappmartin: If possible a unit-test should not require plotting.

src/variational/advi.jl
torfjelde: I actually think this should be dependent on VI algo in many cases, and that the signature should instead be `(elbo:::ELBO)(vi::ADVI, q::MeanField, ...)`. For example, the logabsdet-jac term isn't included in general when computing the ELBO, right? What I was thinking is to have a generic implementation `(elbo::ELBO)(vi, q, ...)` in the same file as impl of `ELBO` (`objectives.jl` right now), and then the different VI algorithms can override this if necessary, e.g. `ADVI`.

BUT this is why if we get a `Transformed` type for distributions, we can actually implement the `ELBO` for general `Transformed` by adding the logabsdet-jac term and implementing `entropy(q::VariationalPosterior)` instead of hard-coding into `ELBO`.

I was under the impression that `link` is supposed to be a bijective transformation such that we get a distribution with support on the entirety of the real line; so I believe for a `Normal` it's the identity transformation, not the `center_diag_gaussian_inv`. The _reparametrization_ we do using `center_diag_guassian_inv` here is to be able to compute gradients of the parameters (`μ` and `ω`).

This is actually something I forgot to mention in the "Notes" section! There are some methods, e.g. Semi-Implicit VI, where they're not only using MC estimation of the ELBO, but instead estimating a _lower bound_ of the ELBO. It's not 100% clear to me if this should then be a separate `VariationalObjective` or we simply make `(::ELBO)(vi::SIVI, ...)` an implementation that actually computes this lower-bound of the ELBO rather than the ELBO itself.
I'm partial to a different `VariationalObjective` so as to not confuse the user.

src/variational/advi.jl
torfjelde: Interested! Could you elobarate? I can only find the gradients of `logp`. 

Also, the comment I've written there is referring to the fact that I should follow the interface suggested in this PR (see above). Nonetheless, in _that_ function I'll need to compute the gradient.

src/variational/advi.jl
torfjelde: Cheers! Didn't know this was a thing.

src/variational/advi.jl
torfjelde: Agree. I was thinking that when we get more of these "custom" distributions, we move it into a different file, e.g. `distributions.jl` or something. But really no reason not to just do it right away!

test/skipped/advi_demo.jl
torfjelde: This was suggested by @xukai92, so I guess he'll have to chime in here. I was under the understanding that this was supposed to be "non-unit tests" and only used as a small script to demo the algo, hence why they're kept in the `skipped` folder.

I do have a simple unit-test in `tests/inference/vi.jl` btw (which should be separated into a `variational` one when if we agree on this submodule-structure).

test/skipped/advi_demo.jl
torfjelde: 100% agree; see previous comment regarding this file.

src/variational/advi.jl
trappmartin: Sure, I was referring to looking into the grad function of Turing and to adjust / make a new one similar to the grad function we have. This would help to make your code less dependent on the AD framework.

test/skipped/advi_demo.jl
trappmartin: I see. I didn't see it was skipped.

src/variational/advi.jl
trappmartin: Yes thats a good point. Having `elbo()` dispatch on the AD sounds like a reasonable idea. You are right, `link` is a bijective, see `link` and `invlink`. This transformation is only necessary if your prior is constrained, e.g. prior of scale of a Normal. For a Normal or the location of a normal you wouldn't need a transformation as it is already unconstrained.

Regarding Semi-Implicit VI, I guess I would stick to dispatching on `elbo` and makes sure its clear in the doc-string that this function actually returns a lower bound. I think we can change this later if necessary but it is probably better to first workout a consistent API that makes sense for most VI algos.

src/variational/advi.jl
torfjelde: Ah, follow :+1:  Somehow thought "this could be adjusted" referred to adjusting the above implementation to _use_ the already existing function; but you're suggesting adjusting existing gradient functions to _this_. Makes much more sense!:) 

I'll definitively try and do that; thanks!

torfjelde: So I've managed to support both `ForwardDiff` and `Tracker`, but the `Tracker` implementation runs _very_ slowly.

For the `ADVI` case I believe the optimization can be sped up quite a bit by simply removing the `grad!` call and performing the gradient computation directly in `optimize!`. This way we can define the `θ_tracked` outside of the loop, and use this for inplace gradient computation using `Tracker`. For `ForwardDiff`, we can define the objective function outside of the loop capturing the parameters by closing over them. This circumvents redefining the objective function upon each step.

Anyways, the current implementation is slow but works as expected.

src/variational/advi.jl
torfjelde: Great!
Hmm, so you agree that `link` and `center_diag_gaussian_inv` have different purposes?

That sounds like a good approach :+1:

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+abstract type VariationalObjective end
+
+const VariationalPosterior = Distribution{Multivariate, Continuous}
+
+"""
+    rand(vi::VariationalInference, num_samples)
+
+Produces `num_samples` samples for the given VI method using number of samples equal to `num_samples`.
+"""
+function rand(vi::VariationalPosterior, num_samples) end
```
xukai92: We probably want to throw an error as callback.

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+
+"""
+    objective(vi::VariationalInference, q::VariationalPosterior, model::Model, args...)
+
+Computes the variational objective to be optimized for a given VI method.
+"""
+function objective(vi::VariationalInference, q::VariationalPosterior, model::Model, num_samples) end
```
xukai92: Similar here.

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+
+"""
+    grad!(vo, vi::VariationalInference, q::VariationalPosterior, model::Model, θ, out, args...)
+
+Computes the gradients used in `optimize!`. Default implementation is provided for `VariationalInference{AD}` where `AD` is either `ForwardDiffAD` or `TrackerAD`. This implicitly also gives a default implementation of `optimize!`.
+
+Variance reduction techniques, e.g. control variates, should be implemented in this function.
+"""
+grad!(vo, vi::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ, out, args...) where AD =
+    error("Turing.Variational.grad!: unmanaged variational inference algorithm: $(typeof(alg))")
```
xukai92: Just like this. Also consider change "unmanaged xxx" to "xxx is not implemented"? How do you think?

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+    DiffResults.gradient!(out, Tracker.grad(θ_tracked))
+end
+
+"""
+    optimize!(vo, alg::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ; optimizer = ADAGrad())
+
+Iteratively updates parameters by calling `grad!` and using the given `optimizer` to compute the steps.
+"""
+function optimize!(vo, alg::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ; optimizer = ADAGrad()) where AD
+    # TODO: this default constructor for `optimizer` is probably dangerous because of the `acc` field being an `IdDict`!!!
```
xukai92: Not sure what does this mean. Let's discuss this tmr.

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+
+    # TODO: should we always assume `samples_per_step` and `max_iters` for all algos?
```
xukai92: It's OK to assume so now. Or maybe we can set them as keyword arguments to `vi(;max_iters=???)`?

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+    # setup
+    # var_info = Turing.VarInfo()
+    # model(var_info, Turing.SampleFromUniform())
+    # num_params = size(var_info.vals, 1)
+    num_params = length(q)
+
+    # # buffer
+    # θ = zeros(2 * num_params)
+
+    # HACK: re-use previous gradient `acc` if equal in value
```
xukai92: What does this mean?

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+            @info "[$alg_name] Re-using previous optimizer accumulator"
+            θ .= vs[idx]
+        end
+    else
+        @info "[$alg_name] Already present in optimizer acc"
+    end
+    
+    diff_result = DiffResults.GradientResult(θ)
+
+    # TODO: in (Blei et al, 2015) TRUNCATED ADAGrad is suggested; this is not available in Flux.Optimise
```
xukai92: Optimizer can also be an argument to pass in.

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+
+    Turing.DEBUG && @debug "Optimizing ADVI..."
+    θ = optimize(elbo, alg, q, model; optimizer = optimizer)
+    μ, ω = θ[1:length(q)], θ[length(q) + 1:end]
+
+    # TODO: make mutable instead?
+    MeanField(μ, ω, dists, ranges) 
+end
+
+# TODO: implement optimize like this?
```
xukai92: Let's stick to the function call version for now.

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+# end
+
+function optimize(elbo::ELBO, alg::ADVI, q::MeanField, model::Model; optimizer = ADAGrad())
+    θ = randn(2 * length(q))
+    optimize!(elbo, alg, q, model, θ; optimizer = optimizer)
+
+    return θ
+end
+
+function (elbo::ELBO)(alg::ADVI, q::MeanField, model::Model, θ::AbstractVector{T}, num_samples) where T <: Real
```
xukai92: Instead of a caller version, maybe we should call this function `estimate`?

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+    q = MeanField(zeros(num_params), zeros(num_params), dists, ranges)
+
+    # construct objective
+    elbo = ELBO()
+
+    Turing.DEBUG && @debug "Optimizing ADVI..."
+    θ = optimize(elbo, alg, q, model; optimizer = optimizer)
+    μ, ω = θ[1:length(q)], θ[length(q) + 1:end]
+
+    # TODO: make mutable instead?
```
xukai92: Do you mean update `q` above?

test/inference/vi.jl
```diff
@@ -0,0 +1,26 @@
+using Turing, Random, Test
+
+include("../test_utils/AllUtils.jl")
+
+@testset "vi.jl" begin
```
xukai92: Think about adding an example for which the true posterior is in the variational family.

test/runtests.jl
```diff
@@ -26,6 +26,7 @@ include("test_utils/AllUtils.jl")
             include("inference/mh.jl")
             include("inference/sghmc.jl")
             include("inference/AdvancedSMC.jl")
+            include("inference/vi.jl")
```
xukai92: Can we have some tests which we know the analytical answers for?

src/variational/VariationalInference.jl
willtebbutt: This will just throw an error if there's nothing implemented though, right?

src/variational/VariationalInference.jl
willtebbutt: Same here

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+function objective(vi::VariationalInference, q::VariationalPosterior, model::Model, num_samples) end
+
+# (::VariationalObjective)(vi::VariationalInference, model::Model, num_samples, args...; kwargs...) = begin
+# end
+
+
+"""
+    grad!(vo, vi::VariationalInference, q::VariationalPosterior, model::Model, θ, out, args...)
+
+Computes the gradients used in `optimize!`. Default implementation is provided for `VariationalInference{AD}` where `AD` is either `ForwardDiffAD` or `TrackerAD`. This implicitly also gives a default implementation of `optimize!`.
```
willtebbutt: Please reformat for 92 char limit. See style guide :)

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+
+Constructs the variational posterior from the `model` and performs the optimization following the configuration of the given `VariationalInference` instance.
+"""
+vi(model::Model, alg::VariationalInference) = 
+    error("Turing.Variational.vi: unmanaged variational inference algorithm: $(typeof(alg))")
+vi(model::Model, alg::VariationalInference, q::VariationalPosterior) =
+    error("Turing.Variational.vi: unmanaged variational inference algorithm: $(typeof(alg))")
+
+# default implementations
+function grad!(vo, alg::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ::AbstractVector{T}, out::DiffResults.MutableDiffResult, args...) where {T <: Real, AD <: ForwardDiffAD}
```
willtebbutt: Please see the `Method Definitions` section of the [style guide](https://turing.ml/docs/style-guide/) for method definitions spread over multiple lines.

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+grad!(vo, vi::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ, out, args...) where AD =
+    error("Turing.Variational.grad!: unmanaged variational inference algorithm: $(typeof(alg))")
+
+"""
+    vi(model::Model, alg::VariationalInference)
+    vi(model::Model, alg::VariationalInference, q::VariationalPosterior)
+
+Constructs the variational posterior from the `model` and performs the optimization following the configuration of the given `VariationalInference` instance.
+"""
+vi(model::Model, alg::VariationalInference) =
```
willtebbutt: Please see the Method Definitions section of the [style guide](https://turing.ml/docs/style-guide/) for method definitions of this size.

src/variational/VariationalInference.jl
```diff
@@ -0,0 +1,171 @@
...
+    DiffResults.value!(out, Tracker.data(y))
+    DiffResults.gradient!(out, Tracker.grad(θ_tracked))
+end
+
+"""
+    optimize!(vo, alg::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ; optimizer = ADAGrad())
+
+Iteratively updates parameters by calling `grad!` and using the given `optimizer` to compute the steps.
+"""
+function optimize!(vo, alg::VariationalInference{AD}, q::VariationalPosterior, model::Model, θ; optimizer = ADAGrad()) where AD
```
willtebbutt: Is there a particular reason that we're defaulting to `AdaGrad` here?

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+    samples_per_step # number of samples used to estimate the ELBO in each optimization step
+    max_iters        # maximum number of gradient steps used in optimization
+end
+
+ADVI(args...) = ADVI{ADBackend()}(args...)
+ADVI() = ADVI(10, 5000)
+
+alg_str(::ADVI) = "ADVI"
+
+vi(model::Model, alg::ADVI; optimizer = ADAGrad()) = begin
```
willtebbutt: Please see [style guide](https://turing.ml/docs/style-guide/) for function definitions. In short, please use
```
function foo(args...)

end
```
form, rather than this one.

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+
+    # construct objective
+    elbo = ELBO()
+
+    Turing.DEBUG && @debug "Optimizing ADVI..."
+    θ = optimize(elbo, alg, q, model; optimizer = optimizer)
+    μ, ω = θ[1:length(q)], θ[length(q) + 1:end]
+
+    # TODO: make mutable instead?
+    MeanField(μ, ω, dists, ranges)
```
willtebbutt: Please `return` this.

src/variational/advi.jl
willtebbutt: Why?

src/variational/advi.jl
```diff
@@ -0,0 +1,105 @@
...
+            
+            # inverse-transform back to domain of original priro
+            θ = invlink(prior, ζ)
+
+            # update
+            var_info.vals[r] = θ
+
+            # add the log-det-jacobian of inverse transform;
+            # `logabsdet` returns `(log(abs(det(M))), sign(det(M)))` so return first entry
+            elbo_acc += logabsdet(jac_inv_transform(prior, ζ))[1] / num_samples
```
willtebbutt: Why is `logabsdet` required here, rather than just `logdet`? Is `M` not guaranteed to be positive definite for some reason?

src/variational/distributions.jl
```diff
@@ -0,0 +1,57 @@
...
+    ForwardDiff.jacobian(x -> invlink(dist, x), x)
+end
+
+function jac_inv_transform(dist::Distribution, x::TrackedArray{T} where T <: Real)
+    Tracker.jacobian(x -> invlink(dist, x), x)
+end
+
+function center_diag_gaussian(x, μ, σ)
+    # instead of creating a diagonal matrix, we just do elementwise multiplication
+    (σ .^(-1)) .* (x - μ)
```
willtebbutt: What are the expected types of `x` and `μ` here? Should this whole thing not be fused?
Also, assuming that everything _should_ be fused, could this please be written as the cleaner
```julia
center_diag_gaussian(x, μ, σ) = (x .- μ) ./ σ
```

src/variational/distributions.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+function center_diag_gaussian_inv(η, μ, σ)
+    (η .* σ) + μ
```
willtebbutt: See previous comment

src/variational/distributions.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+# Mean-field approximation used by ADVI
+struct MeanField{T, TDists <: AbstractVector{<: Distribution}} <: VariationalPosterior where T <: Real
+    μ::Vector{T}
```
willtebbutt: Is there a particular reason to make the types here `Vector`s, rather than some subtype of `AbstractVector`?

src/variational/distributions.jl
```diff
@@ -0,0 +1,57 @@
...
+    ω::Vector{T}
+    dists::TDists
+    ranges::Vector{UnitRange{Int}}
+end
+
+Base.length(advi::MeanField) = length(advi.μ)
+
+_rand!(rng::AbstractRNG, q::MeanField{T, TDists}, x::AbstractVector{T}) where {T<:Real, TDists <: AbstractVector{<: Distribution}} = begin
```
willtebbutt: Style

willtebbutt:
This looks like really fantastic work so far. I've left some nit-picky style-related comments that will need to be addressed before merging. The existing Turing code-base isn't always completely consistent with the style guide, but it would be very nice if new contributions could be :)

src/variational/VariationalInference.jl
torfjelde: Glad you brought up this because I deliberately didn't implement throwing an error for these as I'm not sure we should even have them.

`rand` should already be "enforced" by the fact that `VariationalPosterior <: Distribution`. I put it there before making `VariationalPosterior` an alias to `Distribution{Multivariate, Continuous}`. Should I just remove it for now? 

We can add it back if we move away from aliasing `Distribution`.

src/variational/VariationalInference.jl
torfjelde: Similar to `rand`. I'm now of the opinion that we don't need to force the user to make an `objective` function, but instead allow any arbitrary _callable_ work as an objective.

Remove it for now?

src/variational/VariationalInference.jl
torfjelde: Sure!

src/variational/VariationalInference.jl
torfjelde: That's definitively a possible solution. The only problem is that not _all_ variational inference problems will need this, e.g. ELBO and it's gradients exists in closed form. Then it becomes a question of where the boundary is for a parameter to be a keyword argument, as opposed to being attached to the `VariationalInfernece` instance.

src/variational/VariationalInference.jl
torfjelde: So instead of being a keyword argument, we make it a positional argument? I'm also coming to the same opinion because of what I mentioned before.

Might make sense to let the `vi(model, alg)` choose a default optimizer for the different methods, rather than having a default optimizer in `optimize!`.

src/variational/advi.jl
torfjelde: Sounds good :+1:

src/variational/advi.jl
torfjelde: Yep!

test/inference/vi.jl
torfjelde: Ah, good idea!

src/variational/VariationalInference.jl
torfjelde: Usually the one (+ `RMSProp`) used for `ADVI` in the literature from what I've seen. I remember just trying a couple of different optimizers to make sure the code worked.  The few others I tried didn't converge so I just defaulted to something I knew worked.

src/variational/advi.jl
torfjelde: Yeah sorry about this; I even knew about the style guide, but then somewhere down the road I forgot about it... Sloppy on my part.

src/variational/advi.jl
torfjelde: As far as I'm aware, generally these transformations do not need to result in a positive-definite jacobian. The only requirement is that the matrix is _invertible_, or rather the mapping is a smooth map with a smooth inverse (by the inverse function theorem, this guarantees the jacobian at any point is full rank and thus invertible).

src/variational/distributions.jl
torfjelde: That's much nicer!

I'm wondering if there's a better abstract for  these "reparameterizations" though. Something like having a `<: Reparameterization` type which also have an `inverse` implemented.

src/variational/VariationalInference.jl
torfjelde: Multiple of the optimisers have an `IdDict` attached to them as they need to track certain quantities. Therefore, if the user performs VI by repeatedly calling `optimize!` but is for some reason not using the same `θ` ("same" in the sense that it has the same `Id` not just value), you end up creating a different quantity in the optimizer for each call. In this case the algorithm is probably not doing what the user _thinks_ it's doing.

src/variational/VariationalInference.jl
torfjelde: How do you do this when the signature of the method is spread > 92 chars? E.g.
```
    grad!(vo, vi::VariationalInference, q::VariationalPosterior, model::Model, θ, out, args...)
```
is > 92 chars.

willtebbutt: Yeah, I would also spread this over multiple lines -- in general prefer to break up stuff over multiple lines than have a single long line, or have a single short line that loses information (i.e. please don't sacrifice information for the sake of brevity)

Definitely don't exceed the 92 char limit, but do feel free to write a docstring that runs over multiple lines. e.g.
```
"""
    foo(args...)

Some descriptive text that ...
runs over 92 characters.

# Arguments:
- 
"""
```
is totally fine.

src/variational/VariationalInference.jl
willtebbutt: Hmm interesting. Did you try Adam / AMSGrad?

src/variational/advi.jl
willtebbutt: No problem :)

src/variational/advi.jl
willtebbutt: Yeah, of course, you're totally right.

src/variational/VariationalInference.jl
torfjelde: I remember trying ADAM and pretty sure it failed to converge. I believe it's because ADAM attempts to estimate second-order information and in ADVI you usually use quite few samples to estimate the ELBO. This then leads to high variance for the gradient estimate, in which case estimates second-order information can be quite unreliable.

Definitively something to explore further though.

src/variational/advi.jl
xukai92: As we are actually estimating it?

willtebbutt: Sounds reasonable. How does this play with the `objective` function? Do we need both?

torfjelde: We could do something like `estimate(elbo::ELBO, ...)` instead of having this `objective`. My concern is that this potentially makes it harder to allow arbitrary functions as objectives since the user then basically has to subtype `VariationalObjective`?

And writing generic `optimize!` becomes more difficult since we now don't know whether or not to directly call the `VariationalObjective` or call `estimate`. This was my original reason for not making some `estimate` function.

src/variational/VariationalInference.jl
torfjelde: Alright, I'll do that then.

A general question: I believe I've addressed the style issue (other than this one). Should I just resolve your comments as I see fit, or is the idea that I say "I've fixed it" and then you go through and resolve?

src/variational/advi.jl
```diff
@@ -0,0 +1,112 @@
...
+
+            # add the log-det-jacobian of inverse transform;
+            # `logabsdet` returns `(log(abs(det(M))), sign(det(M)))` so return first entry
+            # add `eps` to ensure SingularException does not occurr in `logabsdet`
+            elbo_acc += logabsdet(jac_inv_transform(prior, ζ) .+ eps(T))[1] / num_samples
+        end
+
+        # compute log density
+        model(var_info)
+        elbo_acc += var_info.logp / num_samples
```
xukai92: Also a tiny thing. I believe in Julia, we tends to squash words in variable names when possible (https://docs.julialang.org/en/v1/manual/style-guide/index.html#Use-naming-conventions-consistent-with-Julia-base/-1). Maybe `var_info` should be `varinfo`?

src/variational/distributions.jl
xukai92: @torfjelde I found this one is not resolved. Can you quick fix it?

torfjelde: Ah, cheers! I've addressed this now:)

src/variational/advi.jl
torfjelde: Ah, didn't know; cheers! And fixed:)

xukai92: Nice thanks

xukai92:
Thanks for the hard work for this @torfjelde!

Issue #795, opened 2019-05-26
Documentation: Code error in Guide


zenon:
In https://turing.ml/docs/guide/ I found some issues:

"Note that the key (:var_1) should be a symbol. For example, to fetch x[1], one would need to do chn[Symbol(:x[1])." 

However, Symbol(:x[1]) doesn't work. Symbol("x[1]") works, but I can't judge whether it does the right thing.

Lower on the same page:

    xs = generated[:x]
    xs[15]

The latter line doesn't work:

ERROR: BoundsError: attempt to access 0-element Array{Any,1} at index [1]

There I'm still stuck.

zenon:
Thank you!

cpfiffer:
Closed by #797.

cpfiffer:
Also @zenon, let me know if the updates to the guide didn't help. I'm happy to add a bit more to explain whatever's not clear.

Issue #786, opened 2019-05-13
"sample from the prior" vs "sample from generative model'


scheidan:
I'm trying to understand how to sample from the prior and how this contrasts to sampling from a "generative model'.

Do I interpret the example from the documentation correctly?
```julia
# define joint distribution p(s, m, x, y)
@model gdemo(x, y) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0,sqrt(s))
    x ~ Normal(m, sqrt(s))
    y ~ Normal(m, sqrt(s))
    return x, y
end

# this samples from p(x,y)  
g_prior_sampler = gdemo()
g_prior_sampler()

# this samples p(x|y=-3.4)
c = sample(gdemo(missing, -3.4), NUTS(5000, 0.65))
```
Based on this I played around with the second example:
```Julia
# define p(s, m, [x_1, ..., x_n])
@model gdemo2(x) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))
    for i in eachindex(x)
        x[i] ~ Normal(m, sqrt(s))
    end
    return x
end

# sample from  p(s, m, x)
g_prior_sampler2 = gdemo2()
g_prior_sampler2()   # -> ERROR: MethodError: no methods matching keys(::Symbol)
# I guess the error is because the length of x in not defined.
# However passing the length as extra parameter does not resolve this.

# sample from p(s, m, [x_1, x_2, x_3])  (that's also sampling from the prior)
c = sample(gdemo2([missing, missing, missing]), NUTS(5000, 0.65))
# -> works :)

# sample from p(s, m, [x_1, x_2] | x_3=2.2)
c = sample(gdemo2([missing, missing, 2.2]), NUTS(5000, 0.65))
# -> ERROR: MethodError: no methods matching (::Colon)(::Int64, ::Base.OneTo{Int64})
```

The last case would be very useful in applications with missing data.

(probably related to #446.)

cpfiffer:
You can fix the first case by passing in a default vector type for `x`, as below:

```julia
@model gdemo2(x=repeat(Union{Missing, Float64}[missing], 3)) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))
    for i in eachindex(x)
        x[i] ~ Normal(m, sqrt(s))
    end
    return x
end
```

This allows Turing's compiler to infer both type and length for `x`.

[...]

PR #797, opened 2019-05-27
Guide updates

cpfiffer:
Addresses both #795 and #786. Includes updates on 

1. How to access values in a chain
2. Some more clarity on `missing` values

Does my language on conditional/unconditional seem appropriate?

### DIFF
--- a/docs/src/_docs/guide.md
+++ b/docs/src/_docs/guide.md
@@ -168,7 +168,7 @@ Output:
 (0.685690547873451, -1.1972706455914328)
 ```
 
-### Sampling from the Conditional Prior
+### Sampling from a Conditional Distribution
 
 #### Using `Missing`
 
--- a/docs/src/_docs/guide.md
+++ b/docs/src/_docs/guide.md
@@ -97,7 +97,9 @@ The returned chain contains samples of the variables in the model.
 var_1 = mean(chn[:var_1]) # Taking the mean of a variable named var_1.
 ```
 
-Note that the key (`:var_1`) should be a symbol. For example, to fetch `x[1]`, one would need to do `chn[Symbol(:x[1])`.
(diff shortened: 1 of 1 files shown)


docs/src/_docs/guide.md
```diff
@@ -97,7 +97,9 @@ The returned chain contains samples of the variables in the model.
 var_1 = mean(chn[:var_1]) # Taking the mean of a variable named var_1.
 ```
 
-Note that the key (`:var_1`) should be a symbol. For example, to fetch `x[1]`, one would need to do `chn[Symbol(:x[1])`.
+Tthe key (`:var_1`) can be a `Symbol` or a `String`. For example, to fetch `x[1]`, one can use `chn[Symbol("x[1]")` or `chn["x[1]"]`.
```
yebai: `Tthe` ==> `The`?

docs/src/_docs/guide.md
```diff
@@ -140,7 +142,7 @@ Having multiple chains in the same object is valuable for evaluating convergence
 
 Please note that Turing does not have native support for chains sampled in parallel.
 
-### Sampling from the Prior
+### Sampling from the Unconditional Prior
```
yebai: It sounds a bit strange to say "Unconditional Prior". Since in most cases, "Prior" refers to the distribution before conditioning on any data. Perhaps consider rewording the title to something else, e.g. "Sampling from the Unconditional Distribution (i.e. the prior)".

docs/src/_docs/guide.md
```diff
@@ -165,11 +168,11 @@ Output:
 (0.685690547873451, -1.1972706455914328)
 ```
 
-### Generative Models
+### Sampling from a Conditional Distribution
```
yebai: Consider "Sampling from a Conditional Distribution (e.g. the posterior)"?

cpfiffer:
Thanks for catching the weird language I had in there, @yebai.

PR #798, opened 2019-05-28
Hacks to make MvNormal work with Tracker

willtebbutt:
This turned out to be a bit involved than I had hoped. Unfortunately, it turned out to be straightforward to roll our own `TuringMvNormal` and `TuringDiagNormal` types to work with `Tracker` than implement the appropriate methods to make `MvNormal` and `DiagNormal` (?) work with `Tracker` -- I'm not even sure that the latter is possible. Fortunately, the user should never have to see these new types as they only get created if `Tracker` is being used (see new methods of `Distributions.MvNormal`).

There are a few aspects of this that I need to improve upon before merging:

- Checking Tracker doesn't have method ambiguities with certain combinations of tracked / untracked types.
- Checking that I'm not accidentally converting `TrackedMatrix` -> `Matrix{<:TrackedReal}` anywhere, as this will kill performance.

I could use feedback on the following:

- Are there any bits of code that are unclear? Which bits could do with extra comments?
- Have I implemented `rand` and `logpdf` correctly for `Turing`'s purposes? Should I be doing `_rand`, or multi-sample versions of `rand` / `logpdf` for any reason?
- Ideas for extra tests -- everyone uses `MvNormal` so it would be nice to do as much testing up front as possible.
- Anything else you think I might have missed.


willtebbutt:
Thanks for the feedback @mohamed82008 

> It would be nice to test logdet separately.

Good call

> I think the case of equal variance MvNormal is not supported.

Could you elaborate on this a bit? I'm not quite sure what you're referring to.

> Why do we need 2 types for TuringMvNormal, can't we use type parameters to distinguish them?

Absolutely could do. Will attempt to refactor to make this work. Shouldn't be too tricky.

> It would be nice to summarize the difficulties you had with Distributions.MvNormal and PDMats while it is still fresh. These can be revisited again in the future on the Distributions side. I ran into similar issues before but I can't recall much now.

It's more an issue with the way `Tracker` does its `TrackedArray` type, rather than `Distrbutions` or `PDMat`s tbh. Sometimes the element-type type-parameter doesn't seem to get a sensible value, which causes problems when working with objects that wrap `TrackedArrays`, and have type parameters involving their element types.

willtebbutt:
Here are some benchmarking results on a 101-dimensional Gaussian with dense covariance

Set up:
```julia
using Turing, Random, BenchmarkTools, Tracker

rng, N = MersenneTwister(123456), 101
B = randn(rng, N, N)

m, Σ = randn(rng, N), B' * B + I
x = rand(MvNormal(m, Σ))

logpdf_func = (x, m, Σ)->logpdf(MvNormal(m, Σ), x)
```

For reference:
```julia
@benchmark logpdf_func($x, $m, $Σ)
BenchmarkTools.Trial: 
  memory estimate:  81.77 KiB
  allocs estimate:  11
  --------------
  minimum time:     85.202 μs (0.00% GC)
  median time:      100.433 μs (0.00% GC)
  mean time:        195.271 μs (3.58% GC)
  maximum time:     35.038 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Forwards pass:
```julia
@benchmark Tracker.forward(logpdf_func, $x, $m, $Σ)

# Results on master
BenchmarkTools.Trial: 
  memory estimate:  55.41 MiB
  allocs estimate:  1712029
  --------------
  minimum time:     102.502 ms (71.46% GC)
  median time:      187.646 ms (81.35% GC)
  mean time:        220.831 ms (84.58% GC)
  maximum time:     495.369 ms (92.13% GC)
  --------------
  samples:          23
  evals/sample:     1

# Results with this PR
BenchmarkTools.Trial: 
  memory estimate:  249.58 KiB
  allocs estimate:  112
  --------------
  minimum time:     98.783 μs (0.00% GC)
  median time:      115.957 μs (0.00% GC)
  mean time:        139.343 μs (6.00% GC)
  maximum time:     26.276 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Results of reverse-pass:
```julia
y, back = Tracker.forward(logpdf_func, x, m ,Σ)
@benchmark $back(1.0)

# Results on master
BenchmarkTools.Trial: 
  memory estimate:  985.54 MiB
  allocs estimate:  5987052
  --------------
  minimum time:     6.602 s (86.61% GC)
  median time:      6.602 s (86.61% GC)
  mean time:        6.602 s (86.61% GC)
  maximum time:     6.602 s (86.61% GC)
  --------------
  samples:          1
  evals/sample:     1

# Results with this PR
BenchmarkTools.Trial: 
  memory estimate:  495.80 KiB
  allocs estimate:  160
  --------------
  minimum time:     319.320 μs (0.00% GC)
  median time:      363.014 μs (0.00% GC)
  mean time:        438.892 μs (4.74% GC)
  maximum time:     27.671 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Clearly, the performance is several orders of magnitude better.

yebai:
> Clearly, the performance is several orders of magnitude better, especially for the reverse pass. Pleasingly, the forwards-pass with this PR has minimal overhead relative to usual evaluation.

This looks great - many thanks, @willtebbutt.

mohamed82008:
Nice speedup! Sorry for late response. By constant variance `MvNormal`, I mean this case `MvNormal([1.0, 1.0], s)`. Inside, I believe it is stored as a single value.

willtebbutt:
The most recent PR substantially improves unit testing, and adds support for all of the `MvNormal` methods that don't involve a `PDMat` (which we are explicitly _not_ handling, as they don't currently work with `Tracker`)

src/core/ad.jl
```diff
@@ -258,25 +258,174 @@ function poislogpdf(v::ForwardDiff.Dual{T}, x::Int) where {T}
...
+MvNormal(m::TrackedVector{<:Real}, σ::AbstractVector{<:Real}) = TuringDiagNormal(m, σ)
+MvNormal(m::TrackedVector{<:Real}, σ::Vector{<:Real}) = TuringDiagNormal(m, σ)
+MvNormal(m::AbstractVector{<:Real}, σ::TrackedVector{<:Real}) = TuringDiagNormal(m, σ)
+
+# dense mean, constant variance
+MvNormal(m::TrackedVector{<:Real}, σ::TrackedReal) = MvNormal(m, fill(σ, length(m)))
+MvNormal(m::TrackedVector{<:Real}, σ::Real) = MvNormal(m, fill(σ, length(m)))
+MvNormal(m::AbstractVector{<:Real}, σ::TrackedReal) = MvNormal(m, fill(σ, length(m)))
+
+# dense mean, constant variance
```
yebai: nice!

PR #803, opened 2019-06-01
TS 3: Hook sample to TypedVarInfo

mohamed82008:
This PR does some small changes to hook `sample` to `TypedVarInfo` for all the samplers of Turing. For this PR to be maximally useful, the following needs to be done possibly in separate PRs:
1. `space` should be part of the type signature of `alg` such that `getspace(alg)` is known at compile time.
2. The construction of `vi::VarInfo` should be separated from the rest of the sampling process by a function barrier. So first construct `vi` using `vi = VarInfo(model)` or `vi = empty!(VarInfo(model))` then pass it to a function to do the sampling. This is relevant to your PR @cpfiffer. Inside the sampling function, you can do `empty!(vi)` to empty the `vi`, or `deepcopy(vi)` if need be; both of these don't hurt the inference.
3. `Sample` should not use `Dict{Symbol, Any}` for its `value` field. The `Any` should be replaced by a concretely typed struct preferably.
4. `spl.info` should be replaced by a concretely typed struct specialized for each sampler.

If all of these changes are completed, Turing should be about as fast as claimed [here](https://github.com/TuringLang/AdvancedHMC.jl/issues/26#issuecomment-475702331) assuming AHMC is type stable and fast.

### DIFF
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -1,7 +1,7 @@
 module RandomVariables
 
 using ...Turing: Turing, CACHERESET, CACHEIDCS, CACHERANGES, Model,
-    AbstractSampler, Sampler, SampleFromPrior,
+    AbstractSampler, Sampler, SampleFromPrior, SampleFromUniform,
     Selector, getspace
 using ...Utilities: vectorize, reconstruct, reconstruct!
 using Bijectors: SimplexDistribution, link, invlink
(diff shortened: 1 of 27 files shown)


cpfiffer:
Regarding number two, the interface currently has this structure:

```julia
function sample(
    rng::AbstractRNG,
    ℓ::ModelType,
    s::SamplerType,
    N::Integer;
    kwargs...
) where {ModelType<:Sampleable, SamplerType<:AbstractSampler}
    # Perform any necessary setup.
    sample_init!(rng, ℓ, s, N; kwargs...)

    # Preallocate the TransitionType vector.
    ts = transitions_init(rng, ℓ, s, N; kwargs...)

    # Step through the sampler.
    for i=1:N
        if i == 1
            ts[i] = step!(rng, ℓ, s, N; kwargs...)
        else
            ts[i] = step!(rng, ℓ, s, N, ts[i-1]; kwargs...)
        end
    end

    # Wrap up the sampler, if necessary.
    sample_end!(rng, ℓ, s, N, ts; kwargs...)

    return Chains(ts)
end
```

My modifications to the sampler code has changed the return value of `SMC(. . .)` from `InferenceAlgorithm` to `Sampler{SMC, SMCState}`, where `SMCState` contains the `VarInfo`. As it is right now, I would set the `SMCState` to hold the `TypedVarInfo` in the `sample_init!` phase above.

Is this problematic? Most of the actual inference work happens in the `step!` call, and by then everything would be type aware, but the actual `sample` function would not be type aware (I think). Do you have any thoughts on that?

mohamed82008:
> My modifications to the sampler code has changed the return value of SMC(. . .) from InferenceAlgorithm to Sampler{SMC, SMCState}, where SMCState contains the VarInfo.

Will `SMC(..)` have access to the model? This will be a big change in the API.

> As it is right now, I would set the SMCState to hold the TypedVarInfo in the sample_init! phase above.

So is the `VarInfo` constructed inside `sample` using `sample_init!` or outside? If inside, then we should have a function barrier between the main loop and `sample_init!`. `sample` would not be type stable but that's because it is a thin wrapper of a type stable function. 

If constructed outside as part of `sampler`, then the API change is necessary because we need `model` to construct `TypedVarInfo`. I prefer this approach because `sample` itself can then be made type stable. But this requires including the `model` as a field in the `SamplerType`, which means the `model` won't be passed to `sample`, but to `SMC(...)` for example.

cpfiffer:
Alright, I've mostly fixed this on my side. I'm now supporting a wrapper function around `sample` that takes an inference algorithm instead of a sampler, so it's possible to construct the sampler, type the `VarInfo`, and do everything we need to before we hit the main interface calls to `sample_init!` and `step!`. I think it's a non-issue now.

cpfiffer:
Something else I forgot to mention: currently we have a smattering of `Sampler` functions that only accept `Sampler(alg, selector)`, which is I think not what you need. My branch is removing the ability to call `Sampler(alg, selector)`, and only allowing calls including a model, as with `Sampler(alg, model, selector)`. 

That's where I'd imagine the `VarInfo` becomes a `TypedVarInfo`, and then I could even put `TypedVarInfo` in the sampler state so there's no possibility of bad typing.

mohamed82008:
I don't quite get the details but just make sure `TypedVarInfo` is constructed once only at the beginning. You can copy and empty it freely, but constructing it messes up the inference of the code that uses it.

cpfiffer:
Yeah, I think I can manage that. Thanks!

mohamed82008:
There are some correctness tests failing for particle samplers, IS and MH beside a few other tests. I will hunt them down one by one.

yebai:
@mohamed82008 Is this ready for a review?

mohamed82008:
Yes pretty much. I was going to give it one more read, but that can happen during review.

cpfiffer:
I've got nothing to add, looks great!

PR #811, opened 2019-06-13
Minor update for #786

cpfiffer:
Adds more detail on the `return` statement and clarifies the role of `x` and `y` in the guide.


yebai:
Thanks @cpfiffer!

PR #819, opened 2019-06-21
[WIP] Heterogeneous vector priors

mohamed82008:
This PR adds support for heterogeneous vectors of prior distributions. The following works as expected:
```julia
using Turing

@model demo() = begin
    a ~ [Normal(), Beta()]
end
model = demo();
alg = HMC(10000, 0.1, 5)
chain = sample(model, alg)
```
This is only supported for parameters so far, but observations can be supported in a similar fashion.

CC: @ChrisRackauckas please try this branch

### DIFF
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -975,10 +975,12 @@ function getindex(vi::AbstractVarInfo, vn::VarName)
 end
 function getindex(vi::AbstractVarInfo, vns::Vector{<:VarName})
     @assert haskey(vi, vns[1]) "[Turing] attempted to replay unexisting variables in VarInfo"
-    dist = getdist(vi, vns[1])
-    return copy(istrans(vi, vns[1]) ?
-        invlink(dist, reconstruct(dist, getval(vi, vns), length(vns))) :
-        reconstruct(dist, getval(vi, vns), length(vns)))
+    n = length(vns)
+    return map(1:n) do i
+        dist = getdist(vi, vns[i])
+        val = reconstruct(dist, getval(vi, vns[i]))
+        istrans(vi, vns[i]) ? invlink(dist, val) : val
+    end
 end
 
 """
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -118,7 +118,7 @@ function assume(spl::A,
 end
 
 function assume(spl::Union{SampleFromPrior, SampleFromUniform},
-    dists::Vector{<:Distribution},
+    dists::AbstractVector{<:Distribution},
     vn::VarName,
     vi::VarInfo)
 
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -117,15 +117,12 @@ function assume(spl::A,
 
 end
 
-function assume(spl::A,
-    dists::Vector{T},
+function assume(spl::Union{SampleFromPrior, SampleFromUniform},
+    dists::Vector{<:Distribution},
     vn::VarName,
-    var::Any,
-    vi::VarInfo) where {T<:Distribution, A<:Union{SampleFromPrior, SampleFromUniform}}
+    vi::VarInfo)
 
-    @assert isa(var, Vector) "Turing.assume: unsupported variable container."
-
-    n = length(var)
+    n = length(dists)
     vns = map(i -> VarName(vn, "[$i]"), 1:n)
 
     if haskey(vi, vns[1])
@@ -144,7 +141,7 @@ function assume(spl::A,
         var = rs
     end
 
-    @assert length(var) == length(rs) "Turing.assume: variable and random number dimension unmatched"
+    @assert n == length(rs) "Turing.assume: variable and random number dimension unmatched"
 
     # acclogp!(vi, logp)
     logp = sum(1:n) do i 
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -123,45 +123,35 @@ function assume(spl::A,
     var::Any,
     vi::VarInfo) where {T<:Distribution, A<:Union{SampleFromPrior, SampleFromUniform}}
 
-    @assert length(dists) == 1 "Turing.assume only support vectorizing i.i.d distribution"
-    dist = dists[1]
-    n = size(var)[end]
+    @assert isa(var, Vector) "Turing.assume: unsupported variable container."
 
+    n = length(var)
     vns = map(i -> VarName(vn, "[$i]"), 1:n)
 
     if haskey(vi, vns[1])
         rs = vi[vns]
     else
-        rs = isa(spl, SampleFromUniform) ? init(dist, n) : rand(dist, n)
-
-        if isa(dist, UnivariateDistribution) || isa(dist, MatrixDistribution)
-            for i = 1:n
-                push!(vi, vns[i], rs[i], dist, spl)
-            end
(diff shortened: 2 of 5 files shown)


mohamed82008:
The old "vectorization" support was removed in the process, but I will see how to add it back again.

mohamed82008:
For the case where we have a vector of parameters all of which have the same prior, we can use `fill` or the lazy `FillArrays`:
```julia
using FillArrays, Turing

@model demo() = begin
    a ~ Fill(Normal(), 2)
end
```

src/core/RandomVariables.jl
```diff
@@ -975,10 +975,12 @@ function getindex(vi::AbstractVarInfo, vn::VarName)
...
-    dist = getdist(vi, vns[1])
-    return copy(istrans(vi, vns[1]) ?
-        invlink(dist, reconstruct(dist, getval(vi, vns), length(vns))) :
-        reconstruct(dist, getval(vi, vns), length(vns)))
+    n = length(vns)
+    return map(1:n) do i
+        dist = getdist(vi, vns[i])
+        val = reconstruct(dist, getval(vi, vns[i]))
+        istrans(vi, vns[i]) ? invlink(dist, val) : val
+    end
```
xukai92: Why do we replace the use on `vns` with `map`?

src/inference/Inference.jl
```diff
@@ -117,51 +117,38 @@ function assume(spl::A,
...
+            rs = isa(spl, SampleFromUniform) ? init.(dists) : rand.(dists)
+        else
+            throw("The distribution and values vectors have mis-matching lengths.")
         end
-    end
 
-    # acclogp!(vi, sum(logpdf_with_trans(dist, rs, istrans(vi, vns[1]))))
+        push!.(Ref(vi), vns, rs, dists, Ref(spl))
+        var = rs
+    end
```
xukai92: Do we still handle different formats of multiple samples from multivariate distribution with this change? E.g. `Vector{Vector}` and `Matrix`.

xukai92: Same question below.

src/core/RandomVariables.jl
mohamed82008: This is doing pretty much the same thing but considering every `vns[i]` can have a different distribution `dists[i]`. Also the output will be a vector of reconstructed values always, not a matrix sometimes. Each reconstructed value can be a scalar, vector or matrix depending on the distribution of the `vns[i]`.

src/inference/Inference.jl
mohamed82008: No the output of `vi[vns]` is now always a vector.

yebai:
This looks like a generalisation of the vectorisation support when combined with the `FillArrays ` trick. One minor comment is how this interacts with the `.~` (see https://github.com/TuringLang/Turing.jl/issues/476). 

I am happy with the changes otherwise.

mohamed82008:
I like the following proposal:
1. When the matrix `M` is column-wise distributed with an `MvNormal` distribution, the user can do `M ~ MvNormal(...)`. This can lower to `sum(logpdf(dist, val))` where `dist` is the `MvNormal` and `val` is `M`, whether `M` is a matrix or not. When the result of `logpdf` is scalar, `sum` will be a no-op. So this basically directly supports the "matrix syntax" of `Distributions`. This also means that the whole `M` will have a single `VarName` associated.
2. I think `M .~ MvNormal(...)` should always have Julia broadcast semantics. This means that it needs to lower to `sum(logpdf.(dist, val))`. 
3. Finally, `M ~ [MvNormal(...), ...]` should always have the Julia `map` semantics, where the LHS and RHS are vectors of the same size, and each `val[i]` is distributed according to `dists[i]`.

Comments?

mohamed82008:
In this PR, I can support points 1 and 3 if you like. Point 2 can be left for another PR.

yebai:
> In this PR, I can support points 1 and 3 if you like. Point 2 can be left for another PR.

Supporting 1 and 3 sounds good to me. For 1, maybe also consider support Multivariate distributions other than `MvNormal`, since they are similar to each other. See https://juliastats.github.io/Distributions.jl/stable/multivariate/

  @willtebbutt @xukai92 do you have any comments?

src/inference/Inference.jl
```diff
@@ -117,51 +117,38 @@ function assume(spl::A,
 
 end
 
-function assume(spl::A,
-    dists::Vector{T},
+function assume(spl::Union{SampleFromPrior, SampleFromUniform},
```
willtebbutt: Style. Please see guide

willtebbutt:
Looks totally reasonable to me :slightly_smiling_face:. Definitely the kind of building block that we need to be able to do Bayesian Deep Learning efficiently -- any chance that we could generalise this to matrices so that I can place an iid prior over the elements of a matrix?

mohamed82008:
> any chance that we could generalise this to matrices / arbitrary dimensional arrays so that I can place an iid prior over the elements of a matrix?

This is possible to support with the `map` semantics, which point 3 above, but is most naturally supported as a broadcast. Either way my implementations will be generic enough to handle arbitrary dimensional arrays for points 2 and 3 above. So `M ~ Fill(dist, size(M))` will do what you want. When `.~` is supported you could also do `M .~ dist`.

mohamed82008:
> For 1, maybe also consider support Multivariate distributions other than MvNormal, since they are similar to each other.

Yes of course, `MvNormal` was just an example :)

src/inference/Inference.jl
xukai92: Then how does it work with multivariate distribution?

xukai92: Like will `x ~ [Fill(MvNormal(), 2)]` works both for `x::Matrix` and `x::Vector{Vecotr}`?

mohamed82008: For `x::Matrix` that is column-wise, `~ MvNormal(zeros(2), 1)`, the user would need to do `x ~ MvNormal(zeros(2), 1)`. This is case 1 in the proposal given below https://github.com/TuringLang/Turing.jl/pull/819#issuecomment-504994714. This will be covered by the `dist::Distribution` method (to be changed in this PR) not the `dist::Vector{<:Distribution}` method which this PR currently modifies.

`x ~ [Fill(MvNormal(zeros(2), 1), 2)]` should arguably error. The rhs is a `Vector{<:Vector{<:Distribution}}`. I think we can should only support `Array{<:Distribution}`. This can be through `map` semantics  using `lhs ~ rhs::Array{<:Distribution}` where `size(lhs) == size(rhs)` (otherwise error), case 3 in the proposal below. Or it can be through broadcast semantics using `.~` which broadcasts `lhs` and `rhs` to matching sizes, case 2 in the proposal below.

mohamed82008:
This is not needed anymore. There are the following alternatives being worked on or already in-place:
1. The `.~` syntax in #965 which enables one to write:
```julia
a = Vector{T}(undef, 2) # or array
a .~ [Normal(), Beta()]
```
2. The `For` combinator/distribution in https://github.com/TuringLang/DistributionsAD.jl/pull/13 which enables the definition of arbitrarily sized lazy multivariate distributions from univariate ones.
3. `Distributions.Product` which can be used as a persistent version of `For` as:
```julia
a ~ Product([Normal(), Beta()])
```

yebai:
@mohamed82008 It might be useful to add this information to the docs, e.g. the customised distribution section or a new page.

mohamed82008:
Sure, will do in #965.

PR #793, opened 2019-05-25
Interface Changeover

cpfiffer:
# To do

Supported samplers (`src/inference/`)
- [x] IS
- [x] MH
- [x] Gibbs
- [x] HMC
  - [x] HMCDA
  - [x] NUTS
- [x] SMC
  - [x] PG
  - [x] CSMC

Contributed samplers (`src/contrib/inference`)
- [x] DynamicNUTS
- ~~[ ] SGLD~~
- ~~[ ] SGHMC~~
- ~~[ ] PIMH~~
- ~~[ ] PMMH~~
- ~~[ ] IPMCMC~~

# Summary

This will be the comprehensive PR containing all the code to change us over to using the new interface. Since the changes must be done for all the samplers pretty much all at once (we don't want half to use the new interface and the other half to use the old version), this will probably be a large PR. There is an `hmc-interface.jl` file which is not really ready for consumption yet, so you can ignore it.

Currently, I have a (very) rough proof-of-concept for SMC to get myself on stable ground for the rest of the samplers. The following interface is supported:

```julia
using Turing

@model test() = begin
    x ~ Normal(0, 1)
end

chn = sample(test(), SMC(50), 500)
```

I've introduced a couple of new things since the discussion in #746.

First, I'm attempting to unify all the samplers under the existing `Sampler` type we have. The `info` field is not going to be removed in this PR because of all the `VarInfo` stuff -- that will be a separate undertaking. The `Sampler` type now has an additional field called `state`, which contains a mutable struct that replaces the sampler-specific information like `logevidence` or `eval_num`. The ultimate goal is to have this go away at some point and move to a more general sampler state, but changing over to the front-end interface and changing the entire backend would be too much for one PR.

Second, this isn't yet benchmarked for performance. One issue I have is how to package up the `TransitionType`s into something that has name & value information together, without too much overhead. At the moment, each step bundles up all the names from `vi` and all the `vi.vals`, but I think there's probably a better design.

More to come on this.

PR #826, opened 2019-06-27
Fix #665, #802 and #829

### LINKED ISSUE #665: Completely eliminating `Real` from the model and pre-allocating
In this issue, I discuss a performance improvement that can be made after #660 goes in. Consider the following model:
```julia
@model vdemo() = begin
    x = Vector{Real}(undef, N)
    for i = 1:N
        x[i] ~ Normal(0, sqrt(4))
    end
end
```
There are 2 things in this model that can be improved:
1. Every time this model is called, the vector `x` is allocated. This is unnecessary in the case where the length of the vector does not change, and the memory is not modified elsewhere.
2. The `eltype` of `x` is not a concrete type.

As I explained in #660, in that PR, the need for `Real` inside `TypedVarInfo` was eliminated by creating a new `TypedVarInfo` of the correct autodiff `eltypes` before passing it to the model. However, #660 doesn't do anything to the `Real` in the model definition above which is passed by the user. This means that a call to the model is still not fully type stable which limits the speedup achievable.

My proposal for solving the 2 problems above is to have a new macro `@init`, such that the model definition can optionally be:
```julia
@model vdemo() = begin
    @init x = Vector{Real}(undef, N)
    for i = 1:N
        x[i] ~ Normal(0, sqrt(4))
    end
end
```

If the user does the above, what this signals to us is that:
1. The LHS is a parameter.
2. The length of `x` is constant for all the runs.
3. This line only needs to run in the initial model call with `UntypedVarInfo` input.
4. In the later runs, we can simply define `x` as `x = vi.vis.x.vals`, which will be a vector of the correct and concrete `eltype`.

The above lets us:
1. Allocate the vector only once when `UntypedVarInfo` is passed, and
2. Piggyback on the `TypedVarInfo` workaround to eliminate the use of `Real` inside the model entirely.

This can all be done at no run-time overhead since we can branch on the type of the input `vi`, which will be compiled away. Notice that this will be an optional feature, so it will be a non-breaking change, just a pure optimization feature. Feedback is appreciated.

### LINKED ISSUE #802: Make alg.space type info of the alg
This is critical for the type stability of AD with `TypedVarInfo`. Essentially, `space` should be another type parameter as opposed to a field.

### LINKED ISSUE #829: Error with tracker based gradient for `MvNormal`
A gradient error is met when running the following model on branch [`mt/665_fix`](https://github.com/TuringLang/Turing.jl/tree/mt/665_fix). This seems like a `VarInfo` related issue since the same error is met for both univariate `Normal` and multivariate `MvNormal` distributions. 

```julia
using Distributions, LinearAlgebra
using Turing

@model mdemo(d, N) = begin
    Θ = Vector(undef, N)
   for n=1:N
      Θ[n] ~ d
   end
end

# Example 1

dim2 = 250
A   = rand(Wishart(dim2, Matrix{Float64}(I, dim2, dim2)))
d   = MvNormal(zeros(dim2), A)

Turing.setadbackend(:reverse_diff)
chain = sample(mdemo(d, 1), HMC(5000, 0.1, 5)) 

# Example 2
dim2 = 250
d = Normal(0, 10*rand())

Turing.setadbackend(:reverse_diff)
chain = sample(mdemo(d, dim2), HMC(5000, 0.1, 5)) 

```

Stack trace:

```julia

julia> chain = sample(mdemo(d, 1), HMC(5000, 0.1, 5)) 
ERROR: MethodError: no method matching Float64(::Tracker.TrackedReal{Float64})
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:194
  Float64(::T<:Number) where T<:Number at boot.jl:741
  Float64(::Int8) at float.jl:60
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::Tracker.TrackedReal{Float64}) at ./number.jl:7
 [2] setindex!(::Array{Float64,1}, ::Tracker.TrackedReal{Float64}, ::Int64) at ./array.jl:767
 [3] setindex! at ./subarray.jl:269 [inlined]
 [4] _setindex! at ./abstractarray.jl:1050 [inlined]
 [5] setindex! at ./abstractarray.jl:1020 [inlined]
 [6] copyto! at ./multidimensional.jl:877 [inlined]
 [7] copyto! at ./broadcast.jl:838 [inlined]
 [8] copyto! at ./broadcast.jl:797 [inlined]
 [9] materialize! at ./broadcast.jl:756 [inlined]
 [10] _setindex! at /Users/hg344/.julia/dev/Turing/src/core/RandomVariables.jl:1051 [inlined]
 [11] _setindex! at /Users/hg344/.julia/dev/Turing/src/core/RandomVariables.jl:1043 [inlined]
 [12] setindex!(::Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64}, ::TrackedArray{…,Array{Float64,1}}, ::Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}) at /Users/hg344/.julia/dev/Turing/src/core/RandomVariables.jl:1038
 [13] (::getfield(Turing.Core, Symbol("#f#34")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}})(::TrackedArray{…,Array{Float64,1}}) at /Users/hg344/.julia/dev/Turing/src/core/ad.jl:157
 [14] #20 at /Users/hg344/.julia/packages/Tracker/RRYy6/src/back.jl:148 [inlined]
 [15] forward(::getfield(Tracker, Symbol("##20#22")){getfield(Turing.Core, Symbol("#f#34")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}}}, ::Tracker.Params) at /Users/hg344/.julia/packages/Tracker/RRYy6/src/back.jl:135
 [16] forward(::Function, ::Array{Float64,1}) at /Users/hg344/.julia/packages/Tracker/RRYy6/src/back.jl:148
 [17] gradient_logp_reverse(::Array{Float64,1}, ::Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64}, ::Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}, ::Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}) at /Users/hg344/.julia/dev/Turing/src/core/ad.jl:163
 [18] gradient_logp at /Users/hg344/.julia/dev/Turing/src/core/ad.jl:82 [inlined]
 [19] (::getfield(Turing.Inference, Symbol("#∂logπ∂θ#29")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}})(::Array{Float64,1}) at /Users/hg344/.julia/dev/Turing/src/inference/hmc.jl:500
 [20] ∂H∂θ at /Users/hg344/.julia/dev/AdvancedHMC/src/hamiltonian.jl:13 [inlined]
 [21] phasepoint(::AdvancedHMC.Hamiltonian{AdvancedHMC.Adaptation.UnitEuclideanMetric,getfield(Turing.Inference, Symbol("#logπ#30")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}},getfield(Turing.Inference, Symbol("#∂logπ∂θ#29")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}}}, ::Array{Float64,1}, ::Array{Float64,1}) at /Users/hg344/.julia/dev/AdvancedHMC/src/hamiltonian.jl:47
 [22] #sample#10(::Bool, ::Bool, ::Function, ::Random.MersenneTwister, ::AdvancedHMC.Hamiltonian{AdvancedHMC.Adaptation.UnitEuclideanMetric,getfield(Turing.Inference, Symbol("#logπ#30")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}},getfield(Turing.Inference, Symbol("#∂logπ∂θ#29")){Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64},Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}},Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}}}, ::AdvancedHMC.StaticTrajectory{AdvancedHMC.Leapfrog{Float64}}, ::Array{Float64,1}, ::Int64, ::AdvancedHMC.Adaptation.NoAdaptation, ::Int64) at /Users/hg344/.julia/dev/AdvancedHMC/src/sampler.jl:34
 [23] #sample at ./int.jl:0 [inlined]
 [24] #sample at ./none:0 [inlined] (repeats 2 times)
 [25] #steps!#27(::Random.MersenneTwister, ::Bool, ::Function, ::Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}, ::Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}, ::Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64}, ::Array{Turing.Utilities.Sample,1}) at /Users/hg344/.julia/dev/Turing/src/inference/hmc.jl:447
 [26] (::getfield(Turing.Inference, Symbol("#kw##steps!")))(::NamedTuple{(:rng, :verbose),Tuple{Random.MersenneTwister,Bool}}, ::typeof(Turing.Inference.steps!), ::Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}, ::Turing.Sampler{HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}}, ::Turing.Core.RandomVariables.VarInfo{NamedTuple{(:Θ,),Tuple{Turing.Core.RandomVariables.Metadata{Dict{Turing.Core.RandomVariables.VarName{:Θ},Int64},Array{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},1},Array{Turing.Core.RandomVariables.VarName{:Θ},1},Array{Float64,1},Array{Set{Turing.Selector},1}}}},Float64}, ::Array{Turing.Utilities.Sample,1}) at ./none:0
 [27] #sample#19(::Bool, ::Nothing, ::Int64, ::Nothing, ::Nothing, ::Random.MersenneTwister, ::Bool, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(sample), ::Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}, ::HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}) at /Users/hg344/.julia/dev/Turing/src/inference/hmc.jl:280
 [28] sample(::Turing.Model{Tuple{:Θ},Tuple{:d,:N},getfield(Main, Symbol("###inner_function#529#41")),NamedTuple{(:d, :N),Tuple{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}},Int64}},NamedTuple{(:d, :N),Tuple{Symbol,Symbol}}}, ::HMC{Turing.Core.TrackerAD,(),AdvancedHMC.Adaptation.UnitEuclideanMetric}) at /Users/hg344/.julia/dev/Turing/src/inference/hmc.jl:229
 [29] top-level scope at none:0
```

mohamed82008:
This fixes #665 by enabling the specialization of type parameters for different sampler types. The following are still pending:

- [x] Make `eltype(vi[spl])` lazy.
- [x] Verify the type stability
- [x] Make tests
- [x] Update docstrings
- [x] Fix #802
- [x] Fix #829
- [x] Make the `vals` vectors `TrackedArray`s for reversediff AD
- [x] Fix perf of `TuringMvNormal` for `ForwardDiff.Dual`

The following works as expected:
```julia
using Turing

@model demo(y, N, K, ::Type{TV1}=Vector{Float64}, ::Type{TV2}=Vector{Vector{Float64}}) where {TV1, TV2} = begin
    m = TV1(undef, K)
    T = TV2(undef, K)
    for i = 1:K
        T[i] ~ Dirichlet(ones(K)/K)
        m[i] ~ Normal(i, 0.01)
    end
    for i = 2:N
        y[i] ~ Normal(m[i], 0.1)
    end
    return m
end
N = 3; K = 3
y = rand(N)
model = demo(y, 3, 3)
alg = HMC(1000, 0.1, 5)
sample(model, alg)
alg = SMC(50)
sample(model, alg)
```

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -89,6 +89,9 @@ Robust initialization method for model parameters in Hamiltonian samplers.
 struct SampleFromUniform <: AbstractSampler end
 struct SampleFromPrior <: AbstractSampler end
 
+getspace(::SampleFromPrior) = ()
+getspace(::SampleFromUniform) = ()
+
 """
     Sampler{T}
 
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -72,7 +72,7 @@ end
 end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
-getspace(alg) = Tuple(alg.space)
+function getspace end
 
 struct Selector
     gid :: UInt64
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -483,18 +483,18 @@ function _getidcs(vi::AbstractVarInfo, spl::Sampler)
         spl.info[:idcs]
     else
         spl.info[:cache_updated] = spl.info[:cache_updated] | CACHEIDCS
-        spl.info[:idcs] = _getidcs(vi, spl.selector, spl.alg.space)
+        spl.info[:idcs] = _getidcs(vi, spl.selector, getspace(spl.alg))
     end
 end
-function _getidcs(vi::UntypedVarInfo, s::Selector, space)
+function _getidcs(vi::UntypedVarInfo, s::Selector, space::Tuple)
     filter(i -> (s in vi.gids[i] || isempty(vi.gids[i])) &&
         (isempty(space) || in(vi.vns[i], space)), 1:length(vi.gids))
 end
-function _getidcs(vi::TypedVarInfo, s::Selector, space)
+function _getidcs(vi::TypedVarInfo, s::Selector, space::Tuple)
     return __getidcs(vi.metadata, s, space)
 end
 # Get a NamedTuple for all the indices belonging to a given selector for each symbol
-@inline function __getidcs(metadata::NamedTuple{names}, s::Selector, space) where {names}
+@inline function __getidcs(metadata::NamedTuple{names}, s::Selector, space::Tuple) where {names}
     # Check if `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
     # Take the first sybmol
@@ -532,23 +532,26 @@ end
 end
 
 # Get the index (in vals) ranges of all the vns of variables belonging to spl
-function _getranges(vi::AbstractVarInfo, spl::Sampler)
-    if ~haskey(spl.info, :cache_updated) spl.info[:cache_updated] = CACHERESET end
-    if haskey(spl.info, :ranges) && (spl.info[:cache_updated] & CACHERANGES) > 0
-        spl.info[:ranges]
-    else
-        spl.info[:cache_updated] = spl.info[:cache_updated] | CACHERANGES
-        spl.info[:ranges] = _getranges(vi, spl.selector, spl.alg.space)
-    end
+@inline function _getranges(vi::AbstractVarInfo, spl::Sampler)
+    ## Uncomment the spl.info stuff when it is concretely typed, not Dict{Symbol, Any}
+    #if ~haskey(spl.info, :cache_updated) spl.info[:cache_updated] = CACHERESET end
+    #if haskey(spl.info, :ranges) && (spl.info[:cache_updated] & CACHERANGES) > 0
+    #    spl.info[:ranges]
+    #else
+        #spl.info[:cache_updated] = spl.info[:cache_updated] | CACHERANGES
+        ranges = _getranges(vi, spl.selector, getspace(spl.alg))
+        #spl.info[:ranges] = ranges
+        return ranges
+    #end
 end
 # Get the index (in vals) ranges of all the vns of variables belonging to selector `s` in `space`
-function _getranges(vi::AbstractVarInfo, s::Selector, space::Set=Set())
+@inline function _getranges(vi::AbstractVarInfo, s::Selector, space::Tuple=())
     __getranges(vi, _getidcs(vi, s, space))
 end
-function __getranges(vi::UntypedVarInfo, idcs)
+@inline function __getranges(vi::UntypedVarInfo, idcs)
     union(map(i -> vi.ranges[i], idcs)...)
 end
-__getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
+@inline __getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
 @inline function __getranges(metadata::NamedTuple{names}, idcs) where {names}
     # Check if `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
@@ -656,20 +659,25 @@ Symbol(vn::VarName) = Symbol(string(vn, all=false))  # simplified symbol
 
 Returns `true` if `vn`'s symbol is in `space` and `false` otherwise.
 """
-function in(vn::VarName, space::Set)::Bool
+function in(vn::VarName, space::Tuple)::Bool
     if vn.sym in space
         return true
     else
-        # Collect expressions from space
-        exprs = filter(el -> isa(el, Expr), space)
-        # Filter `(` and `)` out and get a string representation of `exprs`
-        expr_strs = Set((replace(string(ex), r"\(|\)" => "") for ex in exprs))
         # String representation of `vn`
         vn_str = string(vn, all=false)
-        # Check if `vn_str` is in `expr_strs`
-        valid = filter(str -> occursin(str, vn_str), expr_strs)
-        return length(valid) > 0
+        return _in(vn_str, space)
     end
+end
+@inline function _in(vn_str::String, space::Tuple)
+    length(space) === 0 && return false
+    el = space[1]
+    # Collect expressions from space
+    expr = isa(el, Expr) ? el : return _in(vn_str, Base.tail(space))
+    # Filter `(` and `)` out and get a string representation of `exprs`
+    expr_str = replace(string(expr), r"\(|\)" => "")
+    # Check if `vn_str` is in `expr_strs`
+    valid = occursin(expr_str, vn_str)
+    return valid || _in(vn_str, Base.tail(space))
 end
 
 # VarInfo
@@ -1266,7 +1274,7 @@ If `vn` doesn't have a sampler selector linked and `vn`'s symbol is in the space
 `spl`, this function will set `vn`'s `gid` to `Set([spl.selector])`.
 """
 function updategid!(vi::AbstractVarInfo, vn::VarName, spl::Sampler)
-    if ~isempty(spl.alg.space) && isempty(getgid(vi, vn)) && getsym(vn) in spl.alg.space
+    if ~isempty(getspace(spl.alg)) && isempty(getgid(vi, vn)) && getsym(vn) in getspace(spl.alg)
         setgid!(vi, spl.selector, vn)
     end
 end
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -330,7 +330,7 @@ end
 Returns all the indices of `vns` in `vi.metadata.vals`.
 """
 function getranges(vi::AbstractVarInfo, vns::Vector{<:VarName})
-    return union(map(vn -> getrange(vi, vn), vns)...)
+    return mapreduce(vn -> getrange(vi, vn), vcat, vns, init=Int[])
 end
 
 """
@@ -549,7 +549,7 @@ end
     __getranges(vi, _getidcs(vi, s, space))
 end
 @inline function __getranges(vi::UntypedVarInfo, idcs)
-    union(map(i -> vi.ranges[i], idcs)...)
+    mapreduce(i -> vi.ranges[i], vcat, idcs, init=Int[])
 end
 @inline __getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
 @inline function __getranges(metadata::NamedTuple{names}, idcs) where {names}
@@ -558,7 +558,7 @@ end
     # Take the first symbol
     f = names[1]
     # Collect the index ranges of all the vns with symbol `f`
-    v = union(map(i -> getfield(metadata, f).ranges[i], getfield(idcs, f))..., Int[])
+    v = mapreduce(i -> getfield(metadata, f).ranges[i], vcat, getfield(idcs, f), init=Int[])
     # Make a single-pair NamedTuple to merge with the result of the recursion
     nt = NamedTuple{(f,)}((v,))
     # Recurse using the remaining of `metadata`
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -192,7 +192,7 @@ function VarInfo(old_vi::TypedVarInfo, spl, x::AbstractVector)
     md = newmetadata(old_vi.metadata, spl, x, 0)
     VarInfo(md, Base.RefValue{eltype(x)}(old_vi.logp), Ref(old_vi.num_produce))
 end
-function newmetadata(metadata::NamedTuple{names}, spl, x, offset) where {names}
+@inline function newmetadata(metadata::NamedTuple{names}, spl, x, offset) where {names}
     # Check if the named tuple is empty and end the recursion
     length(names) === 0 && return ()
     # Take the first key in the NamedTuple
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -1046,6 +1046,10 @@ end
     return _setindex!(_tail(metadata), val, ranges, start)
 end
 
+function Base.eltype(vi::AbstractVarInfo, spl::Union{AbstractSampler, SampleFromPrior})
+    return eltype(Core.Compiler.return_type(getindex, Tuple{typeof(vi), typeof(spl)}))
+end
+
 """
 `haskey(vi::VarInfo, vn::VarName)`
 
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -183,8 +183,11 @@ function VarInfo(model::Model)
     return TypedVarInfo(vi)
 end
 
-VarInfo(old_vi::UntypedVarInfo, spl, x::AbstractVector) = old_vi
-
+function VarInfo(old_vi::UntypedVarInfo, spl, x::AbstractVector)
+    new_vi = deepcopy(old_vi)
+    new_vi[spl] = x 
+    return new_vi
+end
 function VarInfo(old_vi::TypedVarInfo, spl, x::AbstractVector)
     md = newmetadata(old_vi.metadata, spl, x, 0)
     VarInfo(md, Base.RefValue{eltype(x)}(old_vi.logp), Ref(old_vi.num_produce))
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -454,9 +454,9 @@ function _getidcs(vi::UntypedVarInfo, ::SampleFromPrior)
 end
 # Get a NamedTuple of all the indices belonging to SampleFromPrior, one for each symbol
 function _getidcs(vi::TypedVarInfo, ::SampleFromPrior)
-    return __getidcs(vi.metadata)
+    return _getidcs(vi.metadata)
 end
-@inline function __getidcs(metadata::NamedTuple{names}) where {names}
+@inline function _getidcs(metadata::NamedTuple{names}) where {names}
     # Check if the `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
     # Take the first key/symbol
@@ -468,7 +468,7 @@ end
     # Make a single-pair NamedTuple to merge with the result of the recursion
     nt = NamedTuple{(f,)}((v,))
     # Recurse using the remaining of metadata
-    return merge(nt, __getidcs(_tail(metadata)))
+    return merge(nt, _getidcs(_tail(metadata)))
 end
 
 # Get all indices of variables belonging to a given sampler
@@ -491,10 +491,10 @@ function _getidcs(vi::UntypedVarInfo, s::Selector, space::Tuple)
         (isempty(space) || in(vi.vns[i], space)), 1:length(vi.gids))
 end
 function _getidcs(vi::TypedVarInfo, s::Selector, space::Tuple)
-    return __getidcs(vi.metadata, s, space)
+    return _getidcs(vi.metadata, s, space)
 end
 # Get a NamedTuple for all the indices belonging to a given selector for each symbol
-@inline function __getidcs(metadata::NamedTuple{names}, s::Selector, space::Tuple) where {names}
+@inline function _getidcs(metadata::NamedTuple{names}, s::Selector, space::Tuple) where {names}
     # Check if `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
     # Take the first sybmol
@@ -507,7 +507,7 @@ end
     # Make a single-pair NamedTuple to merge with the result of the recursion
     nt = NamedTuple{(f,)}((v,))
     # Recurse using the remaining of metadata
-    return merge(nt, __getidcs(_tail(metadata), s, space))
+    return merge(nt, _getidcs(_tail(metadata), s, space))
 end
 
 # Get all vns of variables belonging to spl
@@ -515,10 +515,10 @@ _getvns(vi::UntypedVarInfo, spl::AbstractSampler) = view(vi.vns, _getidcs(vi, sp
 function _getvns(vi::TypedVarInfo, spl::AbstractSampler)
     # Get a NamedTuple of the indices of variables belonging to `spl`, one entry for each symbol
     idcs = _getidcs(vi, spl)
-    return __getvns(vi.metadata, idcs)
+    return _getvns(vi.metadata, idcs)
 end
 # Get a NamedTuple for all the `vns` of indices `idcs`, one entry for each symbol
-@inline function __getvns(metadata::NamedTuple{names}, idcs) where {names}
+@inline function _getvns(metadata::NamedTuple{names}, idcs) where {names}
     # Check if `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
     # Take the first symbol
@@ -528,7 +528,7 @@ end
     # Make a single-pair NamedTuple to merge with the result of the recursion
     nt = NamedTuple{(f,)}((v,))
     # Recurse using the remaining of `metadata`
-    return merge(nt, __getvns(_tail(metadata), idcs))
+    return merge(nt, _getvns(_tail(metadata), idcs))
 end
 
 # Get the index (in vals) ranges of all the vns of variables belonging to spl
@@ -546,13 +546,13 @@ end
 end
 # Get the index (in vals) ranges of all the vns of variables belonging to selector `s` in `space`
 @inline function _getranges(vi::AbstractVarInfo, s::Selector, space::Tuple=())
-    __getranges(vi, _getidcs(vi, s, space))
+    _getranges(vi, _getidcs(vi, s, space))
 end
-@inline function __getranges(vi::UntypedVarInfo, idcs)
+@inline function _getranges(vi::UntypedVarInfo, idcs::Vector{Int})
     mapreduce(i -> vi.ranges[i], vcat, idcs, init=Int[])
 end
-@inline __getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
-@inline function __getranges(metadata::NamedTuple{names}, idcs) where {names}
+@inline _getranges(vi::TypedVarInfo, idcs::NamedTuple) = _getranges(vi.metadata, idcs)
+@inline function _getranges(metadata::NamedTuple{names}, idcs::NamedTuple) where {names}
     # Check if `metadata` is empty to end the recursion
     length(names) === 0 && return NamedTuple()
     # Take the first symbol
@@ -562,7 +562,7 @@ end
     # Make a single-pair NamedTuple to merge with the result of the recursion
     nt = NamedTuple{(f,)}((v,))
     # Recurse using the remaining of `metadata`
-    return merge(nt, __getranges(_tail(metadata), idcs))
+    return merge(nt, _getranges(_tail(metadata), idcs))
 end
 
 """
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -183,13 +183,13 @@ function VarInfo(model::Model)
     return TypedVarInfo(vi)
 end
 
-VarInfo(old_vi::UntypedVarInfo, spl, T) = old_vi
+VarInfo(old_vi::UntypedVarInfo, spl, x::AbstractVector) = old_vi
 
-function VarInfo(old_vi::TypedVarInfo, spl, T)
-    md = newmetadata(old_vi.metadata, spl, T)
-    VarInfo(md, Base.RefValue{T}(old_vi.logp), Ref(old_vi.num_produce))
+function VarInfo(old_vi::TypedVarInfo, spl, x::AbstractVector)
+    md = newmetadata(old_vi.metadata, spl, x, 0)
+    VarInfo(md, Base.RefValue{eltype(x)}(old_vi.logp), Ref(old_vi.num_produce))
 end
-function newmetadata(metadata::NamedTuple{names}, spl, ::Type{T}) where {T, names}
+function newmetadata(metadata::NamedTuple{names}, spl, x, offset) where {names}
     # Check if the named tuple is empty and end the recursion
     length(names) === 0 && return ()
     # Take the first key in the NamedTuple
@@ -200,18 +200,18 @@ function newmetadata(metadata::NamedTuple{names}, spl, ::Type{T}) where {T, name
         idcs = mdf.idcs
         vns = mdf.vns
         ranges = mdf.ranges
-        vals = similar(mdf.vals, T)
+        vals = x[(offset+1):(offset+length(mdf.vals))]
         dists = mdf.dists
         gids = mdf.gids
         orders = mdf.orders
         flags = mdf.flags
         md = Metadata(idcs, vns, ranges, vals, dists, gids, orders, flags)
 	    nt = NamedTuple{(f,)}((md,))
-        return merge(nt, newmetadata(_tail(metadata), spl, T))
+        return merge(nt, newmetadata(_tail(metadata), spl, x, offset+length(mdf.vals)))
     else
         md = getfield(metadata, f)
 	    nt = NamedTuple{(f,)}((md,))
-        return merge(nt, newmetadata(_tail(metadata), spl, T))
+        return merge(nt, newmetadata(_tail(metadata), spl, x, offset))
     end
 end
 
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -103,8 +103,7 @@ function gradient_logp_forward(
     # Define function to compute log joint.
     logp_old = vi.logp
     function f(θ)
-        new_vi = VarInfo(vi, sampler, eltype(θ))
-        new_vi[sampler] = θ
+        new_vi = VarInfo(vi, sampler, θ)
         logp = runmodel!(model, new_vi, sampler).logp
         vi.logp = ForwardDiff.value(logp)
         return logp
@@ -142,8 +141,7 @@ function gradient_logp_reverse(
 
     # Specify objective function.
     function f(θ)
-        new_vi = VarInfo(vi, sampler, eltype(θ))
-        new_vi[sampler] = θ
+        new_vi = VarInfo(vi, sampler, θ)
         logp = runmodel!(model, new_vi, sampler).logp
         vi.logp = Tracker.data(logp)
         return logp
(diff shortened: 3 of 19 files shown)


mohamed82008:
This PR's speedup will once again be inhibited by #802 which I can fix after #793 is merged to avoid confusion. If @cpfiffer doesn't mind, I can also do it before.

mohamed82008:
For the speedup of this PR to kick in, `vi[spl]` needs to be type stable, which means that from the type of `spl`, I should be able to figure out the type of its values in `vi`. For this, I need the symbols comprising the space of `spl` to be part of its type to look at the corresponding fields in `vi.metadata` and find what their `eltype(vi.metadata.var.vals)` is. If this is concretely typed, then we get the speedup.

cpfiffer:
You can go ahead and add the type parameter. To my knowledge, I think that won't affect much on my side in a real way.

mohamed82008:
This code is now fully type stable!
```julia
using Turing
@model demo(::Type{TV1}=Vector{Float64}) where {TV1} = begin
   m = TV1(undef, 3)
   for i = 1:3
       m[i] ~ Normal(2.0, 0.1)
   end
   return m
end
model = demo()
alg = HMC(1000, 0.1, 5)
spl = Turing.Sampler(alg)
vi = Turing.VarInfo(model)
@code_warntype model.f(vi, spl, model)
```

mohamed82008:
As well as:
```julia
using ForwardDiff
function f(θ)
    new_vi = Turing.VarInfo(vi, spl, eltype(θ))
    new_vi[spl] = θ
    logp = Turing.runmodel!(model, new_vi, spl).logp
    @code_warntype model.f(new_vi, spl, model)
    vi[spl] = ForwardDiff.value.(θ)
    vi.logp = ForwardDiff.value(logp)
    return logp
end
θ = rand(3);
∂l∂θ = ForwardDiff.gradient(f, θ);
```

mohamed82008:
I think this is ready for a review.

yebai:
Thanks, @mohamed82008. I'll take a look at this PR soon.

src/core/compiler.jl
```diff
@@ -191,13 +192,21 @@ function model_generator(x = nothing, y = nothing)
     function inner_function(vi::Turing.VarInfo, sampler::Turing.AbstractSampler, model)
         local x
         if isdefined(model.data, :x)
```
xukai92: I probably miss something in previous PRs but why did we use `:x` and `:y` specifically here?

src/core/compiler.jl
```diff
@@ -228,14 +237,38 @@ function build_model_info(input_expr)
     warn_empty(modeldef[:body])
     # Construct model_info dictionary
 
-    arg_syms = [(arg isa Symbol) ? arg : arg.args[1] for arg in modeldef[:args]]
+    arg_syms = map(modeldef[:args]) do arg
```
xukai92: Can you give me a summary here of what the following changes in doing?

src/core/compiler.jl
```diff
@@ -121,10 +120,10 @@ function _tilde(vsym, left, dist, model_info)
     main_body_names = model_info[:main_body_names]
     model_name = main_body_names[:model]
 
+    model_info[:tent_dvars_list] = copy(model_info[:arg_syms])
```
xukai92: What does `tent_dvars_list` short for?

xukai92:
Correct me if I'm wrong. The solution here is to move the `space` as part of the type of sampler algs, thus we can infer them during the unpacking of `@model`.

src/Turing.jl
```diff
@@ -72,7 +72,7 @@ end
 end
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
 function runmodel! end
-getspace(alg) = Tuple(alg.space)
+function getspace end
```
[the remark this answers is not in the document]
mohamed82008: It just declares the function in `Turing`'s namespace since it is used in `Core` and `Inference`. I can also move it to `Core`.

src/core/RandomVariables.jl
```diff
@@ -532,23 +532,26 @@ end
...
-function _getranges(vi::AbstractVarInfo, spl::Sampler)
-    if ~haskey(spl.info, :cache_updated) spl.info[:cache_updated] = CACHERESET end
-    if haskey(spl.info, :ranges) && (spl.info[:cache_updated] & CACHERANGES) > 0
-        spl.info[:ranges]
-    else
-        spl.info[:cache_updated] = spl.info[:cache_updated] | CACHERANGES
-        spl.info[:ranges] = _getranges(vi, spl.selector, spl.alg.space)
-    end
+@inline function _getranges(vi::AbstractVarInfo, spl::Sampler)
+    ## Uncomment the spl.info stuff when it is concretely typed, not Dict{Symbol, Any}
```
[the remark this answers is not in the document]
mohamed82008: I commented out the `spl.info` lines as they were messing up the type stability of `_getranges` and then `vi[spl]`. `spl.info` is `Dict{Symbol, Any}` so the compiler doesn't know what type `spl.info[:ranges]` will be. When `spl.info` becomes a normal mutable struct with concretely defined fields, these lines can be uncommented for additional performance.

src/core/compiler.jl
mohamed82008: It is an example with `x` and `y` as data variables defined with the `@model` macro.

src/core/compiler.jl
mohamed82008: I am extracting all the argument symbols. When the argument is `demo(x)` (branch 1), `demo(::Type{T}) where {T}` (branch 2) or `demo(x = 1)` (branch 3).

src/core/compiler.jl
mohamed82008: Tentative data variables list. It is basically the arguments unless they are passed as `missing` later when constructing the models. The `findfirst` lines are useless, need to be removed.

mohamed82008:
> Correct me if I'm wrong. The solution here is to move the space as part of the type of sampler algs, thus we can infer them during the unpacking of @model.

Yes pretty much. The goal is to make `vi[spl]` type stable. That is the exact type of the output should be inferrable by the compiler knowing only the types of `vi` and `spl`. To do this, we need to know which symbols in `vi.metadata::NamedTuple` to access when doing `vi[spl]`. This will dictate the eltype of the output. For example, some variables can have the eltype of their `vals` be `Float64`, others `Dual`, others `Int`, etc. If we don't know which symbols will be accessed from the types of `vi` and `spl`, then the compiler's best guess is the supertype of all eltypes of the symbols' `vals` in `vi.metadata`. Realistically though it just gives up and uses `Any` as the inferred type of `vi[spl]`. This is type instability.

So the change here means that the type is inferrable but it doesn't guarantee the type inferred and that of the output won't be `Vector{Any}` for example. But as long as each sampler `spl` is associated with just one concrete type, e.g. `Float64`, even if across multiple symbols, the eltype of `vi[spl]` will be concrete, e.g. `typeof(vi[spl]) == Vector{Float64}` which is also the inferred type, thanks to this PR. This is total type stability.

src/Turing.jl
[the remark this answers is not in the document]
xukai92: OK. Is this a comment practice? If not maybe either make a comment or move it to `Core`.

src/core/compiler.jl
xukai92: Oh yes sorry for my stupidness.

src/core/compiler.jl
xukai92: Got it. Maybe make this as a comment.

xukai92:
Thanks for the explanation!

src/Turing.jl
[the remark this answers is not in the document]
mohamed82008: I have seen it a couple of times before. It's really a matter of preference.

src/core/RandomVariables.jl
```diff
@@ -532,23 +532,26 @@ end
...
-function _getranges(vi::AbstractVarInfo, s::Selector, space::Set=Set())
+@inline function _getranges(vi::AbstractVarInfo, s::Selector, space::Tuple=())
     __getranges(vi, _getidcs(vi, s, space))
 end
-function __getranges(vi::UntypedVarInfo, idcs)
+@inline function __getranges(vi::UntypedVarInfo, idcs)
     union(map(i -> vi.ranges[i], idcs)...)
 end
-__getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
+@inline __getranges(vi::TypedVarInfo, idcs) = __getranges(vi.metadata, idcs)
```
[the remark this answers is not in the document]
mohamed82008: The `_f` pattern is definitely common. Here are a few cases of its usage in the Julia ecosystem:
1. Julia Base: https://github.com/JuliaLang/julia/blob/95fcde9dda4261ebd9dd2eb694b46fc74e472ef6/base/tuple.jl#L31
2. Tim Holy: https://github.com/timholy/ArrayIteration.jl/blob/master/src/core.jl#L110
3. Andy Ferris: https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/det.jl#L1
4. Kristoffer and Fredrik: https://github.com/KristofferC/JuAFEM.jl/blob/master/src/Dofs/DofHandler.jl#L273
5. Patrick and Asbjorn: https://github.com/JuliaNLSolvers/NLSolversBase.jl/blob/master/src/objective_types/constraints.jl#L192

The `__f` is just me taking that one step further. I am not sure if I have seen it before in other places or not. The idea of `_f` or `__f` is to take away the burden of reasoning about dispatch of `f` when reading the `_f` or `__f` functions or at least it has its own little dispatch rules. `_f` is supposed to be a more "internal" function than `f`, typically local to this one place in the files, and not to be used elsewhere hence no need to "waste" a new nice name on this function. 

Notice for example that `_getranges` and `__getranges` both have a 2 argument method. Trying to guess which method is called can be tricky without knowing the code inside out in order to know the types of the arguments and which method it will dispatch to. A trivially different name can therefore increase readability for foreigners reading the code. That's just my preference though. I can see the confusion if one is not familiar with this pattern. Perhaps, I can merge either `getranges` and `_getranges` or the other 2. Would you prefer that?

src/core/compiler.jl
mohamed82008: Sure thing!

src/inference/Inference.jl
```diff
@@ -85,6 +85,20 @@ include("is.jl")
...
 include("gibbs.jl")
 
+for alg in (:SMC, :PG, :PMMH, :IPMCMC, :MH)
+    @eval getspace(::$alg{space}) where {space} = space
+end
+for alg in (:HMC, :HMCDA, :NUTS, :SGLD, :SGHMC)
+    @eval getspace(::$alg{<:Any, space}) where {space} = space
+end
+
+@inline floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
```
trappmartin: 👍

src/Turing.jl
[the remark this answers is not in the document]
yebai: yes, this is recommended by the Julia documentation https://pkg.julialang.org/docs/julia/THl1k/1.1.1/manual/methods.html#Empty-generic-functions-1

yebai:
Here is a simple benchmarking based on the `vdemo` example. Surprisingly, the runtime numbers are similar for type-stable and type-unstable version of the model.  Maybe this is not a good benchmarking case, any thoughts?

```julia

julia> using Turing

julia> @model vdemo1(N=2, ::Type{T} = Float64) where {T<:Real} = begin
           x = Vector{T}(undef, N)
           for i = 1:N
               x[i] ~ Normal(0, sqrt(4))
           end
       end
vdemo1 (generic function with 3 methods)

julia> @model vdemo2(N=2) = begin
           x = Vector{Real}(undef, N)
           for i = 1:N
               x[i] ~ Normal(0, sqrt(4))
           end
       end
vdemo2 (generic function with 2 methods)

julia> chain1 = sample(vdemo(), HMC(5000, 0.1, 5));
┌ Info: Finished 5000 sampling steps in 2.62845281 (s)
│   typeof(h.metric) = AdvancedHMC.Adaptation.UnitEuclideanMetric
│   typeof(τ) = AdvancedHMC.StaticTrajectory{AdvancedHMC.Leapfrog{Float64}}
│   EBFMI(Hs) = 4186.739585328327
└   mean(αs) = 0.9999258042394696

julia> chain2 = sample(vdemo2(), HMC(5000, 0.1, 5));
┌ Info: Finished 5000 sampling steps in 2.721567027 (s)
│   typeof(h.metric) = AdvancedHMC.Adaptation.UnitEuclideanMetric
│   typeof(τ) = AdvancedHMC.StaticTrajectory{AdvancedHMC.Leapfrog{Float64}}
│   EBFMI(Hs) = 4191.654064315244
└   mean(αs) = 0.9999195294647887
```

src/core/RandomVariables.jl
```diff
@@ -656,21 +659,26 @@ Symbol(vn::VarName) = Symbol(string(vn, all=false))  # simplified symbol
...
-        valid = filter(str -> occursin(str, vn_str), expr_strs)
-        return length(valid) > 0
+        return _in(vn_str, space)
     end
 end
+@inline function _in(vn_str::String, space::Tuple)
+    length(space) === 0 && return false
+    el = space[1]
+    # Collect expressions from space
+    expr = isa(el, Expr) ? el : return _in(vn_str, Base.tail(space))
```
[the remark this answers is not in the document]
mohamed82008: Yes, the `@inline` causes the unwrapping of the recursion so there is no function call overhead.

src/Turing.jl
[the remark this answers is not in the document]
xukai92: Great thanks.

trappmartin:
I tried it with a larger `N` and also get the same performance for those models.

mohamed82008:
What you do with the array makes a lot of difference. Try:
```julia
using Turing, LinearAlgebra, BenchmarkTools

@model demo1(N, ::Type{TV1}=Matrix{Float64}) where {TV1} = begin
   m = TV1(undef, N, N)
   for i = 1:N^2
       m[i] ~ Normal(2.0, 0.1)
   end
   return det(m)
end
@model demo2(N) = begin
   m = Matrix{Real}(undef, N, N)
   for i = 1:N^2
       m[i] ~ Normal(2.0, 0.1)
   end
   return det(m)
end

alg = HMC(1000, 0.1, 5);
spl = Turing.Sampler(alg);

N = 10;
model1 = demo1(N);
vi1 = Turing.VarInfo(model1);
model2 = demo2(N);
vi2 = Turing.VarInfo(model2);

@btime $model1($vi1, $spl);
@btime $model2($vi2, $spl);
```

In your example, you are mostly timing the difference between type stable and type unstable `setindex!` which in the grand scheme of things seems to be negligible.

trappmartin:
OK, so a logistic regression example would actually be more interesting to look at. I'll give it a try!

yebai:
The Bayesian Logistic Regression model used by Hoffman+ (2011) the can be found at

https://github.com/TuringLang/TuringExamples/blob/old-models/old-models/nuts-paper/lr_nuts.jl

src/core/compiler.jl
```diff
@@ -72,15 +72,14 @@ function generate_assume(var::Union{Symbol, Expr}, dist, model_info)
         varname_expr = quote
             $sym, $idcs, $csym = Turing.@VarName $var
             $csym = Symbol($(QuoteNode(model_info[:name])), $csym)
-            $syms = Symbol[$csym, $(QuoteNode(var))]
-            $varname = Turing.VarName($syms, "")
+            $varname = Turing.VarName{$sym}($csym, "")
```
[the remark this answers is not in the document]
mohamed82008: It was necessary for type stability. Making the `syms` vector and then accessing it inside the `VarName(syms, "")` was hiding the fact that `var`, aka `sym`, is known at compile time. With the above, the compiler could figure out the type of `varname` is `VarName{sym}`.

src/core/RandomVariables.jl
```diff
@@ -656,21 +659,26 @@ Symbol(vn::VarName) = Symbol(string(vn, all=false))  # simplified symbol
...
-        expr_strs = Set((replace(string(ex), r"\(|\)" => "") for ex in exprs))
         # String representation of `vn`
         vn_str = string(vn, all=false)
-        # Check if `vn_str` is in `expr_strs`
-        valid = filter(str -> occursin(str, vn_str), expr_strs)
-        return length(valid) > 0
+        return _in(vn_str, space)
     end
 end
+@inline function _in(vn_str::String, space::Tuple)
```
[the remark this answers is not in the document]
mohamed82008: Not sure how. Most of the time this function won't be called anyways. I think it only gets called when `space` is somewhat funny, not just symbols. Not sure who wrote this but I don't fully understand when it is triggered. From our tests, mh.jl file, I believe the following alg uses this part of the code: 
```
gibbs = Gibbs(1000, CSMC(10, 1, :z1, :z2, :z3, :z4), MH(10, (:mu1,GKernel(1)), (:mu2,GKernel(1))))
```
I can come back to this later since it is a niche use case. If you have an improvement in mind, please open an issue.

src/core/RandomVariables.jl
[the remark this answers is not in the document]
mohamed82008: Sure.

trappmartin:
```julia
using Turing
using StatsFuns: logistic

@model lr(x, y) = begin
           N,D = size(x)
           w = Vector{Real}(undef, D)
           for d = 1:D
               w[d] ~ Normal()
           end
           v = x*w
           for n = 1:N
               y[n] ~ Bernoulli(logistic(v[n]))
           end
       end

@model lr2(x, y, ::Type{TV1}=Vector{Float64}) where {TV1} = begin
           N,D = size(x)
           w = TV1(undef, D)
           for d = 1:D
               w[d] ~ Normal()
           end
           v = x*w
           for n = 1:N
               y[n] ~ Bernoulli(logistic(v[n]))
           end
       end

x = cat(randn(100, 3), randn(100,3).+2, dims=1);
y = vcat(zeros(100), ones(100));

sample(lr(x,y), HMC(1000, 0.1, 5));
┌ Info: Finished 1000 sampling steps in 12.033267139 (s)
│   typeof(h.metric) = AdvancedHMC.Adaptation.UnitEuclideanMetric
│   typeof(τ) = AdvancedHMC.StaticTrajectory{AdvancedHMC.Leapfrog{Float64}}
│   EBFMI(Hs) = 444.0631044469953
└   mean(αs) = 0.9429567286615578

sample(lr2(x,y), HMC(1000, 0.1, 5));
┌ Info: Finished 1000 sampling steps in 2.703317447 (s)
│   typeof(h.metric) = AdvancedHMC.Adaptation.UnitEuclideanMetric
│   typeof(τ) = AdvancedHMC.StaticTrajectory{AdvancedHMC.Leapfrog{Float64}}
│   EBFMI(Hs) = 323.28904114998767
└   mean(αs) = 0.9488512525299668

```

This is amazing!

mohamed82008:
Cool!

src/core/ad.jl
```diff
@@ -287,7 +287,12 @@ end
 
 # Tracker's implementation of ldiv isn't good. We'll use Zygote's instead.
 const TrackedVecOrMat = Union{Tracker.TrackedVector, Tracker.TrackedMatrix}
-zygote_ldiv(A::AbstractMatrix, B::AbstractVecOrMat) = A \ B
+function zygote_ldiv(A::AbstractMatrix, B::AbstractVecOrMat)
```
willtebbutt: @mohamed82008 could you please explain why this solves the perf issues?

mohamed82008: This is a side issue that @yebai noticed. `TuringMvNormal` was slower than `MvNormal`, double more so for `ForwardDiff.Dual`. The reason can be found here https://github.com/JuliaStats/PDMats.jl/issues/99. `A \ B` seems to copy the matrix `A` first for some reason not clear to me. This causes a slowdown, even more so when the elements of `A` need to be converted to `Dual` in the copy. This change here avoids that.

mohamed82008: It was slower than `MvNormal` with the latest release of `PDMats` but `MvNormal` itself got slower in `PDMats` master which is why I opened that issue. It seems that `c::Cholesky \ x::Vector` used the in-place version which made it faster than `c.U' \ x`, which made me raise an eyebrow!

willtebbutt: I'm confused as to why `TuringMvNormal` was ever getting called with `ForwardDiff.Dual`s -- `TuringMvNormal` should only be used with `Tracker` when we're doing reverse-mode. Is there a good reason that this should ever be getting called?

mohamed82008: Not really! It's just nice to be fast anyways!

mohamed82008: If somebody decides to use `TuringMvNormal` directly for whatever reason, and then use `forward_diff` as the backend this will be useful. That's what @yebai did. But I know this is not the intended use of `TuringMvNormal`.

willtebbutt: Fair enough :)

mohamed82008:
The type annotation works now.

mohamed82008:
Triangular types also work in the arguments but not in the type constraints:
```julia
using Turing

setchunksize(10)
alg = HMC(10, 0.01, 5)
@model vdemo1(::Type{T}=Float64, ::Type{TV}=Vector{T}) where {T, TV <: AbstractVector} = begin
    x = Vector{T}(undef, 10)
    @show TV
    for i = 1:10
        x[i] ~ Normal(0, sqrt(4))
    end
end
sample(vdemo1(), alg)
```

mohamed82008:
It seems that tests pass. If no more comments, I believe this is ready to merge.

PR #845, opened 2019-07-09
Update MAP docs 

mohamed82008:
Just a minor performance update for MAP.

### DIFF
--- a/docs/src/_docs/advanced.md
+++ b/docs/src/_docs/advanced.md
@@ -143,34 +143,39 @@ using Turing
 
 # Define the simple gdemo model.
 @model gdemo(x, y) = begin
- s ~ InverseGamma(2,3)
- m ~ Normal(0,sqrt(s))
- x ~ Normal(m, sqrt(s))
- y ~ Normal(m, sqrt(s))
- return s, m
+    s ~ InverseGamma(2,3)
+    m ~ Normal(0,sqrt(s))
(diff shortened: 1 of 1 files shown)


docs/src/_docs/advanced.md
```diff
@@ -143,34 +143,39 @@ using Turing
...
+    return s, m
+end
+
+function get_nlogp(model)
+    # Set up the model call, sample from the prior.
+    vi = Turing.VarInfo(model)
+
+    # Define a function to optimize.
+    function nlogp(sm)
+        spl = Turing.SampleFromPrior()
```
yebai: Maybe add an explanation here why this is more efficient.

mohamed82008: It's just that it uses the `TypedVarInfo` instead of the `UntypedVarInfo`. I also made sure that no global variables are wrapped in the example, so that will make evaluating the `nlogp` function faster.

mohamed82008: `Turing.VarInfo()` makes an `UntypedVarInfo`. `Turing.VarInfo(model)` makes a `TypedVarInfo`.

yebai: I see - thanks. It's worth to add this information to the `advanced` section of our docs. But maybe in a separate PR?

mohamed82008: Sure

PR #849, opened 2019-07-10
Get ready for AHMC v0.2.0

xukai92:
AHMC now assumes the grad function to return a val-grad tuple.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -8,9 +8,9 @@ version = "0.4.2"
 
 [[AdvancedHMC]]
 deps = ["ArgCheck", "InplaceOps", "LazyArrays", "LinearAlgebra", "Parameters", "ProgressMeter", "Random", "Statistics", "StatsBase"]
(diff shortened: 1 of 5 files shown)


xukai92:
Please don't merge until AHMC v0.2.0 is there.

xukai92:
Mainly just updated the gradient function that passes to AHMC to return both value and gradient, plus also passing `progress` keyword argument to AHMC.

xukai92:
There are many tests failed but I'm not quite sure which ones are known to fail? (Should we actually mark them as known to fail by `@test_broken`

xukai92:
There seems to be something wrong - one of the `hmc.jl` test fails. I'm looking into it.

yebai:
> There seems to be something wrong - one of the hmc.jl test fails. I'm looking into it.

I think we'd better make a new release of Turing to accommodate AHMC 0.2. I'll take a look at the new AHMC PR today.

PR #851, opened 2019-07-11
Uniform interface of Benchmark

KDr2:
- set all configurations via global var `CONFIG::Dict`
- use `DataFrame` or `Dict` type result

The benchmark CI job will fail due to it is based on the master branch, and the code in that branch is not updated to the new protocol. I think this PR can be merged ignoring that failure.


trappmartin:
Thanks for the PR. I'll have a look and review it asap.

Issue #827, opened 2019-06-30
MvNormal AD and cov symmetry


mohamed82008:
The following code fails on master:
```julia
using ForwardDiff, Tracker, Turing, Test

function test_ad(f, at = 0.5; rtol = 1e-8, atol = 1e-8)
    isarr = isa(at, AbstractArray)
    reverse = Tracker.gradient(f, at)[1]
    forward = ForwardDiff.gradient(f, at)
    @test isapprox(reverse, forward, rtol=rtol, atol=atol)
end

s = rand(2,2); s = s' * s
test_ad(s->logpdf(MvNormal(zeros(2), s), [1.0, 1.0]), s)
```
The reversediff implementation uses the whole matrix, whereas the ForwardDiff implementation uses only the upper triangle. Is there any negative consequence to this?

CC: @willtebbutt

yebai:
Fixed by https://github.com/FluxML/Zygote.jl/pull/268

mohamed82008:
We need a new release in Zygote before this is truly fixed on our end.

PR #852, opened 2019-07-11
Bring back the generated functions

mohamed82008:
This PR solves some overlooked type instability in certain internal functions. Relying on recursion, inlining and constant propagation is apparently too much magic to work reliably. The compiler inferred the types properly in some simple functions but in more complicated cases, it failed. The generated function is a sure-fire approach so here it is.

More interestingly though, here is some speedup as a result!
```julia
using Turing

@model gdemo(x) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))
    for i in eachindex(x)
        x[i] ~ Normal(m, sqrt(s))
    end
end
sample(gdemo(rand(100)), NUTS(10000, 0.65))
# master - 2.06 s
# This PR - 1.43 s
```

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -89,8 +89,9 @@ Robust initialization method for model parameters in Hamiltonian samplers.
 struct SampleFromUniform <: AbstractSampler end
 struct SampleFromPrior <: AbstractSampler end
 
-getspace(::SampleFromPrior) = ()
-getspace(::SampleFromUniform) = ()
+getspace(::Union{SampleFromPrior, SampleFromUniform}) = ()
+getspace(::Type{<:SampleFromPrior}) = ()
+getspace(::Type{<:SampleFromUniform}) = ()
 
 """
     Sampler{T}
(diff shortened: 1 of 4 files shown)


cpfiffer:
After this PR merges over, I should be able to replace all the cases where I have (in the `csp/interface-pt2` branch)

```julia
spl.state.vi[spl]
```
with 

```julia
getparams(spl.state.vi, spl)
```

Is that correct?

mohamed82008:
Correct

cpfiffer:
I'm getting a similar error on `csp/interface-pt2` to the one that is causing this branch to fail Travis.

One of the tests is 

```julia
test_ad(s->logpdf(MvNormal(zeros(2), s), [1.0, 1.0]), s)
```

```julia
AD compatibility: Test Failed at /home/cameron/.julia/dev/Turing/test/test_utils/ad_utils.jl:9
  Expression: isapprox(reverse, forward, rtol=rtol, atol=atol)
   Evaluated: isapprox([80.2763 -55.3825; -55.3825 37.4044] (tracked), [80.2763 -110.765; -0.0 37.4044]; rtol=1.0e-8, atol=1.0e-8)
```

Any ideas on what's causing the difference between forward & reverse?

mohamed82008:
https://github.com/TuringLang/Turing.jl/issues/827

cpfiffer:
I'd like to merge this one over then @yebai, if the only test failure is everywhere. It will be very helpful for #746.

yebai:
Thanks, @mohamed82008!

PR #858, opened 2019-07-22
Fix all tests

mohamed82008:
This PR fixes all the Turing tests except the PMMH and IPMCMC numerical tests. It relies on https://github.com/TuringLang/Bijectors.jl/pull/33. So after the Bijectors PR is merged and a release is made, the Project and Manifest files here should be updated then this PR should pass tests on Travis.

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -368,16 +368,6 @@ getchol(m::PDMats.AbstractPDMat) = m.chol
 getchol(m::PDMats.PDiagMat) = cholesky(Diagonal(m.diag))
 getchol(m::PDMats.ScalMat) = cholesky(Diagonal(fill(m.value, m.dim)))
 
-# Deal with ambiguities.
-function Base.:*(
-    A::Tracker.TrackedMatrix,
-    B::Adjoint{T, V} where V<:LinearAlgebra.AbstractTriangular{T} where {T},
-)
-    return Tracker.track(*, A, B)
-end
-
-
-
 """
     TuringDiagNormal{Tm<:AbstractVector, Tσ<:AbstractVector} <: ContinuousMultivariateDistribution
 
@@ -498,8 +488,6 @@ for F in (:link, :invlink)
         end
     end
 end
-
-Base.:*(x::Adjoint{T, <:AbstractMatrix{T}} where {T}, y::TrackedVector) = Tracker.track(*, x, y)
 
 for F in (:link, :invlink)
     @eval begin
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -242,28 +242,61 @@ end
 
 
 #
-# Make Tracker work with MvNormal. This is a bit nasty.
+# Make Tracker work with MvNormal and unsafe Cholesky. This is quite nasty.
 #
-
-using Zygote
 
 LinearAlgebra.UpperTriangular(A::Tracker.TrackedMatrix) = Tracker.track(UpperTriangular, A)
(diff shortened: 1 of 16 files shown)


yebai:
Thanks, @mohamed82008. Is this ready for a review?

mohamed82008:
Yes this is ready for a first review. The WIP is to reflect that tests fail even though the PR is called "Fix all tests" :) This is because of the Bijectors PR linked. I won't have time to work on either until this weekend though. But the Bijectors PR is simple enough and probably just needs the Tracker benchmark to get it merged.

mohamed82008:
I merged with master locally and updated AHMC and there seems to be new test failures. I will look into it. Tests also seem to be taking excessively long. I think we may have a speed regression somewhere.

mohamed82008:
It could also be related to the long compilation time that Chris reported on Slack.

src/core/ad.jl
```diff
@@ -433,3 +475,53 @@ MvNormal(m::TrackedVector{<:Real}, A::UniformScaling{<:Real}) = MvNormal(m, A.λ
...
+        Tracker.@grad function $F(
+            dist::Dirichlet, 
+            x::Tracker.TrackedArray, 
+            ::Type{Val{proj}}
+        ) where {proj}
+            x_data = Tracker.data(x)
+            T = eltype(x_data)
+            y = $F(dist, x_data, Val{proj})
+            return  y, Δ -> begin
+                out = (ForwardDiff.jacobian(x -> $F(dist, x, Val{proj}), x_data)::Matrix{T})' * Δ
```
yebai: Is this a duplicate fix for https://github.com/TuringLang/Bijectors.jl/pull/28?

yebai:
@mohamed82008 It seems Travis test halts after `AHMC` is compiled, see https://travis-ci.org/TuringLang/Turing.jl/jobs/561920312#L679. There might be something else slow in the test.

mohamed82008:
This PR requires a new release of Tracker and Zygote. Some tests in stdlib/distributions.jl are still taking a very long time, so I need to look into it. I suspect it has to do with Dirichlet.

mohamed82008:
After https://github.com/FluxML/Tracker.jl/issues/43 and https://github.com/FluxML/Zygote.jl/issues/302 are resolved, this PR should pass almost all tests. Currently, one SGLD test is giving an error, and four numerical PMMH and IPMCMC tests are failing on my machine.

yebai:
> After FluxML/Tracker.jl#43 and FluxML/Zygote.jl#302 are resolved, this PR should pass almost all tests. 

Perhaps @willtebbutt can help with making a new release of Tracker and Zygote.

mohamed82008:
Actually, I think the Zygote release is not important, only the Tracker.

mohamed82008:
So the tests are passing now but are taking some time. I can look into the time issue in another PR.

src/core/ad.jl
yebai: @mohamed82008

mohamed82008: No we won't need `ForwardDiff`, we can replace it with the `link_jacobian` or `invlink_jacobian` functions.

yebai: Do you want to remove these functions from this PR, or keep them until TuringLang/Bijectors.jl#28 is merged?

mohamed82008: I would keep them for now as a stop gap. This fixes the problem but can be improved when the Bijectors PR is merged.

mohamed82008: We are depending on ForwardDiff anyways so nothing to lose.

PR #872, opened 2019-07-27
dispatch-based AD, updated manifest

matbesancon:
Very minor improvement replacing branching with dispatch for `Val{S}`


yebai:
Many thanks, @matbesancon!

PR #875, opened 2019-07-31
The docs link is broken

Red-Portal:
https://turing.ml/docs/ leads to a 404 error.
Fixed it to redirect to https://turing.ml/docs/using-turing/  (which is not broken)


yebai:
Thanks, @Red-Portal!

PR #879, opened 2019-08-02
Increase default tree depth of NUTS from 5 to 10

xukai92:


xukai92:
Just so we match the default setting of Stan.

PR #882, opened 2019-08-06
Fix Base.rand Signature

bhgomes:
The default `Base.rand` is `rand(rng::AbstractRNG, ...)` and the custom Turing distributions were not using this. Since all of the custom distributions are `Sampleable`, the Distributions.jl package already provides the necessary defaults for `rand(...)` without the `AbstractRNG` argument. This also covers the vectorized case since `rand(s::Sampleable, n::Integer)` is defined by Distributions.jl.


yebai:
Many thanks, @bhgomes!

PR #894, opened 2019-08-22
README links broken again

Red-Portal:
The `doc` badge and getting-started link are broken again.
Updated links.

Why is the URL changing so frequently?


xukai92:
We are updating the documentation and a version support is recently added.

xukai92:
Thanks for the catch!

PR #896, opened 2019-08-27
added OrderedLogistic distribution

elizavetasemenova:


elizavetasemenova:
I am not sure how tests work. Have added a line to https://github.com/elizavetasemenova/Turing.jl/blob/master/test/stdlib/distributions.jl. How can I learn whether the test is successful?

xukai92:
You should implement a test more like the one for `BinomialLogit` (copied below)

```julia
    @turing_testset "distributions functions" begin
        ns = 10
        logitp = randn()
        d1 = BinomialLogit(ns, logitp)
        d2 = Binomial(ns, logistic(logitp))
        k = 3
        @test logpdf(d1, k) ≈ logpdf(d2, k)
    end
```

Just use the `@test` to compare your implementaion and some analytical results that you know are correct.

elizavetasemenova:
Have added such a test. Does it look better?

test/stdlib/distributions.jl
```diff
@@ -70,6 +84,7 @@ include(dir*"/test/test_utils/AllUtils.jl")
             # test is broken
             Weibull(2, 1),
             # Cauchy(0, 1),  # mean and variance are undefined for Cauchy
+            OrderedLogistic(0.5, [-1, 1])
```
xukai92: We don't need this any more. Please remove it from the list.

xukai92:
Also please add `OrderedLogistic` to the export list in end of `src/Turing.jl` so the people will be able to use it when doing `using Turing`.

test/stdlib/distributions.jl
elizavetasemenova: done

xukai92:
I will merge this in 24 hours if no rejection.

mohamed82008:
I am not familiar with the `OrderedLogistic` distribution but I find it peculiar that the `rand` function is fairly deterministic in this case. Shouldn't we be sampling according to the pmf (https://mc-stan.org/docs/2_18/functions-reference/ordered-logistic-distribution.html)? Also would be nice to have a method for `rand` that takes `rng` as argument.

xukai92:
@mohamed82008 Good point.

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+
+    if !issorted(cutpoints)
+        error("cutpoints are not sorted")
+    end
+
+    K = length(cutpoints)+1
+    c = vcat(-Inf, cutpoints, Inf)
+    l = [i for i in zip(c[1:(end-1)],c[2:end])]
+    ps = [logistic(η - l[i][1]) - logistic(η - l[i][2]) for i in 1:K]
+    k = findall(ps.== maximum(ps))[1]
```
xukai92: Seems that this should be `k = rand(Categorical(ps))`

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+        logp= log(logistic(c[k]-d.η) - logistic(c[k-1]-d.η))
+    else
+        logp= log(1-logistic(c[k-1]-d.η))
+    end
+
+    return(logp)
+end
+
+# rand of the OrderedLogistic distribution
+function Distributions.rand(d::OrderedLogistic)
```
xukai92: `Distributions.rand(d::OrderedLogistic)` -> `Distributions.rand(rng::AbstractRNG, d::OrderedLogistic)`

src/stdlib/distributions.jl
xukai92: And then `k = rand(Categorical(ps))` -> `k = rand(rng, Categorical(ps))`

test/stdlib/distributions.jl
```diff
@@ -15,6 +15,20 @@ include(dir*"/test/test_utils/AllUtils.jl")
         k = 3
         @test logpdf(d1, k) ≈ logpdf(d2, k)
     end
+
+    @turing_testset "distributions functions" begin
```
xukai92: This test seems to be not enough given that the previous bug was not identified. Wonder if there is any ordered logistic implmeneted in R that we can use for test purpose?

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+
+# logpdf of the OrderedLogistic distribution
+function Distributions.logpdf(d::OrderedLogistic, k::Int)
+
+    K = length(d.cutpoints)+1
+
+    c =  d.cutpoints
+
+    if k==1
+        logp= log(logistic(c[k]-d.η))
```
willtebbutt: If I'm not mistaken, these `log(logistic(x))` functions could be replaced with `log1pexp` a.k.a. `softplus` from StatsFuns.jl

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
     return sum(logpdf_binomial_logit.(d.n, d.logitp, ks))
 end
+
+# OrderedLogistic distribution
+struct OrderedLogistic{T1, T2} <: DiscreteUnivariateDistribution
+   η::T1
+   cutpoints::Vector{T2}
+end
+
+# logpdf of the OrderedLogistic distribution
```
willtebbutt: This is probably redundant tbh. The code self-documents nicely.

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+    elseif k<K
+        logp= log(logistic(c[k]-d.η) - logistic(c[k-1]-d.η))
+    else
+        logp= log(1-logistic(c[k-1]-d.η))
+    end
+
+    return(logp)
+end
+
+# rand of the OrderedLogistic distribution
```
willtebbutt: Same here. The code is nicely self-documenting

src/stdlib/distributions.jl
trappmartin: Correct, I did'n see that too.

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+    cutpoints = d.cutpoints
+    η = d.η
+
+    if !issorted(cutpoints)
+        error("cutpoints are not sorted")
+    end
+
+    K = length(cutpoints)+1
+    c = vcat(-Inf, cutpoints, Inf)
+    l = [i for i in zip(c[1:(end-1)],c[2:end])]
```
willtebbutt: Could these two lines
```julia
l = [i for i in zip(c[1:(end-1)],c[2:end])]
ps = [logistic(η - l[i][1]) - logistic(η - l[i][2]) for i in 1:K]
```
be re-written as
```julia
ps = [logistic(η - i[1]) - logistic(η - i[2]) for i in zip(c[1:(end-1)],c[2:end])]
```
?

Doing this would avoid ever creating the intermediate array `l`, which would be nice.

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+
+    if k==1
+        logp= log(logistic(c[k]-d.η))
+    elseif k<K
+        logp= log(logistic(c[k]-d.η) - logistic(c[k-1]-d.η))
+    else
+        logp= log(1-logistic(c[k-1]-d.η))
+    end
+
+    return(logp)
```
willtebbutt: Convention is just to write
```julia
return logp
```

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,49 @@ end
...
+end
+
+# rand of the OrderedLogistic distribution
+function Distributions.rand(d::OrderedLogistic)
+    cutpoints = d.cutpoints
+    η = d.η
+
+    if !issorted(cutpoints)
```
willtebbutt: Could this check be moved to construction-time? Would be nice to avoid having to check this each time that `rand` is called.

test/stdlib/distributions.jl
willtebbutt: The other option is to check the empirical moments against the known moments (if there are nice expressions for these for the `OrderedLogistic` distribution)

elizavetasemenova: Would using R and rstan library be a good solution? I could use Stan's functionality to generate random numbers.

willtebbutt: We're already using `CmdStan.jl` somewhere in our tests, so perhaps give that a go?

xukai92: `CmdStan.jl` is only set up for interface test but the actual command line backend is not installed. 

As this is a discrete distribution, maybe we can validate `rand` using `logpdf`? I.e. draw a lot of r.v.s and check if the emperical distribution is same as what's from `logpdf`.

src/stdlib/distributions.jl
elizavetasemenova: This check is redundant now, when `Categorical` distribution will be used. In case the vector of cutpoints is not sorted, the error message will come from `Categorical`: `ArgumentError: Categorical: the condition isprobvec(p) is not satisfied.`

xukai92: I think we still want the check as the error message from `Categorical` is not informative.
Can we move the check before constructing `Categorical` maybe?

willtebbutt: Would it not be better to make a PR to distributions to fix the error message then? Doesn't seem like that code really belongs here if `Categorical` should be doing the checks for us.

xukai92: I don't think it's the case here. It just a special case here that when `issorted(cutpoints) == false`, `isprobvec(p)` is also false. For us we want to know where the error is from - it's because the cut points are not sorted. And `Categorical` should not be unaware of the reason why the passed-in `p` is not normalised.

src/stdlib/distributions.jl
```diff
@@ -49,3 +49,52 @@ end
 function Distributions.logpdf(d::VecBinomialLogit{<:Real}, ks::Vector{<:Integer})
     return sum(logpdf_binomial_logit.(d.n, d.logitp, ks))
 end
+
+
+struct OrderedLogistic{T1, T2} <: DiscreteUnivariateDistribution
```
willtebbutt: Since we're storing the `cutpoints` regardless their type, might want to consider parametrising the entire type of `cutpoints`. i.e.
```julia
struct OrderedLogistic{T1, T2<:AbstractVector} <: DiscreteUnivariateDistribution
   η::T1
   cutpoints::T2

   function OrderedLogistic(η, cutpoints)
        if !issorted(cutpoints)
            error("cutpoints are not sorted")
        end
        return new{typeof(η), typeof(cutpoints)}(η, cutpoints)
   end

end
```

elizavetasemenova: This gives `invalid redefinition of constant OrderedLogistic`. I tried to make some minor corrections, still getting error messages.

willtebbutt: Hmmm weird, seems to work fine for me. Is this a `Revise.jl` issue? If so, try restarting your session.

xukai92:
Will merge in 24 hours if no rejection.

PR #900, opened 2019-08-30
Improve particle samplers warning message

yebai:
This PR adds a bit of warning information for illy defined posteriors, e.g. 

```julia

  @model normal() = begin
    a ~ Normal(4,5)
    b ~ Normal(a,1)
    n ~ Categorical(3)
    for i=1:n # random number of observations
       n ~ Normal(a,b)
    end
  end

  tested = sample(normal(), SMC(10))

```


src/core/container.jl
```diff
@@ -43,16 +43,23 @@ function Trace(m::Model, spl::T, vi::AbstractVarInfo) where {T <: AbstractSample
     res = Trace{T}(m, spl, deepcopy(vi));
     # CTask(()->f());
     res.vi.num_produce = 0
-    res.task = CTask( () -> begin vi_new=m(vi, spl); produce(Val{:done}); vi_new; end )
-    if isa(res.task.storage, Nothing)
+    res.task = CTask(() -> m(vi, spl))
```
xukai92: Do we need to do the same change for L35?

trappmartin: Could you also briefly explain why this is needed?

src/core/container.jl
```diff
@@ -166,14 +173,16 @@ function Libtask.consume(pc :: ParticleContainer)
         elseif score == Val{:done}
             num_done += 1
         else
-            error("[consume]: error in running particle filter.")
+            error("[consume]: error in running particle filter: likelihood=$(score)")
```
trappmartin: I don't really understand what the return value of consume is. It seems it is either `<: Real`, `Val{:done}` or `nothing`. Is that correct? If so we should think about refactoring this bit at some point as it will likely improve performance.

test/inference/smc.jl
```diff
@@ -0,0 +1,17 @@
...
+@turing_testset "smc.jl" begin
+  @model normal() = begin
+    a ~ Normal(4,5)
+    3 ~ Normal(a,2)
+    b ~ Normal(a,1)
+    1.5 ~ Normal(b,2)
+    a, b
+  end
+
+  tested = sample(normal(), SMC(10))
```
trappmartin: Maybe add a failing test example?

src/core/container.jl
xukai92: In other places (e.g. SMC https://github.com/TuringLang/Turing.jl/blob/master/src/inference/AdvancedSMC.jl#L51), `Val{:done}` is used to check if the running of the particles is finished.

src/core/container.jl
xukai92: It's either `<:Real` (https://github.com/TuringLang/Turing.jl/blob/master/src/inference/AdvancedSMC.jl#L261) or `Val{:done}` (as we dicussed above). I don't think there is `nothing` procuded.

xukai92: Need verfication from @yebai

src/core/container.jl
yebai: Nice catch! 

This is a small clean up after recent updates to `Libtask.jl`. In short, there were some caveats of previous task copying code, which prevents copied tasks exit normally. This is now fixed by `CTask`.

src/core/container.jl
yebai: Kai is correct, it can only be `<:Real` or `Val{:done}`.

test/inference/smc.jl
yebai: The above example (in PR description) will fail, but I'm not sure how to test it. Any suggestions?

trappmartin: Would it work with the `@test_throws` macro?

src/core/container.jl
trappmartin: Thanks for the clarification.

yebai:
Prerequisite: https://github.com/TuringLang/Libtask.jl/issues/41

yebai:
Ready for another review @trappmartin

yebai:
I'm merging now since there is no functionality in this PR and all the tests are passing.

PR #902, opened 2019-09-02
VI & ADVI improvements

torfjelde:
# Overview
This PR aims to improve VI, in particular ADVI, by adapting the new Bijectors.jl interface together with `Stacked<:Bijector` from [this PR](https://github.com/TuringLang/Bijectors.jl/pull/36), introducing a new optimiser `DecayedADAGrad`, and general improvements. See the full list below.

Worth noting that there are things to do to improve this further, but as we will soon separate out the VI parts of Turing into AdvancedVI.jl, I think this is a good-enough impl for Turing and then we can improve further as we make the separation.

# Changes
1. `meanfield(model::Model)` which returns a mean-field approx of the model which in turn can be passed to `vi(model, alg, q)`. This accommodates future plans of allowing "arbitrary" variational posterior q.
2. This branch doesn't perform assignments in the same way we did before, so now we can do `TrackedArray` instead of `Array{TrackedReal}` (with the help of a bit of behavior-piracy from Tracker.jl to make it work with `SubArray`)
3. Added `DecayedADAGrad` which is currently what's in use in Stan and suggested in [2]. `TruncatedADAGrad` is an impl of what they proposed in their paper on ADVI, but looking at the source code it seems they've moved away from this.[1] `DecayedADAGrad` seems to do better for more complex models, e.g. MNIST example it seems like choosing the truncation length for `TruncatedADAGrad` affects the convergence quite a bit (+ it's much more memory-intensive than `DecayedADAGrad`). `TruncatedADAGrad` works very nicely for smaller models though. Now we can use both :)
4. Using `softplus` instead of `exp` when enforcing positiveness of variances of `TuringDiagNormal` which prevents a numerical errors I've been running into with `exp` in certain cases. There was a brief discussion about `exp` vs. `softplus` on our Slack, and the consensus seemed to be that there's no clear winner in general. Nonetheless I think it's better to go for the most numerically stable one given that we've already run into numerical issues with `exp`. Also worth noting that in [Section 3.3, 2] they actually observe better `softplus` end results than with  `exp`.
5. `setadbackend` now also sets the `ADBackend` of Bijectors.jl, which was introduced in the interface PR.
6. `MeanField` has been fully replaced by `TransformedDistribution{<:TuringDiagNormal, <:Stacked}`. This results in several improvements:
    - _Flexibility._ With this change, the only thing part of VI / ADVI that is hard-coded is the handling of the parameters of variational posterior `q`. This we do because we need to enforce the positive-constraint for variance of `TuringDiagNormal`. As soon as we have a way of dealing with constraints on the parameters of the distributions themselves (not the support) in a more generic manner, we'll be able to plug in any `q` with the correct dimensionality and run ADVI using this instead of a `TuringDiagNormal` as a base distribution. Even now, if you have a custom base distribution, e.g. normalizing flow, we can implement `vi(model::Model, alg::ADVI, q::TransformedDistribution{<:NormalizingFlow})` and similarily for `(elbo::ELBO)(...)` and we're good. With current `advi` we can't do VI on LDA because of singular Jacobians when using AD. When [this PR](https://github.com/TuringLang/Bijectors.jl/pull/28) with analytical expressions for jacobian of `SimplexBijector` is merged in Bijectors.jl, we'll able to do ADVI on LDA without any changes to the Turing-side of things.
    - _Type-stability._ Though my testing hasn't been extensive, from what I can tell, everything _but_ `Variational.meanfield(model::Model)` is now type-stable (same as for samplers, etc. in current Turing.jl). The `optimize` call inside `vi(model, advi)` works as a function-barrier, thus everything other than the first line of `vi(model, advi)` is type-stable.
    - _Performance._ All of the above gives us a perf-increase. Moreover, since everything involving the actual transformations, etc. is now outsourced to Bijectors.jl it's much easier to optimize performance in the future + gives us analytical expressions for `logabsdetjac` for all standard distributions. Should also benefit from eventual support for batch computation in Bijectors.jl (see [this issue](https://github.com/TuringLang/Bijectors.jl/issues/35).
    - _Less code._ The entire `src/variational/distributions.jl` is now redundant and so has been removed.
7. Some extra cleanups of redundant code & comments left over from the previous ADVI PR.
8. ~Added `weight` argument to `(elbo::ELBO)(...)`. This can be used to weight the ` 𝔼_q[log p(x, z)]` term of the ELBO. This is necessary for stuff like _stochastic variational inference (SVI)_ (see SVI PR), more complex initialization, e.g. Eq. (20) in [3].~ This `weight` argument is now baked into `make_logjoint` which makes more sense.
9. Now `vi` have the following definitions:
    - `vi(model, alg)`: uses default variational posterior `q` which is `alg` dependent (in case of `ADVI` this is a mean-field approx with unconstrained-to-constrained `Bijector`)
    - `vi(model, alg, q, θ_init)`: allows the user to pass in an arbitrary variational posterior `q` as long as `Variational.update(d::typeof(q), θ::AbstractArray)::typeof(q)` is defined.
    - `vi(model, alg, getq, θ_init)`: don't need to pass in an initial variational posterior  `q` but instead pass a function `getq(θ::AbstractArray)::VariationalPosterior`, i.e. maps from params to distribution.
    - In all of the above, we can also replace `model` with `logπ(z::AbstractArray)::Real` which  represents the map z → log p(x, z), where `x` are the observations, making this applicable even for non-Turing models (will prob. be more useful when we separate out VI into AdvancedVI.jl).

# Examples
### Old
```julia
# Everything is hard-coded for mean-field Gaussian
elbo_acc = 0.0

for i = 1:num_samples
    # iterate through priors, sample and update
    idx = 0
    z = zeros(T, num_params)
    
    for sym ∈ keys(varinfo.metadata)
        md = varinfo.metadata[sym]
        
        for i = 1:size(md.dists, 1)
            prior = md.dists[i]
            r = md.ranges[i] .+ idx

            # mean-field params for this set of model params
            μ_i = μ[r]
            ω_i = ω[r]

            # obtain samples from mean-field posterior approximation
            η = randn(length(μ_i))
            ζ = center_diag_gaussian_inv(η, μ_i, exp.(ω_i))
            
            # inverse-transform back to domain of original priro
            z[r] .= invlink(prior, ζ)

            # add the log-det-jacobian of inverse transform;
            # `logabsdet` returns `(log(abs(det(M))), sign(det(M)))` so return first entry
            # add `eps` to ensure SingularException does not occurr in `logabsdet`
            elbo_acc += logabsdet(jac_inv_transform(prior, ζ) .+ eps(T))[1] / num_samples
        end

        idx += md.ranges[end][end]
    end
    
    # compute log density
    varinfo = VarInfo(varinfo, SampleFromUniform(), z)
    model(varinfo)
    elbo_acc += varinfo.logp / num_samples
end

# add the term for the entropy of the variational posterior
variational_posterior_entropy = sum(ω)
elbo_acc += variational_posterior_entropy

return elbo_acc
```
### New
```julia
function (elbo::ELBO)(
    rng::AbstractRNG,
    alg::ADVI,
    q::VariationalPosterior,
    logπ,  # ← arbitrary callable z → log p(x, z)
    num_samples
)
    _, z, logjac, _ = forward(rng, q)
    res = (logπ(z) + logjac) / num_samples

    res += entropy(q.dist)
    
    for i = 2:num_samples
        _, z, logjac, _ = forward(rng, q)
        res += (logπ(z) + logjac) / num_samples
    end

    return res
end
```
with
```julia
function make_logjoint(model; weight = 1.0)
    # setup
    ctx = DynamicPPL.MiniBatchContext(
        DynamicPPL.DefaultContext(),
        weight
    )
    varinfo = Turing.VarInfo(model, ctx)

    function logπ(z)
        varinfo = VarInfo(varinfo, SampleFromUniform(), z)
        model(varinfo)
        
        return varinfo.logp
    end

    return logπ
end
```
to make this work nicely with `Turing.Model`.

# TODOs
- [x] Wait for `Stacked<:Bijector` PR has been merged so we don't have to depend on github branch.
- [x] Improve docstrings for the following:
    - `TruncatedADAGrad`
    - `DecayedADAGrad`
    - `ELBO` and `(elbo::ELBO)(...)`
- [x] Tests for
    - `DecayedADAGrad`
    - `TruncatedADAGrad`

# References
[1] Kucukelbir, A., Ranganath, R., Gelman, A., & Blei, D. M., Automatic variational inference in stan, CoRR, (),  (2015). 
[2] Kucukelbir, A., Tran, D., Ranganath, R., Gelman, A., & Blei, D. M., Automatic Differentiation Variational Inference, CoRR, (),  (2016). 
[3] Rezende, D. J., & Mohamed, S., Variational Inference With Normalizing Flows, CoRR, (),  (2015).


torfjelde:
Now that `Stacked` is merged into Bijectors.jl, I'll finish up this and then complete the tutorial on ADVI (which depends on this branch).

xukai92:
What's left to do for this PR?

torfjelde:
> What's left to do for this PR?

There's the two remaining points on the TODO above, and I'd like to get most of the Bijectors.jl PRs merged first and then look into making this work with arbitrary variational posteriors `q`. To make that work, we need some way of doing parameter-updates for any `q`.

I'll have a proper look at this soon, then I'll be able to better judge what's left to do.

yebai:
@torfjelde, maybe focus on this PR for the next 2 weeks, when you got time.

xukai92:
It seems there are some conflicts. Can you fix them?

torfjelde:
On it! I'll also go through it one last time before we merge :+1:

xukai92:
Cool. We plan to do another review tonight.

src/variational/VariationalInference.jl
```diff
@@ -4,8 +4,8 @@ using ..Core, ..Utilities
 using Distributions, Bijectors, DynamicPPL
 using ProgressMeter, LinearAlgebra
 using ..Turing: PROGRESS
-using DynamicPPL: Model, SampleFromPrior, SampleFromUniform
 using ..Turing: Turing
```
xukai92: Shall we merge the two lines
```julia
using ..Turing: PROGRESS
using ..Turing: Turing	
```

src/variational/VariationalInference.jl
```diff
@@ -16,13 +16,17 @@ import ..Core: getchunksize, getADtype
 using Requires
 function __init__()
     @require Flux="587475ba-b771-5e3f-ad9e-33799f191a9c" apply!(o, x, Δ) = Flux.Optimise.apply!(o, x, Δ)
+    @require Flux="587475ba-b771-5e3f-ad9e-33799f191a9c" Flux.Optimise.apply!(o::TruncatedADAGrad, x, Δ) = apply!(o, x, Δ)
```
xukai92: Can we do
```julia
@require Flux="..." begin
...
end
```

src/variational/VariationalInference.jl
```diff
@@ -76,17 +61,29 @@ function grad!(
...
+
+# Arguments
+- `model`: `Turing.Model` or `Function` z ↦ log p(x, z) where `x` denotes the observations
+- `alg`: the VI algorithm used
+- `q`: a `VariationalPosterior` for which it is assumed a specialized implementation of the variational objective used exists.
+- `getq`: function taking parameters `θ` as input and returns a `VariationalPosterior`
+- `θ`: only required if `getq` is used, in which case it is the initial parameters for the variational posterior
 """
-function vi(model::Model, alg::VariationalInference)
+function vi(model, alg::VariationalInference)
```
xukai92: Maybe replace the function with `vi(model, alg:: VariationalInference ) = throw(MethodError(vi, model, alg))`; same for the other two that comes next

src/variational/VariationalInference.jl
```diff
@@ -95,15 +92,17 @@ end
...
+    q,
+    model,
     θ::AbstractVector{<:Real},
     out::DiffResults.MutableDiffResult,
     args...
 )
-    # TODO: this probably slows down executation quite a bit; exists a better way
-    # of doing this?
-    f(θ_) = - vo(alg, q, model, θ_, args...)
+    f(θ_) = if (q isa VariationalPosterior)
```
xukai92: Can we implement a function for this and achive this `if` statement via multiple dispatch?

src/variational/VariationalInference.jl
```diff
@@ -115,51 +114,57 @@ end
...
+    q,
+    model,
     θ::AbstractVector{<:Real},
     out::DiffResults.MutableDiffResult,
     args...
 )
-    θ_tracked = [Tracker.param(θ[i]) for i ∈ eachindex(θ)]
-    y = - vo(alg, q, model, θ_tracked, args...)
+    θ_tracked = Tracker.param(θ)
+    y = if (q isa VariationalPosterior)
```
xukai92: Same here; maybe the function could be reused too?

src/variational/advi.jl
```diff
@@ -34,101 +91,125 @@ function vi(model::Model, alg::ADVI; optimizer = TruncatedADAGrad())
         idx += varinfo.metadata[sym].ranges[end][end]
     end
 
-    q = Variational.MeanField(zeros(num_params), zeros(num_params), dists, ranges)
-    
-    # construct objective
-    elbo = ELBO()
+    # initial params
+    μ = randn(num_params)
+    σ = softplus.(randn(num_params))
```
xukai92: Shall we pass `rng` here somehow?

src/variational/advi.jl
```diff
@@ -34,101 +91,125 @@ function vi(model::Model, alg::ADVI; optimizer = TruncatedADAGrad())
...
+    logπ::Function,
     num_samples
 )
-    # setup
-    varinfo = Turing.VarInfo(model)
-
-    T = eltype(θ)
-    num_params = length(q)
-    μ, ω = θ[1:num_params], θ[num_params + 1: end]
+    #   𝔼_q(z)[log p(xᵢ, z)]
```
xukai92: I suggest we re-write these comments to the docstring of this function in Markdown. Not sure if it support LaTeX though.

src/variational/VariationalInference.jl
devmotion: Can't you remove the lines completely if you just want to throw a `MethodError`? If there exists no suitable method, a method error will be thrown anyways. If some "empty" definition of `vi` is required, you could just write
```julia
function vi end
```
Then users/developers can implement `vi` but you don't have to provide a default implementation (and if there is no suitable method, a `MethodError` will be thrown).

src/variational/VariationalInference.jl
torfjelde: TL;DR: Tried it and it was super-slow for some reason. Current approach is alright IMO, and think we can explore making it nicer in AdvancedVI.jl instead:)

Last week or so I implemented this by adding a "wrap" function, e.g. something along the lines of
```julia
function wrapped_objective(...)
    function f(...)
        # define objective
    end

    return f
end
```
But I ended up with a 10X slow-down because of this, and so I figured "ehh, I'll leave this for now and we can explore this further in AdvancedVI.jl". Also because I wasn't quite happy with the outcome; it's difficult to make this "wrap" approach work in general, i.e. if someone implements a custom `grad!` it's very likely they'll also have to implement their custom `wrap_objective` due to arguments and whatnot.

xukai92: Yes let's defer it to AVI

torfjelde: Great! Then I'll address the other comments and it should be ready to merge

src/variational/advi.jl
torfjelde: So I don't think this is necessarily very useful to expose to the user. It's mainly there to explain why we do that computation internally, e.g. you come across this piece of code 1 year later and wonder why we're not subtracting `logabsdetjac` (it's a bit confusing wrt. the papers, so I messed up the sign at some point and re-did the derivation and left it there for the inevitable future when I'll get confused again:) )

src/variational/advi.jl
torfjelde: Well, actually, is it necessary? This mainly just to initialze the model. If you want to ensure that reproducibility, you should just be instantiating the `MvNormal` yourself and use `bijector(model)` to get the model.

src/variational/VariationalInference.jl
torfjelde: Ah, that's true!
I'll do this for `vi`, `grad!` and `optimize! then:)

src/variational/advi.jl
xukai92: Would it require users to use Bijectors.jl directly then? I think it's still nice if we allow user to pass RNGs via a keyword arguments. We just need to add `rng=GLOBAL_RNG` and pass it to the `randn`

src/variational/VariationalInference.jl
xukai92: Yes sure. Please ping me when it's ready to merge.

src/variational/advi.jl
torfjelde: Nah, you still just do `Turing.Variational.bijector` right?
But yeah, we could just add it as a keyword argument if you think that's the best.

src/variational/VariationalInference.jl
torfjelde: This is resolved right? We're deferring this to AVI

src/variational/advi.jl
xukai92: What I mean is that users need to learn how to use the interface from Bijectors.jl, that contains defining the distribution, how `inv` and `Stacked` works and finally transformation, as you did in Line 98 to Line 103 next. Let's make a keyword argument possible, which is more friendly to end-users.

torfjelde: Done!

xukai92: Cool thanks!

xukai92:
Is it ready to merge?

src/variational/advi.jl
```diff
@@ -68,7 +68,7 @@ end
 
 Creates a mean-field approximation with multivariate normal as underlying distribution.
 """
-function meanfield(model::Model)
+function meanfield(model::Model; rng::AbstractRNG = GLOBAL_RNG)
```
devmotion: The common approach (as far as I have seen in Random, Distributions and some other packages) is to allow a random number generator as first argument, i.e., to define
```julia
meanfield(model::Model) = meanfield(Random.GLOBAL_RNG, model)
function meanfield(rng::AbstractRNG, model::Model)
    ...
end
```
I'm wondering if that could/should be done here as well?

torfjelde: Yeah, so this is what I do for `ELBO` (see `variational/objectives.jl`) and so on. I didn't do it with `meanfield` because I initially didn't think it necessary. What say you @xukai92 ? Should I change from kwarg to this?

xukai92: Non explictly, but we current do both for two different situations
- For internal functions, we stick to the way @devmotion describes above
- For functions to end-users, we use keyword arguments

devmotion: > For functions to end-users, we use keyword arguments

I'm curious, which functions are you referring to? At least the user-facing `sample`, `psample`, and `rand` only allow a random number generator as first argument but not as keyword argument.

xukai92: IIRC our `sample` takes a keyword for `rng`

devmotion: That's not case, I guess that might have changed at some point. The current API of AbstractMCMC and the implementation of `sample` in Turing (https://github.com/TuringLang/Turing.jl/blob/9dc8a297aaed4e898beba698491116c28e40c533/src/inference/Inference.jl#L143-L167) do not take a random number generator as keyword argument but only as first argument.

devmotion: BTW I just notice that this implementation is quite inconsistent since only in the second case the `resume_from` keyword is respected. I'll change that in the updates for AbstractMCMC 0.4.

xukai92: OK cool. If all interfaces take this convention now we should stick to it. Thanks for updating me on this. @torfjelde Let's make it the optional first argument :)

torfjelde: Grreaaat! Done!

xukai92:
I will merge this in 24 hours if no objection.

xukai92:
Merged! Thanks for the great work from @torfjelde and everyone who helped review this PR!

PR #903, opened 2019-09-02
[WIP] Stochastic Variational Inference (SVI)

torfjelde:
# Overview
Introduces **stochastic variational inference (SVI)** algoritm `SVI <: VariationalInference` into the PR #902 . The intention is to wrap any other `VariatonalInference` method with an `SVI` and call `vi` on a model which instantiated only on a subset of the data.

```julia
struct SVI{AD, A, D, F} <: VariationalInference{AD} where {A <: VariationalInference{AD}, D <: NamedTuple, F}
    alg::A
    data::D
    batch::F
    n::Int
end
function SVI(alg::A, data::D, batch::F, n::Int) where {AD, A <: VariationalInference{AD}, D, F}
    return SVI{AD, A, D, F}(alg, data, batch, n)
end
``` 
where `F` is a function with signature `F(model, data)` returning a generator of batches.

The idea is to have something like:
```julia
@model demo(x) = begin
    # ...
end

# returns a generator of batches, e.g. `(x = data[batch_indices])`
gen_batches(model, data) = ...

# instantiate model on an example batch
m = demo(x = data[1:batch_size])

# algos
advi = ADVI(samples_per_step, num_steps)
svi = SVI(advi,  (x = data), gen_batches, length(data))

# perform VI as usual
q = vi(m, svi)
```

# Examples
## ADVI applied to Bayesian NN on MNIST 
[Notebook](https://nbviewer.jupyter.org/gist/torfjelde/bd457142c8a4b5036dd2aca1f6db6a1e)

# Up for discussion
The entire interface:) I think it makes some sense given the rest of the VI interface and Turing in general. It's also very flexible, allowing the user to specify whatever way they want to generate batches. But it could definitively be a bit nicer to work with I think. Anyways, would love feedback :+1:


trappmartin:
Why does the `gen_batches` function need the model? Should it not be sufficient to pass only the data and return batches of data?

willtebbutt:
It's great to see mini-batching being a thing that we're going to be able to do. I think we should discuss names though. To me, SVI corresponds specifically to variational inference (variational Bayes) in which the data is mini-batched, but in which we're working with conjugate exponential family distributions and computing the natural gradients w.r.t. the parameters of the variational distributions. Specifically, this corresponds to [1].

If I've understood correctly, this PR introduces the ability to minibatch variational Bayes, so I guess my issue is the fact that we're calling a thing SVI, but not doing anything natural-gradient-y.

One option would be to name it something like `MinibatchedADVI`? I'm definitely open to suggestions.

[1] - Hoffman, Matthew D., et al. "Stochastic variational inference." The Journal of Machine Learning Research 14.1 (2013): 1303-1347.

torfjelde:
> Why does the gen_batches function need the model? Should it not be sufficient to pass only the data and return batches of data?

Yup, you're completely right:) This is leftover from earlier work that I didn't think to revamp. Originally I had this method alter the `model.data`, but now that this is done in `SVI` (which we should warn the user about) we no longer need to pass the actual model to it. 

Does `gen_batch(data, batch_size)` sound better?

Sidenote: I realize we can also probably provide an alright default `gen_batch` by allowing the user to specify the symbols to change on each batch, always assuming the last index refers to the index of the data. 

> I think we should discuss names though. To me, SVI corresponds specifically to variational inference (variational Bayes) in which the data is mini-batched, but in which we're working with conjugate exponential family distributions and computing the natural gradients w.r.t. the parameters of the variational distributions.

I automatically conflated it with using stochastic optimization since that's basically what they're doing with a single sample (with the exception of using the natural gradient). But now that I'm looking at it, in the paper you mention, at the beginning of Section 2.4 they have a line stating the following _"Stochastic variational inference uses stochastic optimization to fit the global variational parameters."_. Also in [Section 3.1, 1] they write _"Stochastic Variational Inference amounts to applying stochastic optimization to the objective function encountered in VI ..."_. So I think that supports my naming? 

I do see what you mean though; it often seems to be presented hand-in-hand with a conjugate exponential application. It also doesn't matter that much; if `SVI` is confusing, then we could just make it more clear what it does. Not against renaming it :+1:

willtebbutt:
> But now that I'm looking at it, in the paper you mention, at the beginning of Section 2.4 they have a line stating the following "Stochastic variational inference uses stochastic optimization to fit the global variational parameters.". Also in [Section 3.1, 1] they write "Stochastic Variational Inference amounts to applying stochastic optimization to the objective function encountered in VI ...".

Fair point, although by that argument `ADVI` should also be called `SVI`. @yebai @xukai92 any thoughts?

torfjelde:
> Fair point, although by that argument ADVI should also be called SVI. @yebai @xukai92 any thoughts?

Hmm, not quite seeing this. `ADVI` by default evaluates p(x, z) for the _entire_ dataset for gradient estimation, while in `SVI` you only evaluate p(x, z) for a subset of the data in gradient estimation.

willtebbutt:
Another good point, I had forgotten that particular detail. Regardless, I think for the sake of clarity it's probably best just to name this abstraction according to what it does: it mini-batches ADVI, so we should probably refer to it as such.

torfjelde:
> Another good point, I had forgotten that particular detail. Regardless, I think for the sake of clarity it's probably best just to name this abstraction according to what it does: it mini-batches ADVI, so we should probably refer to it as such.

Not opposed renaming :+1: 

But just to re-state: the name shouldn't be specific to `ADVI` as this is designed to wrap any `VariationalInference` to allow mini-batching. Agree? I added this to the above comment, but it was in an edit so you might have missed it.

willtebbutt:
> But just to re-state: the name shouldn't be specific to ADVI as this is designed to wrap any VariationalInference to allow mini-batching. Agree? I added this to the above comment, but it was in an edit so you might have missed it.

Yup, agreed :)

trappmartin:
Why not name it according to what it does and make an interface as follows:

```
advi = ADVI(samples_per_step, num_steps))

# perform ADVI using batch optimisation
q = vi(m, advi, BatchOptimisation())

# perform ADVI using mini-batched optimisation
q = vi(m, advi, MiniBatchOptimisation(gen_batches))
```

torfjelde:
> Why not name it according to what it does and make an interface as follows:

Right now the general interface is `vi(m, advi, q)` to accommodate future specification of the variational posterior. So we'd have to do `vi(m, advi, q, MiniBatchOptimisation())`. We'd also have to carry this through all other function calls, which could make it a bit verbose. It also makes sense to evaluate, say, the ELBO using batches due to memory-constraints. At the moment you can for example do:

```julia
elbo(svi, q, mini_batched_model, 10)
```

which will evaluate the ELBO batch-wise. This way we'd be able to support stuff like getting data from `LazyArray` or loading data from files as you go. Therefore I'd prefer not to tie it down to optimisation alone.

Could we do something like simply renaming `SVI` to `BatchedVI` or `MiniBatchVI` but keep the rest of the interface the same? Then we use the entire dataset by default, but if the `VariationalInference` instance is wrapped in `MiniBatchVI` we do everything batch-wise.

What is `BatchOptimisation()` btw? Is it the entire dataset?

willtebbutt:
> What is BatchOptimisation() btw? Is it the entire dataset?

Yeah, annoying terminological weirdness. Batch === entire data set, mini-batch === mini batches. I've never quite understood why working with the entire data set is referred to as batch...

xukai92:
I think `MiniBatchVI` sounds more reasonable. 

The current `ADVI` is also doing stochastic optimisation as the gradient is computed by MC estimation. In this case it should also be named as `SVI`?

Martin's proposal (https://github.com/TuringLang/Turing.jl/pull/903#issuecomment-527402277) also looks nice.

torfjelde:
> I think MiniBatchVI sounds more reasonable.
:+1:

> The current ADVI is also doing stochastic optimisation as the gradient is computed by MC estimation. In this case it should also be named as SVI?

No because the current ADVI uses _all_ of the dataset, not a subset. You're right that's is stochastic _optimization_ because MC estimation, but SVI refers to using a subset of the data for gradient estimation. The `S` does not mean that the optimization is stochastic:)

So `SVI` can wrap `ADVI` which means we're doing `ADVI` but using minibatches of the data on top of the MC estimation for the gradient on each gradient step. `ADVI` alone will use the entire dataset and MC estimation for the gradient on each gradient step. Does that make sense?

xukai92:
Sure I 100% understand the difference here. 

My point is simply that the name `SVI` itself doesn't relfect where the stochastic is from but `MiniBatchVI` is explicit.

torfjelde:
> My point is simply that the name SVI itself doesn't relfect where the stochastic is from but MiniBatchVI is explicit.

Oh yeah, I completely agree; I remember finding the `SVI` naming confusing when I first saw it in the literature too. So I'm fine with using `MiniBatchVI`:)

yebai:
Since we are now separating variational inference into its own package #1362, maybe we can consider closing this PR and add the code here into `AdvancedVI` instead?

torfjelde:
I think we should just close this PR and open an issue over at AVI. We're going to have to re-design parts of the `SVI` implementation to work with more "general" cases (as the current impl of `optimize!` is heavily tied into Turing). 
It should be pretty straight-forward (maybe just add a field to `SVI` defining the mapping `data ↦ (z ↦ logjoint(z, data))` or something), but I think it requires a small discussion before going ahead.

yebai:
Let’s close this PR then; we can open an issue for discussion on the AVI repo.

Issue #906, opened 2019-09-11
Multivariate proposals for MH current not supported


torfjelde:
With the current implementation of `MH`, it's not possible to specify _multivariate_ proposals. 

MVE:

```julia
julia> using Turing

julia> # Model definition.
       @model demo() = begin
           x ~ MvNormal(zeros(2), ones(2))
       end
demo (generic function with 2 methods)

julia> m = demo()
Turing.Model{Tuple{:x},Tuple{},getfield(Main, Symbol("###inner_function#438#26")),NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}}}(getfield(Main, Symbol("###inner_function#438#26"))(Core.Box(getfield(Main, Symbol("###inner_function#438#26"))(#= circular reference @-2 =#))), NamedTuple(), NamedTuple())

julia> sample(m, MH(10, (:x, x -> MvNormal(x, 0.1))))
ERROR: MethodError: no method matching support(::MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Array{Float64,1}})
Closest candidates are:
  support(::DiscreteNonParametric) at /home/tor/.julia/packages/Distributions/0Wogo/src/univariate/discrete/discretenonparametric.jl:60
  support(::Union{Type{D<:Distribution{Univariate,Continuous}}, D<:Distribution{Univariate,Continuous}}) where D<:Distribution{Univariate,Continuous} at /home/tor/.julia/packages/Distributions/0Wogo/src/univariates.jl:130
  support(::Union{Type{D<:Distribution{Univariate,Discrete}}, D<:Distribution{Univariate,Discrete}}) where D<:Distribution{Univariate,Discrete} at /home/tor/.julia/packages/Distributions/0Wogo/src/univariates.jl:131
Stacktrace:
[...]

PR #910, opened 2019-09-12
Fix for issue 903

torfjelde:
Fix for https://github.com/TuringLang/Turing.jl/issues/906#issue-492186264

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -146,7 +146,7 @@ function assume(spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi::VarInfo
                 spl.state.proposal_ratio -= log(cdf(stdG, (ub-r)/σ) - cdf(stdG,(lb-r)/σ))
             else # Other than Gaussian proposal
                 r = rand(proposal)
-                if (r < support(dist).lb) | (r > support(dist).ub) # check if value lies in support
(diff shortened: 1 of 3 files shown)


src/inference/mh.jl
```diff
@@ -146,7 +146,7 @@ function assume(spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi::VarInfo
                 spl.state.proposal_ratio -= log(cdf(stdG, (ub-r)/σ) - cdf(stdG,(lb-r)/σ))
             else # Other than Gaussian proposal
                 r = rand(proposal)
-                if (r < support(dist).lb) | (r > support(dist).ub) # check if value lies in support
+                if !(insupport(dist, r)) # check if value lies in support
```
xukai92: Thanks for the fix. Looks correct to me. Can we introduce a test for this fix? 

Maybe we can extract the code within the if-statmeent `if typeof(proposal) == Distributions.Normal{Float64}` into a function and perform the "if-else" via multiple dispatch. And by doing we can implement some tests easily too.

torfjelde: It's not quite clear what we want to test :thinking: Sort, what is your `@test the_function(args...) == ?`

yebai: It's probably ok to do the refactoring @xukai92  suggested in a separate PR

PR #915, opened 2019-09-23
Docstring updates, guide improvements.

cpfiffer:
Adds some meager docstrings for the distributions we have, and tacks them onto the API page. I've also changed the name from Library to API for the page that shows off our docstrings.

It doesn't really seem like the docs are being built automatically -- @KDr2 do you know what's up with that?

### DIFF
--- a/docs/src/library/api.md
+++ b/docs/src/library/api.md
@@ -1,11 +1,15 @@
 ---
-title: Library
+title: API
 permalink: /docs/library/
 toc: true
 ---
 
 ```@meta
 CurrentModule = Turing
+```
+## Index
+
+```@index
 ```
 
 ## Modelling
@@ -32,6 +36,17 @@ SGLD
 SMC
 ```
 
+## Distributions
+
+```@docs
+Flat
+FlatPos
+BinomialLogit
+VecBinomialLogit
+OrderedLogistic
+```
+
+
 ## Data Structures
 ```@docs
 TArray
@@ -42,7 +57,3 @@ TArray
 tzeros
 ```
 
-## Index
-
-```@index
-```
--- a/docs/src/using-turing/advanced.md
+++ b/docs/src/using-turing/advanced.md
@@ -244,5 +244,5 @@ sampler = NUTS(0.65)
 model = gdemo([1.2, 3.5]
 
 # Run all samples.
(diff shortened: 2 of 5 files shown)


KDr2:
Hi @cpfiffer, the document will only be automatically built on the master branch or on a commit tagged with a `v*` tag.  After the building, the docs will be pushed to the `gh-pages` branch.

Only allowing master-branch or release tag building could prevent docs in a not-yet-approved commit being published. If you want to preview the update, you can build it on the local machine by run `julia ./docs/make.jl`, then start a simple web server which treats the `docs` directory as its root dir:

```
$cd docs
$ python -mSimpleHTTPServer
Serving HTTP on 0.0.0.0 port 8000 ...
```
The point the browser to http://localhost:8080/dev/ to preview the updates.

After the docs being merged into the master branch, it will be automatically built and published.

KDr2:
The CI job is triggered, but there's an error. I will check it ASAP.
https://travis-ci.org/TuringLang/Turing.jl/jobs/590813375

trappmartin:
It doesn't seem that the update of the webpage actually worked, e.g. Library is still called Library instead of API as changed by Cameron through this PR. I suppose something went wrong.

trappmartin:
Side question: Why do we need `DynamicHMC` to build the docs? This seem strange to me.

trappmartin:
For reference:
https://github.com/TuringLang/Turing.jl/blob/4dbb817e1d1484e9b9763c899dd7e0c7d8d23809/.travis.yml#L116
and 
https://github.com/TuringLang/Turing.jl/blob/4dbb817e1d1484e9b9763c899dd7e0c7d8d23809/docs/make.jl#L1

trappmartin:
Which docstring requires this? Would it make sense to get rid of it? It feels wrong that our documentation build depends on DynamicHMC. But maybe thats just me.

cpfiffer:
Well, ignore me, I guess. I was referring to this docstring

https://github.com/TuringLang/Turing.jl/blob/4dbb817e1d1484e9b9763c899dd7e0c7d8d23809/src/contrib/inference/dynamichmc.jl#L9

but we don't appear to be exporting it to the website, so the documentation build step probably doesn't need to include DynamicHMC. Unless we want to export the docstring.

PR #917, opened 2019-09-26
Use DistributionsAD

mohamed82008:
This PR makes use of DistributionsAD.jl and removes its code from Turing.


mohamed82008:
Tests pass on Julia 1.0 and 1.1.

Issue #886, opened 2019-08-12
API design for non-Turing samplers


yebai:
One important design goal of Turing is to making implementing new sampling algorithms easier. There is already an ongoing effort for decoupling modelling and inference algorithms (see. e.g. https://github.com/TuringLang/Turing.jl/pull/793, https://github.com/TuringLang/Turing.jl/issues/746, https://github.com/TuringLang/Turing.jl/pull/750). For the next steps, it would be helpful to 

- discuss concretely what APIs are required from the compiler to implement a new sampler
- implement a few Turing wrappers for the existing sampler packages, e.g. 
   - https://github.com/anthofflab/RobustAdaptiveMetropolisSampler.jl
   - https://github.com/aplavin/Polychord.jl
   - https://github.com/alan-turing-institute/PDSampler.jl
   - https://github.com/QuantifyingUncertainty/GeneralizedMetropolisHastings.jl
   - https://github.com/mauro3/KissMCMC.jl
   - https://github.com/madsjulia/AffineInvariantMCMC.jl
   - A C++ library of Markov Chain Monte Carlo (MCMC) methods https://www.kthohr.com/mcmclib.html 

More
- https://github.com/madsjulia/Mads.jl - https://lanl.gov/

Related: #895 #889

trappmartin:
I think the following additional functions could be a good start. (I think that the assume and observe statements should be kept.) Please correct me if I’m wrong or if I’m missing something.

* `parameters(model) -> Vector{VarName}`
* `prior(model, vn) -> Distribution`
* `logpdf(model, vi)`
* `logjoint(model, vi)`
* `rand(model) -> VarInfo`

I’m addition it would be good in my opinion to have an way to track the results `logpdf` of each assume and/or each observe statement. This could be stored in an additional data type called `Trace`. 

For this purpose I would add the following macros.
* `@addlogjoint logpdf(d, x)` # add return value to log joint using current VarInfo
* `@track logpdf(d, x)` # track log pdf value
* `@track z = x-1` # track result of expression

cpfiffer:
Are we assuming for this conversation that people are building a new `InferenceAlgorithm`, or do you want the discussion to be broader to include cases where people get rid of Turing's `Sampler` struct?

[...]

PR #920, opened 2019-09-28
Early API functions and the interface guide.

cpfiffer:
This PR includes

1. Some convenience functions for developers, some of which were discussed in https://github.com/TuringLang/Turing.jl/issues/886.
2. A guide on how to use the interface methods on both Turing and non-Turing methods by implementing MH.
3. Miscellaneous changes to internals, such as 
- allowing a default `Sampler` constructor
- storing `VarName` instead of a string in `Transition` structs (for later access)
- a default `observe` statement for all samplers which just uses `SampleFromPrior()` 
- a default `assume` statement which just returns the current value and it's density 

There are some things in here I think we should discuss before merging, because I've made a couple design choices that I hope might enable more people to work with immutable `Transition` structs instead of relying so heavily on mutable states. The guide is an attempt to showcase partial immutable sampling to mirror the non-Turing implementation, and the functions I added in to Turing's code base might make this a little easier.

# New functions

Here are the new tools I've added in for developers. I'd like some feedback on these specifically to see how we feel about them.

- A new `Transition` constructor, which has the signature `Transition(model::Model, spl::Sampler, θ::T, nt::NamedTuple=NamedTuple())`. You can pass in the model and the sampler, it'll run the model, grab the density, and bundle everything up for you into a `Transition`.
- `parameters!` is a function with the signature `parameters!(spl::Sampler, t::Transition)`, which accepts a `Transition` and updates the `VarInfo` to contain the parameters in the `Transition`. It returns a vector of the parameters if you would prefer to work with vectors instead of a `NamedTuple`.
- `parameters` is a function which simply returns a `NamedTuple` of the `vi`.
- `variables(spl::Sampler)` is a function that returns a vector containing the `VarNames` of a sampler, so you can retrieve or set information about them in a `VarInfo`. I would like this to be an iterator or something slightly quicker than what it is now, but I like the idea of giving developers quick access to some iterable of `VarName`.
- `Distributions.logpdf(model::Model, spl::T, θ::P)` accepts a model, a sampler, and a vector, and returns the log density. It's a little clumsy but it should get us closer to models that people can just evaluate.

There's some bugs in here I still have to fix, but this is the core of the PR.

### DIFF
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -1165,7 +1165,7 @@ function parameters!(vi::VarInfo, metadata::NamedTuple{md_names}, nt::NamedTuple
     for key in nt_keys
         if key in md_names
             for (k, v) in zip(metadata[key].vns, nt[key])
-                vi[k] = v isa Array ? v : [v]
+                vi[k] = v isa AbstractArray ? v : [v]
             end
         else
             throw(ArgumentError("Key $key not found in VarInfo metadata, acceptable values are $md_names"))
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -41,7 +41,8 @@ export  VarName,
         istrans,
         link!,
         invlink!,
-        tonamedtuple
+        tonamedtuple,
+        variables
 
 ####
 #### Types for typed and untyped VarInfo
@@ -168,7 +169,7 @@ symbols have been observed. `VarInfo{<:NamedTuple}` is aliased `TypedVarInfo`.
 
 Note: It is the user's responsibility to ensure that each "symbol" is visited at least
 once whenever the model is called, regardless of any stochastic branching. Each symbol
-refers to a Julia variable and can be a hierarchical array of many random variables, e.g. `x[1] ~ ...` and `x[2] ~ ...` both have the same symbol `x`.
+refers to a Julia variable and can be a hierarchical array of many random 7, e.g. `x[1] ~ ...` and `x[2] ~ ...` both have the same symbol `x`.
 """
 struct VarInfo{Tmeta, Tlogp} <: AbstractVarInfo
     metadata::Tmeta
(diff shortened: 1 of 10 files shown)


cpfiffer:
Alright, I've fixed the bugs on this one. We're ready for a review.

src/core/RandomVariables.jl
```diff
@@ -1169,6 +1215,30 @@ function show(io::IO, vi::UntypedVarInfo)
     print(io, vi_str)
 end
 
+function show(io::IO, vi::TypedVarInfo)
```
xukai92: Why the type is strict to `TypedVarInfo`?

src/core/RandomVariables.jl
```diff
@@ -1128,6 +1128,52 @@ end
...
+to access the `metadata` field of the `VarInfo`.
+"""
+function parameters!(vi::VarInfo, nt::NamedTuple)
+    return parameters!(vi, vi.metadata, nt)
+end
+function parameters!(vi::VarInfo, metadata::NamedTuple{md_names}, nt::NamedTuple{nt_keys}) where {md_names, nt_keys}
+    for key in nt_keys
+        if key in md_names
+            for (k, v) in zip(metadata[key].vns, nt[key])
+                vi[k] = v isa Array ? v : [v]
```
xukai92: `isa Array` -> `isa AbstractArray`?

src/core/RandomVariables.jl
cpfiffer: Because they have different structures -- `UntypedVarInfo` has explicit field names, whereas a `TypedVarInfo` splits all the information into various parts of a `NamedTuple`.

src/core/RandomVariables.jl
cpfiffer: Good catch, thanks!

cpfiffer:
Looks like the tests are good. Does anyone have any further comments on this?

yebai:
Ready to merge from my side after the merge conflict is fixed.

cpfiffer:
I'm going to close this for now and split the guide into parts. I will add the guide on just the interface methods in another PR, #1007.

PR #922, opened 2019-09-29
Multithreading support

cpfiffer:
Allows for very basic multithreaded sampling using the `psample(model, alg, samples, chains)` function. Requires Julia 1.3 or higher.

```julia
using Turing

@model gdemo() = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))
    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))
end

chains = psample(gdemo(), MH(), 1000, 4)
```

Output:

```
Object of type Chains, with data of type 1000×3×4 Array{Union{Missing, Real},3}

Iterations        = 1:1000
Thinning interval = 1
Chains            = 1, 2, 3, 4
Samples per chain = 1000
internals         = lp
parameters        = m, s

Summary Statistics

│ Row │ parameters │ mean    │ std      │ naive_se  │ mcse      │ ess     │ r_hat   │
│     │ Symbol     │ Float64 │ Float64  │ Float64   │ Float64   │ Any     │ Any     │
├─────┼────────────┼─────────┼──────────┼───────────┼───────────┼─────────┼─────────┤
│ 1   │ m          │ 1.15737 │ 0.823153 │ 0.0130152 │ 0.0242592 │ 871.953 │ 1.00229 │
│ 2   │ s          │ 2.0536  │ 1.67389  │ 0.0264665 │ 0.0516064 │ 1097.88 │ 1.00373 │

Quantiles

│ Row │ parameters │ 2.5%      │ 25.0%    │ 50.0%   │ 75.0%   │ 97.5%   │
│     │ Symbol     │ Float64   │ Float64  │ Float64 │ Float64 │ Float64 │
├─────┼────────────┼───────────┼──────────┼─────────┼─────────┼─────────┤
│ 1   │ m          │ -0.405005 │ 0.674982 │ 1.13412 │ 1.63625 │ 2.74961 │
│ 2   │ s          │ 0.604815  │ 1.15447  │ 1.62179 │ 2.46583 │ 5.92489 │
```

It seems to work trivially with all the samplers, since the model and the sampler are deepcopied for each thread:

```julia
chains = psample(gdemo(), NUTS(), 1000, 4)
```
Output:
```
Object of type Chains, with data of type 500×11×4 Array{Union{Missing, Real},3}

Iterations        = 1:500
Thinning interval = 1
Chains            = 1, 2, 3, 4
Samples per chain = 500
internals         = acceptance_rate, hamiltonian_energy, is_accept, log_density, lp, n_steps, numerical_error, step_size, tree_depth
parameters        = m, s

Summary Statistics

│ Row │ parameters │ mean    │ std      │ naive_se  │ mcse      │ ess     │ r_hat   │
│     │ Symbol     │ Float64 │ Float64  │ Float64   │ Float64   │ Any     │ Any     │
├─────┼────────────┼─────────┼──────────┼───────────┼───────────┼─────────┼─────────┤
│ 1   │ m          │ 1.11925 │ 0.808271 │ 0.0180735 │ 0.0246458 │ 967.523 │ 1.00123 │
│ 2   │ s          │ 2.04593 │ 1.76884  │ 0.0395526 │ 0.064614  │ 1091.36 │ 1.00281 │

Quantiles

│ Row │ parameters │ 2.5%      │ 25.0%    │ 50.0%   │ 75.0%   │ 97.5%   │
│     │ Symbol     │ Float64   │ Float64  │ Float64 │ Float64 │ Float64 │
├─────┼────────────┼───────────┼──────────┼─────────┼─────────┼─────────┤
│ 1   │ m          │ -0.493255 │ 0.622198 │ 1.10681 │ 1.60489 │ 2.73904 │
│ 2   │ s          │ 0.585433  │ 1.06458  │ 1.55281 │ 2.41384 │ 6.69643 │
```

How should we handle the tests on this one? Restrict the `psample` tests to only 1.3+ or something? Anyone see any obvious parallel problems I may have caused?


mohamed82008:
Is the random number generator thread-safe in Julia 1.3? It isn't currently.

cpfiffer:
I think so? I'm not sure, though. From the [blog post](https://julialang.org/blog/2019/07/multithreading):

> The approach we’ve taken with Julia’s default global random number generator (rand() and friends) is to make it thread-specific. On first use, each thread will create an independent instance of the default RNG type (currently MersenneTwister) seeded from system entropy. All operations that affect the random number state (rand, randn, seed!, etc.) will then operate on only the current thread’s RNG state. This way, multiple independent code sequences that seed and then use random numbers will individually work as expected.
>
> If you need all threads to use a known initial seed, you will need to set it up explicitly. For that kind of more precise control, or better performance, we recommend allocating and passing your own RNG objects (e.g. Rand.MersenneTwister()).

mohamed82008:
That's cool. But the blog post seems to be talking about calling `rand` and its family without the `rng` argument. I think if you pass the `rng` argument explicitly things may be different. Just to be sure, try sampling 4 random numbers in a parallel and serial loop starting from the same deepcopied `rng`. Then check if:
1. Any 2 numbers generated by the parallel loop are the same
2. The numbers generated by the parallel and serial loop are the same or not, allowing a different order.
If 1 is true, then a race condition happened. If 2 is true, then no race condition happened but the same `rng` was used by all threads. I suspect 1 may be false but 2 is true.

cpfiffer:
- Using the non-global `rng` gives the same values for the serial and threaded implementations, without a deep copy.
- With a deep copy, the threads all generate identical values.
- With the global range, they are independent samples.
- Using `RandomDevice` as a `rng` provides completely separate samples in both threaded and serial implementations (with no deepcopy).

Currently this PR doesn't perform any deep copying of the `rng` -- it just uses either the `GLOBAL_RNG` (which is threadsafe) by default, or it accepts whatever the user passes in without making a copy. If they pass in `RandomDevice` then they'll get random samples, but if they set a seed manually with `MersenneTwister` then all the draws will be the same if you sort and compare them between the threaded/serial versions, though the threaded version will have a scalar multiple more draws because of the number of chains sampled.

It looks like the `deepcopy` is the wrong way to go for the `rng`, which is good, because I don't have to change anything. The default `GLOBAL_RNG` is safe, but custom seeds might cause a bit of trouble.

I have a question on the concerns for sample validity on that. Let's say I draw five random numbers two ways. One is parallel, one is serial. In the end, the same set of random numbers are generated but in a different order. Does this invalidate the sample or otherwise cause some kind of numerical issue? 

Perhaps by default we should use the initial range to set one seed for each chain, which will maintain reproducibility regardless of the number of chains. Something like

```julia
seeds = [rand(rng, UInt) for _ in 1:n_chains]

Threads.@threads for i in 1:n_chains
    crng = MersenneTwister(seeds[i])
    # do chain stuff
end
```

cpfiffer:
Alright, I've addressed @mohamed82008's notes by pre-generating a vector of seeds, one for each chain. The seeds are on a per-chain basis because I wanted to ensure that each chain was sampled deterministically (according to it's seed) regardless of how many threads were used -- if you run `psample` with only two threads but you ask it to sample four chains, you will get the exact same result as if you had used 1, 10, or 30 threads. There's some additional code in there to make sure that the chains are all ordered in the same way every time.

There's some tests in here, but they'll only run on 1.3+, so I've hardcoded that restriction into the testing suite.

I'm ready for a formal review.

PR #928, opened 2019-10-05
docs bugfix: context url and version switcher

KDr2:
After v0.7.0 was tagged, I found two other bugs of the docs building process:
1. the `baseurl`(i.e. context path of the URL) was wrong, it should start with a slash
2. the version-switcher has a small glitch, it can not distinguish `v0.N` and `v0.N.0` very well.

This PR fixed them.

Also, these issues on the gh-pages branch were fixed manually.

### DIFF
--- a/docs/assets/js/version-switch.js
+++ b/docs/assets/js/version-switch.js
@@ -26,8 +26,10 @@ $(document).ready(function(){
         });
     }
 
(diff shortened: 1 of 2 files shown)


yebai:
Thanks @KDr2!

PR #931, opened 2019-10-16
Fix histogram plotting on Quick Start example

metanoid:
As per https://discourse.julialang.org/t/how-can-i-solve-no-method-matching-chains-in-turing-julia/27166

I needed this change to get the "Probabilistic Programming in Thirty Seconds" tutorial to run correctly


yebai:
Many thanks, @metanoid!

PR #940, opened 2019-10-27
[WIP] Compiler docs and primer on Bayesian inference for non-experts

mohamed82008:
This is a work-in-progress documentation of the current compiler design in Turing and related work. I finished the detailed description of `@model` macro but I skipped over the details of `VarInfo`, `VarName`, `observe` and `assume`. 

I will now cover the details of `VarInfo` and `VarName` then move on to describe what `assume` and `observe` do given that. I believe this will then complete the documentation of all the boilerplate in Turing. I won't get into the details of the MCMC algorithms though.

While trying to accurately define the notation that I used in the compiler docs, I found myself writing a primer on Bayesian inference for non-experts explaining the basic theory behind what we are doing in Turing. I think this is useful but maybe I can split it to another PR. I believe @mkborregaard asked for something like this on Slack before. Currently, it lacks any runnable code or pretty plots, but it would be nice to get some feedback.

### DIFF
--- a/docs/src/using-turing/compiler.md
+++ b/docs/src/using-turing/compiler.md
@@ -1,0 +1,221 @@
+---
+title: Turing Compiler Design
+---
+
+In this section, I will describe the current design of Turing's model compiler which enables Turing to perform various types of Bayesian inference without changing the model definition. What we call "compiler" is essentially just a macro that transforms the user's code to something that Julia's dispatch can operate on and that Julia's compiler can successfully do type inference on for efficient machine code generation.
+
+# Overview
+
+For a quick primer on Bayesian inference, [see this]({{site.baseurl}}/docs/using-turing/primer.md). The following terminology will be used in this section:
+
+- `D`: observed data variables conditioned upon in the posterior,
+- `P`: parameter variables distributed according to the posterior distribution, these will also be referred to as random variables,
+- `Model`: a fully defined probabilistic model with input data, and
+- `Model` generator: a function that can be used to instantiate a `Model` instance by inputing data `D`.
+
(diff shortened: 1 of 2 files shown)


mkborregaard:
Hi this looks great! Just for the record, what I was requesting was specifically something in the gap between the introduction and the more complex models used in practice. I realize that this gap is probably quite large :-)

mohamed82008:
Thanks for your response. I think some things missing from this primer (well there are a lot but at least some of them) are:
1. The advantages and disadvantages of each MCMC algorithm including the intuitive role of each parameter, 
2. When model re-parameterization is useful and why,
3. MCMC result diagnosis techniques, and
4. Some case studies discussing all the above.

The case studies can be in the form of commentary on some of the interesting models in the https://github.com/TuringLang/TuringTutorials repo. @cscherrer might be interested in this discussion. 

If you have other things you want to add to the list, please mention them. I don't know how or how fast this primer will evolve given everyone's schedules but at the very least I think we can provide pointers to good, compact, user-targeted educational material to answer the FAQs of PPL users; I am sure there is a lot of amazing material out there. For a number of practitioners and researchers trying to use Bayesian inference, reading multiple books with 500+ pages is not really an option. I think these should be our target audience.

I believe the Julia community is a great place to get more penetration for Bayesian inference into all sorts of fields. The differentiable programming support, the synergy we have in the form of code reuse, and the friendly syntax in PPLs and Julia itself make Bayesian inference within reach to a wide variety of people from various fields in the Julia community. So it would be nice if Julia had the best "Get started in Bayesian inference" document out there targeting non-PPL experts, although some mathematical knowledge may be assumed. That's just my opinion though.

mohamed82008:
I will open 2 new PRs, one for the primer and one for the compiler docs.

PR #941, opened 2019-10-28
A few updates of HMC, Gibbs, and NUTS to new API.

andreasnoack:
I think I got most of them but `@tbenchmark` looks like it might need to be updated to the new API so I haven't fixed the places where it's used. Also, from the changes I made to `stan-interface.jl` I'm wondering that it might not be tested but I can see that the coverage info is old. Any chance that you'd be able to upload new information? Finally, it looks like the tests in `test/skipped` have bitrotten so maybe they could just be deleted.

### DIFF
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -80,7 +80,7 @@ The arguments for each sampler are:
   * HMC: leapfrog step size, leapfrog step numbers.
   * Gibbs: component sampler 1, component sampler 2, ...
   * HMCDA: total leapfrog length, target accept ratio.
-  * NUTS: Number of adaptation steps (optional), target accept ratio.
+  * NUTS: number of adaptation steps (optional), target accept ratio.
 
 
 For detailed information on the samplers, please review Turing.jl's [API]({{site.baseurl}}/docs/library) documentation.
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -77,10 +77,10 @@ The arguments for each sampler are:
 
   * SMC: number of particles.
   * PG: number of particles, number of iterations.
-  * HMC: number of samples, leapfrog step size, leapfrog step numbers.
-  * Gibbs: number of samples, component sampler 1, component sampler 2, ...
-  * HMCDA: number of samples, total leapfrog length, target accept ratio.
-  * NUTS: number of samples, target accept ratio.
+  * HMC: leapfrog step size, leapfrog step numbers.
+  * Gibbs: component sampler 1, component sampler 2, ...
+  * HMCDA: total leapfrog length, target accept ratio.
+  * NUTS: Number of adaptation steps (optional), target accept ratio.
 
 
 For detailed information on the samplers, please review Turing.jl's [API]({{site.baseurl}}/docs/library) documentation.
@@ -373,7 +373,7 @@ end
 ```
 
 
-If we are trying to generate random random values from the `generator` model and we call `sample(generator(), HMC(1000, 0.01, 5))`, we will receive an error. This is because there is no way to determine `length(x)`, whether `x` is a vector, and the type of the values in `x`.
+If we are trying to generate random random values from the `generator` model and we call `sample(generator(), HMC(0.01, 5), 1000)`, we will receive an error. This is because there is no way to determine `length(x)`, whether `x` is a vector, and the type of the values in `x`.
 
 
 A sensible default value might be:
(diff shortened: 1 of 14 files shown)


andreasnoack:
Looks like the test failure on 32 bit is because of the use of `Int64` instead of just `Int` in https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/integrator.jl#L11 and https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/integrator.jl#L88

docs/src/using-turing/guide.md
```diff
@@ -77,10 +77,10 @@ The arguments for each sampler are:
...
   * SMC: number of particles.
   * PG: number of particles, number of iterations.
-  * HMC: number of samples, leapfrog step size, leapfrog step numbers.
-  * Gibbs: number of samples, component sampler 1, component sampler 2, ...
-  * HMCDA: number of samples, total leapfrog length, target accept ratio.
-  * NUTS: number of samples, target accept ratio.
+  * HMC: leapfrog step size, leapfrog step numbers.
+  * Gibbs: component sampler 1, component sampler 2, ...
+  * HMCDA: total leapfrog length, target accept ratio.
+  * NUTS: Number of adaptation steps (optional), target accept ratio.
```
xukai92: Can you change `number` -> `Number` so that it's consistent to those above?

xukai92:
Let's wait for https://github.com/TuringLang/AdvancedHMC.jl/pull/121 to be merged and released. Hopefully CI will pass then and I will merge this PR.

cpfiffer:
I would actually be fine with deleting the `skipped` tests -- does anyone have any objection?

xukai92:
Just close and open to re-trigger CI.

xukai92:
Seems that the `Int` fix causes some weird interaction on Windows 32 machines.

xukai92:
From https://dev.azure.com/yebai/TuringLang/_build/results?buildId=683&view=logs&jobId=a846d25a-e32c-5640-1b53-e815fab94407&taskId=777e4b2c-7cad-5027-3b62-2efbb20eb961&lineStart=308&lineEnd=309&colStart=1&colEnd=1, it seems to be that, on Windows 32, `::NamedTuple{(:i, :is_half),Tuple{Int,Bool}}` is assumed as `::NamedTuple{(:i, :is_half),Tuple{Int64,Bool}}`? I don't have a Windows 32 machine to test it but what exactly happens here?

xukai92:
I made a temp fix at https://github.com/TuringLang/AdvancedHMC.jl/pull/126

andreasnoack:
The issue was fixed in https://github.com/TuringLang/AdvancedHMC.jl/pull/121 but, for some reason, https://dev.azure.com/yebai/TuringLang/_build/results?buildId=683&view=logs&jobId=a846d25a-e32c-5640-1b53-e815fab94407&taskId=777e4b2c-7cad-5027-3b62-2efbb20eb961&lineStart=308&lineEnd=309&colStart=1&colEnd=1 is using version AdvancedHMC version 0.2.7 instead of the latest version. Hence, I think https://github.com/TuringLang/AdvancedHMC.jl/pull/126 is redundant.

xukai92:
Yes you're right. I will undo TuringLang/AdvancedHMC.jl#126 later.

andreasnoack:
Just a partial update here. I've added an `Pkg.update` step to the CI script since it was using old, probably cached, versions of the registry. However, updating the registry revealed several issues. There has been a new release of Zygote which requires `SepcialFunctions` version 0.8 so I had to expand the compat entry here as well. However, `Distributions` also needs to be upgraded to allow for `SpecialFunctions` version 0.8 so we need https://github.com/JuliaStats/Distributions.jl/pull/1006 and a new release.

xukai92:
Cool. Thanks.

andreasnoack:
So this ended up way ore complicated that anticipated. Currently, Distributions can't upgrade to SpecialFunctions 0.8 because of https://github.com/JuliaStats/Distributions.jl/pull/891 so I've removed the upgrade of Zygote and submitted https://github.com/TuringLang/Turing.jl/pull/946 to fix the current master. Unfortunately, latest release is still broken by the ZygoteRules release. 

However, that is still not enough to get this PR working. The problem is that AdvancedHMC doesn't allow StatsBase 0.32 which is what Distributions requires so instead the resolver fetches and older version of AdvancedHMC which doesn't have bounds. https://github.com/TuringLang/AdvancedHMC.jl/pull/127 fixes that.

andreasnoack:
Tests pass and the PR now only contains the original changes and the issues have been resolved in other PRs.

yebai:
Thanks @andreasnoack!

PR #946, opened 2019-11-05
Add bound on ZygoteRules because it's not done by Zygote

andreasnoack:
The new release of ZygoteRules broke Turing. This should unbreak it. Once we can upgrade to Zygote 0.4, it should be fine to remove ZygoteRules as a direct dependency.


xukai92:
Thanks! I'm actuall a bit confused on the version `compat` section. I feel that setting the upper bound for packages which are before `v1.0.0` release is hard as the upper bound is very strict. I guess it is intended right?

andreasnoack:
The upper bounds make life harder for the developers but easier for the users since it helps avoiding breaking things so I think we just have to live with the trouble.

The failure on Linux is now a proper but probably harmless test failure
```julia
hmc reverse diff: Test Failed at /home/vsts/work/1/s/test/test_utils/numerical_tests.jl:50
  Expression: ≈(E, val, atol=eps)
   Evaluated: 2.180438164264343 ≈ 2.0416666666666665 (atol=0.1) 
```
I guess we might want to use a relative tolerance here?

The failure on 32 but Windows is now because of DynamicHMC has an incorrect type annotation on 32 bit in the Report struct. Maybe we'll have to put that test behind a 64 bit condition for the time being since we using version 1 of DynamicHMC right now so it's unlikely that the issue will be fixed.

andreasnoack:
So now tests are finally passing. I had to change the handling of the tolerance in the test setup to allow for relative tolerances so now you can either specify `atol` or `rtol` instead of just the absolute tolerance `eps`. I've then changed the failing test mentioned in https://github.com/TuringLang/Turing.jl/pull/946#issuecomment-549893386 to use relative tolerance.

yebai:
Thanks, @andreasnoack!

PR #954, opened 2019-11-09
Export resume function

cpfiffer:
`resume` was not being exported, so I fixed that and made it more intuitive by automatically concatenating samples to the old chain. Tests won't pass until MCMCChains has a new minor release.


Project.toml
```diff
@@ -36,8 +36,8 @@ DistributionsAD = "0.1.0"
 FiniteDifferences = "0.9"
 ForwardDiff = "0.10.3"
 Libtask = "0.3.1"
+MCMCChains = "^0.3.15"
```
devmotion: The [caret specifiers are the default version specifiers](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1), so it's equivalent to writing just "0.3.15" without caret. Maybe you actually wanted to use a tilde specifier?

cpfiffer:
```suggestion
MCMCChains = "0.3.15"
```

Nope, I didn't mean to use a tilde, but I will remove the caret.

cpfiffer: Thanks for catching that, though, I wasn't aware that it was the default version spec.

cpfiffer:
Tests are passing on this one, could someone take a look at this?

PR #955, opened 2019-11-09
Bump Zygote version and drop bound on ZygoteRules

torfjelde:
Bump Zygote version and drop bound on ZygoteRules which is now bounded in Zygote.

Depends on https://github.com/TuringLang/DistributionsAD.jl/pull/15


yebai:
Is this ready to merge?

torfjelde:
> Is this ready to merge?

Yup! Should be good:)

PR #959, opened 2019-11-15
Remove redundant computations and unneeded allocations

devmotion:
This PR gets rid of some redundant computations and an additional allocation.

### DIFF
--- a/src/core/container.jl
+++ b/src/core/container.jl
@@ -185,12 +185,19 @@ function Libtask.consume(pc :: ParticleContainer)
     res
 end
 
-function weights(pc :: ParticleContainer)
+function weights(pc::ParticleContainer)
     @assert pc.num_particles == length(pc)
+
+    # compute the normalized weights
     logWs = pc.logWs
-    Ws = exp.(logWs .- maximum(logWs))
-    logZ = log(sum(Ws)) + maximum(logWs)
-    Ws = Ws ./ sum(Ws)
+    maximum_logWs = maximum(logWs)
+    Ws = @. exp(logWs - maximum_logWs)
+    sum_Ws = sum(Ws)
+    ldiv!(sum_Ws, Ws)
+
+    # compute the log-likelihood estimate, ignoring constant term ``- \log num_particles``
+    logZ = log(sum_Ws) + maximum_logWs
+
     return Ws, logZ
 end
 
--- a/src/core/container.jl
+++ b/src/core/container.jl
@@ -151,7 +151,7 @@ end
 function Libtask.consume(pc :: ParticleContainer)
     @assert pc.num_particles == length(pc)
     # normalisation factor: 1/N
-    _, z1      = weights(pc)
+    z1 = logZ(pc)
     n = length(pc.vals)
 
     particles = collect(pc)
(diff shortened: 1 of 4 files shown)


src/core/container.jl
```diff
@@ -185,12 +185,19 @@ function Libtask.consume(pc :: ParticleContainer)
...
+
+    # compute the normalized weights
     logWs = pc.logWs
-    Ws = exp.(logWs .- maximum(logWs))
-    logZ = log(sum(Ws)) + maximum(logWs)
-    Ws = Ws ./ sum(Ws)
+    maximum_logWs = maximum(logWs)
+    Ws = @. exp(logWs - maximum_logWs)
+    sum_Ws = sum(Ws)
+    ldiv!(sum_Ws, Ws)
```
willtebbutt: Would you mind writing this as `Ws ./= sum_Ws`? Just a matter of taste, but it took me a minute to figure out what `ldiv!(sum_Ws, Ws)` is meant to do.

devmotion: I just assumed that `ldiv!` might be more optimized for such a simple operation than the broadcast machinery.  However, a quick benchmark shows that this seems to be the case only for matrices, for vectors both versions seem to be equally fast:
```julia
using BenchmarkTools
using LinearAlgebra

f!(A, b) = (A ./= b)

A = rand(10_000)
b = rand()

@btime f!($A, $b) # 5.731 μs (0 allocations: 0 bytes)
@btime ldiv!($b, $A) # 5.731 μs (0 allocations: 0 bytes)

A = rand(100, 100)

@btime f!($A, $b) # 11.469 μs (0 allocations: 0 bytes)
@btime ldiv!($b, $A) # 5.731 μs (0 allocations: 0 bytes)
```

willtebbutt: Huh, that's really weird... clearly your choice is better though, so I'm happy to go with it :)

mohamed82008: Same comment as Will, please change it to broadcast for readability. Happy to approve otherwise.

mohamed82008:
Thanks @devmotion!

src/core/container.jl
```diff
@@ -245,21 +243,3 @@ function resample!(
 
     pc
 end
-
-
-########### Auxilary Functions ###################
-
-# ParticleContainer: particles ==> (weight, results)
-function getsample(pc :: ParticleContainer, i :: Int, w :: Float64 = 0.)
```
xukai92: I notice this function is removed in this PR. Any particular reason?

devmotion: When I started cleaning the implementation of `getsample` I noticed that actually it's never called. So I just removed it.

xukai92: Can you just comment them out for now instead of deleting? Unless @trappmartin or @yebai can confirm this is not useful at all.

yebai: I can confirm this is no longer needed after the MCMC interface work #793

src/core/container.jl
```diff
@@ -185,18 +185,16 @@ function Libtask.consume(pc :: ParticleContainer)
...
-function weights(pc :: ParticleContainer)
-    @assert pc.num_particles == length(pc)
-    logWs = pc.logWs
-    Ws = exp.(logWs .- maximum(logWs))
-    logZ = log(sum(Ws)) + maximum(logWs)
-    Ws = Ws ./ sum(Ws)
-    return Ws, logZ
-end
+# compute the normalized weights
+weights(pc::ParticleContainer) = softmax(pc.logWs)
```
yebai: nice!

PR #960, opened 2019-11-16
Update shim to DynamicHMC 2 and LogDensityProblems 0.9.

tpapp:
- move FunctionLogDensity to the DynamicHMC file, as nothing else uses it

- update glue code to LogDensityProblems 0.9

- fix sample extraction

- add sarcastic comment about unreported bug


yebai:
Many thanks, @tpapp!

PR #961, opened 2019-11-16
Update deprecated lbeta to logbeta

itsdfish:
`lbeta` in the `BinomialLogit` function is deprecated in favor of `logbeta`. Fixing this deprecation resulted in about a 25 fold improvement on my system. I'm not sure whether performance was improved or deprecations reduce performance.


yebai:
Thanks @itsdfish!

PR #962, opened 2019-11-17
Compiler and Varinfo docs

mohamed82008:
This PR explains the internals of Turing in a way that I hope makes sense to a reader who is familiar with Bayesian inference and intermediate-level Julia. If some things are not clear, please let me know so I can put more effort into explaining them. Given the coming re-design of the compiler, some of these details may become obsolete but I will try to keep it updated with every compiler PR so that others can read through it while reviewing my PR.

### DIFF
--- a/docs/src/using-turing/compiler.md
+++ b/docs/src/using-turing/compiler.md
@@ -1,0 +1,340 @@
+---
+title: Turing Compiler Design
+---
+
+In this section, I will describe the current design of Turing's model compiler which enables Turing to perform various types of Bayesian inference without changing the model definition. What we call "compiler" is essentially just a macro that transforms the user's code to something that Julia's dispatch can operate on and that Julia's compiler can successfully do type inference on for efficient machine code generation.
+
+# Overview
+
+The following terminology will be used in this section:
+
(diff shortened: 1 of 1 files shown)


docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,340 @@
...
+---
+
+In this section, I will describe the current design of Turing's model compiler which enables Turing to perform various types of Bayesian inference without changing the model definition. What we call "compiler" is essentially just a macro that transforms the user's code to something that Julia's dispatch can operate on and that Julia's compiler can successfully do type inference on for efficient machine code generation.
+
+# Overview
+
+The following terminology will be used in this section:
+
+- `D`: observed data variables conditioned upon in the posterior,
+- `P`: parameter variables distributed according to the posterior distribution, these will also be referred to as random variables,
```
trappmartin: Maybe better: distributed according to some prior distribution

mohamed82008: Sounds good.

mohamed82008:
Now that the compiler is being significantly simplified and changed in #965, I will add the updated docs there when ready.

PR #964, opened 2019-11-17
Update checks for nothing

devmotion:
Using `=== nothing` instead of `== nothing` can be slightly more performant (see https://github.com/JuliaLang/julia/issues/29674#issue-370706380) and it doesn't hurt to use `===` consistently. Additionally, `===` always returns a `Bool` whereas e.g. `missing == nothing` evaluates to `missing`.

### DIFF
--- a/docs/make-utils.jl
+++ b/docs/make-utils.jl
@@ -25,7 +25,7 @@ function remove_yaml(file, key=nothing)
     if !isempty(yaml)
         open(file, "w+") do f
             # write back YAML lines.
-            if key != nothing
+            if key !== nothing
                 for line in yaml
                     startswith(line, key * ":") || write(f, line)
                 end
--- a/src/contrib/inference/AdvancedSMCExtensions.jl
+++ b/src/contrib/inference/AdvancedSMCExtensions.jl
@@ -135,7 +135,7 @@ function sample(  model::Model,
                 )
 
     spl = Sampler(alg, model)
-    if resume_from != nothing
+    if resume_from !== nothing
         spl.selector = resume_from.info[:spl].selector
     end
     alg_str = "PMMH"
@@ -152,7 +152,7 @@ function sample(  model::Model,
     end
 
     # Init parameters
-    vi = if resume_from == nothing
+    vi = if resume_from === nothing
         vi_ = VarInfo(model)
     else
         resume_from.info[:vi]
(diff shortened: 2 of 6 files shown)


mohamed82008:
I don't see the perf issue with `x == nothing`. Especially with constant propagation, these if statements all compile away now:
```julia
julia> f(x) = x == nothing ? 1 : 2;

julia> g(x) = x === nothing ? 1 : 2;

julia> @code_typed f(1)
CodeInfo(
1 ─     goto #3 if not false
2 ─     nothing::Nothing
3 ┄     return 2
) => Int64

julia> @code_typed g(1)
CodeInfo(
1 ─     goto #3 if not false
2 ─     nothing::Nothing
3 ┄     return 2
) => Int64
```

devmotion:
As I wrote, it's not always more performant and in some cases constant propagation is sufficient. However, in general that's not true, and hence for consistency we changed all (in)equality checks with `nothing` (and other singleton types) to `===` and `!==` in JuliaDiffEq (see, e.g., https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/587). Moreover, using `===` or `!==` has the advantage that the expression is guaranteed to evaluate to `true` or `false`.

The original example from base (https://github.com/JuliaLang/julia/issues/28695) is still significantly slower with regular equality checks on Julia 1.2:
```julia
using BenchmarkTools

function f(d)
   it = keys(d)
   res = 0
   elst1 = iterate(it)
   while elst1 != nothing
      s1,ns1 = elst1
      elst2 = iterate(it,ns1)
      while elst2 != nothing
         s2,ns2 = elst2
         res += d[s1]*d[s2]
         elst2 = iterate(it,ns2)
      end
      elst1 = iterate(it,ns1)
   end
   res
end

function g(d)
   it = keys(d)
   res = 0
   elst1 = iterate(it)
   while elst1 !== nothing
      s1,ns1 = elst1
      elst2 = iterate(it,ns1)
      while elst2 !== nothing
         s2,ns2 = elst2
         res += d[s1]*d[s2]
         elst2 = iterate(it,ns2)
      end
      elst1 = iterate(it,ns1)
   end
   res
end

d = Dict(:a=>1, :b=>2, :c=>3, :d=>4)

@btime f($d) # 218.605 ns (10 allocations: 320 bytes)
@btime g($d) # 159.806 ns (0 allocations: 0 bytes)
```

mohamed82008:
I see your point, but I think this example is a bit too special because it uses `iterate` which has an inferred return type `Union{T, Nothing}`, so constant propagation is not enough to compile the check away. I doubt this will be a problem for us in Turing since the way we used `nothing` is pretty basic. Could you please run a benchmark to see if any speedup results from this PR?

trappmartin:
Thanks!

I don't see an issue with merging this. Maybe we can simply run the MCMC benchmarks on this and see what happens.

Issue #754, opened 2019-04-17
Function to add to logp within @model Turing model 


jonasmac16:
Following from a discussion on Slack #turing I am opening this issue to request the possibility  to add to `logp` of sample from within Turing model defined with the macro @model and to discuss what the best implementation would look like.

Two suggestions so far where 
1. Directly adding to the `logp` variable like this `logp += log(1.0)`
2. Having a function that handles the adding e.g.  `acclogpdf!(log(1.0))`

@trappmartin suggested the second to be potentially the better solution as it is independent of how the `logp` value is reflected internally and you don't add to an "undefined" variable like in option 1.

mohamed82008:
@xukai92 suggested before to have some way to get the `VarInfo` inside the model. I think we can some function `__getvarinfo__()` for this purpose. So at the beginning of the model call function body generated by the macro, we can have:
```julia
__getvarinfo__() = $vi_sym
```
where `vi_sym` is the generated symbol for `vi::VarInfo`. `acclogp!` is defined for `VarInfo`.

mohamed82008:
This is also useful for #703 where the user can use their own `try`-`catch` block or any more performant error avoidance mechanism, e.g. checking for signs of individual failure modes, and then if a failure mode is detected, `logp` can be set to `-Inf` and `return` can be called to interrupt.

trappmartin:
> @xukai92 suggested before to have some way to get the `VarInfo` inside the model. I think we can some function `__getvarinfo__()` for this purpose. So at the beginning of the model call function body generated by the macro, we can have:
> 
> ```julia
> __getvarinfo__() = $vi_sym
> ```
> 
> where `vi_sym` is the generated symbol for `vi::VarInfo`. `acclogp!` is defined for `VarInfo`.

I was thinking about replacing `acclogp!` and every `$lp_ += ...` statement with a macro call, e.g. `@acclogp ...` which would remove the need of passing a VarInfo object. I think it would be the best if the user doesn't need to pass a VarInfo object and doesn't need to know about the `vi_sym` or similar. 

In the macro, we would still manipulate the varinfo object but without the user having to know about it.

What do you think?

trappmartin:
> This is also useful for #703 where the user can use their own `try`-`catch` block or any more performant error avoidance mechanism, e.g. checking for signs of individual failure modes, and then if a failure mode is detected, `logp` can be set to `-Inf` and `return` can be called to interrupt.

[...]

Issue #761, opened 2019-04-23
Observe vectorisation issue.


trappmartin:
We increase `vi.num_produce` on `observe` only for a single distribution and not during vectorisation.
Note: This is only required for Particle based samplers.

https://github.com/TuringLang/Turing.jl/blob/65371ce4d5a241f123d00e413f02080b387370b4/src/inference/Inference.jl#L182-L187

mohamed82008:
Closed via #965.

Issue #760, opened 2019-04-23
Assume in vectorisation of HMC bug.


trappmartin:
We call `setgid!` only if `assume` is called for a single distribution during HMC sampling
https://github.com/TuringLang/Turing.jl/blob/65371ce4d5a241f123d00e413f02080b387370b4/src/inference/hmc.jl#L265-L267 

But we do not do the same during vectorisation.

PR #965, opened 2019-11-17
Compiler 3.0

mohamed82008:
This WIP PR will include yet another refactor of the internals of Turing with the following main goals:
1. Readability,
2. Modularity to be able to move the compiler to DynamicPPL at some point,
3. Flexibility and extensibility to enable the adding of features which are now tricky to implement, e.g. querying the likelihood, and finally
4. Interoperability with other PPL approaches to enable interop with Soss, Gen and ProbabilityModels.

This PR currently implements the following features:
- Full-fledged support for partially missing values with automatic differentiation. For example, the following will treat `x[1]` and `x[3]` as random variables and `x[2]` and `x[4]` as observations:
```julia
@model f(x) = begin
    for i in 1:length(x)
        x[i] ~ Normal()
    end
end
sample(f([missing, 1.0, missing, 2.0]), NUTS(10, 0.65), 1000)
``` 
- The way in which arguments to the model are handled was significantly changed. Every argument now needs to be defined either when constructing the model or by setting a default value that works the same way a default value works in Julia. When an input variable can be missing, its input value should be `missing`. This can be input when constructing the model or using the default value. If this missing variable is a vector and we need to loop or broadcast over it inside the model, the user needs to do: 
```julia
@model f(x, ::Type{T} = Float64) where {T} = begin
    if ismissing(x)
        x = Vector{T}(undef, 10)
    end
    for i in 1:length(x)
        x[i] ~ Normal()
    end
end
```
Note that any model argument whose value is not `missing` is not going to be handled as a random variable and will be treated as an observation. If an argument has no default value and is not input when constructing the `Turing.Model`, an error will be thrown. Note that the above `ismissing` if-block is only needed if we need to loop over `x`. If we only need `x ~ MvNormal(...)` or something similar, then there is no need to initialize it when `ismissing(x)` is true.
- Running the model now happens in a context, `DefaultContext` by default to calculate the log joint probability. But we can also run the model in different contexts to evaluate variants of the `logp`. For example, I defined the `LikelihoodContext` which enables the computation of the log likelihood only instead of the log joint probability. Another context defined is the `BatchContext` that computes `logprior + s * loglikelihood`, where `s = npoints / batch_size`. This is important for stochastic (batch-based) gradient algorithms to be optimizing the original objective in the expectation, and not highly weighing the prior.
- The user can now tap into the `VarInfo` object in the model definition using the macro `@varinfo()`. This is not a real macro and will throw if called outside the `@model` definition. However, inside the `@model` definition, one can use this to get a handle to the `VarInfo` object inside the model run for more advanced use cases. There is also `@logpdf()` which simply returns the current accumulated value of the log joint probability or whatever the context is having the model compute.
- Random variable names are now _not_ baked into the `Turing.Model` type and can be passed by the user using the special distribution type `NamedDist(dist, name)`. So `x ~ NamedDist(dist, name)` will have the random variable named `name` not `x`. `name` can be a string or a `Turing.VarName`.
- The `VarName` constructor is now very neat and intuitive. You can do `vn = Turing.@varname x[1,:][i]` to get a `VarName{:x}` with `vn.indexing == "[1,Colon()][$i]"`, whatever `i` is.
- And last but not least, broadcasting the `~` symbol is now supported for both observations and random variables. One can do `x .~ dist` to mean that all elements of `x` are distributed according to `dist`. The main difference between `~` and `.~` in usage is that `~` doesn't assume the left-hand-side to be defined while `.~` does. So using `x .~ dist` when `x` wasn't previously defined will lead to an error. Broadcasting will work column-wise when `dist isa MultivariateDistribution` and `x isa AbstractMatrix` but not `AbstractMatrix{>:AbstractVector}`. If `x isa AbstractMatrix{Any}` for example, the broadcasting semantics here become ambiguous so we throw an error. This should not be an issue for most use cases.

I still need to do more tests and rebase but I thought I should write this summary now because there will be no more features added to this PR, just fixes from now on.


xukai92:
I will review it tomorrow.

mohamed82008:
Fixes https://github.com/TuringLang/Turing.jl/issues/754.

src/Turing.jl
```diff
@@ -31,45 +31,43 @@ const CACHERANGES = 0b01
...
 
 """
-    struct Model{pvars, dvars, F, TData, TDefaults}
-        f::F
-        data::TData
-        defaults::TDefaults
-    end
+struct Model{F, Targs <: NamedTuple}
+    f::F
+    args::TData
```
xukai92: Should be `::Targs`

src/Turing.jl
```diff
@@ -31,45 +31,43 @@ const CACHERANGES = 0b01
...
-end
-get_pvars(m::Model{params}) where {params} = Tuple(params.types)
-get_dvars(m::Model{params, data}) where {params, data} = Tuple(data.types)
-get_defaults(m::Model) = m.defaults
-@generated function in_pvars(::Val{sym}, ::Model{params}) where {sym, params}
-    return sym in params.types ? :(true) : :(false)
+(model::Model)(vi) = model(vi, SampleFromPrior())
+(model::Model)(vi, spl) = model(vi, spl, DefaultContext())
+(model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
+function Base.getproperty(m::Model, f::Symbol)
```
xukai92: I think we should avoid doing this by simplying calling `getmissing(m)` directly.

src/Turing.jl
```diff
@@ -121,6 +119,15 @@ Sampler(alg) = Sampler(alg, Selector())
 Sampler(alg, model::Model) = Sampler(alg, model, Selector())
 Sampler(alg, model::Model, s::Selector) = Sampler(alg, model, s)
 
+abstract type AbstractContext end
```
xukai92: I feel this context thing is like what @yebai wanted to do via `AbstractRunner`. But the difference here is that for context we "dispatch" during compile time while runner "dispatch" during runtime. 
1. Is my understanding correct here?
2. If so, we probably only want to keep such functionality at one place.

src/core/RandomVariables.jl
```diff
@@ -54,30 +54,25 @@ export  VarName,
...
- `indexing == "[Colon(), 1]"`.
+ `indexing == "[Colon(),1]"`.
 - `x[:,1][2] ~ Normal()` will generate a `VarName` with `sym == :x` and
- `indexing == "[Colon(), 1][2]"`.
+ `indexing == "[Colon(),1][2]"`.
 """
 struct VarName{sym}
-    csym      ::    Symbol        # symbol generated in compilation time
-    indexing  ::    String        # indexing
-    counter   ::    Int           # counter of same {csym, uid}
```
xukai92: Seems that here we remove two fileds. Are they in favour of https://github.com/TuringLang/Turing.jl/issues/972#issuecomment-555117648?

src/core/compiler.jl
```diff
@@ -1,150 +1,126 @@
...
-        isdist = if isa($dist, AbstractVector)
-            # Check if the right-hand side is a vector of distributions.
-            all(d -> isa(d, Distribution), $dist)
-        else
-            # Check if the right-hand side is a distribution.
-            isa($dist, Distribution)
-        end
-        @assert isdist @error($(wrong_dist_errormsg(@__LINE__)))
-        $vi.logp += Turing.observe($sampler, $dist, $observation, $vi)
+function vsym(expr::Union{Expr, Symbol})
```
xukai92: Can you add an example of what this function does?

src/core/compiler.jl
```diff
@@ -1,150 +1,126 @@
...
 end
 
-"""
-    tilde(left, right, model_info)
-
-The `tilde` function generates observation expression for data variables and assumption expressions for parameter variables, updating `model_info` in the process.
-"""
-function tilde(left, right, model_info)
-    return generate_observe(left, right, model_info)
+function assert_dist(dist; msg)
```
xukai92: Maybe worth refactor this via multiple dispatch.

src/core/compiler.jl
```diff
@@ -339,106 +322,60 @@ Builds the output expression.
...
             local $var
-            if isdefined($model_name.data, $(QuoteNode(var)))
-                # The `Type{T}` arguments will always show up in the `model.data` named tuple.
-                # This is because we are not replacing the default value of these arguments with `nothing` but keeping the same value defined by by the user, e.g. `Float64`.
-                # So if the value is indeed correct, i.e. a type then it should just work
-                # If the value is not a type, i.e. `::Type{T} = 1` and the user doesn't pass it something for this argument, then it will give an error when constructing the model, which is a correct Julia error.
-                # This means that we don't need an expression for the default value of the `Type{T}` arguments in `model.defaults`.
-                if $model_name.data.$var isa Type && ($model_name.data.$var <: AbstractFloat || $model_name.data.$var <: AbstractArray)
-                    $var = Turing.Core.get_matching_type($sampler_name, $vi_name, $model_name.data.$var)
+            if isdefined($model.args, $(QuoteNode(var)))
```
xukai92: Do we still need `isdefined` if we are going to thorw an error if something is not passed in?

src/Turing.jl
xukai92: Also is this functionality is only possible to be done in compile time?

src/Turing.jl
mohamed82008: Thanks!

src/Turing.jl
mohamed82008: I can do that.

src/Turing.jl
mohamed82008: I think it is similar to the `AbstractRunner` concept but I never quite understood the objective of `AbstractRunner`. The point of the context here is that it introduces another dispatch layer that lets us modify the behavior of `~`. But instead of getting involved with the `assume` and `observe` dispatch on the sampler type, I made the function `assume_or_observe` which dispatches on the output of `@preprocess` to call either `assume` or `observe`, and simultaneously dispatches on the context to modify the inputs or outputs to the `assume` or `observe` functions as needed.

I don't think the concepts of compile-time and run-time are relevant here. We are dispatching on the context to run different code for different contexts. This naturally means that different code will be compiled and run for different contexts.

src/core/RandomVariables.jl
mohamed82008: These 2 are independent. The 2 fields removed were simply never used in all of Turing. The custom variable name was implemented using `NamedDist`.

src/core/compiler.jl
mohamed82008:
```julia
expr = :(x[1,:][2])
vsym(expr)
```
returns `:x`.

src/core/compiler.jl
mohamed82008: Sure

src/core/compiler.jl
mohamed82008: Agreed. The `isdefined` is unnecessary as it is implied.

src/core/compiler.jl
```diff
@@ -288,47 +255,63 @@ function build_model_info(input_expr)
...
+    replace_vi!(model_info)
 
-Translates ~ expressions to observation or assumption expressions, updating `model_info`.
+Replaces @vi() expressions to the VarInfo instance.
 """
-function translate_tilde!(model_info)
+function replace_vi!(model_info)
+    ex = model_info[:main_body]
+    vi = model_info[:main_body_names][:vi]
+    ex = MacroTools.postwalk(x -> @capture(x, @vi()) ? vi : x, ex)
```
torfjelde: Is the `macro vi` defined somewhere or is it just used as a placeholder?

Also, should we use the `vi`, given that we already have a `vi` function used in variational inference, i.e. completely different purpose?

mohamed82008: The macro is not defined. I can define and export it just to give a warning if someone tries to define another macro with the same name. I can change the name of the macro to `varinfo`.

torfjelde: > The macro is not defined. I can define and export it just to give a warning if someone tries to define another macro with the same name. 

Is there a reason why it looks like a macro then? 
And yeah, might be a good idea to define and export as you suggested:)

> I can change the name of the macro to varinfo. Does that sound better?

Sounds good to me!:)

Btw, awesome work on this PR! :heart:

mohamed82008: Thanks! I wanted to make it a macro because I didn't want to conflict with other variables. This use case is rare enough that we don't want to reserve a variable name for it. So I just felt that it is much harder to define a macro `@varinfo` by mistake than it is to define a variable. And if a macro with this name is defined in the user's module, when we have exported `@varinfo` from Turing, (unless in REPL I believe) a warning will be given to the user so there is less possibility of making a mistake and calling `@varinfo` thinking it is doing something when in fact we are just replacing it with our `VarInfo`.

mohamed82008:
Please don't review this PR now except for the fun of it. I will break it down a bit.

cpfiffer:
Could we add a macro `@logpdf` that just wraps around the `@vi` macro and only returns the log density?

mohamed82008:
> Could we add a macro `@logpdf` that just wraps around the `@vi` macro and only returns the log density?

Sure, I can do that.

mohamed82008:
I wrote a summary of the features added in this PR. I will get to more tests and rebasing as soon as I can.

mohamed82008:
The main things left here are:
- [ ] Making the failed tests pass
- [ ] Adding more tests for the new syntax features
- [ ] Modifying and adding the compiler and VarInfo docs from #962

mohamed82008:
So it seems that type stability was hurt in this PR. I will look into why.

yebai:
@mohamed82008, maybe take a look at the following two issues (if not fixed already) in this PR:

- https://github.com/TuringLang/Turing.jl/issues/760
- https://github.com/TuringLang/Turing.jl/issues/760

mohamed82008:
@yebai I pushed a fix to #760. You mentioned 2 issues but I only see one.

yebai:
>@yebai I pushed a fix to #760. You mentioned 2 issues but I only see one.

Sorry, there was a copy-and-paste issue. The other one is https://github.com/TuringLang/Turing.jl/issues/761

mohamed82008:
#761 is already fixed here for non-particle samplers. The dot syntax is not supported yet for particle samplers. I will do that in another PR.

yebai:
> #761 is already fixed here for non-particle samplers. The dot syntax is not supported yet for particle samplers. I will do that in another PR.

I see - thanks for the clarification.

yebai:
Ps. I closed this PR accidentally..

src/Turing.jl
xukai92: I guess we can achieve the same thing by dispatching over a different sampler type directly can we? I would agree having a context is much cleaner though.

test/core/compiler.jl
```diff
@@ -77,12 +76,6 @@ priors = 0 # See "new grammar" test.
         chain = sample(fggibbstest(xs), gibbs, 2);
     end
     @testset "model macro" begin
-        model_info = Dict(:main_body_names => Dict(:vi => :vi, :sampler => :sampler))
-        # unit test model macro
-        expr = Turing.generate_observe(:x, :y, model_info)
-        @test expr.head == :block
-        @test :(vi.logp += Turing.observe(sampler, y, x, vi)) in expr.args
-
         @model testmodel_comp(x, y) = begin
```
xukai92: Can we add a test that only passing one variable throws errors here.

test/core/compiler.jl
```diff
@@ -95,23 +88,26 @@ priors = 0 # See "new grammar" test.
...
             x ~ Normal()
             return x
         end
         f0_mm = testmodel0()
         @test mean(f0_mm() for _ in 1:1000) ≈ 0. atol=0.1
 
         # Test #544
-        @model testmodel0(x = Vector{Float64}(undef, 2)) = begin
+        @model testmodel0(x = missing) = begin
+            if ismissing(x)
```
xukai92: What's the expected behaviour of "a user forget to write this condition while not providing `x` to the model"?

test/core/compiler.jl
mohamed82008: Sure!

test/core/compiler.jl
mohamed82008: `x` will be missing so `x[1]` will error. I think this is pretty intuitive.

yebai:
Thanks, @mohamed82008. I'll take a look at this PR soon.

mohamed82008:
No worries. There are some minor test failures now again. I will fix them soon.

mohamed82008:
I will rebase and fix the tests here before the end of the week.

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+- `md.idcs` is the dictionary that maps each `VarName` instance to its index in
+ `md.vns`, `md.ranges`, `md.dists`, `md.orders` and `md.flags`.
+- `md.vns[md.idcs[vn]] == vn`.
+- `md.dists[md.idcs[vn]]` is the distribution of `vn`.
+- `md.gids[md.idcs[vn]]` is the set of algorithms used to sample `vn`. This is used in
+ the Gibbs sampling process.
+- `md.orders[md.idcs[vn]]` is the number of `observe` statements before `vn` is sampled.
+- `md.ranges[md.idcs[vn]]` is the index range of `vn` in `md.vals`.
+- `md.vals[md.ranges[md.idcs[vn]]]` is the linearized vector of values of corresponding to `vn`.
+- `md.flags` is a dictionary of true/false flags. `md.flags[flag][md.idcs[vn]]` is the
```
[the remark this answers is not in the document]
mohamed82008: This dictionary is used mostly for the `trans` and `del` flags. The `trans` flag is used to check if the variable's value corresponds to its distribution's support or the Euclidean space after transforming it with a bijector. The `del` flag is used for lazy deletion of variables in SMC.

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+sample this variable, its value and whether this value is transformed to real space or
+not. Let `md` be an instance of `Metadata`:
+- `md.vns` is the vector of all `VarName` instances. Let `vn` be an arbitrary element of `md.vns`
+- `md.idcs` is the dictionary that maps each `VarName` instance to its index in
+ `md.vns`, `md.ranges`, `md.dists`, `md.orders` and `md.flags`.
+- `md.vns[md.idcs[vn]] == vn`.
+- `md.dists[md.idcs[vn]]` is the distribution of `vn`.
+- `md.gids[md.idcs[vn]]` is the set of algorithms used to sample `vn`. This is used in
+ the Gibbs sampling process.
+- `md.orders[md.idcs[vn]]` is the number of `observe` statements before `vn` is sampled.
```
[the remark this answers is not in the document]
mohamed82008: I agree, the names can be made clearer. I would rather work on `VarInfo` in a separate PR though. This PR mostly touches the model macro and some inference code to add the contexts and I think it's good to keep it that way.

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+
+## `Metadata`
+
+The `Metadata` struct stores some metadata about the random variables sampled. This helps
+query certain information about a variable such as: its distribution, which samplers
+sample this variable, its value and whether this value is transformed to real space or
+not. Let `md` be an instance of `Metadata`:
+- `md.vns` is the vector of all `VarName` instances. Let `vn` be an arbitrary element of `md.vns`
```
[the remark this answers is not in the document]
mohamed82008: I can work on those in a separate PR.

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+- `md.idcs` is the dictionary that maps each `VarName` instance to its index in
```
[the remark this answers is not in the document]
mohamed82008: I can do it in a separate PR.

mohamed82008:
Thanks @willtebbutt for your comments. `VarInfo` will go through another refactor as I am moving it over to `DynamicPPL`. This PR goes part of the way towards simplifying Turing's internal. In this PR, I removed quite a lot of lines and complexity from the model macro. Most of the code added here was added to `Inference` to support contexts and to better handle broadcasting `~`. So let's move the VarInfo discussion to another thread. @cpfiffer also had a lot of thoughts on `VarInfo` which I would like to simultaneously consider in the other PR.

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+    end
+    p = TV(undef, 2)
+    p[1] ~ InverseGamma(2, 3)
+    p[2] ~ Normal(0, 1.0)
+    @. x[1:2] ~ Normal(p[2], sqrt(p[1]))
+    x[3] ~ Normal()
+    y ~ Normal(p[2], sqrt(p[1]))
+end
+```
+A `model::Model` can be defined using `gauss(rand(3), 1.0)` or `gauss(x = rand(3), y = 1.0)`. While constructing the model, if an argument is not passed in, it will be assigned to its default value. If there is no default value given, an error will be thrown. If an argument has a default value `missing`, when not passed in, it will be treated as a random variable. For variables which require an intialization because we need to loop or broadcast over its elements, such as `x` above, the following needs to be done:
```
yebai: nice explanation!

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,224 @@
...
+- `name`: the model name.
+- `main_body`: the model body excluding the header and `end`.
+- `arg_syms`: the argument symbols, e.g. `[:x, :y, :TV]` above.
+- `args`: a modified version of the arguments changing `::Type{TV}=Vector{Float64}` and `where {TV <: AbstractVector}` to `TV::Type{<:AbstractVector}=Vector{Float64}`. This is `[:(x = missing) :(y = 1.0), :(TV::Type{<:AbstractVector}=Vector{Float64})]` in the example above.
+- `args_nt`: an expression constructing a `NamedTuple` of the input arguments, e.g. :((x = x, y = y, T = T)) in the example above.
+and returns it as a dictionary called `model_info`.
+
+## `replace_tilde!`
+
+After some model information have been extracted, `replace_tilde!` replaces the `L ~ R` lines in the model with the output of `tilde(L, R, model_info)` where `L` and `R` are either expressions or symbols. `L` can also be a constant literal. The `replace_tilde!` function also replaces expressions of the form `@. L ~ R` with the output of `dot_tilde(L, R, model_info)`.
```
yebai: Any reason why `model_info` is passed to `tilde`?

src/contrib/inference/AdvancedSMCExtensions.jl
```diff
@@ -65,19 +65,6 @@ function Sampler(alg::PMMH, model::Model, s::Selector)
         space = union(space, sub_alg.space)
     end
 
-    # Sanity check for space
```
yebai: Why removing sanity checks? Do we have an alternative now?

src/inference/gibbs.jl
```diff
@@ -59,13 +59,6 @@ end
 function Sampler(alg::Gibbs, model::Model, s::Selector)
     # sanity check for space
     space = getspace(alg)
-    pvars = get_pvars(model)
-    @assert issubset(pvars, space) "[Gibbs] symbols specified to samplers ($space) doesn't cover the model parameters ($pvars)"
```
yebai: Again, why sanity checks for Gibbs sampling are being removed?

yebai:
Really nice work thanks @mohamed82008. I am happy with merging this PR asap, except a few questions above.

Project.toml
```diff
@@ -8,6 +8,7 @@ Bijectors = "76274a88-744f-5084-9051-94815aaf08c4"
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DistributionsAD = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
```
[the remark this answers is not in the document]
mohamed82008: I use it https://github.com/TuringLang/Turing.jl/pull/965/files#diff-af86f4f961877d43dd8b04152a37cdbaR666. I will workaround it.

docs/src/using-turing/compiler.md
mohamed82008: This is `Core.tilde` which is different from `Inference.tilde`. `Core.tilde` returns the expression that will be run instead of the `~` line. To write out this expression, we need the variable names generated for `model`, `ctx`, `vi`, and `sampler`. These names are `gensym`ed in `build_model_info` and saved in `model_info` for reuse. One part of the expression generated by `Core.tilde` is a call to `Inference.tilde`. `Core.tilde` is defined in the `compiler.jl` file, while `Inference.tilde` is defined in the `Inference.jl` file.

src/contrib/inference/AdvancedSMCExtensions.jl
mohamed82008: Because `get_pvars` is not defined in the type of the model now; the variable names can be passed in by the user using `NamedDist`.

mohamed82008: Previously, the parameter symbols were part of the type of `model`, so this was simple. Now, we would need to call `VarInfo(model)` and get the symbols of the result to know the symbols of the variables. I can try saving the symbols in the `Model` struct at construction by doing something similar to `VarInfo(model)` but because the user is free to change the variable names of `model` after construction, this approach is not good.

yebai: I see. It is useful to have some sanity checking for variable assignment in case some variables are not assigned. But I agree that we can address this issue in a separate effort. Can you create a new issue for this issue?

docs/src/using-turing/compiler.md
yebai: Thanks for the clarification. Maybe add this distinction to the docs, e.g. in the form of a comment/footnote?

src/Turing.jl
```diff
@@ -119,6 +114,15 @@ Sampler(alg) = Sampler(alg, Selector())
...
+struct DefaultContext <: AbstractContext end
+struct LikelihoodContext <: AbstractContext end
+struct BatchContext{Tctx, T} <: AbstractContext
```
yebai: Maybe consider 

1) renaming `BatchContext` to `MiniBatchContext`, since batched methods often refer to algorithms utilising the full dataset? 
2) adding a comment here, explaining the purpose of `BatchContext`?

yebai: cc @willtebbutt @trappmartin

yebai:
Ready to merge from my side except for a few minor issues above.

willtebbutt:
Will make a proper pass over this this afternoon.

docs/src/using-turing/compiler.md
mohamed82008: Sure

src/contrib/inference/AdvancedSMCExtensions.jl
mohamed82008: Sure

src/Turing.jl
mohamed82008: Sure

cpfiffer:
@mohamed82008 I rebased onto master for you since the changes in #1007 were fairly small.

mohamed82008:
@cpfiffer Thanks!

src/contrib/inference/AdvancedSMCExtensions.jl
mohamed82008: https://github.com/TuringLang/Turing.jl/issues/1012

docs/src/using-turing/compiler.md
mohamed82008: Thanks!

src/Turing.jl
```diff
@@ -114,6 +109,36 @@ Sampler(alg) = Sampler(alg, Selector())
...
+"""
+struct LikelihoodContext <: AbstractContext end
+
+"""
+    struct MiniBatchContext{Tctx, T} <: AbstractContext
+        ctx::Tctx
+        loglike_scalar::T
+    end
+
+The `MiniBatchContext` enables the computation of `log(prior) + s * log(likelihood of a batch)` when running the model, where `s` is the `loglike_scalar` field, typically equal to `the number of data points / batch size`. This is useful in batch-based stochastic gradient descent algorithms to be optimizing `log(prior) + log(likelihood of all the data points)` in the expectation.
```
willtebbutt: Please make line under 90 chars.

src/Turing.jl
```diff
@@ -114,6 +109,36 @@ Sampler(alg) = Sampler(alg, Selector())
...
+    struct DefaultContext <: AbstractContext end
+
+The `DefaultContext` is used by default to compute log the joint probability of the data and parameters when running the model.
+"""
+struct DefaultContext <: AbstractContext end
+
+"""
+    struct LikelihoodContext <: AbstractContext end
+
+The `LikelihoodContext` enables the computation of the log likelihood of the data when running the model.
```
willtebbutt: 90 char lim

src/Turing.jl
```diff
@@ -114,6 +109,36 @@ Sampler(alg) = Sampler(alg, Selector())
 Sampler(alg, model::Model) = Sampler(alg, model, Selector())
 Sampler(alg, model::Model, s::Selector) = Sampler(alg, model, s)
 
+abstract type AbstractContext end
+
+"""
+    struct DefaultContext <: AbstractContext end
+
+The `DefaultContext` is used by default to compute log the joint probability of the data and parameters when running the model.
```
willtebbutt: 90 char lim

src/Turing.jl
```diff
@@ -114,6 +109,36 @@ Sampler(alg) = Sampler(alg, Selector())
...
+        loglike_scalar::T
+    end
+
+The `MiniBatchContext` enables the computation of `log(prior) + s * log(likelihood of a batch)` when running the model, where `s` is the `loglike_scalar` field, typically equal to `the number of data points / batch size`. This is useful in batch-based stochastic gradient descent algorithms to be optimizing `log(prior) + log(likelihood of all the data points)` in the expectation.
+"""
+struct MiniBatchContext{Tctx, T} <: AbstractContext
+    ctx::Tctx
+    loglike_scalar::T
+end
+MiniBatchContext(ctx = DefaultContext(); batch_size, npoints) = MiniBatchContext(ctx, npoints/batch_size)
```
willtebbutt: 90 chars

src/core/RandomVariables.jl
```diff
@@ -54,30 +54,25 @@ export  VarName,
...
+    indexing::String # indexing
```
willtebbutt: redundant comment

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
+macro varinfo()
+    :(throw(_error_msg()))
+end
+macro logpdf()
+    :(throw(_error_msg()))
+end
+macro sampler()
+    :(throw(_error_msg()))
+end
+_error_msg() = "This macro is only for use in the `Turing.@model` macro and not for external use."
```
willtebbutt: 90 chars

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
-# TODO: Replace this macro, see issue #514
 """
-Usage: @VarName x[1,2][1+5][45][3]
-  return: (:x,[1,2],6,45,3)
+Usage: @varname x[1,2][1+5][45][3]
+  return: VarName{:x}("[1,2][6][45][3]")
 """
-macro VarName(expr::Union{Expr, Symbol})
+macro varname(expr::Union{Expr, Symbol})
+    expr |> varname |> esc
```
willtebbutt: Do we really need to pipe?

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
-function var_tuple(sym::Symbol, inds::Expr=:(()))
-    return esc(:($(QuoteNode(sym)), $inds, $(QuoteNode(gensym()))))
-end
-
 
-function wrong_dist_errormsg(l)
-    return "Right-hand side of a ~ must be subtype of Distribution or a vector of " *
-        "Distributions on line $(l)."
+macro vsym(expr::Union{Expr, Symbol})
+    expr |> vsym
```
willtebbutt: Again, piping?

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
-        end
-    end
-
-    lp = gensym(:lp)
-    return quote
-        $varname_expr
-        isdist = if isa($dist, AbstractVector)
-            # Check if the right-hand side is a vector of distributions.
-            all(d -> isa(d, Distribution), $dist)
+function split_var_str(var_str, inds_as = Vector)
```
willtebbutt: Please add docstring

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
-        isdist = if isa($dist, AbstractVector)
-            # Check if the right-hand side is a vector of distributions.
-            all(d -> isa(d, Distribution), $dist)
-        else
-            # Check if the right-hand side is a distribution.
-            isa($dist, Distribution)
-        end
-        @assert isdist @error($(wrong_dist_errormsg(@__LINE__)))
-        $vi.logp += Turing.observe($sampler, $dist, $observation, $vi)
+function vsym(expr::Union{Expr, Symbol})
```
willtebbutt: Please add docstring

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
+    :(throw(_error_msg()))
+end
+_error_msg() = "This macro is only for use in the `Turing.@model` macro and not for external use."
 
-# TODO: Replace this macro, see issue #514
 """
-Usage: @VarName x[1,2][1+5][45][3]
-  return: (:x,[1,2],6,45,3)
+Usage: @varname x[1,2][1+5][45][3]
+  return: VarName{:x}("[1,2][6][45][3]")
```
willtebbutt: Please improve docstring to be more canonical and informative.

src/core/compiler.jl
```diff
@@ -1,150 +1,134 @@
...
-                    msg  *= " (ignoring `$(vsym)` found in global scope)"
+function wrong_dist_errormsg(l)
+    return "Right-hand side of a ~ must be subtype of Distribution or a vector of " *
+        "Distributions on line $(l)."
+end
+
+macro preprocess(data_vars, missing_vars, ex)
+    ex
+end
+macro preprocess(data_vars, missing_vars, ex::Union{Symbol, Expr})
```
willtebbutt: Docstring please :)

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
 end
 
 """
-    translate_tilde!(model_info)
+    replace_vi!(model_info)
+
+Replaces @varinfo() expressions to the VarInfo instance.
```
willtebbutt: Please say what this replaces the `@varinfo()` expressions with.

Also:
```suggestion
Replaces `@varinfo()` expressions in the VarInfo instance.
```

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
+    vi = model_info[:main_body_names][:vi]
+    ex = MacroTools.postwalk(x -> @capture(x, @varinfo()) ? vi : x, ex)
+    model_info[:main_body] = ex
+    return model_info
+end
+
+"""
+    replace_logpdf!(model_info)
+
+Replaces @logpdf() expressions to the VarInfo instance.
```
willtebbutt: See `replace_vi!` comment

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
+    ex = MacroTools.postwalk(x -> @capture(x, @logpdf()) ? :($vi.logp) : x, ex)
+    model_info[:main_body] = ex
+    return model_info
+end
+
+"""
+    replace_sampler!(model_info)
 
-Translates ~ expressions to observation or assumption expressions, updating `model_info`.
+Replaces @sampler() expressions to the VarInfo instance.
```
willtebbutt: See `replace_vi!` comment

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
+function replace_sampler!(model_info)
     ex = model_info[:main_body]
+    spl = model_info[:main_body_names][:sampler]
+    ex = MacroTools.postwalk(x -> @capture(x, @sampler()) ? spl : x, ex)
+    model_info[:main_body] = ex
+    return model_info
+end
+
+"""
+\"""
```
willtebbutt: What does this block do?

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
     return model_info
 end
+""" |> Meta.parse |> eval
 
 """
-    update_args!(model_info)
+    tilde(left, right, model_info)
 
-Extracts default argument values and replaces them with `nothing`.
+The `tilde` function generates observation expression for data variables and assumption expressions for parameter variables, updating `model_info` in the process.
```
willtebbutt: 90 chars. Also, maybe consider replacing "observation" with `observe` and `assumption with `assume` to more tightly reflect internal naming scheme.

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
+            $assert_ex
+            $vi.logp += Turing.Inference.tilde($ctx, $sampler, $temp_right, $left, $vi)
         end
     end
-    if length(fargs_default_values) == 0
-        tent_arg_defaults_nt = :(NamedTuple())
+    return ex
+end
+
+function dot_tilde(left, right, model_info)
```
willtebbutt: docstring please

src/core/compiler.jl
```diff
@@ -288,49 +229,144 @@ function build_model_info(input_expr)
...
+            $vi.logp += Turing.Inference.dot_tilde($ctx, $sampler, $temp_right, $temp_left, $vi)
+        end
     end
-    model_info[:args] = fargs
-    model_info[:tent_arg_defaults_nt] = tent_arg_defaults_nt
-    return model_info
+    return ex
 end
 
+is_number_or_array_type(T::Type{<:AbstractFloat}) = true
```
willtebbutt: Please consider replacing this function with something like
```julia
const FloatOrArray = Union{AbstractFloat, AbstractArray}
```
and `isa FloatOrArray` wherever relevant. It's just a bit cleaner imho.

src/inference/Inference.jl
```diff
@@ -535,9 +508,10 @@ end
 floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
 floatof(::Type) = Real # fallback if type inference failed
 
-@inline Turing.Core.get_matching_type(spl::Turing.Sampler, vi::Turing.RandomVariables.VarInfo, ::Type{<:AbstractFloat}) = floatof(eltype(vi, spl))
-@inline Turing.Core.get_matching_type(spl::Turing.Sampler{<:Hamiltonian}, vi::Turing.RandomVariables.VarInfo, ::Type{Array{T,N}}) where {T, N} = Array{Turing.Core.get_matching_type(spl, vi, T), N}
-@inline Turing.Core.get_matching_type(spl::Turing.Sampler{<:Union{PG, SMC}}, vi::Turing.RandomVariables.VarInfo, ::Type{Array{T,N}}) where {T, N} = TArray{T, N}
+@inline Turing.Core.get_matching_type(spl::AbstractSampler, vi::VarInfo, ::Type{T}) where {T <: AbstractFloat} = floatof(eltype(vi, spl))
```
willtebbutt: Can probably drop the `@inline`s. Please also re-write to meet 90 char limits for these four methods.

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+    return dot_tilde(ctx.ctx, sampler, right, left, vn, vi)
+end
+
+function _dot_tilde(sampler, right, left, vn::VarName, vi)
+    return dot_assume(sampler, right, vn, left, vi)
+end
+
+# Ambiguity error when not sure to use Distributions convention or Julia broadcasting semantics
+function _dot_tilde(sampler, right::Union{MultivariateDistribution, AbstractVector{<:MultivariateDistribution}}, left::AbstractMatrix{>:AbstractVector}, vn::VarName, vi)
+    throw("Ambiguous `lhs .~ rhs` or `@. lhs ~ rhs` syntax. The broadcasting can either be column-wise following the convention of Distributions.jl or element-wise following Julia's general broadcasting semantics. Please make sure that the element type of `lhs` is not a supertype of the support type of `AbstractVector` to eliminate ambiguity.")
```
willtebbutt: 90 char lim

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+    spl::Sampler,
+    ::Any,
+    ::VarName,
+    ::Any,
+    ::VarInfo
+)
+    error("[Turing] $(alg_str(spl)) doesn't support vectorizing assume statement")
+end
+
+@inline function get_and_set_val!(vi, vn::VarName, dist::Distribution, spl)
```
willtebbutt: Is it crucial to inline this?

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+    if haskey(vi, vn)
+        r = vi[vn]
+    else
+        r = isa(spl, SampleFromUniform) ? init(dist) : rand(dist)
+        push!(vi, vn, r, dist, spl)
+    end
+    return r
+end
+@inline function get_and_set_val!(vi, vns::AbstractVector{<:VarName}, dist::MultivariateDistribution, spl)
```
willtebbutt: 90 chars

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+        r = vi[vns]
+    else
+        r = isa(spl, SampleFromUniform) ? init(dist, n) : rand(dist, n)
+        for i in 1:n
+            push!(vi, vns[i], r[:,i], dist, spl)
         end
     end
+    return r
+end
+@inline function get_and_set_val!(vi, vns::AbstractArray{<:VarName}, dists::Union{Distribution, AbstractArray{<:Distribution}}, spl)
```
willtebbutt: 90 chars

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
 
-function observe(
-    spl::Union{SampleFromPrior, SampleFromUniform},
-    dist::Distribution,
-    value,
+function dot_observe(
+    ::Nothing,
+    dist::Union{Distribution, AbstractArray{<:Distribution}},
+    value, 
     vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
 )
+    return dot_observe(SampleFromPrior(), dist, value, vi)
+end
+function dot_observe(
+    spl::Union{SampleFromPrior, SampleFromUniform},
+    dist::MultivariateDistribution,
+    value::AbstractMatrix,
+    vi::VarInfo)
```
willtebbutt:
```suggestion
    vi::VarInfo,
)
```

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+
```
willtebbutt:
```suggestion
```

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
 end
-
-function observe(
+function dot_observe(
     spl::Union{SampleFromPrior, SampleFromUniform},
-    dists::Vector{<:Distribution},
-    value,
+    dists::Union{Distribution, AbstractArray{<:Distribution}},
+    value::AbstractArray,
     vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/inference/Inference.jl
```diff
@@ -573,92 +596,188 @@ function assume(
...
+    vi.num_produce += one(vi.num_produce)
+    Turing.DEBUG && @debug "dists = $dists"
+    Turing.DEBUG && @debug "value = $value"
+    return sum(logpdf.(dists, value))
+end
+function dot_observe(
+    spl::Sampler,
+    ::Any,
+    ::AbstractArray,
+    ::VarInfo
```
willtebbutt:
```suggestion
    ::VarInfo,
```

src/inference/hmc.jl
```diff
@@ -450,50 +450,52 @@ function assume(
...
 
-function assume(
+function dot_assume(
     spl::Sampler{<:Hamiltonian},
-    dists::Vector{<:Distribution},
+    dist::MultivariateDistribution,
     vn::VarName,
-    var,
+    var::AbstractMatrix,
     vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/inference/hmc.jl
```diff
@@ -450,50 +450,52 @@ function assume(
...
+    r = vi[vns]
+    var .= r
+    return var, sum(logpdf_with_trans(dist, r, istrans(vi, vns[1])))
+end
+function dot_assume(
+    spl::Sampler{<:Hamiltonian},
+    dists::Union{Distribution, AbstractArray{<:Distribution}},
+    vn::VarName,
+    var::AbstractArray,
+    vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/inference/hmc.jl
```diff
@@ -450,50 +450,52 @@ function assume(
...
+    return var, sum(logpdf_with_trans.(dists, r, istrans(vi, vns[1])))
 end
 
 function observe(
-    ::Sampler{<:Hamiltonian},
-    d::Union{Distribution,Vector{<:Distribution}},
+    spl::Sampler{<:Hamiltonian},
+    d::Distribution,
     value,
     vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/inference/hmc.jl
```diff
@@ -450,50 +450,52 @@ function assume(
...
 )
-    return observe(d, value, vi)
+    return observe(nothing, d, value, vi)
+end
+
+function dot_observe(
+    spl::Sampler{<:Hamiltonian},
+    ds::Union{Distribution, AbstractArray{<:Distribution}},
+    value::AbstractArray,
+    vi::VarInfo
```
willtebbutt:
```suggestion
    vi::VarInfo,
```

src/stdlib/distributions.jl
```diff
@@ -132,3 +146,26 @@ end
 function Distributions.logpdf(lp::LogPoisson, k::Int)
     return k * lp.logλ - exp(lp.logλ) - loggamma(k + 1)
 end
+
+"""
+A named distribution that carries the name of the random variable with it.
+"""
+struct NamedDist{variate, support, Td <: Distribution{variate, support}, Tn} <: Distribution{variate, support}
```
willtebbutt: 90 char lim

src/stdlib/distributions.jl
```diff
@@ -132,3 +146,26 @@ end
...
+
+struct NoDist{variate, support, Td <: Distribution{variate, support}} <: Distribution{variate, support}
+    dist::Td
+end
+NoDist(dist::NamedDist) = NamedDist(NoDist(dist.dist), dist.name)
+
+Distributions.rand(d::NoDist) = rand(d.dist)
+Distributions.logpdf(d::NoDist{<:Univariate}, ::Real) = 0
+Distributions.logpdf(d::NoDist{<:Multivariate}, ::AbstractVector{<:Real}) = 0
+Distributions.logpdf(d::NoDist{<:Multivariate}, x::AbstractMatrix{<:Real}) = zeros(Int, size(x, 2))
```
willtebbutt: 90 char lim

src/stdlib/distributions.jl
```diff
@@ -132,3 +146,26 @@ end
...
+Distributions.logpdf(d::NoDist{<:Matrixvariate}, ::AbstractMatrix{<:Real}) = 0
+Bijectors.logpdf_with_trans(d::NoDist{<:Univariate}, ::Real) = 0
+Bijectors.logpdf_with_trans(d::NoDist{<:Multivariate}, ::AbstractVector{<:Real}) = 0
+Bijectors.logpdf_with_trans(d::NoDist{<:Multivariate}, x::AbstractMatrix{<:Real}) = zeros(Int, size(x, 2))
```
willtebbutt: 90 char lim

src/stdlib/distributions.jl
```diff
@@ -132,3 +146,26 @@ end
...
+Bijectors.logpdf_with_trans(d::NoDist{<:Matrixvariate}, ::AbstractMatrix{<:Real}) = 0
```
willtebbutt:
```suggestion
Bijectors.logpdf_with_trans(d::NoDist{<:Matrixvariate}, ::AbstractMatrix{<:Real}) = 0

```

docs/src/using-turing/compiler.md
```diff
@@ -0,0 +1,226 @@
...
+    p[1] ~ InverseGamma(2, 3)
+    p[2] ~ Normal(0, 1.0)
+    @. x[1:2] ~ Normal(p[2], sqrt(p[1]))
+    x[3] ~ Normal()
+    y ~ Normal(p[2], sqrt(p[1]))
+end
+```
+A `model::Model` can be defined using `gauss(rand(3), 1.0)` or `gauss(x = rand(3), y = 1.0)`. While constructing the model, if an argument is not passed in, it will be assigned to its default value. If there is no default value given, an error will be thrown. If an argument has a default value `missing`, when not passed in, it will be treated as a random variable. For variables which require an intialization because we need to loop or broadcast over its elements, such as `x` above, the following needs to be done:
+```julia
+if ismissing(x)
```
devmotion: Is there a particular reason for recommending `ismissing(x)` instead of `x === missing`? Even though it might seem they should be equivalent, they're actually not (see, e.g., https://github.com/JuliaLang/julia/issues/27681#issuecomment-399054971; I just reproduced the example there on Julia 1.3).

src/core/compiler.jl
mohamed82008: No but I think it's neater this way.

mohamed82008:
Thanks for your reviews. I will complete addressing the comments tonight.

mohamed82008:
Ok definitely tomorrow morning :)

docs/src/using-turing/compiler.md
mohamed82008: I will recommend `===`, thanks!

src/core/compiler.jl
mohamed82008: This one supports `.~` syntax only if in Julia 1.1 and higher. Writing code with `.~` in Julia 1.0 gives a parsing error even if not run at all.

mohamed82008: I will add a comment.

src/inference/Inference.jl
mohamed82008: Probably not. I will do another inline vs noinline PR at some point for every function in Turing! Will remove this for now.

mohamed82008:
So there seems to be some version conflicts.

mohamed82008:
The tests are running properly now thanks to @goedman 's latest CmdStan release, thanks Rob!

yebai:
Many thanks, @mohamed82008!

goedman:
Mohamed, just to make sure, are you referring to CmdStan v5.3.0 or v5.3.1.

I believe the latter as currently BinDeps and Homebrew are not version-wise compatible.

mohamed82008:
5.3.1 yes

goedman:
Thank you.

PR #966, opened 2019-11-17
Simplify ParticleContainer

devmotion:
This PR simplifies the type parameters of `ParticleContainer`, and removes the `conditional` field which by definition was always `nothing`.

Moreover, instead of appending newly allocated arrays, `push!` now resizes the existing arrays and sets the new values, which is faster:
```julia
using BenchmarkTools

function f(x, n)
    y = zeros(eltype(x), n)
    append!(x, y)
    nothing
end

function g(x, n)
    m = length(x)
    mpn = m + n
    resize!(x, mpn)
    @inbounds for i in (m + 1):mpn
        x[i] = zero(eltype(x))
    end
    nothing
end

a = rand(1_000)

@btime f($(copy(a)), 100) # 339.092 ns (1 allocation: 896 bytes)
@btime g($(copy(a)), 100) # 263.981 ns (0 allocations: 0 bytes)
```

The `empty!` function now defines an array of particles whose element type is based on the previously used array instead of always using the abstract element type `Particle`.

The `copy!` function of `ParticleContainer` creates a set of copied particles by list comprehension instead of pushing them individually to an initially empty array.

Moreover, the definition of `Trace` is simplified by removing the completely uninitialized inner constructor and reordering the fields (making it easier to create new instances without specifying a `task`).

### DIFF
--- a/src/core/container.jl
+++ b/src/core/container.jl
@@ -1,31 +1,22 @@
-mutable struct Trace{Tspl <: AbstractSampler, Tvi <: AbstractVarInfo, Tmodel <: Model}
-    task  ::  Task
-    vi    ::  Tvi
-    spl   ::  Tspl
-    model ::  Tmodel
-    Trace{Tspl, Tvi, Tmodel}() where {Tspl, Tvi, Tmodel} = new()
-    function Trace{SampleFromPrior}(m::Model, spl::AbstractSampler, vi::AbstractVarInfo)
-        res = new{SampleFromPrior, typeof(vi), typeof(m)}()
-        res.vi = vi
-        res.model = m
-        res.spl = SampleFromPrior()
-        return res
+mutable struct Trace{Tspl<:AbstractSampler, Tvi<:AbstractVarInfo, Tmodel<:Model}
+    model::Tmodel
+    spl::Tspl
+    vi::Tvi
+    task::Task
+
+    function Trace{SampleFromPrior}(model::Model, spl::AbstractSampler, vi::AbstractVarInfo)
+        return new{SampleFromPrior,typeof(vi),typeof(model)}(model, SampleFromPrior(), vi)
     end
-    function Trace{T}(m::Model, spl::AbstractSampler, vi::AbstractVarInfo) where T <: Sampler
-        res = new{T, typeof(vi), typeof(m)}()
-        res.model = m
-        res.spl = spl
-        res.vi = vi
-        return res
+    function Trace{S}(model::Model, spl::S, vi::AbstractVarInfo) where S<:Sampler
+        return new{S,typeof(vi),typeof(model)}(model, spl, vi)
     end
 end
+
 function Base.copy(trace::Trace)
-    newtrace = typeof(trace)()
-    newtrace.vi = deepcopy(trace.vi)
-    newtrace.task = Base.copy(trace.task)
-    newtrace.spl = trace.spl
-    newtrace.model = trace.model
-    return newtrace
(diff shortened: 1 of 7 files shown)


cpfiffer:
I'm really glad to see some of this stuff updated. The particle stuff has always been a little rickety.

Anyone know why tests aren't running for this PR?

mohamed82008:
I think there is a problem with tests on all our PRs.

src/core/container.jl
```diff
@@ -81,19 +72,20 @@ Data structure for particle filters
...
 - consume(pc::ParticleContainer): return incremental likelihood
 """
-mutable struct ParticleContainer{T<:Particle, F, Tvals <: Array{T}, TlogW <: Array{Float64}}
-    model :: F
-    num_particles :: Int
-    vals  :: Tvals
-    logWs :: TlogW       # Log weights (Trace) or incremental likelihoods (ParticleContainer)
-    logE  :: Float64     # Log model evidence
-    # conditional :: Union{Nothing,Conditional} # storing parameters, helpful for implementing rejuvenation steps
-    conditional :: Nothing # storing parameters, helpful for implementing rejuvenation steps
```
yebai: Thanks for cleaning this up - this was meant to implement the rejuvenation step in the SMC2 algorithm, but the `VarInfo` mechanism turns out to be a more generic mechanism for evaluation conditional distributions (e.g. those in Gibbs).

yebai:
@devmotion The CI is now fixed on the master branch. Can you try to rebase `master` into this PR? Once all tests pass, I'm happy to merge it.

devmotion:
I rebased on master, but as observed above tests aren't running. Skimming through other PRs it seems that tests don't run for PRs from forks?

yebai:
Yep; seems that's true. I don't know how to activate CI on forks. Since you're a repo developer now, do you want to create a new branch inside this repo?

devmotion:
Sure, I can do that. I just noticed that apparently the actions are run on my fork but not on the main repo: https://github.com/devmotion/Turing.jl/actions

yebai:
> Sure, I can do that. I just noticed that apparently the actions are run on my fork but not on the main repo: https://github.com/devmotion/Turing.jl/actions

That is equivalent - happy to merge once the actions based CI passes.

devmotion:
It's not completely clear to me from the documentation, but maybe the issue for PRs could be solved by replacing lines such as 

https://github.com/TuringLang/Turing.jl/blob/87a3db946a1e63b46f57fdee460acaa7309c4db6/.github/workflows/DynamicHMC.yml#L3

with `on: [push, pull_request]`. To run the Travis CIs for all PRs, one should probably enable `Build pushed pull requests` in the [settings of Travis](https://travis-ci.org/TuringLang/Turing.jl/settings).

yebai:
Thanks, Travis CI is now turned on. I’m not sure about the actions based CI though, cause turning both push and pull request on causes duplicate builds in the past.

devmotion:
I guess we'll face the same issue with Travis CI and end up with duplicate tests for forks from the main repo. However, I think that can be avoided by adding
```yaml
branches:
  only:
    - master
```
to `.travis.yml` (at least that's what Google told me :smile:).

devmotion:
Apparently, in the same way Github Actions can be filtered: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#filtering-for-specific-branches-tags-and-paths

yebai:
That looks quite promising. Would you like to give it a try in this PR? It seems I can't push to this PR easily.

PR #967, opened 2019-11-17
Add numerically stable Poisson logpdf

itsdfish:
I was hoping to add a numerically stable version of the Poisson logpdf suggested by Kai a few months ago. I found it to be useful. So I added it to your distributions.jl file and included a test.


trappmartin:
Thanks!

PR #970, opened 2019-11-17
Simplify calculations of tables for DirichletProcess and PitmanYorProcess

devmotion:
This PR simplifies the calculations of the tables for DirichletProcess and PitmanYorProcess by avoiding redundant summations and loops over the provided cluster counts. Additionally, the explicit definition of the output arrays (and in particular its element types) ensures that the function is type-stable.

Moreover, in the sampling from the Chinese restaurant process the use of the in-place softmax function reduces allocations.

### DIFF
--- a/src/stdlib/RandomMeasures.jl
+++ b/src/stdlib/RandomMeasures.jl
@@ -64,9 +64,7 @@ Parameters:
 * `m`: Cluster counts
 
 """
-function _logpdf_table(d::AbstractRandomProbabilityMeasure, m::AbstractVector{Int})
-    throw(MethodError(_logpdf_table(), (d, m)))
-end
+function _logpdf_table end
 
 function logpdf(d::ChineseRestaurantProcess, x::Int)
     if insupport(d, x)
@@ -143,8 +141,7 @@ function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<
 
     # pre-calculations
     dα = d.α
-    z = log(sum_m) - 1 + dα
-    table_zero = log(dα) - z
+    z = log(sum_m - 1 + dα)
 
     # construct the table
     K = length(m)
@@ -157,7 +154,7 @@ function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<
             if contains_zero
                 table[i] = -Inf
             else
-                table[i] = table_zero
+                table[i] = log(dα) - z
                 contains_zero = true
             end
         else
@@ -166,7 +163,7 @@ function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<
     end
 
     if !contains_zero
-        push!(table, table_zero)
+        push!(table, log(dα) - z)
     end
 
     return table
@@ -231,7 +228,6 @@ function _logpdf_table(d::PitmanYorProcess{T}, m::AbstractVector{Int}) where {T<
     dθ = d.θ
     dd = d.d
     z = log(sum_m + dθ)
-    table_zero = log(dθ + dd * d.t) - z
 
     # construct the table
     K = length(m)
@@ -244,7 +240,7 @@ function _logpdf_table(d::PitmanYorProcess{T}, m::AbstractVector{Int}) where {T<
             if contains_zero
                 table[i] = -Inf
             else
-                table[i] = a
+                table[i] = log(dθ + dd * d.t) - z
                 contains_zero = true
             end
         else
@@ -253,7 +249,7 @@ function _logpdf_table(d::PitmanYorProcess{T}, m::AbstractVector{Int}) where {T<
     end
 
     if !contains_zero
-        push!(table, table_zero)
+        push!(table, log(dθ + dd * d.t) - z)
     end
 
     return table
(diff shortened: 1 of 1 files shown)


torfjelde:
I just did a quick scan and things look nice, but have you checked differentiability? To me the "old" code looks like it's made an explicit effort to make things work with AD, and for example the use of inplace `softmax!` won't. 

Maybe @trappmartin can chime in on whether or not we actually need to support AD in this case:)

devmotion:
Ah, good point, I haven't checked that (I just ran the regular tests of the processes, IIRC). If a specialized implementation for AD is required, then one could dispatch depending on the parameter type of the processes. I guess it would still make sense to optimize the existing code even in that case, e.g., by avoiding redundant summations and using `softmax` (I guess that should be fine and probably be slightly more performant and numerically stable than the current version).

src/stdlib/RandomMeasures.jl
```diff
@@ -134,21 +134,42 @@ function _logpdf(d::SizeBiasedSamplingProcess{DirichletProcess{T}}, x::T) where
...
+function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<:Real}
+    # compute the sum of all cluster counts
+    sum_m = sum(m)
+
+    # shortcut if all cluster counts are zero
+    iszero(sum_m) && return zeros(T, 1)
+
+    # pre-calculations
+    dα = d.α
+    z = log(sum_m) - 1 + dα
```
trappmartin: This looks wrong. It should probably be `z = log(sum_m - 1 + dα)`

src/stdlib/RandomMeasures.jl
```diff
@@ -134,21 +134,42 @@ function _logpdf(d::SizeBiasedSamplingProcess{DirichletProcess{T}}, x::T) where
...
+
+    # construct the table
+    K = length(m)
+    table = Vector{T}(undef, K)
+    contains_zero = false
+    @inbounds for i in 1:K
+        mi = m[i]
+
+        if iszero(mi)
+            if contains_zero
```
trappmartin: Maybe add a comment here, so that people know why `table[i]` will be set to `-Inf`.

src/stdlib/RandomMeasures.jl
```diff
@@ -134,21 +134,42 @@ function _logpdf(d::SizeBiasedSamplingProcess{DirichletProcess{T}}, x::T) where
...
+    # compute the sum of all cluster counts
+    sum_m = sum(m)
+
+    # shortcut if all cluster counts are zero
+    iszero(sum_m) && return zeros(T, 1)
+
+    # pre-calculations
+    dα = d.α
+    z = log(sum_m) - 1 + dα
+    table_zero = log(dα) - z
```
trappmartin: Is this pre-allocation necessary? We anyhow set only one table entry to this value.

src/stdlib/RandomMeasures.jl
```diff
@@ -199,21 +220,43 @@ function _logpdf(d::SizeBiasedSamplingProcess{PitmanYorProcess{T}}, x::T) where
...
+    sum_m = sum(m)
+
+    # shortcut if all cluster counts are zero
+    iszero(sum_m) && return zeros(T, 1)
+
+    # pre-calculations
+    dθ = d.θ
+    dd = d.d
+    z = log(sum_m + dθ)
+    table_zero = log(dθ + dd * d.t) - z
```
trappmartin: Again, do we need this pre-allocation?

src/stdlib/RandomMeasures.jl
devmotion: Oops, yes, I'll fix it.

src/stdlib/RandomMeasures.jl
devmotion: It's not required, I just thought it might be easier to see that the same value is added to the table if `m` contains a 0 (at the index of the first 0) and if it doesn't (then it's pushed to the end). Of course, I can just duplicate the line for both cases, then no precalculation is required.

src/stdlib/RandomMeasures.jl
devmotion: The intention was the same as above.

src/stdlib/RandomMeasures.jl
trappmartin: I don't have a strong opinion about this but it feels somewhat unnecessary.

devmotion: I don't have strong feelings about it either, so I just remove it.

devmotion:
I rechecked the results of the tests in `stdlib/RandomMeasures.jl` (which are commented out by default), on master, the initial commit in this PR which contains two bugs, and the updated version of this PR in which these bugs are fixed. In all cases all tests in the `distributions` testset pass, however, all other testsets `chinese restaurant processes`, `stick breaking`, and `size-based resampling` error due to a `KeyError` when calling `Libtask.consume(::Trace)`. I assume that due to these test errors (which exist on master as well and which I haven't paid too much attention to yet) I did miss the bugs in the initial commit.

trappmartin:
Thanks! I’ll take a look at the tests.
Thanks for the PR.

PR #975, opened 2019-11-18
Bump FiniteDifferences version and fix AD tests

mohamed82008:
This PR tries to fix the CI test failures. Let's see.


mohamed82008:
So it seems another source of failure on Julia 1.0 is that DynamicHMC dropped support for Julia 1.0 https://github.com/tpapp/DynamicHMC.jl/issues/106. I think we can merge this for now and release but we can also consider dropping support for Julia 1.0.

yebai:
Seems the issue is with DynamicHMC now - DHMC 2.0 no longer supports Julia 1.0. I moved the DynamicHMC CI from Travis to Github Actions, so that the failure won't affect the main Turing CI.

PR #979, opened 2019-11-20
Remove num_particles and pre-allocate children

devmotion:
This PR removes the field `pc.num_particles` since it is redundant with `length(pc) = length(pc.vals)`. Moreover, in the resampling step the arrays of children and log weights are pre-allocated.


yebai:
Looks correct to me, thanks @devmotion for the constant improvements!

devmotion:
So far my contributions are heavily biased towards particle filters, but I guess that might be due to my supervisor :smile:

PR #982, opened 2019-11-22
Simplify systematic resampling

devmotion:
This PR simplifies the implementation of systematic resampling by removing the allocation of the ordered random numbers and the cumulative sum of weights. Moreover, it adds some documentation and some checks.


devmotion:
I ran the following simple benchmark (copied from the tests) on my computer:
```julia
using Turing
using BenchmarkTools

D = [0.3, 0.4, 0.3]
num_samples = Int(1e6)

@btime Turing.Inference.resample_systematic($D, $num_samples)
```
It yields on master
```julia
  7.777 ms (8 allocations: 30.52 MiB)
```
and with this PR
```julia
  1.193 ms (2 allocations: 7.63 MiB)
```

devmotion:
To me it seems the test failure in one of the distribution tests should not be related to the changes in this PR? Even more so since it passed on all other platforms and versions?

PR #983, opened 2019-11-22
Simplify stratified resampling

devmotion:
In the same way as https://github.com/TuringLang/Turing.jl/pull/982, this PR simplifies the implementation of systematic resampling by removing the allocation of the random numbers and the cumulative sum of weights. Moreover, it adds some documentation and some checks.


devmotion:
I ran the following simple benchmark (copied from the tests) on my computer:
```julia
using Turing
using BenchmarkTools

D = [0.3, 0.4, 0.3]
num_samples = Int(1e6)

@btime Turing.Inference.resample_stratified($D, $num_samples)
```
It yields on master
```julia
  5.860 ms (5 allocations: 15.26 MiB)
```
and with this PR
```julia
  3.062 ms (2 allocations: 7.63 MiB)
```

PR #990, opened 2019-11-22
Add explicit resampler with ESS threshold

devmotion:
This PR adds a resampling method that is only executed if the estimated ESS drops below a certain threshold. I'm not completely sure about the design choices in this PR, though.

The main motivation for the changes is that currently the design of `SMC` feels a bit restrictive since it always performs ESS-based resampling. To avoid it one would have to set the threshold to 1, and even in that case the ESS would be estimated, leading to unneeded computations and allocations (since the weights used in the computation of the ESS are not reused in the resampling).

The PR also adds a keyword argument `weights` to `resample!` that allows to specify precomputed weights, if available. This change allows to reuse the weights that are computed for the estimation of the ESS in the resampling step.


src/core/container.jl
```diff
@@ -167,18 +167,16 @@ increase_logevidence(pc :: ParticleContainer, logw :: Float64) =
 function resample!(
     pc :: ParticleContainer,
     randcat :: Function = Turing.Inference.resample_systematic,
-    ref :: Union{Particle, Nothing} = nothing
+    ref :: Union{Particle, Nothing} = nothing;
+    weights = weights(pc)
```
xukai92: This overloads the function `weights` by a variable. It's better to rename one of them.

src/core/container.jl
```diff
@@ -219,3 +217,29 @@ function resample!(
 
     pc
 end
+
+struct ResampleWithESSThreshold{R, T<:Real}
```
xukai92: Maybe `T<:AbstractFloat`?

src/inference/AdvancedSMC.jl
```diff
@@ -118,11 +106,8 @@ function sample_init!(
     # create a new particle container
     spl.state.particles = pc = ParticleContainer(model, particles)
 
-    while consume(pc) !== Val{:done}
-        ess = effectiveSampleSize(pc)
-        if ess <= spl.alg.resampler_threshold * N
-            resample!(pc, spl.alg.resampler)
-        end
+    while consume(spl.state.particles) !== Val{:done}
```
xukai92: Either use `pc` here or remove the `pc` definition above.

src/core/container.jl
```diff
@@ -219,3 +217,29 @@ function resample!(
...
+end
+ResampleWithESSThreshold(resampler) = ResampleWithESSThreshold(resampler, 0.5)
+
+function resample!(
+    pc::ParticleContainer,
+    resampler::ResampleWithESSThreshold,
+    ref::Union{Particle,Nothing} = nothing;
+    weights = weights(pc)
+)
+    # compute the effective sample size ``1 / ∑ wᵢ²`` with normalized weights ``wᵢ``
```
xukai92: `# c` -> `# C` as single line comments start with captical letters.

src/core/container.jl
devmotion: Do you have any suggestions for new names? I thought `weights` is a quite self-explanatory name for both the function and the variable, and the function could still be accessed by `Turing.Inference.weights`.

src/inference/AdvancedSMC.jl
devmotion: Oh, I guess I missed that when I merged the master branch.

src/core/container.jl
xukai92: Do you think `getweights` is more informative as the function name?

src/core/container.jl
devmotion: I guess there's no harm in keeping it a bit more general.

src/core/container.jl
devmotion: IMO it's not more informative, personally I think calling it just `weights` indicates a bit more that the weights have to be computed first. But it seems to be in line with some other function names in Turing, such as `getlogp`, so maybe that's the best solution :slightly_smiling_face:

yebai:
Many thanks, @devmotion!

PR #992, opened 2019-11-23
Simplify the acceptance check in MH and update the documentation

devmotion:
This PR simplifies the acceptance check in MH by only returning if the proposal is accepted or not (the other returned value was always discarded). Moreover, the documentation of the acceptance criterion and the MH algorithm is updated. PMMH is updated as well to keep up with these changes (as far as I understand it's not working at the moment?).

To my big surprise, using `randexp()` instead of `log(rand())` in the MH acceptance criterion leads to 3 test errors. Naively I would have assumed that it might be an advantage to generate exponentially distributed random numbers directly. I added a comment about this observation.


devmotion:
It seems the test failures in the distribution test are unrelated.

PR #993, opened 2019-11-23
Implement getspace only for instances

devmotion:
According to the [Julia documentation](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type-1), one should implement methods only for either instances or types, if possibly. This PR removes unneeded implementations of `getspace` for types.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -530,7 +530,7 @@ end
 getspace(::Gibbs) = ()
 
 floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
-floatof(::Type{Any}) = Real # fallback if type inference failed
+floatof(::Type) = Real # fallback if type inference failed
 
 @inline Turing.Core.get_matching_type(spl::Turing.Sampler, vi::Turing.RandomVariables.VarInfo, ::Type{<:AbstractFloat}) = floatof(eltype(vi, spl))
 @inline Turing.Core.get_matching_type(spl::Turing.Sampler{<:Hamiltonian}, vi::Turing.RandomVariables.VarInfo, ::Type{Array{T,N}}) where {T, N} = Array{Turing.Core.get_matching_type(spl, vi, T), N}
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -527,9 +527,10 @@ end
 for alg in (:HMC, :HMCDA, :NUTS, :SGLD, :SGHMC)
     @eval getspace(::$alg{<:Any, space}) where {space} = space
 end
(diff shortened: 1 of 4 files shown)


src/inference/Inference.jl
```diff
@@ -523,21 +523,17 @@ include("../contrib/inference/AdvancedSMCExtensions.jl")
...
 end
 for alg in (:HMC, :HMCDA, :NUTS, :SGLD, :SGHMC)
     @eval getspace(::$alg{<:Any, space}) where {space} = space
-    @eval getspace(::Type{<:$alg{<:Any, space}}) where {space} = space
 end
 getspace(::Gibbs) = Tuple{}()
-getspace(::Type{<:Gibbs}) = Tuple{}()
 
 @inline floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
-@inline floatof(::Type) = Real
```
mohamed82008: This function is needed for type unstable code and when type inference fails. The input to this comes from `Core.Compiler.return_type` which returns `Any` in some cases.

mohamed82008:
Thanks for the PR. Looks good to me, other than the one comment above.

src/inference/Inference.jl
```diff
@@ -523,21 +523,18 @@ include("../contrib/inference/AdvancedSMCExtensions.jl")
...
-    @eval getspace(::Type{<:$alg{<:Any, space}}) where {space} = space
 end
-getspace(::Gibbs) = Tuple{}()
-getspace(::Type{<:Gibbs}) = Tuple{}()
+getspace(::Gibbs) = ()
 
-@inline floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
-@inline floatof(::Type) = Real
+floatof(::Type{T}) where {T <: Real} = typeof(one(T)/one(T))
+floatof(::Type{Any}) = Real # fallback if type inference failed
```
mohamed82008: I would just get rid of `Any` here. For example, `Core.Compiler.return_type` can also return `Union{}` sometimes. `Number` may also not be impossible to get as input here.

devmotion: Ah, I somehow assumed it would always return `Any` if type inference fails.

devmotion:
I don't see how the HMC reverse diff test error would be related to this PR.

mohamed82008:
Ya it's probably just a tolerance thing. Please increase the number of samples in that test and see if it passes.

devmotion:
Maybe we should just fix a seed? The test passes for all other OS and Julia versions. Probably that should be done for the distribution test that fails from time to time.

PR #994, opened 2019-11-24
Add space as type parameter to Gibbs sampler

devmotion:
This PR adds a space type parameter to the Gibbs algorithm, similar to the other algorithms. In that way the recomputation of the space can be avoided when a new sampler is constructed. Moreover, the checks of the types of the component algorithms in the constructor of the sampler are not needed anymore since the type parameters of the Gibbs algorithm are restricted to the supported algorithms.

### DIFF
--- a/src/inference/gibbs.jl
+++ b/src/inference/gibbs.jl
@@ -2,6 +2,8 @@
 ### Gibbs samplers / compositional samplers.
 ###
 
+const GibbsComponent = Union{Hamiltonian,MH,PG}
+
 """
     Gibbs(algs...)
 
@@ -20,7 +22,7 @@ end
 # Use PG for a 'v2' variable, and use HMC for the 'v1' variable.
 # Note that v2 is discrete, so the PG sampler is more appropriate
 # than is HMC.
-alg = Gibbs(HMC(0.2, 3, :v1), PG(20, 1, :v2))
+alg = Gibbs(HMC(0.2, 3, :v1), PG(20, :v2))
 ```
 
 Tips:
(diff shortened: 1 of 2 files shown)


devmotion:
Again the same (seemingly unrelated) distribution test error.

src/inference/gibbs.jl
```diff
@@ -20,19 +22,24 @@ end
...
-    function Gibbs(algs...)
-        return new{typeof(algs)}(algs)
-    end
+struct Gibbs{space, A<:Tuple{Vararg{GibbsComponent}}} <: InferenceAlgorithm
+    algs::A   # component sampling algorithms
+end
+
+function Gibbs(algs::GibbsComponent...)
+    # obtain space of sampling algorithms
+    spacearray = collect(Iterators.flatten(getspace(alg) for alg in algs))::Vector{Symbol}
```
mohamed82008: `Tuple(union(getspace.(algs)...))` is shorter and more readable.

src/inference/gibbs.jl
```diff
@@ -46,52 +53,46 @@ mutable struct GibbsState{V<:VarInfo, S<:Tuple{Vararg{Sampler}}} <: AbstractSamp
...
+        @warn("[Gibbs] extra parameters specified by samplers don't exist in model: $(setdiff(space, pvars))")
     end
 
-    # Sanity check for space
-    @assert issubset(get_pvars(model), space) "[Gibbs] symbols specified to samplers ($space) doesn't cover the model parameters ($(get_pvars(model)))"
-
-    if !(issetequal(get_pvars(model), space))
-        @warn("[Gibbs] extra parameters specified by samplers don't exist in model: $(setdiff(space, get_pvars(model)))")
+    # create tuple of samplers
+    samplers = let algs=alg.algs, model=model
```
mohamed82008: Is the `let` block here necessary?

src/inference/gibbs.jl
devmotion: That one is nice! It yields a slightly different result with duplicates being removed, but I guess that's fine here (since we do not throw an error in case of duplicates anyway). I tried to not use the splatting operator, but the more readable version seems to be even slightly faster for some simple examples I tried :+1:

src/inference/gibbs.jl
devmotion: I guess not since the type of `samplers` can not be inferred in any case. In DiffEq we had some issues with captured variables in closures leading to performance regressions, and hence I sometimes just apply the [let-block trick](https://github.com/JuliaLang/julia/issues/15276#issuecomment-297596373), even though it might not be required.

mohamed82008: I rarely have a problem with closures these days, so I suspect this is overkill. Please check it with `@code_warntype`.

devmotion: Sure, I'll just remove it. I tried some simple examples (e.g., with `gdemo_default`), and in any case the type can't be inferred (it's just `Tuple`), so it doesn't seem to make a difference.

devmotion:
Again the unrelated distribution test is failing.

yebai:
> Again the unrelated distribution test is failing.

Many thanks. It's ok to ignore the distribution test for now - will try to fix it later in a separate PR.

devmotion:
> Many thanks. It's ok to ignore the distribution test for now - will try to fix it later in a separate PR.

Oh sorry, I didn't see your comment. Hopefully https://github.com/TuringLang/Turing.jl/pull/998 fixes the issue.

Issue #989, opened 2019-11-22
Simple syntax for querying the model and posterior sample


mohamed82008:
In this issue, I propose the creation of a new macro `@logprob` to query Turing models as well as the sampled posterior. Let's take the following model as an example using the syntax of #965:
```julia
@model demo(x, y) = begin
    a ~ Normal()
    b ~ Gamma()
    y .~ Normal.(a .* x, b)
end
model = demo(rand(100), rand(100))
chain = sample(model, NUTS(10, 0.65), 10000)
```

I propose the following syntax:
- `@logprob a = 1.0, b = 2.0 | model = model` -> returns log the prior
- `@logprob a = 1.0, b = 2.0, x = rand(100), y = rand(100) | model = model` -> returns log the joint probability
- `@logprob 0.2 <= a <= 0.3, 1.0 <= b <= 1.1 | model = model, chain = chain` -> returns log the ratio of the number of samples in `chain` with `0.2 <= a <= 0.3, 1.0 <= b <= 1.1`. For discrete distributions, we can also use ` a = 2` or `a == 2` for example. Passing `model` here can be optional.
- `@logprob x = rand(10), y = rand(10) | model = model, chain = chain` returns the log likelihood of `x = rand(10), y = rand(10)` for each sample in `chain`
- `@logprob x = rand(10), y = rand(10) | model = model, a = 1.0, b = 2.0` returns the log likelihood of `x = rand(10), y = rand(10)` using `a = 1.0, b = 2.0`.

Let me know if you have comments on the syntax or if I missed any use case.

mohamed82008:
I can also provide an `@prob` macro that returns the probability instead of log the probability.

mohamed82008:
I suggest we also allow `=` and `==` to be used interchangeably in the examples above.

cpfiffer:
I like it, I think it'd be a good tool to have. I have no helpful comments, so I guess you're welcome?

mohamed82008:
I am working on it.

xukai92:
> @logprob a = 1.0, b = 2.0, x = rand(100), y = rand(100) | model = model -> returns log the joint probability

`model` is created by `demo(rand(100), rand(100))` in which `x` and `y` are already provided. It's unclear to me what the log joint is supposed to return.

devmotion:
I agree, it feels a bit strange to me. I think it would be more natural to write
```julia
@logprob a = 1.0, b = 2.0 | model = demo
```
[...]

PR #997, opened 2019-11-26
prob and logprob macros for simple model queries

mohamed82008:
This PR is based on #965. In this PR, I implemented a similar syntax to the one proposed in #989. Here is an example:
```julia
using Turing

@model demo(y, ::Type{T} = Float64) where {T} = begin
	x = Vector{T}(undef, 10)
	x .~ Normal()
	y .~ Normal.(x, 1.0)
end
chain = sample(demo(rand(10)), HMC(0.64, 1), 1000)

# log likelihood
logprob"y = rand(10) | model = demo, x = rand(10)" 

# all log likelihoods, one for each sample in chain
logprob"y = rand(10) | chain = chain"

# log prior
logprob"x = rand(10) | model = demo" 
```
The posterior and joint probabilities are next.

Note the use of a string macro which is handled just like a normal macro except there is no syntax highlighting. The main difference under the hood is that this one doesn't go through the Julia parser to parse the input expression. The precedence of `|` in the Julia parser is quite high compared to how we want it here, so I chose not to go against the flow and just used a string macro.


mohamed82008:
Just some dispatch bugs left in Inference, then I will add tests.

mohamed82008:
Here are the implemented and remaining features using the above model:
- [ ] Query the prior using `prob"x = rand(3) | model = demo"`
- [ ] Query the likelihood using `prob"y = rand(3) | x = rand(3), model = demo"`
- [ ] Query the likelihood given every sample in the chain using `prob"y = rand(3) | chain = chain, model = demo"`
- [ ] Query the joint probability using `prob"x = rand(3), y = rand(3) | model = demo"`
- [ ] Create a prior probability distribution using `prob"x | model = demo"`
- [ ] Create the likelihood probability distribution using `prob"y | x = rand(3), model = demo"`
- [ ] Create a joint probability distribution using `prob"x, y | model = demo"`
- [ ] Create the posterior predictive distribution of `y` using `prob"y | chain = chain, model = demo"`

mohamed82008:
Because this PR is already quite large, I will leave the syntax for creating a distribution from the model for another PR. For now, I will just finish the joint probability querying and update the docs, then this PR will be ready for review.

yebai:
> Because this PR is already quite large, I will leave the syntax for creating a distribution from the model for another PR. For now, I will just finish the joint probability querying and update the docs, then this PR will be ready for review.

Sounds good.

mohamed82008:
Just docs left and maybe a couple more tests.

src/Turing.jl
```diff
@@ -37,22 +37,27 @@ include("stdlib/distributions.jl")
 include("stdlib/RandomMeasures.jl")
 
 """
-struct Model{F, Targs <: NamedTuple}
+struct Model{F, Targs <: NamedTuple, Tmissings <: Val}
```
trappmartin: Could you add a line explaining the fields of Model?

src/Turing.jl
```diff
@@ -37,22 +37,27 @@ include("stdlib/distributions.jl")
...
 (model::Model)(vi) = model(vi, SampleFromPrior())
 (model::Model)(vi, spl) = model(vi, spl, DefaultContext())
 (model::Model)(args...; kwargs...) = model.f(args..., model; kwargs...)
-getmissing(model::Model) = _getmissing(model.args)
-@generated function _getmissing(args::NamedTuple{names, ttuple}) where {names, ttuple}
+
+getmissing(model::Model) = model.missings
+@generated function getmissing(args::NamedTuple{names, ttuple}) where {names, ttuple}
+    length(names) == 0 && return :(Val{()}())
     minds = filter(1:length(names)) do i
```
trappmartin: I'm confused by this syntax. Sorry for the stupid question, but what does `filter(1:length(names))` do? Is this not equivalent to `1:length(names)`?

src/Turing.jl
```diff
@@ -120,12 +125,31 @@ and parameters when running the model.
...
+    end
+
+The `LikelihoodContext` enables the computation of the log likelihood of the data when 
+running the model. `vars` can be used to evaluate the log likelihood for specific values 
+of the model's parameters. If `vars` is `nothing`, the parameter values inside the `VarInfo` will be used by default.
+"""
+struct LikelihoodContext{Tvars} <: AbstractContext
+    vars::Tvars
+end
+LikelihoodContext() = LikelihoodContext(nothing)
```
trappmartin: Cool!

src/core/compiler.jl
```diff
@@ -1,3 +1,14 @@
+struct ModelGen{Targs, F, Tdefaults} <: Function
```
trappmartin: Maybe add some doc strings.

src/core/compiler.jl
```diff
@@ -59,6 +70,22 @@ function vsym(expr::Union{Expr, Symbol})
...
 
+macro vinds(expr::Union{Expr, Symbol})
+    expr |> vinds |> esc
+end
+function vinds(expr::Union{Expr, Symbol})
+    ex = deepcopy(expr)
+    inds = Expr(:tuple)
+    (ex isa Symbol) && return inds
+    (ex.head == :ref) || throw("VarName: Mis-formed variable name $(expr)!")
+    while ex.head == :ref
```
trappmartin: Is this guaranteed to terminate?

src/core/compiler.jl
```diff
@@ -59,6 +70,22 @@ function vsym(expr::Union{Expr, Symbol})
     throw("VarName: Mis-formed variable name $(expr)!")
 end
 
+macro vinds(expr::Union{Expr, Symbol})
+    expr |> vinds |> esc
+end
+function vinds(expr::Union{Expr, Symbol})
```
trappmartin: Add doc string

src/core/compiler.jl
```diff
@@ -135,13 +162,13 @@ macro preprocess(data_vars, missing_vars, ex::Union{Symbol, Expr})
         # This branch should compile nicely in all cases except for partial missing data
         # For example, when `ex` is `x[i]` and `x isa Vector{Union{Missing, Float64}}`
         if !Turing.Core.inparams($sym, $data_vars) || Turing.Core.inparams($sym, $missing_vars)
-            $(varname(ex))
+            $(varname(ex)), $(vinds(ex))
```
trappmartin: 👍

src/core/compiler.jl
```diff
@@ -252,6 +304,19 @@ function build_model_info(input_expr)
     return model_info
 end
 
+function to_nt_expr(syms::Vector, vals = syms)
```
trappmartin: Maybe change to a more telling name? `tonamedtuple_expr` ?

src/core/prob_macro.jl
```diff
@@ -0,0 +1,186 @@
...
+macro logprob_str(str)
+	expr1, expr2 = get_exprs(str)
+    return :(Turing.Core.logprob($expr1, $expr2)) |> esc
+end
+macro prob_str(str)
+	expr1, expr2 = get_exprs(str)
+    return :(exp.(Turing.Core.logprob($expr1, $expr2))) |> esc
+end
+
+function get_exprs(str::String)
```
trappmartin: It feels like it might be necessary to add some error handling here.

trappmartin: Especially if `|` is not existing.

src/core/probmacro.jl
```diff
@@ -0,0 +1,13 @@
+function getexprs(expr)
```
trappmartin: Is this used?

src/Turing.jl
devmotion: The `do` syntax is a convenient way to use longer function definitions, this is equivalent to `filter(i -> ...., 1:length(names))`.

mohamed82008:
Thanks @trappmartin for your review. I will add the missing docstrings and comments while updating the docs, shortly.

src/Turing.jl
mohamed82008: Sure.

src/Turing.jl
mohamed82008: In this case it is not really long, but I find it more readable.

src/core/compiler.jl
mohamed82008: Sure.

src/Turing.jl
mohamed82008: Thanks!

src/core/compiler.jl
mohamed82008: Hmm, good question, it will either terminate or error eventually, assuming the input expression is finite which it is.

src/core/compiler.jl
mohamed82008: Sure.

src/core/compiler.jl
mohamed82008: Sure.

src/core/prob_macro.jl
mohamed82008: I agree.

src/core/probmacro.jl
mohamed82008: Nope, a duplicate from a previous re-write. I will remove it.

mohamed82008:
DistributionsAD-Zygote bug

mohamed82008:
Docs, docs and more docs now.

test/core/prob_macro.jl
```diff
@@ -0,0 +1,78 @@
...
+        loglike = like(mval, xval)
+        logjoint = logprior + loglike
+
+        @test logprob"m = mval | model = demo" == logprior
+        @test logprob"x = xval | m = mval, model = demo" == loglike
+        @test logprob"x = xval, m = mval | model = demo" == logjoint
+
+        varinfo = Turing.VarInfo(demo(xval))
+        @test logprob"m = mval | model = demo, varinfo = varinfo" == logprior
+        @test logprob"x = xval | m = mval, model = demo, varinfo = varinfo" == loglike
```
trappmartin: In my understanding this is equivalent to: `p(x = xval | m = mval, model = demo, varinfo = varinfo) = p(x = xval | m = mval, model = demo, x = xval) = p(x = xval | m = mval, model = demo)`. Is this correct? If so, I feel there should be a comment explaining this here.

Additional test:
* Test that `logprob"x = xval | model = demo, varinfo = varinfo" == loglike` if `m = mval` in varinfo

test/core/prob_macro.jl
```diff
@@ -0,0 +1,78 @@
...
+        # Currently, we cannot easily pre-allocate `VarInfo` for vector data
+
+        chain = sample(demo(xval), HMC(0.5, 1), iters)
+        chain2 = Chains(chain.value, chain.logevidence, chain.name_map, NamedTuple())
+        lps = like.([[chain["m[$i]"].value[j] for i in 1:n] for j in 1:iters], Ref(xval))
+        @test logprob"x = xval | chain = chain" == lps
+        @test logprob"x = xval | chain = chain2, model = demo" == lps
+        @test logprob"x = xval | chain = chain, varinfo = varinfo" == lps
```
trappmartin: `p(x = xval | m = chain[i, :m], x = xval)` ? Does it make sense to combine the4 arguments for varinfo and chain? What would happen if varinfo contains `m = 0`? `p(x = xval | m = chain[i, :m], m = 0)`?

test/core/prob_macro.jl
trappmartin: * Test that `logprob"x = xval | model = demo, m = 0, varinfo = varinfo"` throws an error if `m = mval != 0` in varinfo

test/core/prob_macro.jl
```diff
@@ -0,0 +1,78 @@
...
+        @test logprob"x = xval | chain = chain2, model = demo, varinfo = varinfo" == lps
```
trappmartin: See above.

test/core/prob_macro.jl
```diff
@@ -0,0 +1,78 @@
...
+        loglike = logpdf(Normal(mval, 1), xval)
+        logjoint = logprior + loglike
+
+        @test logprob"m = mval | model = demo" == logprior
+        @test logprob"m = mval | x = xval, model = demo" == logprior
+        @test logprob"x = xval | m = mval, model = demo" == loglike
+        @test logprob"x = xval, m = mval | model = demo" == logjoint
+
+        varinfo = Turing.VarInfo(demo(xval))
+        @test logprob"m = mval | model = demo, varinfo = varinfo" == logprior
```
trappmartin: See comments on test for vectorised `logprob`.

test/core/prob_macro.jl
mohamed82008: Well in the above syntax, `varinfo` is only for pre-allocation and type stability. If this is done for many values of `xval` in a loop, having a single `varinfo` is necessary for performance. I will add this to the docs. So the value inside `varinfo` is overwritten.

test/core/prob_macro.jl
mohamed82008: Again `varinfo` is just for pre-allocation. This is just in case the user didn't save `varinfo` in `chain` for some reason (even though the default now is to save it) and wants to pass their own for some reason.

test/core/prob_macro.jl
trappmartin: I see. In that case I find the syntax confusing. Would it be possible to do this differently. Currently it implies to me that we condition on VarInfo, which we don’t actually do if I understand it correctly.

mohamed82008: True, do you have an alternative in mind?

test/core/prob_macro.jl
trappmartin: Ok, but what happens in the case of `p(x = xval | model = demo, chain = chain, m = mval)`?

test/core/prob_macro.jl
mohamed82008: Perhaps use `;` instead of `,`?

mohamed82008: I take that back :)

test/core/prob_macro.jl
mohamed82008: The `mval` is ignored I believe.

mohamed82008: What is the preferred behavior here?

trappmartin: I would say it’s not well defined what happens. But I think a sensible approach would be that `m = mval` overwrites what is in `chain`.

trappmartin:
This PR feels like a Christmas present. 🎁

test/core/prob_macro.jl
mohamed82008: I say we deal with these cases in a future PR. This use of the syntax is likely to be very rare anyways. I won't even document it in the getting started, just the advanced section of the docs. Sounds good?

test/core/prob_macro.jl
mohamed82008: Let's open an issue and deal with this in a separate PR. What do you think?

trappmartin:
I’m happy to approve once we have the issues which discuss the special cases.

test/core/prob_macro.jl
mohamed82008: https://github.com/TuringLang/Turing.jl/issues/1038

test/core/prob_macro.jl
mohamed82008: https://github.com/TuringLang/Turing.jl/issues/1039

mohamed82008:
Some spacings are off, and the docs need an update. Then a rebase and we should be ready for a merge. Should be hopefully done by tomorrow. Would be cool to merge this on Christmas day!

mohamed82008:
If tests pass and there are no further comments, this is ready for merge.

mohamed82008:
The error is due to a Pkg bug. The only workarounds I know are to either limit our MCMCChains compat to pre-1.0 or to get a new release for CmdStan that supports MCMCChains 1.0.

mohamed82008:
https://github.com/StanJulia/CmdStan.jl/issues/76

mohamed82008:
We can also just drop CmdStan tests for now, or merge anyways and hope for a new CmdStan release soon.

cpfiffer:
We could also drop the MCMCChains compat to 0.4, not much changed between 0.4 and 1.0. Then we can keep the CmdStan tests in.

mohamed82008:
We can do that but that won't fix the issue. There is no version of CmdStan that supports MCMCChains 1.0 yet. My understanding is that Pkg is struggling to downgrade MCMCChains from 1.0 to the perfectly compatible 0.4 because of a bug. So we have to either stop Pkg from thinking MCMCChains 1.0 is valid to begin with (by limiting our MCMCChains compat in Turing to pre-1.0) or provide a version of CmdStan that supports MCMCChains 1.0 and hence Pkg doesn't need to downgrade anything. This is my best understanding of the dependency error we have.

mohamed82008:
> We could also drop the MCMCChains compat to 0.4

Oh I think I misread your comment. If you mean to only support MCMCChains 0.4 then yes this should fix the error IIUC.

mohamed82008:
Pkg error resolved, tests are running.

trappmartin:
Looks good so far

mohamed82008:
As always @goedman comes and saves the day, thanks Rob!

trappmartin:
Looks like this can be merged.

yebai:
Thanks, @mohamed82008!

Issue #991, opened 2019-11-23
Elliptical slice sampling


devmotion:
Some time ago I wrote down a quick and rough implementation of the [elliptical slice sampling algorithm](http://proceedings.mlr.press/v9/murray10a/murray10a.pdf) in the Julia package [EllipticalSliceSampling.jl](https://github.com/devmotion/EllipticalSliceSampling.jl) (actually, recently I rewrote it almost completely but I haven't pushed the changes yet). Since it's a quite simple algorithm I thought it would be fun to implement an elliptical slice sampler for Turing.

I've managed to put together something that seems to work on a small and simple example. However, since I'm quite new to Turing and still don't understand all internals, I'm pretty sure my draft could be improved. Honestly, it took me quite some time to get a rough feeling for which types should be implemented and which methods should be overloaded (I think https://github.com/TuringLang/Turing.jl/issues/895#issuecomment-530942507 was the most helpful resource I could find - thank you @cpfiffer for this write-up!). It was also surprisingly complicated to figure out how to sample from the prior and how to evaluate the log-likelihood (hopefully I actually managed to do it somewhat correctly in the end :smile: ).

I came up with the following implementation: https://gist.github.com/devmotion/6bab2561a9c340bac03c50b3fec73441#file-ess-jl Basically I tried to copy and adapt the MH implementation. However, I don't know if anything important is missing or wrong, or if some parts are not needed at all. I would be very happy if someone with a better understanding of Turing could comment on that! I'm also wondering if there is a way to avoid the repeated allocations caused by sampling from the prior? In the non-Turing implementation I used a cache to avoid these. One thing one would want to add is a check that the prior distribution is a uni- or multivariate normal distribution. Where would be the best place to do that? In the constructor of the sampler? Or in the implementation of `assume`? 

For the simple model 
```julia
@model gdemo(x) = begin
    m ~ Normal(0, 1)
    x ~ MvNormal(fill(m, length(x)), sqrt(σ²) * I)
end
```
and 10 random observations (you can check the [code](https://gist.github.com/devmotion/6bab2561a9c340bac03c50b3fec73441#file-demo-jl)) the sampler produced the following MCMC chain of 20000 samples:
![demo](https://user-images.githubusercontent.com/26102/69498300-b8910480-0ee6-11ea-92c3-174b7f09bf9a.png)
[...]

PR #1000, opened 2019-11-27
Add elliptical slice sampling algorithm

devmotion:
This is an updated version of the ESS in https://github.com/TuringLang/Turing.jl/issues/991. I noticed that the bias visible in the plots in https://github.com/TuringLang/Turing.jl/issues/991 is real and the mean of the MCMC chain does not converge to the true posterior mean. The problem was that I computed the joint probabilities although ESS requires the likelihood. I tried to fix this by changing the assume function. Moreover, I included some checks for normality and allowed only one parameter.

The example in the original issue and two other examples included in the tests seem to converge now, but the `gdemo` example still fails. I'm not sure why this is the case and assume there is some problem with how ESS is integrated in the Gibbs sampler - I guess I forgot to implement something related to Turing's internals or implemented it incorrectly.

You can check the MATLAB implementation from Ian Murray, one of the original authors, for a comparison: https://homepages.inf.ed.ac.uk/imurray2/pub/10ess/elliptical_slice.m


src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+    model::Model,
+    spl::Sampler{<:ESS},
+    ::Integer,
+    ::Transition;
+    kwargs...
+)
+    # recompute log-likelihood in logp
+    vi = spl.state.vi
+    if spl.selector.tag !== :default
+        runmodel!(model, vi, spl)
```
devmotion: I'm not sure - when should I use `runmodel!(...)` and when just `model(...)`? And do I have to pass `spl` to `runmodel!` here? It's not passed in the MH implementation.

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+
+    # obtain previous sample
+    f = copy(vi[spl])
+
+    # sample log-likelihood threshold for the next sample
+    threshold = getlogp(vi) - randexp(rng)
+
+    # sample from the prior
+    runmodel!(model, vi, spl)
+    ν = copy(vi[spl])
```
devmotion: Again, is `runmodel!` correct here? I just want to obtain a new sample from the prior.

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+    θₘₐₓ = θ
+
+    while true
+        # compute proposal
+        sinθ, cosθ = sincos(θ)
+        @. vi[spl] = f * cosθ + ν * sinθ
+
+        # recompute log-likelihood and check if threshold is reached
+        resetlogp!(vi)
+        model(vi, spl)
```
devmotion: Is this the correct way to just recompute the log-likelihood for the updated sample? Is `resetlogp!` needed?

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+
+    return Transition(spl)
+end
+
+isnormal(dist) = false
+isnormal(::Normal) = true
+isnormal(::NormalCanon) = true
+isnormal(::AbstractMvNormal) = true
+
+function assume(spl::Sampler{<:ESS}, dist::Distribution, vn::VarName, vi::VarInfo)
```
devmotion: Is the implementation of `assume` correct in the sense that I will get the log-likelihood for my parameter of interest?

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+    space = getspace(spl)
+    if space === () || space === (vn.sym,)
+        isnormal(dist) ||
+            error("[ESS] does only support normally distributed prior distributions")
+
+        r = rand(dist)
+        vi[vn] = vectorize(dist, r)
+        setgid!(vi, spl.selector, vn)
```
devmotion: Is that the correct way to handle the Gibbs sampler? I just copied this from the MH sampler.

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+        return r, zero(Base.promote_eltype(dist, r))
```
devmotion: I'm a bit that this will lead to type instabilities, since I guess the check `space === (vn.sym,)` can not be compiled away. But it seems also wasteful to actually compute the logpdf and then throw it away, in particular if the prior distribution is high dimensional.

test/inference/ess.jl
```diff
@@ -0,0 +1,44 @@
...
+    @numerical_testset "ESS inference" begin
+        Random.seed!(1)
+        alg = ESS()
+        chain = sample(demo_default, alg, 5_000)
+        check_numerical(chain, [:m], [0.8], atol = 0.1)
+
+        Random.seed!(100)
+        alg = Gibbs(CSMC(15, :s), ESS(:m))
+        chain = sample(gdemo_default, alg, 5_000)
+        @test_broken check_numerical(chain, [:s, :m], [49/24, 7/6], atol=0.1)
```
devmotion: This test still fails.

src/inference/ess.jl
cpfiffer: I would default to using `runmodel!`, since it resets the `logp` and then runs the model. If your sampler also tracks the number of times the model has been evaluated, `runmodel!` will also update it.

cpfiffer: Also if you only want to sample from the prior you don't have to pass `spl` to the `runmodel!` call. If you have custom `observe`/`assume` behavior then you should pass the sampler in.

src/inference/ess.jl
cpfiffer: In this case you can just use `runmodel!(model, vi)`

src/inference/ess.jl
cpfiffer: If you consolidate these to `runmodel!` you don't need `resetlogp!`. You're essentially doing all of that function in two lines right here, it would be slightly more idiomatic to use `runmodel!`.

src/inference/ess.jl
```diff
@@ -0,0 +1,132 @@
...
+isnormal(dist) = false
+isnormal(::Normal) = true
+isnormal(::NormalCanon) = true
+isnormal(::AbstractMvNormal) = true
+
+function assume(spl::Sampler{<:ESS}, dist::Distribution, vn::VarName, vi::VarInfo)
+    space = getspace(spl)
+    if space === () || space === (vn.sym,)
+        isnormal(dist) ||
+            error("[ESS] does only support normally distributed prior distributions")
```
cpfiffer:
```suggestion
            error("[ESS] only supports normally distributed prior distributions")
```

src/inference/ess.jl
cpfiffer: Looks correct to me.

src/inference/ess.jl
devmotion: OK, thanks! So since I want to get the log-likelihood (and hence have to use my custom assume function, haven't I?) I have to specify `spl`. If I would be interested only in the log joint probability (as, e.g., in the MH algorithm) that would not be necessary, it seems?

src/inference/ess.jl
devmotion: Again, I need `spl` since I want the log likelihood?

src/inference/ess.jl
devmotion: OK, thanks!

cpfiffer:
I believe that's correct, yes.

src/inference/ess.jl
cpfiffer: I believe that's correct, yes.

src/inference/ess.jl
devmotion: Thanks! Another thing to note here: actually the current implementation is only correct for Gaussian distributions with zero means (as assumed in the original algorithm). Hence that could be checked here as well.

However, I think it's more useful to just perform the change of variables and use a modified version of the original algorithm, as done in https://github.com/devmotion/EllipticalSliceSampling.jl/blob/e10d21da2096717068c10877522ba19e59bb6479/src/iterator.jl#L43-L44. What's the most idiomatic way to access the mean of the distribution in the `step!` function?

src/inference/ess.jl
devmotion: Good :+1: And you don't know of any more idiomatic way to compute the log-likelihood than what I've done in the `assume` function?

src/inference/ess.jl
cpfiffer: As far as I'm aware we don't have anything to do that yet other than manually extracting parameters and distributions from `vi`, which is kind of a pain. If this is something that would be useful here I think we could probably extend some kind of convenience function for accessing distributions ahead of `observe`/`assume`.

src/inference/ess.jl
cpfiffer: I think this is as close to idiomatic as it gets in Turing, so you're good.

src/inference/ess.jl
```diff
@@ -0,0 +1,141 @@
...
+    vi = spl.state.vi
+    if spl.selector.tag !== :default
+        runmodel!(model, vi, spl)
+    end
+
+    # obtain mean of distribution
+    vns = _getvns(vi, spl)
+    length(vns) == 1 || error("[ESS] does only support one parameter")
+    dist = getdist(vi, vns[1][1])
+    μ = vectorize(dist, mean(dist))
```
devmotion: Are there better ways to retrieve the mean of the prior distribution of the variable of interest? BTW, IMO having to write `vns[1][1]` is quite unintuitive.

cpfiffer: Not that I know of. @mohamed82008, do you know of a quicker way to get to a variable's distribution in a `VarInfo` outside of the `observe`/`assume` pattern?

src/inference/ess.jl
mohamed82008: `space === (vn.sym,)` should compile away IIUC.

src/inference/ess.jl
mohamed82008: `Turing.Core.RandomVariables.getdist(vi, vn)`

mohamed82008: Oh he is already using `getdist`.

src/inference/ess.jl
mohamed82008: If you just want the likelihood, you might want to wait for the compiler3.0 PR which makes the likelihood computation easier and more efficient.

src/inference/ess.jl
mohamed82008: `assume` is for parameters, this is technically the log prior, not likelihood.

devmotion:
I updated the PR and changed in particular the implementation of `assume` and `observe` (basically, they are not doing anything special anymore apart from not evaluating the `logpdf` of the prior distribution since it is never needed. Now MCMC converges for both models
```julia
@model demo(x) = begin
    m ~ Normal(0, 1)
    x ~ MvNormal(fill(m, length(x)), sqrt(σ²))
end
```
and
```julia
@model gdemo(x, y) = begin
    s ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s))
    x ~ Normal(m, sqrt(s))
    y ~ Normal(m, sqrt(s))
    return s, m
end
```
successfully. I created plots of the trajectories and the true (dashed blue) and approximated (orange) posterior ([code](https://gist.github.com/devmotion/601d95112df0920cdb9098cd2f2942e2)):

ESS (demo): demo_ESS
NUTS (demo): demo_NUTS

ESS (gdemo): gdemo_ESS
HMC (gdemo): gdemo_HMC

src/core/RandomVariables.jl
```diff
@@ -186,7 +186,7 @@ function VarInfo(old_vi::UntypedVarInfo, spl, x::AbstractVector)
     return new_vi
 end
 function VarInfo(old_vi::TypedVarInfo, spl, x::AbstractVector)
-    md = newmetadata(old_vi.metadata, getspaceval(spl), x)
```
trappmartin: This seems unrelated to the actual PR. Is this change necessary?

src/inference/ess.jl
```diff
@@ -0,0 +1,134 @@
...
+
+    # sample log-likelihood threshold for the next sample
+    threshold = getlogp(vi) - randexp(rng)
+
+    # sample from the prior
+    ν = vectorize(dist, rand(rng, dist))
+
+    # sample initial angle
+    θ = 2 * π * rand(rng)
+    θₘᵢₙ = θ - 2 * π
```
trappmartin: I would suggest to not use too many Unicode characters. It makes maintaining the code more difficult and doesn't seem necessary here.

src/inference/ess.jl
```diff
@@ -0,0 +1,134 @@
...
+    θₘₐₓ = θ
```
trappmartin: Same as above.

src/inference/ess.jl
```diff
@@ -0,0 +1,134 @@
...
+
+    while true
+        # compute proposal and apply correction for distributions with nonzero mean
+        sinθ, cosθ = sincos(θ)
+        a = 1 - (sinθ + cosθ)
+        vi[vn] = @. f * cosθ + ν * sinθ + μ * a
+
+        # recompute log-likelihood and check if threshold is reached
+        runmodel!(model, vi, spl)
+        if getlogp(vi) > threshold
```
trappmartin: I feel it would be good add a max tries condition here, to ensure the while loop will always terminate. You could give a warning if this occurs and return the initial theta.

src/inference/hmc.jl
```diff
@@ -488,7 +488,7 @@ function observe(
     value,
     vi::VarInfo,
 )
-    return observe(nothing, d, value, vi)
```
trappmartin: Why did this change?

src/inference/hmc.jl
```diff
@@ -497,7 +497,7 @@ function dot_observe(
     value::AbstractArray,
     vi::VarInfo,
 )
-    return dot_observe(nothing, ds, value, vi)
```
trappmartin: Why did this change?

src/inference/mh.jl
```diff
@@ -159,7 +159,7 @@ function assume(spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi::VarInfo
 end
 
 function observe(spl::Sampler{<:MH}, d::Distribution, value, vi::VarInfo)
-    return observe(nothing, d, value, vi)  # accumulate pdf of likelihood
```
trappmartin: Why did this change?

src/inference/mh.jl
```diff
@@ -168,5 +168,5 @@ function dot_observe(
     value,
     vi::VarInfo,
 )
-    return dot_observe(nothing, ds, value, vi) # accumulate pdf of likelihood
```
trappmartin: Why did this change?

src/inference/ess.jl
mohamed82008: The `LikelihoodContext` is now the idiomatic way in Turing to compute the log likelihood.

src/inference/ess.jl
mohamed82008: Please use the `LikelihoodContext` instead.

mohamed82008:
I would suggest keeping lateral changes to another PR. Also if you are trying to compute the log likelihood, I recommend overloading `tilde` and `dot_tilde` for the ESS sampler with `ctx::DefaultContext` and call the `tilde` or `dot_tilde` functions using the `spl::SampleFromPrior` and `ctx::LikelihoodContext`. This will significantly simplify this PR.

src/inference/hmc.jl
devmotion: That was fixed in https://github.com/TuringLang/Turing.jl/pull/995, but when I merged the master branch I noticed that apparently it was reverted by the compiler change https://github.com/TuringLang/Turing.jl/pull/965. However, I guess it's better to move that fix to a separate PR.

src/inference/hmc.jl
devmotion: Same here.

src/inference/mh.jl
devmotion: And here.

src/inference/mh.jl
devmotion: And here.

src/core/RandomVariables.jl
devmotion: Some of the changes in RandomVariables.jl are needed but I'll move the rest to a separate PR.

trappmartin: Thanks!

src/inference/hmc.jl
trappmartin: Yes, that would be great and make PR more clear. Thanks!

src/inference/ess.jl
devmotion: I just tried to use the notation in the original paper to make it easier to understand, and assumed that it's easier to read with Unicode characters. I don't have a strong opinion on this, so I'll change it to `theta`, `thetamin`, `pi` etc.

src/inference/ess.jl
trappmartin: Or adjust the threshold instead of returning the initial theta. Whatever is more sensible.

src/inference/ess.jl
trappmartin: Thanks. We had a lot of Unicode in the old NUTS sampler for the same reasons. This turned out to be harder to maintain than using clear variable names. It’s often easier to understand what is happening if the variable names are telling, which is not necessarily the case for the notation used in the paper.

trappmartin: BTW: I’m fine with the \theta but I referred to the variable you assign a value to.

src/inference/ess.jl
devmotion: I'm not sure, both approaches feel wrong since they seem to impact the correctness of the sampling algorithm. I think, either one could error if the threshold is not reached within a certain number of iterations or one could loop until the threshold is reached but output a warning after a certain number of iterations.

trappmartin: If I understand it correctly, return the initial theta would simply mean you rejected all proposals. This should still be correct. 

Is it guaranteed that the threshold will be reached? I don’t see why that should be the case. Anyways, throwing an error sounds sensible.

trappmartin: Disclaimer: I read the paper a while ago, so please ignore my comments if they don’t make sense.

cpfiffer: I think there should be an error and not a warning, just to make it very obvious that the max iteration limit needs to be adjusted or that there is some numerical issue with the inference (if that actually happens with ESS).

devmotion: > I feel it would be good add a max tries condition here, to ensure the while loop will always terminate.

I shouldn't have commented on that while being on a flight and without checking it more carefully - actually by definition the loop will always terminate. The old theta satisfies the exit condition and hence if the bracket is shrunk enough a theta close/equal to the old theta will be picked as the next theta (but the chain might be stuck and mix slowly). Due to this behaviour ESS can be inefficient (see, e.g., some comments in the original paper and https://arxiv.org/pdf/1210.7477.pdf) but I'm not sure we should address this in our implementation, since the sampling step is guaranteed to terminate. Murray's MATLAB code does not impose a maximum number of iterations, and it feels more like the main issue would be that the chain mixes slowly.

devmotion:
> Also if you are trying to compute the log likelihood, I recommend overloading tilde and dot_tilde for the ESS sampler with ctx::DefaultContext and call the tilde or dot_tilde functions using the spl::SampleFromPrior and ctx::LikelihoodContext. This will significantly simplify this PR.

If I understand you correctly, then you suggest using
```julia
function tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, left, vi)
    if left isa VarName && left in getspace(sampler)
        return tilde(LikelihoodContext(), SampleFromPrior(), right, left, vi)
    else
        return tilde(ctx, SampleFromPrior(), right, left, vi)
    end
end

function dot_tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, left, vn::VarName, vi)
    if vn in getspace(sampler)
        return dot_tilde(LikelihoodContext(), SampleFromPrior(), right, left, vn, vi)
    else
        return dot_tilde(ctx, SampleFromPrior(), right, left, vn, vi)
    end
end
```
instead of the `assume` and `observe` implementation. Everything else would be unaffected it seems, so I'm not sure if this is actually the significant simplification you're referring to. I still think it would be nice to use and it seems to work as well.

However, in general it seems the approach of sampling from the prior and computing the proposals in the `step!` function is problematic since it can't cope with arrays of distributions. I'm not sure what would be the best way to handle this, it seems ideally one would like to use `runmodel!` also for sampling from the prior - in that case, however, one can't forward to `SampleFromPrior` since that sampler does not overwrite existing variables. Maybe one solution would be to always sample the variables of interest in the `DefaultContext` and use `ctx::LikelihoodContext` instead of `ctx::DefaultContext` in the examples above. But even in that case I would have to retrieve the mean of all different distributions to correct for them...

cpfiffer:
Perhaps it would be nice if we exposed something similar to `vi[spl]` that returned a vector of distributions -- would this help at all?

mohamed82008:
I meant something like this (only valid after #997 goes in).
```julia
# assume
function tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, vn::VarName, inds, vi)
    return tilde(LikelihoodContext(), SampleFromPrior(), right, vn, inds, vi)
end
# observe
function tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, left, vi)
    return tilde(LikelihoodContext(), SampleFromPrior(), right, left, vi)
end

# dot_assume
function dot_tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, left, vn::VarName, inds, vi)
    return dot_tilde(LikelihoodContext(), SampleFromPrior(), right, left, vn, inds, vi)
end
# dot_observe
function dot_tilde(ctx::DefaultContext, sampler::Sampler{<:ESS}, right, left, vi)
    return dot_tilde(LikelihoodContext(), SampleFromPrior(), right, left, vi)
end
```

This takes care of the `assume`, `observe`, `dot_assume` and `dot_observe` for you. It's also clearer than having to define those functions ourselves.

mohamed82008:
> Perhaps it would be nice if we exposed something similar to vi[spl] that returned a vector of distributions -- would this help at all?

If it's helpful, we can have a wrapper around `vi` that gives us the distributions when we use `getindex`. For example, `dists(vi) isa DistsWrapper` and `dists(vi)[spl]` gives us what we want. `dists(vi)[vn]` also does the obvious.

src/inference/ess.jl
trappmartin: I wouldn’t address it if it is guaranteed to terminate. This was my only concern.

trappmartin: Thanks for clarifying!

devmotion:
> I meant something like this 

Ah OK, but then it seems it's basically what I outlined above. Your example wouldn't work, however, since the loglikelihood should be computed only based on the variable of interest. But it seems https://github.com/TuringLang/Turing.jl/pull/997 would provide a simpler way to achieve what I sketched above by specifying the variables of interest `vns` in `LikelihoodContext(vns)`?

yebai:
@mohamed82008 This PR contains some modification to `RandomVariables`. It might be useful to sync these changes with DynamicPPL.

mohamed82008:
@yebai sure.

mohamed82008:
@devmotion in your implementation, the log prior will be counted in for the random variables that under the ESS sampler. I assume this is your intention.

devmotion:
> @devmotion in your implementation, the log prior will be counted in for the random variables that are not under the ESS sampler. I assume this is your intention. 

Yes, that's exactly my intention. Basically, I'm interested in the log joint minus the log prior of the random variables that are sampled by the ESS sampler. A motivating example is the hierarchical model
```julia
@model demo(x) = begin
    m ~ Normal()
    k ~ Normal(m, 0.5)
    x ~ Normal(k, 0.5)
end
```
in which we might want to sample from the posterior distribution of `m` and `k` using a Gibbs sampler with ESS for both variables. In this model, if `k` is given, the log likelihood of `x` (as evaluated by Turing's `LikelihoodContext` without the log priors) is independent of the choice of `m`. Clearly, that's not useful for evaluating which sample on the elliptical slice to accept, but instead we are interested in how the likelihood of `k` changes with the choice of `m`. Hence we have to evaluate the log priors of all variables that are not under the ESS sampler since they might be affected by a change in `m`.

yebai:
Great work, many thanks @devmotion!

PR #1003, opened 2019-11-27
[WIP] Particle gibbs extension

donhausk:
Please read trough the code and propose tests/ corrections. Note that the AdvancedPS package is not yet merged. This is a Work in Progress pull request!


Project.toml
```diff
@@ -1,64 +0,0 @@
-name = "Turing"
```
devmotion: `Project.toml` should not be deleted :stuck_out_tongue:

src/Turing.jl
```diff
@@ -1,202 +1,202 @@
...
-
-        turnprogress,           # debugging
-
-        Flat,
-        FlatPos,
-        BinomialLogit,
-        VecBinomialLogit,
-        OrderedLogistic,
-        LogPoisson
+        ##############
```
devmotion: I guess the indentation changes are not intended? AFAIK the standard in Julia (and in Turing it seems) is to not indent the top level inside a module.

src/core/RandomVariables.jl
```diff
@@ -1252,6 +1255,29 @@ function setorder!(mvi::TypedVarInfo, vn::VarName{sym}, index::Int) where {sym}
     return mvi
 end
 
+"""
+`getorder(vi::VarInfo, vn::VarName)`
+
+Some proper checks are missing!!
+"""
+
```
devmotion: The documentation should go right above the function definition (without additional lines in between). Also the standard in Julia is to indent the first line in the documentation block with 4 spaces.

src/inference/AdvancedSMC.jl
```diff
@@ -327,153 +533,38 @@ function observe( spl::Sampler{A},
...
-            sample > m &&
-                error("sample could not be selected (are the weights normalized?)")
-
-            # update the cumulative sum of weights (scaled by `n`)
-            v += n * weights[sample]
-        end
-
-        # save the next sample
-        samples[i] = sample
+function get_srvndbs(spl::Sampler{A}) where A<:Union{PG,SMC}
```
devmotion: Just a general comment: The `where` clause is not needed here, it seems. According to the [documentation](https://docs.julialang.org/en/v1/manual/style-guide/#Don't-use-unnecessary-static-parameters-1) one should try to avoid unnecessary static parameters, if possible. The better alternative seems to be
```julia
function get_srvndbs(spl::Sampler{<:Union{PG,SMC}})
```

However, in general I doubt that this function is actually needed. It seems you could just use an anonymous function and replace the lines above with, e.g.,
```julia
particles.manipulators["set_retained_vns_del_by_spl!"] = x -> set_retained_vns_del_by_spl!(x, spl)
```
?

devmotion:
Nice, I'm looking forward to the refactored version and the PGAS implementation :slightly_smiling_face: I haven't studied the code carefully yet, and just commented on some things that I noticed while skimming through the changes. Hope that's helpful!

src/Turing.jl
trappmartin: Yes, the standard is no indent.  Maybe this was done automatically by the editor? We also have style guidelines which might be useful to look at: https://turing.ml/dev/docs/contributing/style-guide

yebai:
I'll open a new PR transferring the particle sampler implementations into [`AdvancedPS.jl`](https://github.com/TuringLang/AdvancedPS.jl/)

PR #1007, opened 2019-12-02
Migrate to AbstractMCMC, add interface manual

cpfiffer:
This PR moves Turing over to using AbstractMCMC as the home of the interface methods. I have also added a guide on the use of AbstractMCMC's interface methods (a modified version of the one in #920), but currently it does not include any details on how to link up a foreign algorithm to Turing. I will follow up with that document in a separate PR after #997 and #965 go through, since they seem to be providing a couple of useful tools I want to be able to point people at.

One thing to note is that I have removed the `<:Sampleable` typing on model structs in favor of a much simpler `AbstractModel`, since we have not been using `Sampleable` and it's a bit of a headache to work with.


docs/src/using-turing/interface.md
```diff
@@ -0,0 +1,321 @@
...
+Note that we only have to do this because we are not yet integrating the sampler with Turing -- Turing has a very sophisticated modelling engine that removes the need to define custom model structs. 
+
+```julia
+# Define a model type. Stores the log density function.
+struct DensityModel{F<:Function} <: AbstractModel
+    ℓπ :: F
+end
+
+# Default density constructor.
+DensityModel(π::Function, data::T) where T = DensityModel{VariateForm, ValueSupport, T}(π, data)
```
devmotion: That seems wrong and does not match the `struct` definition above? Moreover, the static parameter could/should be removed (https://docs.julialang.org/en/v1/manual/style-guide/#Don't-use-unnecessary-static-parameters-1).

docs/src/using-turing/interface.md
```diff
@@ -0,0 +1,321 @@
...
+```julia
+# Create a very basic Transition type, only stores the 
+# parameter draws and the log probability of the draw.
+struct Transition{T<:Union{Vector{<:Real}, <:Real}, L<:Real} <: AbstractTransition
+    θ :: T
+    lp :: L
+end
+
+# Store the new draw and its log density.
+Transition(model::M, θ::T) where {M<:DensityModel, T} = Transition(θ, ℓπ(model, θ))
```
devmotion: The static parameters are not needed.

docs/src/using-turing/interface.md
```diff
@@ -0,0 +1,321 @@
...
+    Transition(model, θ + rand(spl.proposal))
+propose(spl::MetropolisHastings, model::DensityModel, t::Transition) = propose(spl, model, t.θ)
+
+# Calculates the probability `q(θ | θcond)`, using the proposal distribution `spl.proposal`.
+q(spl::MetropolisHastings, θ::Real, θcond::Real) = logpdf(spl.proposal, θ - θcond)
+q(spl::MetropolisHastings, θ::Vector{<:Real}, θcond::Vector{<:Real}) = logpdf(spl.proposal, θ - θcond)
+q(spl::MetropolisHastings, t1::Transition, t2::Transition) = q(spl, t1.θ, t2.θ)
+
+# Calculate the density of the model given some parameterization.
+ℓπ(model::DensityModel, θ::T) where T = model.ℓπ(θ)
```
devmotion: Same here, `T` could be removed.

docs/src/using-turing/interface.md
```diff
@@ -0,0 +1,321 @@
...
+# A basic chains constructor that works with the Transition struct we defined.
+function bundle_samples(
+    rng::AbstractRNG, 
+    ℓ::DensityModel, 
+    s::MetropolisHastings, 
+    N::Integer, 
+    ts::Vector{T}; 
+    param_names=missing,
+    kwargs...
+) where {ModelType<:AbstractModel, T<:AbstractTransition}
```
devmotion: `ModelType` seems wrong and should be removed, I guess, and `T` is not needed.

src/inference/Inference.jl
```diff
@@ -140,28 +139,28 @@ const TURING_INTERNAL_VARS = (internals = [
...
-function Interface.sample(
+function AbstractMCMC.sample(
     rng::AbstractRNG,
     model::ModelType,
     alg::AlgType,
     N::Integer;
     kwargs...
 ) where {
-    ModelType<:Sampleable,
+    ModelType<:AbstractModel,
```
devmotion: It seems all static parameters could/should be removed?

src/inference/Inference.jl
```diff
@@ -140,28 +139,28 @@ const TURING_INTERNAL_VARS = (internals = [
...
-function Interface.sample(
+function AbstractMCMC.sample(
     model::ModelType,
     alg::AlgType,
     N::Integer;
     resume_from=nothing,
     kwargs...
 ) where {
-    ModelType<:Sampleable,
+    ModelType<:AbstractModel,
```
devmotion: Same here.

PR #1014, opened 2019-12-08
Increment version number

cpfiffer:


cpfiffer:
My bad. This should be good to go. I'll push a minor release out once this goes through.

Issue #1024, opened 2019-12-15
Recommend memoization in docs for expensive functions in models when Gibbs sampling


mohamed82008:
The problem of recomputing expensive functions when performing Gibbs sampling can be mostly overcome if we use memoization packages like https://github.com/JuliaCollections/Memoize.jl or https://github.com/marius311/Memoization.jl. The expensive function is annotated with `@memoize` to create a cache for it. When called, the function then checks if the inputs are the same as the cached inputs and if so, the cached output is returned without recomputing the function. This is a neat solution to the Gibbs sampling efficiency issue discussed in #833 without having to change the Turing compiler in any way. The question now is which memoization tool is the best tool, i.e. doesn't hurt type inference. Last I checked Memoize.jl wasn't type stable.

mohamed82008:
The run-time checking of the arguments may be sub-optimal but it's certainly worth it for expensive functions with a time complexity higher than that of the checking of the arguments. The advantage here though is that we don't need to make any special assumptions about the model, e.g. being static or anything like that. The function needs to not have side-effects but this is an intuitive property for any function to be memoize-friendly.

We can also roll our own version of memoization at some point that associates the memoized function with specific symbols in the model using a syntax similar to that proposed in https://github.com/TuringLang/Turing.jl/issues/1001. This way, we can make the memoization check happen at compile-time.

PR #1025, opened 2019-12-15
Add memoization when Gibbs sampling to performance tips

mohamed82008:
This PR adds another performance tip, telling users about Memoization.jl which can be used to reuse expensive computations when performing Gibbs sampling.

### DIFF
--- a/docs/src/using-turing/performancetips.md
+++ b/docs/src/using-turing/performancetips.md
@@ -7,6 +7,7 @@ title: Performance Tips
 This section briefly summarises a few common techniques to ensure good performance when using Turing.
 We refer to [julialang.org](https://docs.julialang.org/en/v1/manual/performance-tips/index.html) for general techniques to ensure good performance of Julia programs.
 
+
 ## Use multivariate distributions
 
 It is generally preferable to use multivariate distributions if possible.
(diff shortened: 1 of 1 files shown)


mohamed82008:
Closes #1024

docs/src/using-turing/performancetips.md
```diff
@@ -94,4 +99,40 @@ spl = Turing.SampleFromPrior();
...
+sample(demo(zeros(10)), alg, 1000)
+```
+when only updating `a` in a Gibbs sub-iteration, keeping `b` the same, the value of `d` doesn't change. And when only updating `b`, the value of `c` doesn't change. However, if `function1` and `function2` are expensive and are both run in every Gibbs sub-iteration, a lot of time would be spent computing values that we already computed before. Such a problem can be overcome using `Memoization.jl`. Memoizing a function lets us store and reuse the output of the function for every input it is called with. This has a slight time overhead but for expensive functions, the savings will be far greater. 
+
+To use `Memoization.jl`, simply define memoized versions of `function1` and `function2` as such:
+```julia
+using Memoization
+
+@memoize memoized_function1(args...) = function1(args...)
+@memoize memoized_function2(args...) = function2(args...)
```
yebai: nice!

mohamed82008:
Automating the caching is possible but I will come back to it after DynamicPPL comes to life.

PR #1026, opened 2019-12-15
Minor website changes & GitHub actions setup

cpfiffer:
Summary of changes:

- Adds a first attempt at a GitHub action to build the documentation site -- I'll probably need to revise something after this PR goes through.
- Adds a project file to the `docs` folder. Any new package needed for the documentation build should be added to that environment.
- Fixes a broken link on the splash page, where one of the cards was not having a hyperlink inserted.

PR #1030, opened 2019-12-17
Remove .travis.yml site build

cpfiffer:
Removes the stage of the Travis build that generates the documentation. Documentation builds are soon to be handled by GitHub actions, see #1026.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,10 +51,6 @@ jobs:
       env: STAGE=benchmark
       script:
         - julia ./benchmarks/runbenchmarks.jl
-after_success:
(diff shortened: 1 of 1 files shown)


devmotion:
IMO you have to remove https://github.com/TuringLang/Turing.jl/blob/9750809dbb320b03e234710a1d8b75c2e0bed1d4/.travis.yml#L39-L47. I also noticed that there are still duplicate code coverage commands - I guess one should be sufficient.

cpfiffer:
Man, I'm not doing a great job today. Good catch.

PR #1031, opened 2019-12-17
Replace deprecated option

devmotion:
It seems the changes introduced by https://github.com/TuringLang/Turing.jl/pull/1029 work, but somehow I used a deprecated option. According to the [README](https://github.com/actions/setup-ruby) the version in this PR should be correct.


devmotion:
The test failure is unrelated: https://travis-ci.org/TuringLang/Turing.jl/jobs/626358576?utm_medium=notification&utm_source=github_status

PR #1036, opened 2019-12-21
Do not allow `nothing` as sampler

devmotion:
In https://github.com/TuringLang/Turing.jl/pull/995 the support of `nothing` as a sampler was removed, but this change was reverted by the compiler changes in https://github.com/TuringLang/Turing.jl/pull/965. In contrast to the previous approach, with this PR the interface for `assume` and `observe` always expects a sampler as first argument but `nothing` is not allowed anymore. Instead one has to specify `SampleFromPrior()` explicitly, which IMO is also more informative.


yebai:
Thanks @devmotion!

PR #1037, opened 2019-12-23
Fix docstrings and remove static type parameters

devmotion:
This PR
- removes unneeded static type parameters in RandomVariables.jl
- fixes indentation and use of imperative form in the docstrings in RandomVariables.jl according to the [manual](https://docs.julialang.org/en/latest/manual/documentation/)
- adds a `getmetadata(vi, vn)` method that allows to remove many redundant function definitions for `UntypedVarInfo` and `TypedVarInfo`
- fixes the inconsistency between `getproperty` and `setproperty!`: currently `getproperty` forwards every property access of `vi`, apart from `num_produce`, `logp` and `metadata`, to `vi.metadata` whereas `setproperty!` only handles `num_produce` and `logp` in a special way. Apart from the inconsistency between `getproperty` and `setproperty!`, a problem of the current definition of `getproperty` is that forwarding the access is only reasonable for `UntypedVarInfo` but not for `TypedVarInfo`. Thus this PR redefines `getproperty` to only handle `num_produce` and `logp` in a special way.
- removes `vn.sym` since it is redundant with `getsym`

### DIFF
--- a/src/core/RandomVariables.jl
+++ b/src/core/RandomVariables.jl
@@ -337,7 +337,7 @@ getdist(vi::VarInfo, vn::VarName) = getmetadata(vi, vn).dists[getidx(vi, vn)]
 
 Return the value(s) of `vn`.
 
-The values may or may not be transformed to Euclidian space.
+The values may or may not be transformed to Euclidean space.
 """
 getval(vi::VarInfo, vn::VarName) = view(getmetadata(vi, vn).vals, getrange(vi, vn))
 
@@ -346,7 +346,7 @@ getval(vi::VarInfo, vn::VarName) = view(getmetadata(vi, vn).vals, getrange(vi, v
 
 Set the value(s) of `vn` in the metadata of `vi` to `val`.
 
-The values may or may not be transformed to Euclidian space.
+The values may or may not be transformed to Euclidean space.
 """
 setval!(vi::VarInfo, val, vn::VarName) = getmetadata(vi, vn).vals[getrange(vi, vn)] = val
 
@@ -355,7 +355,7 @@ setval!(vi::VarInfo, val, vn::VarName) = getmetadata(vi, vn).vals[getrange(vi, v
 
 Return the value(s) of `vns`.
 
-The values may or may not be transformed to Euclidian space.
+The values may or may not be transformed to Euclidean space.
 """
 function getval(vi::AbstractVarInfo, vns::Vector{<:VarName})
     return mapreduce(vn -> getval(vi, vn), vcat, vns)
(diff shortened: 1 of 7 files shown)


mohamed82008:
Thanks a lot for this! I will take a closer look soon.

src/core/RandomVariables.jl
```diff
@@ -274,115 +275,100 @@ end
...
+getdist(vi::VarInfo, vn::VarName) = getmetadata(vi, vn).dists[getidx(vi, vn)]
 
 """
-`getval(vi::VarInfo, vn::VarName)`
+    getval(vi::VarInfo, vn::VarName)
+
+Return the value(s) of `vn`.
 
-Returns the value(s) of `vn`. The values may or may not be transformed to Eucledian space.
+The values may or may not be transformed to Euclidian space.
```
mohamed82008: Typo: Euclidean

src/core/RandomVariables.jl
```diff
@@ -568,77 +546,69 @@ end
...
+Return a `Symbol` represenation of the variable identifier `VarName`.
 """
 Symbol(vn::VarName) = Symbol(string(vn))  # simplified symbol
 
 """
-`in(vn::VarName, space::Set)`
+    in(vn::VarName, space::Set)
 
-Returns `true` if `vn`'s symbol is in `space` and `false` otherwise.
+Check wheher `vn`'s symbol is in `space`.
```
mohamed82008: Type: whether

src/inference/AdvancedSMC.jl
```diff
@@ -267,7 +267,7 @@ end
 
 function assume(spl::Sampler{<:Union{PG,SMC}}, dist::Distribution, vn::VarName, ::VarInfo)
     vi = current_trace().vi
-    if isempty(getspace(spl.alg)) || vn.sym in getspace(spl.alg)
+    if vn in getspace(spl)
```
mohamed82008: What about the case where the space is `()`?

devmotion:
> Also please add a type inference test using Tests.@inferred

I'm not sure what exactly you mean - what function you want to be tested for type inference?

src/inference/AdvancedSMC.jl
devmotion: There's no need to take of this anywhere (e.g., in the `assume` implementations) since the implementation of `in` in `RandomVariables.jl` checks
https://github.com/TuringLang/Turing.jl/blob/7e2e159184c2e1846fbb780c6b7bcd38b1944029/src/core/RandomVariables.jl#L611

mohamed82008: Cool, forgot that! Thanks.

mohamed82008:
Testing that the model call infers. Use some model that returns a function of its random variables. Then test:
```julia
vi = VarInfo(model)
spl1 = SampleFromPrior()
@inferred model(vi, spl1)`
spl2 = Sampler(HMC(0.1, 5), model)
new_vi = VarInfo(vi, spl2, ForwardDiff.Dual{Nothing}.((rand(2)))) # if 2 parameters
@inferred model(new_vi, spl2)`
```
for example.

devmotion:
Ah OK, I see! I just added some `@inferred` statements to the existing test set.

yebai:
Many thanks, @devmotion!

PR #1042, opened 2019-12-25
 Switch to DynamicPPL

mohamed82008:
This is a WIP PR that makes Turing depend on DynamicPPL.


mohamed82008:
There seems to be issues with the new AdvancedHMC and HMCDA in Turing. CC: @xukai92

mohamed82008:
I am getting the same failures on master locally.

xukai92:
Since which AHMC version we have this error?

mohamed82008:
I will check.

mohamed82008:
NUTS also seems to have issues.

mohamed82008:
Since 0.2.15

xukai92:
Thanks for checking. I will work on it when I'm with my laptop later.

xukai92:
It's weird that I cannot reproduce the error locally.

xukai92:
> It's weird that I cannot reproduce the error locally.

Sorry my tests were influnced by the Manifest.toml file - since when we had it back?

mohamed82008:
> Sorry my tests were influnced by the Manifest.toml file - since when we had it back?

We don't it's gitignored, it is just generated automatically on your system when you add a dependency using `Pkg` in the package's environment.

mohamed82008:
Or if you instantiate the package's environment.

xukai92:
Oh you are right. Thanks for sharing this :) I can reproduce the error now and let me figure out what's going on here.

xukai92:
I pushed some commits to fix deprecations. The error will need https://github.com/TuringLang/AdvancedHMC.jl/pull/156. Basically I forget to check cases where `n_adapts` is smaller than the default adpataion buffer of the Stan HMC adaptor and this causes the error.

xukai92:
We might not need https://github.com/TuringLang/Turing.jl/pull/1042/commits/ffbc169371bf915b115cae02cac7fed7219297f7 to fix the error any more but I think it's good to have it. How do you think?

Also cc @cpfiffer

cpfiffer:
ffbc169 may have the unintended consequence of not allowing us to automatically set the number of adaptation samples, which might be fine. Currently, if the number of adaptation samples is set to zero, it'll be changed to something reasonable in `sample_init!`:

https://github.com/TuringLang/Turing.jl/blob/ffbc169371bf915b115cae02cac7fed7219297f7/src/inference/hmc.jl#L125

It's probably fine to do this kind of thing, but it makes convenience functions like just `NUTS()` use no adaptation by default -- is that something we'd prefer?

src/utilities/stan-interface.jl
```diff
@@ -66,6 +66,7 @@ function sample(
     end
 end
 
+# FIXME: where is `spl` in the code below?
```
mohamed82008: Seriously where is `spl` in the code below? 😆

mohamed82008: Is this function even used?

xukai92:
I think the current order of automatically settings the number of adaptation is flawed. What would happen if one is intended to set it to `0`? The current behaviour will still change it to something else. 

I think what we had before could handle it correctly, i.e. setting it during the construction of algorithm.

src/utilities/stan-interface.jl
xukai92: Might not be correctly tested since whatever.

cpfiffer:
> I think the current order of automatically setting the number of adaptation is flawed. What would happen if one is intended to set it to `0`? The current behaviour will still change it to something else.

Yeah, but this was also our old logic carried in from the original Hamiltonian implementations -- I mostly just moved it around. 

The problem with handling this during the construction of the algorithm is that currently the inference algorithm constructor does not know the total number of iterations because `N` is part of the outer `sample` call. We could probably get around that, though.

Maybe a better way to do this would be to move the number of adapatation samples to the keyword side:

```julia
sample(model, NUTS(500, 0.65), 1000)
```

would become

```julia
sample(model, NUTS(0.65), 1000; n_adapts=500)
```

Then we could handle all this more robustly at the inference constructor by passing keywords around. Having the number of adaptations in the inference constructor always seemed a little weird to me anyway.

mohamed82008:
I say we merge this after a rebase, and open a new issue and PR to discuss adaptation issues. Do you agree?

cpfiffer:
Agreed, I'll open a new issue.

PR #1043, opened 2019-12-26
Make _params_to_array faster

mohamed82008:
This PR optimizes the `_params_to_array` function. In the following example, the running time of `sample` was cut down from 22 s to 8 s. Of these times, < 2 s is spent doing the sampling and the rest is saving the chain. There is a lot of room for improvement here. The 50000 iterations case terminates in 134 s now of which 16 s is sampling. Before this PR, it was too slow to even wait and time it.

```julia
using Turing

fake_dat = [5*ones(Int64, 30) for j=1:18]
fake_N = fill(5*30, 18) #Total N

@model multinomial_RE(data, N, ::Type{T}=Vector{Float64}) where {T} = begin
    B = length(data)
    R = length(data[1,1])
    theta = [T(undef, R) for j = 1:B]
    beta ~ Dirichlet(R, 1)
    alpha ~ truncated(Normal(1, 100), 0, Inf)
    for b in 1:B
    	theta[b] ~ Dirichlet(beta * alpha)
    	if any(isnan, theta[b])
    		@logpdf() = -Inf
    		return
    	end
    	data[b] ~ Multinomial(N[b], theta[b])
    end
end
model2 = sample(multinomial_RE(fake_dat, fake_N), MH(), 5000);
@time model2 = sample(multinomial_RE(fake_dat, fake_N), MH(), 5000);
```

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -242,7 +242,7 @@ end
 function _params_to_array(ts::Vector{<:AbstractTransition}, spl::Sampler)
     names_set = Set{String}()
     # Extract the parameter names and values from each transition.
-    dicts = map(enumerate(ts)) do (i, t)
+    dicts = map(ts) do t
         nms, vs = flatten_namedtuple(t.θ)
         for nm in nms
             push!(names_set, nm)
@@ -257,8 +257,8 @@ function _params_to_array(ts::Vector{<:AbstractTransition}, spl::Sampler)
     return names, vals
 end
 
-function flatten_namedtuple(nt::NamedTuple{pnames}) where {pnames}
-    temp = map(keys(nt)) do k
+function flatten_namedtuple(nt::NamedTuple)
+    names_vals = mapreduce(vcat, keys(nt)) do k
         v = nt[k]
         if length(v) == 1
             return [(string(k), v)]
@@ -268,7 +268,6 @@ function flatten_namedtuple(nt::NamedTuple{pnames}) where {pnames}
             end
         end
     end
-    names_vals = vcat(temp...)
     return [vn[1] for vn in names_vals], [vn[2] for vn in names_vals]
 end
 
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -240,85 +240,37 @@ end
 ##########################
 
 function _params_to_array(ts::Vector{<:AbstractTransition}, spl::Sampler)
-    names = Set{String}()
-    dicts = Vector{Dict{String, Any}}()
-
+    names_set = Set{String}()
     # Extract the parameter names and values from each transition.
-    for t in ts
+    dicts = map(enumerate(ts)) do (i, t)
         nms, vs = flatten_namedtuple(t.θ)
-        push!(names, nms...)
-
+        for nm in nms
+            push!(names_set, nm)
+        end
         # Convert the names and values to a single dictionary.
-        d = Dict{String, Any}()
-        for (k, v) in zip(nms, vs)
-            d[k] = v
-        end
-        push!(dicts, d)
+        return Dict(nms[j] => vs[j] for j in 1:length(vs))
     end
-
-    # Convert the set to an ordered vector so the parameter ordering
-    # is deterministic.
-    ordered_names = collect(names)
-    vals = Matrix{Union{Real, Missing}}(undef, length(ts), length(ordered_names))
-
-    # Place each element of all dicts into the returned value matrix.
-    for i in eachindex(dicts)
-        for (j, key) in enumerate(ordered_names)
-            vals[i,j] = get(dicts[i], key, missing)
-        end
-    end
-
-    return ordered_names, vals
-end
-
-function flatten_namedtuple(nt::NamedTuple{pnames}) where {pnames}
-    vals = Vector{Real}()
-    names = Vector{AbstractString}()
-
-    for k in pnames
-        v = nt[k]
-        if length(v) == 1
-            flatten(names, vals, string(k), v)
-        else
-            for (vnval, vn) in zip(v[1], v[2])
-                flatten(names, vals, vn, vnval)
-            end
-        end
-    end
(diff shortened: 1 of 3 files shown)


src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
...
-    # Convert the set to an ordered vector so the parameter ordering 
-    # is deterministic.
-    ordered_names = collect(names)
-    vals = Matrix{Union{Real, Missing}}(undef, length(ts), length(ordered_names))
-
-    # Place each element of all dicts into the returned value matrix.
-    for i in eachindex(dicts)
-        for (j, key) in enumerate(ordered_names)
-            vals[i,j] = get(dicts[i], key, missing)
+        foreach(nms) do nm
```
devmotion: Why `foreach` instead of a regular `for` loop?

src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
 ##########################
 
 function _params_to_array(ts::Vector{T}, spl::Sampler) where {T<:AbstractTransition}
-    names = Set{String}()
-    dicts = Vector{Dict{String, Any}}()
-
+    names_set = Set{String}()
     # Extract the parameter names and values from each transition.
-    for t in ts
+    dicts = map(enumerate(ts)) do (i, t)
```
devmotion: `i` is never used it seems?

devmotion:
```suggestion
    dicts = map(ts) do t
```

src/inference/Inference.jl
devmotion:
```suggestion
        for nm in nms
```

src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
 ##########################
 
 function _params_to_array(ts::Vector{T}, spl::Sampler) where {T<:AbstractTransition}
```
devmotion:
```suggestion
function _params_to_array(ts::Vector{<:AbstractTransition}, spl::Sampler)
```

src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
...
     end
+    names = collect(names_set)
+    vals = [get(dicts[i], key, missing) for i in eachindex(dicts), 
+        (j, key) in enumerate(names)]
 
-    return ordered_names, vals
+    return names, vals
 end
 
 function flatten_namedtuple(nt::NamedTuple{pnames}) where {pnames}
```
devmotion:
```suggestion
function flatten_namedtuple(nt::NamedTuple)
```

src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
...
-    vals = Vector{Real}()
-    names = Vector{AbstractString}()
-
-    for k in pnames
+    temp = map(keys(nt)) do k
```
devmotion: It seems you could just use `mapreduce`?

```suggestion
    names_vals = mapreduce(vcat, keys(nt)) do k
```

src/inference/Inference.jl
```diff
@@ -259,86 +259,38 @@ end
...
-            for (vnval, vn) in zip(v[1], v[2])
-                flatten(names, vals, vn, vnval)
+            return mapreduce(vcat, zip(v[1], v[2])) do (vnval, vn)
+                return collect(FlattenIterator(vn, vnval))
             end
         end
     end
-
-    return names, vals
+    names_vals = vcat(temp...)
```
devmotion: Corresponding to the suggestion above:
```suggestion
```

src/utilities/helper.jl
```diff
@@ -73,3 +73,63 @@ function reconstruct!(r, d::MultivariateDistribution, val::AbstractVector, n::In
     r .= val
     return r
 end
+
+struct FlattenIterator{Tname, Tvalue}
+    name::Tname
+    value::Tvalue
+end
+
+Base.length(iter::FlattenIterator) where {T} = _length(iter.value)
```
devmotion: `T` is not defined it seems?
```suggestion
Base.length(iter::FlattenIterator) = _length(iter.value)
```

src/utilities/helper.jl
```diff
@@ -73,3 +73,63 @@ function reconstruct!(r, d::MultivariateDistribution, val::AbstractVector, n::In
...
+@inline _length(a::AbstractArray) = sum(_length, a)
```
devmotion: Are the `@inline` statements actually needed? And maybe define a special case for `AbstractArray{<:Number}` to avoid summing up 1s?

```suggestion
_length(a::AbstractArray) = sum(_length, a)
_length(a::AbstractArray{<:Number}) = length(a)
```

src/utilities/helper.jl
```diff
@@ -73,3 +73,63 @@ function reconstruct!(r, d::MultivariateDistribution, val::AbstractVector, n::In
...
+@inline _length(::Number) = 1
+
+Base.eltype(iter::FlattenIterator{String}) = Tuple{String, _eltype(typeof(iter.value))}
+@inline _eltype(::Type{TA}) where {T, TA <: AbstractArray{T}} = _eltype(T)
```
devmotion: Again, is `@inline` needed? And it seems `T` is not needed?
```suggestion
_eltype(::Type{TA}) where {TA <: AbstractArray} = _eltype(eltype(TA))
```

src/utilities/helper.jl
mohamed82008: Sounds good.

mohamed82008:
Thanks @devmotion! I have this impulse to add `@inline` next to small recursive functions I write which I should probably control :)

yebai:
Looks like we need a rebase before merging.

mohamed82008:
Yes I was planning to try to make this even faster, but we can merge for now after the rebase and I will make another PR.

mohamed82008:
I messed up the rebase. Let me fix the mess.

PR #1044, opened 2019-12-27
Remove static type parameters

devmotion:
This PR removes many unneeded `where` clauses.

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -26,31 +26,25 @@ end
 
 abstract type ADBackend end
 struct ForwardDiffAD{chunk} <: ADBackend end
-getchunksize(::T) where {T <: ForwardDiffAD} = getchunksize(T)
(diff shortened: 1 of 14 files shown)


src/core/ad.jl
```diff
@@ -26,31 +26,25 @@ end
...
 getADtype(alg)
 
 Finds the autodifferentiation type of the algorithm `alg`.
 """
-getADtype(::Nothing) = getADtype(Nothing)
-getADtype(::Type{Nothing}) = getADtype()
-getADtype() = ADBackend()
-getADtype(s::Sampler) = getADtype(typeof(s))
-getADtype(s::Type{<:Sampler{TAlg}}) where {TAlg} = getADtype(TAlg)
+getADtype(::Nothing) = ADBackend()
```
mohamed82008: Do we need this now that the algorithm cannot be `nothing`?

devmotion: I guess you're right and it can be removed :+1:

PR #1046, opened 2019-12-27
Simplify random processes

devmotion:
This PR simplifies `rand` and `logpdf` for random processes by defining the corresponding (scaled) Beta distribution and calling `rand` and `logpdf` for it instead of defining `rand` and `logpdf` from scratch. Moreover, it removes some static type parameters.

### DIFF
--- a/src/stdlib/RandomMeasures.jl
+++ b/src/stdlib/RandomMeasures.jl
@@ -25,8 +25,8 @@ struct SizeBiasedSamplingProcess{T<:AbstractRandomProbabilityMeasure,V<:Abstract
     surplus::V
 end
 
-logpdf(d::SizeBiasedSamplingProcess, x::Real) = _logpdf(d, x)
-rand(rng::AbstractRNG, d::SizeBiasedSamplingProcess) = _rand(rng, d)
+logpdf(d::SizeBiasedSamplingProcess, x::Real) = logpdf(distribution(d), x)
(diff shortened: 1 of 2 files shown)


src/stdlib/RandomMeasures.jl
```diff
@@ -118,29 +118,26 @@ struct DirichletProcess{T<:Real} <: AbstractRandomProbabilityMeasure
...
 end
 
-function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<:Real}
+function _logpdf_table(d::DirichletProcess, m::AbstractVector{Int})
     # compute the sum of all cluster counts
     sum_m = sum(m)
 
     # shortcut if all cluster counts are zero
+    dα = d.α
+    T = typeof(dα)
```
trappmartin: I wonder if this makes any performance difference to the original code. I would assume the original implementation should be easier to optimise for the compiler. What do you think?

devmotion: I think the compiler is smart enough to compile it away. In the example
```julia
struct A{T}
    a::T
end

f(a::A) = typeof(a.a)
g(::A{T}) where T = T
```
`f` and `g` seem to produce the same native assembly instructions, as
```julia
@code_native f(A(rand(3)))
@code_native g(A(rand(3)))
```
shows.

PR #1050, opened 2019-12-28
Make MH in Gibbs faster

mohamed82008:
This PR makes MH in Gibbs faster by skipping the re-running of the model when the previous alg had computed the `logjoint`. The following example is twice as fast:
```julia
using Turing
@model gdemo() = begin
  sum(ones(1000, 1000) * ones(1000))
  s ~ InverseGamma(2, 3)
  m ~ Normal(0, sqrt(s))
  1.5 ~ Normal(m, sqrt(s))
  2.0 ~ Normal(m, sqrt(s))
  return s, m
end
alg = Gibbs(MH(:m), MH(:s))
sample(gdemo(), alg, 1000)

#master: 17 s
#This PR: 8 s
```
I added some useless computation in the model to make the model's run time dominate the total sampling time. We need something similar for HMC at some point or a more systematic way to decide when to re-run the model and when not to.

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -77,9 +77,14 @@ function getspace end
 struct Selector
     gid :: UInt64
     tag :: Symbol # :default, :invalid, :Gibbs, :HMC, etc.
+    rerun :: Bool
 end
-Selector() = Selector(time_ns(), :default)
-Selector(tag::Symbol) = Selector(time_ns(), tag)
+function Selector(tag::Symbol = :default, rerun = tag != :default)
+    return Selector(time_ns(), tag, rerun)
+end
+function Selector(gid::Integer, tag::Symbol = :default)
+    return Selector(gid, tag, tag != :default)
+end
 hash(s::Selector) = hash(s.gid)
 ==(s1::Selector, s2::Selector) = s1.gid == s2.gid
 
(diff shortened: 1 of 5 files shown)


devmotion:
I just had a different idea: since it seems we just need need this for Gibbs sampling (at the moment) and maybe we would like something similar for other algorithms than `MH`, maybe it's easier to extend (and easier for implementing custom samplers) if this is not part of `Selector` and the different `sample!` implementations but rather we just check some new function `needrerun(lastsampler, nextsampler)` (defaults to `true`) and then rerun (or not) the model inside the `sample!` implementation of the Gibbs sampler (using the next sampler).

devmotion:
Maybe that could/should be optimized a bit further, otherwise it could become tricky to implement `needrerun` for all possible choices of samplers. Maybe the default implementation of `needrerun` could make use of traits such as e.g. `needslogjoint` and `saveslogjoint` (and maybe `needsloglikelihood` and `savesloglikelihood`?).

mohamed82008:
@devmotion I agree that we need to do this in a more systematic way. This will require a refactoring of Gibbs sampling in Turing which I am not attempting here per se. Also it's not just about re-running the model. For example in HMC, we do the domain transformation bijection before and after the run. So MH for example would be sampling without transformation but then HMC does the transformation. Whether we should be doing that or not, versus doing all the sampling in the transformed (Euclidean) space, is probably questionable. May be we can expose that as an option in the Gibbs sampler. All in all, I think we need to re-think Gibbs sampling for performance without sacrificing flexibility. I agree with Cameron that this can be done in a separate PR.

mohamed82008:
Also, I agree that traits would be perfect to use for this refactor.

PR #1056, opened 2020-01-05
Relax constraints for Requires

itsdfish:
On Slack, I mentioned to Mohamed that many packages (e.g. Atom, JSON and others) have not adopted Requires 1.0 and this has caused problems with installing Turing. I ran two tests on Turing:

[Requires](https://travis-ci.com/itsdfish/Turing.jl/builds/143137609) "0.5.2, 1.0" which used v1.0 
[Requires](https://travis-ci.com/itsdfish/Turing.jl/builds/143136017) "0.5.2"

Not all tests passed, but the same tests failed in both cases. So I do not believe that changing the version constraints introduced new problems. If this change looks good, would it be possible to merge and release v0.8.1?


Project.toml
```diff
@@ -42,7 +42,7 @@ MacroTools = "0.5.1"
 PDMats = "0.9.9"
 ProgressMeter = "1.0.0"
 Reexport = "0.2.0"
-Requires = "1.0"
+Requires = "0.5.2, 1.0"
```
mohamed82008: I think `0.5` is better here.

mohamed82008:
Thanks for the PR @itsdfish! I can confirm that tests pass locally with Requires 0.5.2.

Project.toml
mohamed82008:
```suggestion
Requires = "0.5, 1.0"
```

itsdfish:
No problem. I am glad to help in what little ways that I can. 

I commited your suggested change. Let me know if you need anything else.

yebai:
Thanks @itsdfish!

PR #1062, opened 2020-01-07
added NUTS test

trappmartin:


benchmarks/nuts/hlr.jl
```diff
@@ -0,0 +1,25 @@
+using Turing, TuringBenchmarks.TuringTools
```
yebai: Kai and I plan to move all the micro-benchmarks into `MCMCBenchmarks`, and depreciate `TuringBenchmarks` together with `ContinousBenchmarks.jl`.  Given that this transition will happen soon, maybe 

- remove the dependency on `TuringBenchmarks.TuringTools`
- and replace `@tbenchmark` with `@benchmark` from `BenchmarkTools`

mschauer:
@trappmartin Do you still remember where this example originated?

trappmartin:
Those where taken from the NUTS paper by Hoffman and Gelman (Section 4). I think the data comes from their repo, but I'm not 100% sure.

PR #1083, opened 2020-01-25
Use AdvancedMH

cpfiffer:
This PR removes Turing's original Metropolis-Hastings algorithm and replaces it with the `StaticMH` variant from AdvancedMH.jl. There's still a couple things left to do, but for the most part it's functioning if you don't use it as part of Gibbs sampling.

Tests are not going to pass until I push some stuff upstream to AdvancedMH.

I have also added in a couple of things to keep up with AbstractMCMC's recent developments, namely, the ability to select the type of chain you want to receive at the end of sampling. You can now request `chain_type = NamedTuple` to get a vector of `NamedTuple`s, or `chain_type=Vector` which returns the raw `Transition` objects, especially if you are annoyed by the slowness of `MCMCChains`. 

## Todo

- [x] Make it work
- [x] Fix Gibbs sampling
- [x] Allow custom proposals in the same way that the original implementation did
- [x] Test expansion
- [x] Rebase
- [x] Remove old MH code

## Speed

Currently, the performance of the AdvancedMH implementation is slightly improved over the vanilla version for small sample sizes. Using the classic Turing toy model:

```julia
using Turing
using BenchmarkTools
using Random

Random.seed!(1234)

@model gdemo(xs) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, s)

    for i in 1:length(xs)
        xs[i] ~ Normal(m, s)
    end
end

data = rand(Normal(0,1), 30)
model = gdemo(data)

N = 10_000

sampler_new = MH()
b_new = @benchmark chain_new = sample(model, sampler_new, N)

sampler_old = Turing.Inference.MHOld()
b_old = @benchmark chain_old = sample(model, sampler_old, N)
```

AdvancedMH:

```julia
  memory estimate:  88.65 MiB
  allocs estimate:  1981221
  --------------
  minimum time:     82.960 ms (14.52% GC)
  median time:      98.216 ms (23.36% GC)
  mean time:        99.338 ms (20.38% GC)
  maximum time:     115.144 ms (25.57% GC)
  --------------
  samples:          51
  evals/sample:     1
```

Vanilla MH:

```julia
  memory estimate:  106.01 MiB
  allocs estimate:  3572170
  --------------
  minimum time:     121.415 ms (15.52% GC)
  median time:      131.936 ms (16.37% GC)
  mean time:        133.293 ms (15.98% GC)
  maximum time:     149.723 ms (16.84% GC)
  --------------
  samples:          38
  evals/sample:     1
```

If you increase the sampled data size from 30 to 300, the gains are a lot more significant:

AdvancedMH:

```julia
  memory estimate:  88.71 MiB
  allocs estimate:  1984461
  --------------
  minimum time:     104.611 ms (11.88% GC)
  median time:      119.269 ms (19.72% GC)
  mean time:        119.940 ms (16.65% GC)
  maximum time:     137.361 ms (20.06% GC)
  --------------
  samples:          42
  evals/sample:     1
```

Vanilla MH:

```julia
  memory estimate:  395.08 MiB
  allocs estimate:  19788228
  --------------
  minimum time:     455.833 ms (12.88% GC)
  median time:      480.949 ms (14.12% GC)
  mean time:        483.483 ms (14.21% GC)
  maximum time:     533.009 ms (14.52% GC)
  --------------
  samples:          11
  evals/sample:     1
```


src/inference/Inference.jl
```diff
@@ -317,7 +317,8 @@ function AbstractMCMC.bundle_samples(
     model::AbstractModel,
     spl::Sampler,
     N::Integer,
-    ts::Vector{<:AbstractTransition};
+    ts::Vector{<:AbstractTransition},
+    chain_type::Type{Any};
```
devmotion: IMO the more intuitive approach would be to use
```suggestion
    chain_type::Type{Chains};
```
here, since this implementation returns a `Chains` object. If this is supposed to be the default behaviour in Turing, I think one should just set these defaults in the Turing implementations of `sample`, i.e., one should use something similar to
```julia
sample([rng::AbstractRNG, ]model::AbstractModel, alg::InferenceAlgorithm, N::Integer; kwargs...) =
    sample(rng, model, Sampler(alg, model), N; chain_type = Chains, kwargs...)
```
Then there would be also no need for defining the method for `chain_type = Vector`, since one could just use `sample(rng, model, alg, N; chain_type = Any)` to get the default behaviour of returning just the set of all transitions.

src/inference/Inference.jl
```diff
@@ -369,6 +370,49 @@ function AbstractMCMC.bundle_samples(
...
+    return ts
+end
+
+function AbstractMCMC.bundle_samples(
+    rng::AbstractRNG,
+    model::AbstractModel,
+    spl::Sampler,
+    N::Integer,
+    ts::Vector{<:AbstractTransition},
+    chain_type::Type{NamedTuple};
```
devmotion: It seems more accurate would be
```suggestion
    chain_type::Type{Vector{NamedTuple}};
```
?

src/inference/Inference.jl
```diff
@@ -369,6 +370,49 @@ function AbstractMCMC.bundle_samples(
...
+    discard_adapt::Bool=true,
+    save_state=true,
+    kwargs...
+)
+    nts = Vector{NamedTuple}(undef, N)
```
devmotion: This creates an array with an abstract element type - I guess the different elements are not guaranteed to be of the same type? Maybe it could be useful to return `map(i -> nts[i], 1:N)` instead of `nts` at the end, similar to what is done in DiffEqBase for vectors with abstract element types such as in https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/2386f211a2228bb302a0145c039ffa9ca2d2b450/src/ensemble/basic_ensemble_solve.jl#L180?

src/inference/amh.jl
```diff
@@ -0,0 +1,278 @@
...
+
+    # Get `init_theta`
+    initialize_parameters!(spl; verbose=verbose, kwargs...)
+end
+
+function step!(
+    rng::AbstractRNG,
+    model::Model,
+    spl::Sampler{<:MH},
+    N::Integer;
```
devmotion: I think it might be better/safer to add a `::Nothing` argument as well, and implement the method that is further down in the call stack. Otherwise calling `step!(..., N; kwargs...)` and `step!(..., N, nothing; kwargs...)` return different results.

cpfiffer:
Test failures are due to bounds on DynamicPPL, which doesn't allow AbstractMCMC 0.3. 

The PR [here](https://github.com/TuringLang/DynamicPPL.jl/pull/34) needs to go through with a release for this to clear. I'll fix the test failures in DynamicPPL where I can tomorrow.

cpfiffer:
This is ready for a formal review by everyone. The only thing that has not made it over is support checking for custom proposals -- I haven't found a good procedure yet to handle this automatically. My preferred method would be to have transforms between R and R+ or whatever when generating proposals for constrained distributions. I'm open to suggests on this, but for the moment I've changed the tests so that they use static proposals for constrained distributions.

src/inference/Inference.jl
```diff
@@ -369,12 +374,55 @@ function AbstractMCMC.bundle_samples(
...
+    spl::Sampler,
+    N::Integer,
+    ts::Vector{<:AbstractTransition},
+    chain_type::Type{Any};
+    discard_adapt::Bool=true,
+    save_state=true,
+    kwargs...
+)
+    return ts
+end
```
devmotion: I thought the `Any` case should be covered by the [default implementation in AbstractMCMC](https://github.com/TuringLang/AbstractMCMC.jl/blob/0641b816c91027f0f1ffaf3c515ac103b5b45b68/src/AbstractMCMC.jl#L272-L282)?

cpfiffer: You are correct, good catch!

src/inference/Inference.jl
```diff
@@ -431,7 +467,7 @@ include("../contrib/inference/AdvancedSMCExtensions.jl")
 # Typing tools #
 ################
 
-for alg in (:SMC, :PG, :PMMH, :IPMCMC, :MH, :IS, :ESS, :Gibbs)
+for alg in (:SMC, :PG, :PMMH, :IPMCMC, :MH, :MHOld, :IS, :ESS, :Gibbs)
```
yebai: Is the `MHOld` algorithm still available?

test/inference/mh.jl
```diff
@@ -8,7 +8,7 @@ include(dir*"/test/test_utils/AllUtils.jl")
         Random.seed!(0)
         N = 500
         s1 = MH(
-            (:s, GKernel(3.0)),
+            (:s, InverseGamma(2,3)),
             (:m, GKernel(3.0)))
```
yebai: Is `GKernel` equivalent with `Normal`? If so, maybe consider replacing it with `Normal`.

src/inference/Inference.jl
```diff
@@ -369,12 +374,41 @@ function AbstractMCMC.bundle_samples(
...
+            push!(vs, v[1])
+        end
+
+        push!(k, :lp)
+        
+        
+        nts[i] = NamedTuple{tuple(k...)}(tuple(vs..., t.lp))
+    end
+
+    return map(x -> x, nts)
```
devmotion: Maybe
```suggestion
    return map(identity, nts)
```
?

src/inference/mh.jl
```diff
@@ -1,173 +1,308 @@
-mutable struct MHState{V<:VarInfo} <: AbstractSamplerState
-    proposal_ratio        ::   Float64
-    prior_prob            ::   Float64
-    violating_support     ::   Bool
-    vi                    ::   V
+import AdvancedMH
+const AMH = AdvancedMH
+
+import DynamicPPL: VarName, _getranges, _getindex, getval, _getvns
```
devmotion: Maybe move the import statements to `Inference.jl`?

src/inference/Inference.jl
cpfiffer: Nope, thanks for catching that one.

test/inference/mh.jl
cpfiffer: `GKernel` returns a function that you can draw proposals from:

https://github.com/TuringLang/Turing.jl/blob/97e9eb534f68e45b9558bab5e4f5ab047c920f20/test/test_utils/testing_functions.jl#L3

So this is a random-walk style proposal. I left this method in for backwards compatibility, as you can replicate this behavior in a more direct way by constructing

```julia
MH(:m:  => AMH.Proposal(RandomWalk(), Normal(0,3), . . .)
```

src/inference/mh.jl
cpfiffer: Sure.

yebai:
Ready to merge from my side once the tests pass.

cpfiffer:
I'll have to comb through all these test failures. None of then were occurring on my local build, so this'll be fun.

cpfiffer:
Looks like tests are passing! Should be good to merge for now.

yebai:
Great work, thanks @cpfiffer!

PR #1089, opened 2020-01-31
Add method for pdf(::OrderedLogistic, x)

baggepinnen:
Before fix
```julia
julia> pdf(OrderedLogistic(1,[0.2,0.5]), 1.0)
ERROR: StackOverflowError:
Stacktrace:
 [1] pdf(::OrderedLogistic{Int64,Array{Float64,1}}, ::Int64) at /home/fredrikb/.julia/packages/Distributions/KjaXI/src/univariates.jl:333 (repeats 80000 times)
```
after fix
```julia
julia> pdf(OrderedLogistic(1,[0.2,0.5]), 1.0)
0.3100255188723875
```

### DIFF
--- a/src/stdlib/distributions.jl
+++ b/src/stdlib/distributions.jl
@@ -103,6 +103,8 @@ function Distributions.logpdf(d::OrderedLogistic, k::Int)
     return logp
 end
 
+Distributions.pdf(d::OrderedLogistic, k::Int) = exp(logpdf(d,k))
+
 function Distributions.rand(rng::AbstractRNG, d::OrderedLogistic)
     cutpoints = d.cutpoints
     η = d.η


devmotion:
I'm not sure this problem should be fixed in Turing, since the main problem seems to be the default definition of `pdf` for discrete univariate distributions in Distributions in https://github.com/JuliaStats/Distributions.jl/blob/b637c8aecbca62fb1a1fa41febd6378c489de1e1/src/univariates.jl#L333. IMO one should maybe rather change the default implementation in Distributions. I'm not sure what's the most reasonable approach - one idea would be that `pdf` should maybe rather be defined as `exp(logpdf(..))` instead of the other way around (since I guess usually one wants to implement the logpdf anyway) and then one could define
```julia
logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : - partype(d)(Inf)
```
(or something else since that seems prone to type instability...)

devmotion:
One additional argument for changing it in Distributions by defining `pdf(...) = exp(logpdf(...))` is that the current way is inconsistent with the default definitions for multivariate distributions. There the default implementation uses `_pdf(...) = exp(_logpdf(...))` in https://github.com/JuliaStats/Distributions.jl/blob/b637c8aecbca62fb1a1fa41febd6378c489de1e1/src/multivariates.jl#L190.

xukai92:
I think the fix makes sense. The issue is really if `pdf(d::OrderedLogistic, k::T)` is defined for `T<:Int`, as it would be triggered in https://github.com/JuliaStats/Distributions.jl/blob/b637c8aecbca62fb1a1fa41febd6378c489de1e1/src/univariates.jl#L333, which is to correctly handle a non integreter `k` passed to `pdf`.

devmotion:
> I think the fix makes sense. The issue is really if pdf(d::OrderedLogistic, k::T) is defined for T<:Int, as it would be triggered in https://github.com/JuliaStats/Distributions.jl/blob/b637c8aecbca62fb1a1fa41febd6378c489de1e1/src/univariates.jl#L333, which is to correctly handle a non integreter k passed to pdf.

Sure, the PR fixes the problem and it might make sense to handle integers in Distributions (although discrete distributions in general don't have to use integers as support...). However, IMO integers should be taken care of on the level of `logpdf` since I guess (and the multivariate implementation in Distribution indicates) usually one implements `logpdf` and assumes that `pdf(...) = exp(logpdf(...))` just works. So I think one should fix the inconsistent and unexpected behaviour in Distributions (which fixes the problem here as well) instead of adding a very specific fix to Turing.

cpfiffer:
I think since we define `OrderedLogistic` here it's fine to add this method for now.

yebai:
Thanks, @baggepinnen!

PR #1093, opened 2020-02-05
fixes and improvements on CRP implementations

trappmartin:


src/stdlib/RandomMeasures.jl
```diff
@@ -82,7 +82,7 @@ function rand(rng::AbstractRNG, d::ChineseRestaurantProcess)
 end
 
 minimum(d::ChineseRestaurantProcess) = 1
-maximum(d::ChineseRestaurantProcess) = length(d.m) + 1
+maximum(d::ChineseRestaurantProcess) = any(i -> i == 0, d.m) ? length(d.m) : length(d.m)+1
```
devmotion: A bit more general (and simpler IMO) would be
```suggestion
maximum(d::ChineseRestaurantProcess) = any(iszero, d.m) ? length(d.m) : length(d.m) + 1
```

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
     return LocationScale(zero(α), d.surplus, Beta(one(α), α))
 end
 
-function _logpdf_table(d::DirichletProcess, m::AbstractVector{Int})
+function _logpdf_table(d::DirichletProcess{T}, m::AbstractVector{Int}) where {T<:AbstractFloat}
```
devmotion: The static type parameter is not needed (since `T = typeof(d.α)`), but in any case I guess it should be `T<:Real` since `d.α isa Real` but not necessarily an `AbstractFloat`.

trappmartin: I think we should make `T<:AbstractFloat`. `Real` doesn't actually make sense to me. I'll change the type signature.

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
     sum_m = sum(m)
 
-    # shortcut if all cluster counts are zero
-    dα = d.α
-    T = typeof(dα)
-    iszero(sum_m) && return zeros(T, 1)
-
     # pre-calculations
-    z = log(sum_m - 1 + dα)
+    z = log(sum_m - 1 + d.α)
```
devmotion: This will error if `sum_m - 1 + d.α < 0`, so I guess one should check this condition?

src/stdlib/RandomMeasures.jl
trappmartin: Cool, thanks!

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
-            if contains_zero
-                table[i] = -Inf
-            else
-                table[i] = log(dα) - z
-                contains_zero = true
-            end
-        else
-            table[i] = log(mi) - z
-        end
+    has_zeros = any(i -> i == 0, m)
```
devmotion: Similar to above, it might be easier to use `iszero`. Moreover, maybe one could add a shortcut here in case `iszero(sum_m)`, i.e.
```suggestion
    has_zeros = (iszero(sum_m) && length(m) > 0) || any(iszero, m)
```
I'm not sure if the case `length(m) == 0` can actually occur/should be handled.

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
+
+    if iszero(sum_m)
+        table[1] = T(0)
+        return table
     end
 
-    if !contains_zero
-        push!(table, log(dα) - z)
+    contains_zero = false
+    @inbounds for i in (has_zeros ? findall(j -> j>0, m) : 1:(K-1))
```
devmotion: The creation of a new array by `findall` could be changing this to
```suggestion
    @inbounds for i in 1:length(m)
        m[i] > 0 || continue
```

trappmartin: I'll time what is faster. Thanks.

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
-            else
-                table[i] = log(dα) - z
-                contains_zero = true
-            end
-        else
-            table[i] = log(mi) - z
-        end
+    has_zeros = any(i -> i == 0, m)
+    K = has_zeros ? length(m) : length(m)+1
+    table = ones(T, K) * T(-Inf)
```
devmotion:
```suggestion
    table = fill(T(-Inf), K)
```

src/stdlib/RandomMeasures.jl
trappmartin: I think we can actually drop the normalisation constant.

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
     end
 
-    if !contains_zero
-        push!(table, log(dα) - z)
+    contains_zero = false
+    @inbounds for i in (has_zeros ? findall(j -> j>0, m) : 1:(K-1))
+        table[i] = T(log(m[i]) - z)
     end
 
+    k_new = has_zeros ? findfirst(j -> j == 0, m) : K
```
devmotion: Again I think one could use `iszero` here. Moreover, I would suggest computing `first_zero = findfirst(iszero, m)` at the beginning of the function, since that yields `has_zeros` naturally, without checking `m` twice. I.e., I would recommend to change the implementation above to
```julia
first_zero = findfirst(iszero, m)
has_zeros = first_zero !== nothing
```
and just defining
```julia
k_new = first_zero === nothing ? K : first_zero
```
here.

src/stdlib/RandomMeasures.jl
devmotion: See my comment below for a better suggestion, IMO.

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
     return LocationScale(zero(d_rpm_d), d.surplus, dist)
 end
 
-function _logpdf_table(d::PitmanYorProcess, m::AbstractVector{Int})
+function _logpdf_table(d::PitmanYorProcess{T}, m::AbstractVector{Int}) where {T<:AbstractFloat}
```
devmotion: Same as above, in case you want to add a static type parameter, it should probably be `T<:Real`.

src/stdlib/RandomMeasures.jl
```diff
@@ -128,41 +128,31 @@ function distribution(d::SizeBiasedSamplingProcess{<:DirichletProcess})
...
+    table = ones(T, K) * T(-Inf)
+
+    if iszero(sum_m)
+        table[1] = T(0)
+        return table
     end
 
-    if !contains_zero
-        push!(table, log(dα) - z)
+    contains_zero = false
```
devmotion: This is not needed it seems and can be removed?
```suggestion
```

src/stdlib/RandomMeasures.jl
devmotion: Actually, I think `has_zeros` is not needed anymore, if you follow this suggestion and change also the `for` loop above since `K` can be defined based on `first_zero` as
```julia
K = first_zero === nothing ? length(m) + 1 : length(m)
```

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
...
-            if contains_zero
-                table[i] = -Inf
-            else
-                table[i] = log(dθ + dd * d.t) - z
-                contains_zero = true
-            end
-        else
-            table[i] = log(mi - dd) - z
-        end
+    has_zeros = any(i -> i == 0, m)
```
devmotion: As above, I would recommend changing this to
```suggestion
    first_zero = findfirst(iszero, m)
    K = first_zero === nothing ? length(m) + 1 : length(m)
```
If the `for` loop below is simplified and one uses
```julia
k_new = first_zero === nothing ? K : first_zero
```
below, `has_zeros` is not needed anymore.

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
...
+    K = has_zeros ? length(m) : length(m)+1
+
+    # sanity check
+    @assert d.t == sum(i -> i>0, m)
+    table = ones(T, K) * T(-Inf)
```
devmotion:
```suggestion
    table = fill(T(-Inf), K)
```

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
...
+
+    if iszero(sum_m)
+        table[1] = T(0)
+        return table
     end
 
-    if !contains_zero
-        push!(table, log(dθ + dd * d.t) - z)
+    contains_zero = false
```
devmotion: Can be removed, it seems?
```suggestion
```

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
...
+    @inbounds for i in (has_zeros ? findall(j -> j>0, m) : 1:(K-1))
```
devmotion: As above, this could be simplified to
```suggestion
    @inbounds for i in 1:length(m)
        m[i] > 0 || continue
```
which would avoid the creation of a new array and one additional run over `m`.

src/stdlib/RandomMeasures.jl
```diff
@@ -211,42 +201,35 @@ function distribution(d::SizeBiasedSamplingProcess{<:PitmanYorProcess})
...
+        table[i] = T(log(m[i] - d.d) - z)
     end
 
+    k_new = has_zeros ? findfirst(j -> j == 0, m) : K
```
devmotion: As mentioned above, if `first_zero` is obtained at the beginning of this function this could be changed to
```suggestion
    k_new = first_zero === nothing ? K : first_zero
```

src/stdlib/RandomMeasures.jl
devmotion: IMO one should keep `Real` - using `AbstractFloat` would remove the possibility to use `Integer`s, dual numbers from ForwardDiff, and tracked reals from Tracker as concentration parameter.

src/stdlib/RandomMeasures.jl
trappmartin: This turns out to be significantly slower. So I'll keep the current version.

devmotion: Ah interesting. Maybe that's due to most of the entries being positive actually? One could both avoid the unneeded checks and the use of `findall` by instead using
```julia
if first_zero === nothing
    @inbounds for i in 1:length(m)
        table[i] = ....
    end
else
    @inbounds for i in 1:length(m)
        mi = m[i]
        mi > 0 && (table[i] = ....)
    end
end
```
That should not lead to any performance regressions for `first_zero === nothing` and I would assume that is faster for `first_zero !== nothing` (since `findall` has to perform those checks as well and additionally creates an array).

trappmartin: This seems to depend a lot on the number of zero entries. If it's 50/50, findall performs better than your proposal while it, of course, slows down in case we have very few zeros.

trappmartin: This feels like over optimising stuff.

devmotion: > This seems to depend a lot on the number of zero entries. If it's 50/50, findall performs better than your proposal while it, of course, slows down in case we have very few zeros.

I'm confused - as expected, in my benchmarks `findall` is always significantly slower than just checking each entry in a for loop, regardless of the number of zeros:
```julia
julia> using BenchmarkTools

julia> function f_findall!(out, x)
          @inbounds for i in findall(xi -> xi > 0, x)
              out[i] = log(x[i])
           end
           out
       end

julia> function f_if!(out, x)
          @inbounds for i in 1:length(x)
              xi = x[i]
              xi > 0 && (out[i] = log(xi))
           end
           out
       end
julia> out = Vector{Float64}(undef, 10);

julia> x = vcat(zeros(5), rand(5));

julia> @btime f_findall!($out, $x);
  194.050 ns (7 allocations: 352 bytes)

julia> @btime f_if!($out, $x);
  56.534 ns (0 allocations: 0 bytes)

julia> y = rand(10);

julia> @btime f_findall!($out, $y);
  275.958 ns (8 allocations: 496 bytes)

julia> @btime f_if!($out, $y);
  90.970 ns (0 allocations: 0 bytes)

julia> z = vcat(0, rand(9));

julia> @btime f_findall!($out, $z);
  267.887 ns (8 allocations: 496 bytes)

julia> @btime f_if!($out, $z);
  83.254 ns (0 allocations: 0 bytes)

julia> a = zeros(10);

julia> @btime f_findall!($out, $a);
  60.132 ns (4 allocations: 144 bytes)

julia> @btime f_if!($out, $a);
  16.722 ns (0 allocations: 0 bytes)

julia> out_large = Vector{Float64}(undef, 100);

julia> x_large = vcat(zeros(50), rand(50));

julia> @btime f_findall!($out_large, $x_large);
  910.286 ns (10 allocations: 1.28 KiB)

julia> @btime f_if!($out_large, $x_large);
  412.361 ns (0 allocations: 0 bytes)

julia> y_large = rand(100);

julia> @btime f_findall!($out_large, $y_large);
  1.677 μs (11 allocations: 2.34 KiB)

julia> @btime f_if!($out_large, $y_large);
  746.392 ns (0 allocations: 0 bytes)

julia> z_large = vcat(0, rand(99));

julia> @btime f_findall!($out_large, $z_large);
  1.829 μs (11 allocations: 2.34 KiB)

julia> @btime f_if!($out_large, $z_large);
  745.020 ns (0 allocations: 0 bytes)

julia> a_large = zeros(100);

julia> @btime f_findall!($out_large, $a_large);
  132.950 ns (4 allocations: 144 bytes)

julia> @btime f_if!($out_large, $a_large);
  81.840 ns (0 allocations: 0 bytes)
```
So to me these benchmarks seem to indicate that one shouldn't use `findall` in this case, since the alternative is significantly faster.

Issue #1098, opened 2020-02-09
DynamicHMC errors on 1.3


sethaxen:
On Julia 1.3, when I call

```julia
julia> using Turing

julia> using Soss
┌ Warning: Package Turing does not have Pkg in its dependencies:
│ - If you have Turing checked out for development and have
│   added Pkg as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Turing
│ Loading Pkg into Turing from project dependency, future warnings for Turing are suppressed.
└ @ nothing nothing:904
┌ Warning: Error requiring DynamicHMC from Turing:
│ KeyError: key "DynamicHMC" not found
```

There seem to be a few problems here. The relevant Turing code:

```julia
@init @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" @eval Inference begin
    using Pkg; 
    Pkg.installed()["DynamicHMC"] < v"2.0" && error("Please upgdate your DynamicHMC, v1.x is no longer supported")
    using ..Turing.DynamicHMC: DynamicHMC, mcmc_with_warmup
    include("contrib/inference/dynamichmc.jl")
end
```

As the message notes, `Pkg` is not in Turing's dependencies. Also, DynamicHMC is a dependency of Soss and is not directly installed, hence it triggers Requires but is not in `Pkg.installed()`.

devmotion:
I guess one simple solution would be to just check if `DynamicHMC.mcmc_with_warmup` exists, which is the new entry point for DynamicHMC introduced in version 2.0.0 and used by Turing. I.e., a fix might be something like
```julia
@init @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" @eval Inference begin
    isdefined(..Turing.DynamicHMC, :mcmc_with_warmup) || error("Please update DynamicHMC, v1.x is no longer supported")
    using ..Turing.DynamicHMC: DynamicHMC, mcmc_with_warmup
    include("contrib/inference/dynamichmc.jl")
end
```

PR #1099, opened 2020-02-09
Fix errors with DynamicHMC

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1098, tested it locally on my computer.

@sethaxen Can you confirm that this PR fixes your issue?


sethaxen:
I can't mutually install this PR and Soss#master on my machine due to incompatible compatibility requirements with DistributionsAD, but I can confirm that when DynamicHMC is imported with this PR, everything seems to work fine. Thanks for the fix!

yebai:
Looks good, many thanks, @devmotion!

PR #1116, opened 2020-02-17
Update to new AbstractMCMC API

devmotion:
This PR requires a new release of AdvancedMH (see https://github.com/TuringLang/AdvancedMH.jl/pull/18).

Moreover, it supersedes https://github.com/TuringLang/Turing.jl/pull/1112, https://github.com/TuringLang/Turing.jl/pull/1114, and https://github.com/TuringLang/Turing.jl/pull/1111.


devmotion:
Hmm I don't understand why the `resume` tests would fail on Julia 1.1 and 1.2 but pass on Julia 1.0 and 1.3 with the same sets and versions of packages?!

cpfiffer:
I'm not sure either, but I don't know how much it matters. Honestly I think we should consider dropping CI for 1.1 and 1.2, since those are neither LTS nor the most current release. Anyone have thoughts on that?

devmotion:
The problem seems to be that, e.g., on Julia 1.2 `c` and `newchain` are both a `Chains` but still `vcat(c, newchain)` returns an array in https://github.com/devmotion/Turing.jl/blob/407594230a8dc0a397bad8cf9dc28d586bb2bc51/src/inference/Inference.jl#L416. I have no clue why this is happening since at the same time with Julia 1.2 I get
```julia
julia> using MCMCChains: Chains

julia> a = Chains(rand(10, 3, 1));

julia> b = Chains(rand(5, 3, 1));

julia> vcat(a, b)
Object of type Chains, with data of type 15×3×1 Array{Float64,3}

...
```
as expected.

devmotion:
Seems I can fix the error on Julia 1.2 by replacing `vcat(c, newchain)` with `cat(c, newchain; dims=1)`. However, I still don't understand what's the underlying problem.

src/inference/AdvancedSMC.jl
```diff
@@ -238,18 +238,15 @@ function AbstractMCMC.sample_end!(
...
-    # Set the default for resuming the sampler.
-    resume_from = get(kwargs, :resume_from, nothing)
-
     # Exponentiate the average log evidence.
     # loge = exp(mean([t.le for t in ts]))
     loge = mean(t.le for t in ts)
 
     # If we already had a chain, grab the logevidence.
-    if resume_from !== nothing   # concat samples
-        @assert resume_from isa Chains "resume_from needs to be a Chains object."
```
cpfiffer: I think we should leave in the `@assert` statement, or at least throw an error when the value of `resume_from` is not a `Chains`, so people know that resume functionality only works for MCMCChains objects.

devmotion:
OK, don't merge this. I figured out what's going on and will fix it in MCMCChains.

Project.toml
```diff
@@ -29,19 +29,18 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
...
 Distributions = "0.21.11, 0.22"
 DistributionsAD = "0.2"
-DynamicPPL = "0.3"
-FiniteDifferences = "0.9"
+DynamicPPL = "0.4"
 ForwardDiff = "0.10.3"
 Libtask = "0.3.1"
 LogDensityProblems = "^0.9, 0.10"
-MCMCChains = "~0.4.0, 1.0"
+MCMCChains = "2"
```
cpfiffer:
```suggestion
MCMCChains = "3"
```

devmotion:
The test errors seem to be the same as on master, which are supposed to be fixed by https://github.com/TuringLang/Turing.jl/pull/1128.

Issue #1115, opened 2020-02-16
Failure when dimension changes


cpfiffer:
From Slack, this causes an error in Turing 8.3:

```julia
using Turing

data =  [
    0.388235  0.388235  0.388235  0.388235;
    0.388235  0.388235  0.388235  0.388235;
    0.388235  0.388235  0.388235  0.388235;
]

@model mixture(x) = begin
    n = size(x, 2)
    γ = 1
    # Number of components and component attributions:
    k ~ Geometric(.1)
    π ~ Dirichlet(k+1, γ)
    y = tzeros(Int, n)
    for i in 1:n
        y[i] ~ Categorical(π)
    end
    # Component parameters:
    u = tzeros(Float64, (3,k+1))
    v = tzeros(Float64, (3,k+1))
    for i in 1:3
        for j in 1:k+1
            u[i,j] ~ Uniform()
            v[i,j] ~ Gamma()
        end
    end

    # Draw observation:
    for i in 1:3
        for j in 1:n
            α = u[i,y[j]] * v[i,y[j]]
            β = (1 - u[i,y[j]]) * v[i,y[j]]
            x[i,j] ~ Beta(α, β)
        end
    end
end

model = mixture(data)
chain = sample(model, MH(), 2)
```

Error:

```julia
ERROR: LoadError: DimensionMismatch("Inconsistent array dimensions.")                                                                                                                                                                                          
Stacktrace:                                                                                                                                                                                                                                                     [1] logpdf at /home/cameron/.julia/packages/Distributions/kPXE9/src/multivariates.jl:193 [inlined]                                                                                                                                                             [2] assume(::DynamicPPL.Sampler{MH{()},Turing.Inference.MHState{DynamicPPL.VarInfo{NamedTuple{(:k, :π, :y, :u, :v),Tuple{DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:k},Int64},Array{Geometric{Float64},1},Array{DynamicPPL.VarName{:k},1},Array{Int64,1},Ar$
ay{Set{DynamicPPL.Selector},1}},DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:π},Int64},Array{Dirichlet{Float64},1},Array{DynamicPPL.VarName{:π},1},Array{Float64,1},Array{Set{DynamicPPL.Selector},1}},DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:y},Int64},$
[...]

PR #1117, opened 2020-02-17
Use reset_num_produce!

devmotion:
Fixes the bug I mentioned in https://github.com/TuringLang/Turing.jl/issues/1115#issuecomment-586934847.


yebai:
Thanks @devmotion for the fix!

PR #1119, opened 2020-02-19
Reduce benchmark related comment reply

KDr2:
Now, reply only when `runbenchmarks(...)` command received.


.github/workflows/MicroBenchmarks.yml
```diff
@@ -5,8 +5,8 @@ on:
     types: [created]
   pull_request_review_comment:
     types: [created]
-  pull_request:
-    types: [opened, synchronize, reopened]
+  # pull_request:
+  #   types: [opened, synchronize, reopened]
```
KDr2: test!

Issue #1113, opened 2020-02-15
Build an offical docker for Turing


xukai92:
- Make it available at https://hub.docker.com/

KDr2:
@xukai92 had you started on this issue? If not I could do this. Do we have an Org account on hub.docker.com now? My account is `kdr2`, I'll appreciate it if you add me as a member 😄

xukai92:
I haven't. Please feel free to take over this issue. We created one org for Turing and added you :)

KDr2:
OK, I will do it this or next week.

PR #1120, opened 2020-02-20
Dockerfile for the Turing (latest, 0.8.3)

KDr2:
#1113

### DIFF
--- a/docker/latest/Dockerfile
+++ b/docker/latest/Dockerfile
@@ -1,0 +1,12 @@
+FROM julia:1.3.1-buster
+LABEL maintainer="ZHUO Qingliang <zhuo.dev@gmail.com>"
+
+RUN true \
+    && julia -e 'using Pkg; Pkg.add("Turing");' \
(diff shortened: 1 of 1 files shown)


docker/latest/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM julia:1.3.1-buster
```
yebai: Any reason for `docker/latest/Dockerfile` instead of `docker/Dockerfile`?

KDr2: the subfolder is recognized as the tag of the docker image, for this image, the tag will be `latest`, which is the default one when `docker pull turinglang/turing` is used.

We can provide other folders for different versions of Turing, like `0.8.4`, `0.8.5`, if our docker file is accepted by the official docker image repo, these tags will appear on the main page of the image, like this:
https://hub.docker.com/_/perl, the section "Supported tags and respective Dockerfile links".

PR #1126, opened 2020-02-24
Update to AbstractMCMC 0.4 and MCMCChains 4.0

cpfiffer:
Various internal fixes to keep Turing up-to-date with upstream practices. Tests won't pass until MCMCChains 0.4 goes through and is released.

I've updated Turing's version to 0.9, but I don't know if it should be 0.8.4 instead. I don't think there's any breaking changes in Turing but there were breaking changes in AbstractMCMC 0.4.


devmotion:
Did you see https://github.com/TuringLang/Turing.jl/pull/1116? That should be fine to merge after reverting the `vcat` change, but maybe you added more changes here?

cpfiffer:
Nope, I had forgotten! I like yours better.

PR #1128, opened 2020-02-24
Remove TrackedArray implementation and update DistributionsAD

devmotion:
This PR removes a quite general definition of TrackedArray that causes test failures on master. Moreover, it removes some methods that were moved to DistributionsAD. Unfortunately, not all of them could be removed since the implementation of `params` was removed again from DistributionsAD. After https://github.com/TuringLang/DistributionsAD.jl/pull/35 is merged and released, this implementation can be removed from Turing as well.


yebai:
Thanks, @devmotion!

torfjelde:
I'm running into the following issue now:
```julia
using Turing

xs = randn(100);

@model demo(x) = begin
    μ ~ MvNormal(zeros(length(x)), ones(length(x)))
    for i in eachindex(x)
        x[i] ~ Normal(μ[i], 1.)
    end
end

m = demo(xs)


Turing.setadbackend(:tracker)
advi = ADVI(10, 1000)
q = vi(m, advi)
```
results in the following error:
```
MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}
Closest candidates are:
  convert(::Type{T}, !Matched::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, !Matched::Factorization) where T<:AbstractArray at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/factorization.jl:55
  convert(::Type{T}, !Matched::T) where T at essentials.jl:171
  ...
Stacktrace:
 [1] setproperty!(::Tracker.Tracked{SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}, ::Symbol, ::Array{Float64,1}) at ./Base.jl:34
 [2] back(::Tracker.Tracked{SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}, ::Array{Float64,1}, ::Bool) at /home/tor/.julia/packages/Tracker/cpxco/src/back.jl:53
 [3] (::Tracker.var"#13#14"{Bool})(::Tracker.Tracked{SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}, ::Array{Float64,1}) at /home/tor/.julia/packages/Tracker/cpxco/src/back.jl:38
...
```

This is the original issue which made me introduce the implementation of custom adjoint for `SubArray`. But I feel like we discussed this at some point @devmotion, right? I can't find our discussion though, so uncertain where it was.

devmotion:
> But I feel like we discussed this at some point @devmotion, right?

Hmm yes I have the same impression. Maybe it was on Slack?

I'm wondering why you end up with subarrays in the first place. Do you have any clue?

torfjelde:
> Hmm yes I have the same impression. Maybe it was on Slack?

Was thinking the same but haven't been able to find it :confused: 

> Do you have any clue?

Can it be something with `VarInfo`? There's no use of `view` in the VI-parts of the code, so not clear to me what the reason is.

torfjelde:
So this came up again over at TuringTutorials. Has this been fixed somewhere?

devmotion:
Sorry, haven't had time to look into it (and also forgot about it :grimacing:)

devmotion:
If it's not in the VI part, the view might originate from DynamicPPL, e.g., https://github.com/TuringLang/DynamicPPL.jl/blob/477308cc159235e3afc79dea1f62baa81eef1ea8/src/varinfo.jl#L213 or https://github.com/TuringLang/DynamicPPL.jl/blob/477308cc159235e3afc79dea1f62baa81eef1ea8/src/varinfo.jl#L273.

torfjelde:
Aight, aight. But is this something we should put in DistributionsAD then? It seems like this adjoint fix for `SubArray` is necessary though, no?

PR #1131, opened 2020-02-25
Use concrete types for fields of ADVI

devmotion:


src/variational/advi.jl
```diff
@@ -104,8 +104,10 @@ end
 Automatic Differentiation Variational Inference (ADVI) for a given model.
 """
 struct ADVI{AD} <: VariationalInference{AD}
-    samples_per_step # number of samples used to estimate the ELBO in each optimization step
-    max_iters        # maximum number of gradient steps used in optimization
+    "Number of samples used to estimate the ELBO in each optimization step."
```
torfjelde: Should we maybe just have these docstrings in the docstring of `ADVI` itself? If you're wondering what fields `ADVI` has you generally do
```julia
?ADVI
```
and expect an answer. With these "per-field" docstrings, you have to actually instantiate `ADVI` before you can hit TAB to make auto-complete show you what the fields are, and _then_ you do `?ADVI.samples_per_step` :confused:

devmotion: Sure, that's fine with me, just kept them at their old place without thinking about it.

torfjelde: Yeah, I should have put them in the doc-string myself ages ago. But awesome! Thanks man:)

yebai: It is ok to keep docstrings in the type definition, there is an automated way of extracting them for documentations, see https://github.com/JuliaDocs/DocStringExtensions.jl/blob/e4bfe3537cc033c161d7e65bff776c15dcf3e9a4/src/abbreviations.jl#L37

devmotion: So far, however, Turing does not use DocStringExtensions - maybe it's a bit too much to change that just for ADVI?

yebai: That's true, but maybe it's a good time to start using it in other places? 

The API documentation for Turing is not very high-standard at the moment and could use some improvements. Any reason why dependency on `DocStringExtensions` might be a bad thing?

devmotion: I'm fine with using it in other places as well, just felt weird to only use it here. DocStringExtensions does not even depend on any external packages, so that should be fine as well.

devmotion: I added DocStringExtensions, resulting in
```julia
help?> ADVI
search: ADVI readavailable ischardev GeneralizedExtremeValue

  struct ADVI{AD} <: Turing.Variational.VariationalInference{AD}

  Automatic Differentiation Variational Inference (ADVI) with automatic differentiation backend AD.

  Fields
  ≡≡≡≡≡≡≡≡

    •    samples_per_step::Int64
      
        Number of samples used to estimate the ELBO in each optimization step.

    •    max_iters::Int64
      
        Maximum number of gradient steps.

  ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  ADVI([samples_per_step=1, max_iters=1000])

  Create an ADVI with the currently enabled automatic differentiation backend ADBackend().
```

yebai:
Thanks, @devmotion!

Issue #799, opened 2019-05-30
Juno Progress Metres


MikeInnes:
Turing's HMC sampler currently logs its progress in the terminal. While this is nice in a pure REPL, inside Juno it'd be even better to use the progress API. This has a number of nice features, including a built in ETA, as well as being able to work well with nested progress metres (e.g. in deep learning we have progress for both individual epochs and the overall training).

willtebbutt:
When you say the `Progress API`, are you referring to `ProgressMeters`, or something else?

yebai:
https://docs.junolab.org/latest/man/juno_frontend/#Progress-Meters-1

cpfiffer:
How does this work outside Juno? Couldn't quite tell from the link.

MikeInnes:
Currently it does nothing outside of Juno, so you'd want to keep the current behaviour depending on whether Juno is active.

cpfiffer:
Ah, okay. I think it's pretty cool, might be worth building into the sampling API for #746 once the bulk of the work is done there. It'll get rid of a lot of replicated code and help integrate us a bit better with the IDE ecosystem, however slightly.

yebai:
@MikeInnes Does progress API work for Jupyter notebooks?

devmotion:
Usually I don't work with notebooks so I can't say anything about the progress API support in that setting, but in general it seems it would be nice to make use of https://github.com/JunoLab/ProgressLogging.jl. With Juno one gets a nice progress bar, and on the REPL one could get the current `ProgressmMeter` output with the https://github.com/tkf/ConsoleProgressMonitor.jl backend. I used that combination successfully in a project some weeks ago.

devmotion:
This was solved by https://github.com/TuringLang/Turing.jl/pull/1134.

PR #1134, opened 2020-02-26
Use ProgressLogging instead of ProgressMeter

devmotion:
This PR replaces ProgressMeter with [ProgressLogging](https://github.com/JunoLab/ProgressLogging.jl), which is compatible with both Juno and [TerminalLoggers](https://github.com/c42f/TerminalLoggers.jl) in the terminal. It addresses https://github.com/TuringLang/Turing.jl/issues/799.

The branch works locally together with https://github.com/TuringLang/AbstractMCMC.jl/pull/17. The only regression so far is that the HMC statistics are not displayed anymore during sampling. I'm not sure if there is a natural or good way to achieve this with ProgressLogging, at least there is no discussion of such functionality in the documentation. Another regressions for users that do not use Juno might be that they have to create, e.g., a TerminalLogger explicitly.

Note that this implementation does not use the updated ProgressLogging API since it does not seem to be fully supported by the progress monitors yet.


cpfiffer:
Can we remove the draft flag for this PR now that MCMCChains is using this change as well?

devmotion:
I'm planning to do that within the next hours. I would like to enforce AbstractMCMC 0.5, but that requires a new release of DynamicPPL that is compatible with AbstractMCMC. I've updated the Turing tests in DynamicPPL with the ones from this PR, but I have to wait for the new MCMCChains release due to compatibility issues.

cpfiffer:
I think this is good to merge in.

Issue #1133, opened 2020-02-26
Failure working with MH sampler and `MvNormal`


kir19890817:
A simple model which used to work before v0.9.0:
```julia
using Turing
@model M(mu, sigma, observable) = begin
       z ~ MvNormal(mu, sigma)
       observable ~ Bernoulli(cdf(Normal(), z.*z))
end
```
 which I call with the MH sampler
```julia
sample(M(zeros(2), ones(2), 1), MH(), 1000)
```
and I obtain the following:
```julia
ERROR: MethodError: no method matching logpdf(::MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Array{Float64,1}}, ::Float64)
Closest candidates are:
  logpdf(::Chernoff, ::Real) at /home/kir/.julia/packages/Distributions/kPXE9/src/univariate/continuous/chernoff.jl:169
  logpdf(::Flat, ::Real) at /home/kir/.julia/packages/Turing/azHIm/src/stdlib/distributions.jl:13
  logpdf(::Arcsine{T}, ::Real) where T<:Real at /home/kir/.julia/packages/Distributions/kPXE9/src/univariate/continuous/arcsine.jl:83
  ...
Stacktrace:
 [1] logpdf(::AdvancedMH.Proposal{AdvancedMH.Static,MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Array{Float64,1}}}, ::Float64) at /home/kir/.julia/packages/AdvancedMH/ieEVJ/src/proposal.jl:17
 [2] q(::AdvancedMH.Proposal{AdvancedMH.Static,MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Array{Float64,1}}}, ::Float64, ::Array{Float64,1}) at /home/kir/.julia/packages/AdvancedMH/ieEVJ/src/proposal.jl:92
[...]

PR #1135, opened 2020-02-26
Fix MH indexing bug

cpfiffer:
This PR fixes https://github.com/TuringLang/Turing.jl/issues/1133 by fixing a case where the Turing-side code was not passing scalars correctly to AdvancedMH. 

Currently the way I handle passing NamedTuples into Turing is not very robust (and could be more performant), so I'm doubly open to comments on this one. `set_namedtuple!` for example is a very messy and ugly function that relies far too much on heuristics.

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -68,13 +68,17 @@ function set_namedtuple!(vi::VarInfo, nt::NamedTuple)
         n_vns = length(vns)
         n_vals = length(vals)
         v_isarr = vals isa AbstractArray
-
+        
         if v_isarr && n_vals == 1 && n_vns > 1
             for (vn, val) in zip(vns, vals[1])
                 vi[vn] = val isa AbstractArray ? val : [val]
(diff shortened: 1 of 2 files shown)


devmotion:
I'm not sure if this can be handled in a completely "nice" way at the moment. Maybe something along the lines of https://github.com/TuringLang/DynamicPPL.jl/issues/5 would be helpful.

cpfiffer:
Looks like the test failures are due to nondeterminism in the VI suite. I'll open a separate PR to fix the test.

cpfiffer:
I think this can go through, at the very least to fix #1133. If the comments in https://github.com/TuringLang/DynamicPPL.jl/issues/5 are addressed, I can come back to the general approach that AdvancedMH uses to integrate into Turing.

yebai:
Maybe add a test for the error case?

cpfiffer:
Updated to include the test for the error.

PR #1136, opened 2020-02-26
Add seed to VI test

cpfiffer:
Adds a seed to a VI test that's failing non-deterministically.


cpfiffer:
Nevermind, @devmotion has this in #1134.

PR #1138, opened 2020-02-27
Update dockerfile for new release

KDr2:


docker/0.8.3/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM julia:1.3.1-buster
+LABEL maintainer="The Turing team"
```
tianon: See https://github.com/docker-library/official-images/issues/3540#issuecomment-530925319, especially:

> We don't actively recommend using labels. If an image maintainer wants to have labels, that is fine, but label names should adhere to the image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md

KDr2: labels removed.

PR #1146, opened 2020-03-05
Tidy up benchmark files

### LINKED ISSUE #1144: Tidy up benchmarking scripts
The following scripts can be simplified/removed after the recent adoption of Nanosoilders in #1142 

- https://github.com/TuringLang/Turing.jl/blob/master/benchmarks/ns-utilities.jl
- https://github.com/TuringLang/Turing.jl/blob/master/benchmarks/local-runner
- https://github.com/TuringLang/Turing.jl/blob/master/benchmarks/config.toml
- https://github.com/TuringLang/Turing.jl/blob/master/benchmarks/BenchmarkHelper.jl

KDr2:
Fix #1144.


KDr2:
@BayesBot `runbenchmarks(ALL, vs=":master")`

benchmarks/BenchmarkHelper.jl
```diff
@@ -1,129 +1,21 @@
...
-    elseif haskey(event_data["comment"], "commit_id")
-        comment_kind = :commit
-        comment_parent = event_data["comment"]["commit_id"]
-    end
-
-    create_comment(repo, comment_parent, comment_kind;
-                   params=params, auth=auth)
-end
-
 function benchmark_excluded(file)
```
yebai: Do we still need this, now that we can use `tags` from BenchmarkTools to select benchmarks?

benchmarks/config.toml
```diff
@@ -1,9 +1,7 @@
 # Excludes
 EXCLUDES = [
+    "turing_benchmark.jl",
```
yebai: Similar to above, maybe remove this since we can now use BenchmarkTools' tag to select benchmarks.

KDr2: I think we'd better keep this:

1. if there are broken benchmarks (by broken I mean it will error even only load it), we can use this to temporarily ignore the file, then wait for someone to fix it
2. some helper files, like "tutring_benchmark.jl" and "BenchmarkHelper.jl", are not necessary to load in the process which benchmarks are running, for performance and for pkg dependency concerns, we'd better ignore them as well.

PR #1149, opened 2020-03-06
Provide default progress loggers

devmotion:
This PR updates the progress logging of VI to the updated ProgressLogging API and makes use of the new `@ifwithprogresslogger` macro in AbstractMCMC for selecting a default visualization backend in case the none is provided by user.


yebai:
thanks @devmotion!

PR #1151, opened 2020-03-09
Move default model evaluation code to DynamicPPL

phipsgabler:
Corresponding changes to [DynamicPPL#43](https://github.com/TuringLang/DynamicPPL.jl/pull/43#).  Removes the fallback code for `assume`/`observe`/`tilde`, `NoDist` and `NamedDist`, and fixes the leftovers.

`assume`/`observe` and `NamedDist` are still reexported.

I tested this with my local DynamicPPL in dev mode, but there still might be something overlooked concerning package things or exports.


phipsgabler:
Of course the tests fail, since they have to be performed against the corresponding updates in DPPL.

mohamed82008:
I think this is ready to merge.

yebai:
Many thanks, @phipsgabler!

PR #1156, opened 2020-03-11
Disable progress logs in tests

devmotion:
It seems the output in the tests is too verbose.


devmotion:
Progress logs are all disabled now but tests still fail since they use Libtask 0.3.2 (the issues are supposed to be fixed in 0.3.3).

yebai:
thanks @devmotion!

PR #1160, opened 2020-03-15
Add compiler doc to website

### LINKED ISSUE #1159: Add compiler doc to navigation
It seems the compiler doc is not currently listed anywhere on our doc site. Maybe we should add a link in the navigation bar. 

https://github.com/TuringLang/Turing.jl/blob/e0ccf328d07bc89d36953a468844477affd9a45d/docs/src/using-turing/compiler.md

cpfiffer:
Adds the compiler manual to the TOC on the website, fixes #1159.

### DIFF
--- a/docs/_data/toc.yml
+++ b/docs/_data/toc.yml
@@ -13,9 +13,6 @@
   - title: "Advanced Usage"
     url: "docs/using-turing/advanced"
 
-  - title: "Interface Guide"
-    url: "docs/using-turing/interface"
-
   - title: "Automatic Differentiation"
     url: "docs/using-turing/autodiff"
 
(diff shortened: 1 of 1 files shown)


docs/_data/toc.yml
```diff
@@ -28,6 +28,9 @@
   - title: "Sampler Visualization"
     url: "docs/using-turing/sampler-viz"
 
+  - title: "Turing Compiler Design"
```
yebai: The compiler docs are fairly advanced and may not be useful for users. Maybe add an `Internal` or `For Developers` section, instead of putting compiler docs under `Using Turing`?

cpfiffer: I added a separate section to the TOC, so this will hold the interface guide and the compiler guide.

PR #1166, opened 2020-03-18
Save transitions of samplers in Gibbs sampling

devmotion:
I noticed that currently the Gibbs sampler just forwards the transition of the last sampling step (of type `Transition`) to all subsamplers. However, this is problematic if a subsampler expects a certain type of sampler-specific transitions. To resolve this issue I added a `GibbsTransition` that saves the transitions of the subsamplers and forwards them to the corresponding subsamplers.

### DIFF
--- a/src/inference/gibbs.jl
+++ b/src/inference/gibbs.jl
@@ -98,6 +98,30 @@ function Sampler(alg::Gibbs, model::Model, s::Selector)
     return spl
 end
 
+"""
+    GibbsTransition
+
+Fields:
+- `θ`: The parameters for any given sample.
+- `lp`: The log pdf for the sample's parameters.
+- `transitions`: The transitions of the samplers.
+"""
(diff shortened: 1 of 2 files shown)


devmotion:
I just thought of it now, but maybe we should implement `transitions_save!` and `transitions_init` so that actually only a vector of `Transition`s with the parameters and log probabilities after each full sampling step is created during sampling but not all transitions of the subsamplers are stored?

cpfiffer:
As in, just storing the samples at `i-1` at the Gibbs level? I think that's probably easier on memory usage, so I'm fine with that.

devmotion:
Basically, in https://github.com/TuringLang/AbstractMCMC.jl/blob/dc5a76f45acc3f3b1025ea5856808edc3f07c10a/src/AbstractMCMC.jl#L145-L151 only `transition` would be a `GibbsTransition` but `transitions` would be a vector of `Transition`s.

PR #1167, opened 2020-03-19
Fix MH sampler for arrays of distributions

devmotion:
I noticed that the test example for the MH sampler
```julia
@model M(mu, sigma, observable) = begin
    z ~ MvNormal(mu, sigma)

    m = Vector{Any}(undef, 2)
    m[1] ~ Normal(0,1)
    m[2] ~ InverseGamma(2,1)
    s ~ InverseGamma(2,1)

    observable ~ Bernoulli(cdf(Normal(), z'*z))

    1.5 ~ Normal(m[1], m[2])
    -1.5 ~ Normal(m[1], m[2])

    1.5 ~ Normal(m[1], s)
    2.0 ~ Normal(m[1], s)
end
```
that was added in https://github.com/TuringLang/Turing.jl/pull/1135 does not work correctly with Turing at the moment. The problem is that the MH sampler uses the prior distribution `Normal(0, 1)` of `m[1]` for `m`, i.e., also for `m[2]` (in the `logpdf` evaluation this leads to a deprecation warning that can be seen in the tests).

I changed the implementation so that an array of distributions with more than one element is forwarded completely to AdvancedMH. Moreover, I switched to `DynamicPPL.reconstruct` for unpacking samples of scalar-valued distributions and reshaping samples of matrix-valued distributions. Although these changes fix the bug described above (I added some tests for it), I'm pretty sure that one can still come up with other problematic examples. So far we just try to work around the issues that arise from DynamicPPL putting all samples into vectors.

Additionally, I removed the `MHState` which was only needed for keeping track of an `AdvancedMH.DensityModel`. Instead now the log density function is implemented explicitly as a callable struct that captures the model and the sampler and is not saved in the state anymore. That allows use to reuse a sampler with different models without having to define abstract fields in the state (additionally, the state had to be initialized with a dummy density model which is unsatisfying IMO).

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -114,15 +114,16 @@ function gen_logπ_mh(spl::Sampler, model)
     return logπ
 end
 
-function scalar_map(vi::VarInfo, vns::Vector{V}) where V<:VarName
-    vals = getindex(vi, vns)
-    if length(vals) == length(vns) == 1
-        # It's a scalar!
-        return vals[1]
-    else 
-        # Go home, vector, you're drunk.
-        return vals
-    end
+# unpack a vector if possible
+unvectorize(dists::AbstractVector) = length(dists) == 1 ? first(dists) : dists
+
+# possibly unpack and reshape samples according to the prior distribution 
+reconstruct(dist::Distribution, val::AbstractVector) = DynamicPPL.reconstruct(dist, val)
+function reconstruct(
+    dist::AbstractVector{<:UnivariateDistribution},
+    val::AbstractVector
+)
+    return val
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/inference/mh.jl
```diff
@@ -132,78 +150,41 @@ Returns two `NamedTuples`. The first `NamedTuple` has symbols as keys and distri
...
 end
 
-@generated function _val_tuple(metadata::NamedTuple, vi::VarInfo, vns::NamedTuple{names}) where {names}
-    length(names) === 0 && return :(NamedTuple())
-    expr = Expr(:tuple)
-    map(names) do f
-        push!(expr.args, Expr(:(=), f, :(scalar_map(vi, metadata.$f.vns))))
-    end
-    return expr
+function _val_tuple(
```
cpfiffer: Any reason the generated functions were removed? My understanding was that they were significantly faster for named tuple generation (perhaps a faulty understanding).

devmotion: I haven't benchmarked it, I guess I should do that :stuck_out_tongue: I just usually try to avoid generated functions (for reasons such as mentioned in [this discourse discussion](https://discourse.julialang.org/t/why-not-use-generated/29531/4?u=devmotion)), and it didn't feel necessary to use generated functions here since the same functionality could be achieved using `map`.

cpfiffer: I know @mohamed82008 has worked a lot with these. Mohamed, do you know how well generated functions benchmark when returning NamedTuples?

devmotion: I noticed that it's not possible to compare the old and the new version in a fair way since the logic changed (and is type unstable now since we want to unpack single distributions that are wrapped in an array but must not unpack arrays with multiple distributions). Instead, I compared a simple (but somewhat similar) example
```julia
@generated function square_generated(a::NamedTuple{names}, b::NamedTuple) where names
    length(names) === 0 && return :(NamedTuple())
    expr = Expr(:tuple)
    map(names) do f
        push!(expr.args, Expr(:(=), f, :(b.$f^2)))
    end
    return expr
end

function square_map(a::NamedTuple{names}, b::NamedTuple) where names
    return NamedTuple{names}(
        map(names) do f
            b[f]^2
        end
    )
end

function square_broadcast(a::NamedTuple{names}, b::NamedTuple) where names
    return NamedTuple{names}(
        getfield.(Ref(b), names).^2
    )
end
```
with
```julia
using BenchmarkTools
using Random

names = [Symbol.(randstring(4)) for _ in 1:100]
a = NamedTuple{Tuple(names)}(rand(100))
b = NamedTuple{Tuple(names[randperm(100)])}(rand(100)
```
I got
```julia
julia> @btime square_generated($a, $b);
  62.283 ns (0 allocations: 0 bytes)

julia> @btime square_map($a, $b);
  17.847 μs (303 allocations: 85.28 KiB)

julia> @btime square_broadcast($a, $b);
  14.124 μs (200 allocations: 81.25 KiB)
```
so at least in this example clearly the implementation with a generated function is the most efficient one, by far.

I'll revert it to an implementation with generated functions! :+1:

cpfiffer: I'm honestly surprised at how crazy good the generated function is -- I expected maybe twice as fast as the mapped version, but this is just crazy. Plus, no allocations!

PR #1172, opened 2020-03-22
[WIP] Gibbs conditionals

phipsgabler:
Introduces a sampler `GibbsConditional` that can be used to manually specify a -- surprise! -- Gibbs conditional for a variable:

```
sample(model, Gibbs(GibbsConditional(:λ, cond_λ), GibbsConditional(:m, cond_m)), N)
```

See also [this discussion](https://github.com/phipsgabler/LittleGibbs.jl/issues/3).

where the `cond_v` are functions from a `NamedTuple` of the conditioned variables (`(λ,)` for `m` and vice versa in the above example) to a `Distribution` (or probably just a `Sampleable` is enough).  [Example code](https://gist.github.com/phipsgabler/dfac25deceedd30c6642b8b6333797a0).

Currently, only one variable per sampler object is handled (as opposed to some other samplers).  Of course extension is possible, but I believe it is better to leave it this way and let `Gibbs` handle the sweeps.

Since this seemed too simple, I'm quite sure there's work left.  At least:

- [x] There's a lot of stuff (esp. `assume`/`observe`.) simply copied from MH -- can this be simplified or at least abstracted?
- [x] Documentation!
- [ ] Tests, naturally
  - [x] Basic feasibility (gdemo, mixture of Gaussians)
  - [ ] Check whether indexed variables really work
  - [ ] Mixing Gibbs conditionals with other samplers
- [ ] Doctests (discuss first)


src/inference/gibbs.jl
```diff
@@ -2,7 +2,7 @@
 ### Gibbs samplers / compositional samplers.
 ###
 
-const GibbsComponent = Union{Hamiltonian,MH,ESS,PG}
+const GibbsComponent = Union{Hamiltonian,MH,ESS,PG,GibbsConditional}
```
devmotion: I know it's not part of this PR but IMO we should definitely use traits here.

devmotion:
I'm wondering, isn't that somewhat similar to the proposals in the MH sampler design-wise? And only supposed to be used inside of a Gibbs sampler? If that's the case, I think the GibbsConditional is useful but should maybe not be a full sampler on its own.

phipsgabler:
> I'm wondering, isn't that somewhat similar to the proposals in the MH sampler design-wise? 

Absolutely, and in fact I wrote it by stripping down the code in MH.jl.  I initially thought about making it just a "pseudo sampler" to dispatch on, but I think in the current way the implementation is much more readable.

In principle, one could do dispatch in MH, but then there you have to provide the conditional in a different way, and make some twists to always accept (or calculate the acceptance probability anyway, for nothing).  

> And only supposed to be used inside of a Gibbs sampler? If that's the case, I think the GibbsConditional is useful but should maybe not be a full sampler on its own.

Yeah, @trappmartin already [has argued similarly](https://github.com/phipsgabler/LittleGibbs.jl/issues/3), too, and I agree in principle, but this was the easiest way that occured to me.  How could such a "pseudo-sampler" be implemend if not as a full sampler?  Dispatch and extra code within the `Gibbs` implementation?

src/inference/gibbs.jl
phipsgabler: True.  Are traits used anywhere else already?  If not, would make an issue or something to discuss the style or package we should use in the future?

phipsgabler:
(To be honest, this implementation was more of an exercise for me, to get my hands dirty with sampler implementations.  I actually don't have any particular interests/features I want to get to, really.)

trappmartin:
As Philipp said, this was also meant as an exercise. I think we can integrate the code into Gibbs later on.

devmotion:
> In principle, one could do dispatch in MH, but then there you have to provide the conditional in a different way, and make some twists to always accept (or calculate the acceptance probability anyway, for nothing) -- I guess it's an advantage to not have to evaluate the likelihood in every step?

Oh it just reminded me of the code I've seen there but I haven't actually thought about merging/converging anything in the implementations of Gibbs and MH samplers yet. I completely agree with you, to me it seems both are too different and one really wants to exploit the different structure and possibilities for optimization.

> How could such a "pseudo-sampler" be implemend if not as a full sampler? Dispatch and extra code within the Gibbs implementation?

I just had a look at the code of the Gibbs sampler again, and I think one possibility would be to replace all parts in gibbs.jl that currently assume that each subsampler is a fully fledged sampler with calls to Gibbs-specific functions, e.g., replace the call of `step!` in https://github.com/TuringLang/Turing.jl/blob/8cafa3114089d21070790f2b4e09e3d23e4bcbb8/src/inference/gibbs.jl#L175-L176 with `gibbsstep!`. Then one could have a specific implementation for GibbsConditional and one for samplers. Maybe one could/would want to actually wrap samplers in another struct to offload keeping track of the transition of the last step of each sampler to the samplers (to fix that issue I just introduced a GibbsTransition type, but actually it feels a bit more natural if that would not be part of the general Gibbs interface).

src/inference/gibbs.jl
devmotion: Hmm no I don't think so, there were some discussion about refactoring and improving the design of the Gibbs sampler but it might have been part of an older PR. Actually I think for now it could be as simple as defining a `isgibbscomponent(component)` function (or maybe `isgibbsconditional` or some other self-explaining name?) that returns `false` as default and is then checked in the constructor of `Gibbs` since I don't think we dispatch on `GibbsComponent` anywhere in the code. Then every (potentially user-defined) sampler that could be used as part of the Gibbs sampler could implement this function (and potentially in the future a full Gibbs interface).

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,168 @@
...
+
+function GibbsConditionalTransition(spl::Sampler{<:GibbsConditional})
+    θ = tonamedtuple(spl.state.vi)
+    return GibbsConditionalTransition(θ)
+end
+
+
+mutable struct GibbsConditionalState{V<:VarInfo} <: AbstractSamplerState
+    vi::V
+end
```
devmotion: You also don't need this, you could just use a simple `SamplerState` https://github.com/TuringLang/Turing.jl/blob/8cafa3114089d21070790f2b4e09e3d23e4bcbb8/src/inference/Inference.jl#L442-L444.

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,149 @@
+struct GibbsConditional{space, C} <: InferenceAlgorithm
```
trappmartin:
```suggestion
struct GibbsConditional{space, C<:Distribution} <: InferenceAlgorithm 
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,149 @@
...
+    verbose::Bool=true,
+    resume_from=nothing,
+    kwargs...
+)
+    # Resume the sampler.
+    set_resume!(spl; resume_from=resume_from, kwargs...)
+
+    # Get `init_theta`
+    initialize_parameters!(spl; verbose=verbose, kwargs...)
+end
```
devmotion: Actually I think you can remove `sample_init!` as well.
```suggestion
```
I think the default implementation in https://github.com/TuringLang/Turing.jl/blob/bcd5ab99c26ebb111d0accf9be2e51ce725abbc3/src/inference/Inference.jl#L184-L196 is sufficient.

src/inference/gibbs_conditional.jl
phipsgabler: That doesn't work -- the conditional is a function from the conditioned values to a `Distribution`.

phipsgabler:
Have a look now, I stripped down everything.  `GibbsConditional` is not an `InferenceAlgorithm` anymore, and only `gibbs_step!`, exactly as David suggested, remained.

src/inference/gibbs.jl
phipsgabler: I have made a branch for this solution: https://github.com/phipsgabler/Turing.jl/compare/phg/gibbs_conditionals...phipsgabler:phg/gibbs_component_trait.  But it doesn't feel better.  `GibbsComponent` isn't used anywhere except in the `Gibbs` constructor, but there it adds type safety that is lost with a trait. 

We could add a second `Gibbs` constructor dispatching on the trait, but that would only lead to more confusing type errors IMHO...

trappmartin:
Cool. Could you please add some tests.

src/inference/gibbs.jl
devmotion: I still think it makes sense if we check `isgibbscomponent` in an inner constructor of `Gibbs`, even if that's the only place where we (currently) would use it. The main advantage would be that users can allow their custom samplers to be used with the Gibbs algorithm - that can't be done right now since it is not possible to alter or redefine the type union `GibbsComponent` whereas after the change everyone could just implement `isgibbscomponent`.

trappmartin: I agree but I think this could be done in an extra PR.

phipsgabler: Sure, that was my plan, hence the extra branch.

devmotion: Sounds good!

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,131 @@
...
+    λ ~ Gamma(α₀, θ₀)
+    m ~ Normal(0, √(1 / λ))
+    x .~ Normal(m, √(1 / λ))
+end
+
+m = gdemo(x)
+
+sample(m, Gibbs(GibbsConditional(:λ, gdemo_cond_λ), GibbsConditional(:m, gdemo_cond_m)), 10)
+```
+"""
```
devmotion: Very good documentation!

Do you think it would make sense to write the code example as a doctest, as suggested in the [Julia docs](https://docs.julialang.org/en/latest/manual/documentation/)? BTW apparently it's also suggested to put code examples in a `# Examples` section, so maybe that could be changed.

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,131 @@
...
+
+
+"""
+    conditioned(θ::NamedTuple, ::Val{S})
+
+Extract a `NamedTuple` of the values in `θ` conditioned on `S`; i.e., all names of `θ` except for
+`S`, mapping to their respecitve values.
+
+`θ` is assumed to come from `tonamedtuple(vi)`, which returns a `NamedTuple` of the form
+```
```
devmotion:
```suggestion
```julia
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,131 @@
...
+t = (m = ([0.234, -1.23], ["m[1]", "m[2]"]), λ = ([1.233], ["λ"])
+```
+so this function does both the cleanup of indexing and filtering by name. `conditioned(t, Val{m}())`
+and `conditioned(t, Val{λ}())` will therefore return
+```
```
devmotion:
```suggestion
```julia
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,131 @@
...
+(λ = 1.233,)
+```
+and
+```
```
devmotion:
```suggestion
```julia
```

src/inference/gibbs_conditional.jl
phipsgabler: Aren't doctests a bad idea for programs with random output?  Or are they working with a fixed seed automatically?

trappmartin: Let's keep it as it is and have a discussion about using doctest in the code base on slack.

test/inference/gibbs.jl
```diff
@@ -106,4 +106,90 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+                z ~ arraydist(Categorical.(fill(π, N)))
+                x ~ arraydist(Normal.(μ[z], σ))
+                
+                return x
+            end
+
+            # see http://www.cs.columbia.edu/~blei/fogm/2015F/notes/mixtures-and-gibbs.pdf
+            function cond_z(c)
+                μ = c.μ
+                function ϖ(ξ)
```
trappmartin: Please use a slightly less unicode heavy notation here.

test/inference/gibbs.jl
```diff
@@ -106,4 +106,90 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+                end
+                return arraydist([Categorical(ϖ(x[n])) for n = 1:N])
+            end
+
+            function cond_μ(c)
+                z = c.z
+                n = [count(z .== k) for k = 1:K]
+                x̄ = [sum(x[z .== k]) / (n[k] == 0 ? Inf : n[k]) for k = 1:K]
+                λ̂ = [inv(n[k] / σ^2 + 1/λ^2) for k = 1:K]
+                μ̂ = [x̄[k] * (n[k]/σ^2) * λ̂[k] for k = 1:K]
```
trappmartin: Same here. and the lines above.

test/inference/gibbs.jl
phipsgabler: Sure! Now why have I learn Greek then :(

trappmartin: I would suggest to make a trade-off between readability and pretty notation. Using zeta is fine with me but more complex unicode letters (mu with hat and whatever) might be unnecessary.

phipsgabler: But are mu, sigma, and lambda OK?

trappmartin: Of course. I would say, everything that is simply a greek letter should be fine. But I would restrain from using `μ̂` and so on.

trappmartin: And I'm not sure it is necessary to use greek letters for function names. This was my initial reason for the comment here.

src/inference/gibbs_conditional.jl
devmotion: > Aren't doctests a bad idea for programs with random output? Or are they working with a fixed seed automatically?

It's not reproducible automatically but of course one can explicitly specify a RNG. From the docs:

> Calling rand and other RNG-related functions should be avoided in doctests since they will not produce consistent outputs during different Julia sessions. If you would like to show some random number generation related functionality, one option is to explicitly construct and seed your own MersenneTwister (or other pseudorandom number generator) and pass it to the functions you are doctesting.

test/inference/gibbs.jl
phipsgabler: > Using zeta is fine with me but more complex unicode letters (mu with hat and whatever) might be unnecessary.

That's xi.  Which is reason enough to change it :smile:

cpfiffer: Let's just agree that xi should never be used for anything, code or math.

phipsgabler: It's one of the nicest integration variables I know... way better than "anything prime" :P

trappmartin:
What is still open until the PR can be merged?

phipsgabler:
Some more tests with fancy indexing, and documentation for the user side (the stuff that will be rendered on the homepage).  How does the latter work?  Is it reusing docstrings, or does is need to be put somewhere separately?

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,113 @@
...
+        Random.seed!(100)
+        alg = Gibbs(GibbsConditional(:z, cond_z), GibbsConditional(:μ, cond_μ))
+        chain = sample(mixture(x), alg, 10000)
+        
+        check_numerical(chain, [:μ], [[0, 1]], atol=0.1)
+
+        # we can't be sure about the order of cluster assignment, so we check both
+        ẑ = dropdims(mean(chain[:z].value, dims=1), dims=(1, 3))
+        z_true = [fill(1, 10); fill(2, 10)]
+        @test isapprox(ẑ, z_true, atol=0.2, rtol=0.0) || isapprox(ẑ, z_true[[11:20; 1:10]], atol=0.2, rtol=0.0)
```
trappmartin, line 111:
```suggestion
        @test isapprox(ẑ, z_true, atol=0.2, rtol=0.0) || isapprox(ẑ, z_true[[11:20; 1:10]], atol=0.2, rtol=0.0)
        
        Random.seed!(100)
        alg = Gibbs(GibbsConditional(:z, cond_z), HMC(0.05, 4, :μ,))
        chain = sample(mixture(x), alg, 10000)

        check_numerical(chain, [:μ], [[0, 1]], atol=0.1)

        # we can't be sure about the order of cluster assignment, so we check both
        ẑ = dropdims(mean(chain[:z].value, dims=1), dims=(1, 3))
        z_true = [fill(1, 10); fill(2, 10)]
        @test isapprox(ẑ, z_true, atol=0.2, rtol=0.0) || isapprox(ẑ, z_true[[11:20; 1:10]], atol=0.2, rtol=0.0)
```

phipsgabler: On my system, this works for Tracker, but not ReverseDiff or ForwardDiff -- so I suspect AD is broken?  What would be an appropriate non-Hamiltonian sampler to compare HMC to in this case?

trappmartin: @mohamed82008 do you know why this would be broken for ForwardDiff?

trappmartin: You could use `MH`, but ideally we want to test against `HMC` as this will be the most frequently used scenario.

trappmartin:
> Some more tests with fancy indexing, and documentation for the user side (the stuff that will be rendered on the homepage). How does the latter work? Is it reusing docstrings, or does is need to be put somewhere separately?

Docstrings will be fine for now. I hope to be able to use this for the infinite mixture model tutorial as an example. We can add more examples later.

test/inference/gibbs_conditional.jl
phipsgabler: MH passes.  Which strongly indicates that AD is in fact the problem, I'd say.  I'll have a look at the implementation again to see whether I find anything suspicious.

phipsgabler: Ah, I'm using `foldl(vcat, arrays)`, which is discussed [over in ReverseDiff](https://github.com/JuliaDiff/ReverseDiff.jl/issues/122#issuecomment-609075734) already.

phipsgabler: Or maybe not... `f(xs) = sum(foldl(vcat, [xs[1:2]; xs[3:3]]))` does have a correct ReverseDiff gradient.

phipsgabler: I missread the test results, it's actually Reversediff and Tracker that fail, and ForwardDiff which works.  Both get the value of the first entry of \mu right, but the second one is off to the negative (wiggling around -3).

trappmartin:
107 files changed.. I guess something went wrong here..

phipsgabler:
Aaahhh rebase hell.

phipsgabler:
**Moved to https://github.com/TuringLang/Turing.jl/pull/1275 to get rid of the rebase mess.**

PR #1175, opened 2020-03-23
Catch Array{Any} return value

cpfiffer:
I haven't been able to run the infinite mixture model tutorial because it contains integer cluster assignments, and Julia is unable to determine that a more concrete type for `Array{Any, 2}` is `Array{Union{Real, Missing}}` or `Array{Union{Int64, Float74, Missing}}`. This is a quick fix to tell MCMCChains and others that the vector being given is a numeric type if the compiler is otherwise unable to determine the type of the array.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -334,7 +334,7 @@ function AbstractMCMC.bundle_samples(
 
     # Extract names & construct param array.
     nms = [nms; extra_params]
-    parray = hcat(vals, extra_values)
+    parray = map(x -> x, hcat(vals, extra_values))
 
     # If the state field has average_logevidence or final_logevidence, grab that.
     le = missing
@@ -355,9 +355,6 @@ function AbstractMCMC.bundle_samples(
     else
         info = NamedTuple()
     end
-
-    # Ensure that the chain is not of type Any.
-    parray = eltype(parray) == Any ? convert(Array{Union{Real, Missing}}, parray) : parray
 
     # Chain construction.
     return Chains(
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -356,6 +356,9 @@ function AbstractMCMC.bundle_samples(
         info = NamedTuple()
     end
 
+    # Ensure that the chain is not of type Any.
+    parray = eltype(parray) == Any ? convert(Array{Union{Real, Missing}}, parray) : parray
+
     # Chain construction.
     return Chains(
         parray,


devmotion:
I guess you could use `MCMCChains.concretize` here alternatively? That should not allocate in case the type is concrete (`map` always allocates).

cpfiffer:
`concretize` doesn't work, unfortunately -- I still get `Array{Any, 2}`.

src/inference/Inference.jl
```diff
@@ -356,6 +356,9 @@ function AbstractMCMC.bundle_samples(
         info = NamedTuple()
     end
 
+    # Ensure that the chain is not of type Any.
+    parray = eltype(parray) == Any ? convert(Array{Union{Real, Missing}}, parray) : parray
```
devmotion: Actually, I don't think this is a proper solution that should be part of Turing since there's no way for a user to correct this in case `Array{Union{Real,Missing}}` is not a suitable type. It's not guaranteed to work, and hence I think something like `map(identity, parray)` (which is included above already?) or `concretize` is the best you can do in such a general function. As long as the type inference problems are not fixed where they arise, IMO users should cast to the correct types themselves afterwards by adding a line such as `concretize(chains)` or `convert`. I think we should rather add a `convert(MCMCChains{T}, ...)` method to MCMCChains that would just cast the underlying array to the correct new type than forcing `Array{Union{Real,Missing}}` in all cases where type inference fails.

devmotion:
I'm surprised, the following works for me:
```julia
julia> using MCMCChains

julia> a = convert(Array{Any}, rand(4,5))
4×5 Array{Any,2}:
 0.973438   0.682132  0.368751  0.533482  0.592842 
 0.0320415  0.895776  0.855807  0.523576  0.550649 
 0.336734   0.941327  0.48445   0.191544  0.0839327
 0.485901   0.074632  0.597695  0.490765  0.560896 

julia> MCMCChains.concretize(a)
4×5 Array{Float64,2}:
 0.973438   0.682132  0.368751  0.533482  0.592842 
 0.0320415  0.895776  0.855807  0.523576  0.550649 
 0.336734   0.941327  0.48445   0.191544  0.0839327
 0.485901   0.074632  0.597695  0.490765  0.560896 

julia> map(identity, a)
4×5 Array{Float64,2}:
 0.973438   0.682132  0.368751  0.533482  0.592842 
 0.0320415  0.895776  0.855807  0.523576  0.550649 
 0.336734   0.941327  0.48445   0.191544  0.0839327
 0.485901   0.074632  0.597695  0.490765  0.560896 
```
Actually not too surprising that both work since `concretize` ends up calling `map(identity, a)` if the type is not concrete.

cpfiffer:
Your example doesn't catch the full use case, which is mixed integer and float types.

devmotion:
For me mixed integer and float types work as well, i.e., as expected I obtain `Array{Real}` in both cases:
```julia
julia> a = hcat(convert(Array{Any}, rand(4,2)), convert(Array{Any}, ones(Int, 4,3)))
4×5 Array{Any,2}:
 0.779229  0.70086   1  1  1
 0.221815  0.38549   1  1  1
 0.715204  0.187649  1  1  1
 0.112325  0.31379   1  1  1

julia> MCMCChains.concretize(a)
4×5 Array{Real,2}:
 0.779229  0.70086   1  1  1
 0.221815  0.38549   1  1  1
 0.715204  0.187649  1  1  1
 0.112325  0.31379   1  1  1

julia> map(identity, a)
4×5 Array{Real,2}:
 0.779229  0.70086   1  1  1
 0.221815  0.38549   1  1  1
 0.715204  0.187649  1  1  1
 0.112325  0.31379   1  1  1
```

cpfiffer:
It appears that the combination of `Missing`, `Int64`, and `Float64` make this difficult.

Here's a MWE:

```julia
z = Any[0.5 1; 0.5 missing]
MCMCChains.concretize(z)
```

Result:
```julia
2×2 Array{Any,2}:
 0.5  1
 0.5   missing
```

devmotion:
I investigated this a bit more, and I think there's a better solution if we want to solve exactly the problem of not obtaining `Union{Missing,Real}`.

https://discourse.julialang.org/t/small-union-type-failure/18640/8?u=devmotion provides an explanation for what is happening with `map` and `collect`. Both functions use `Base.promote_typejoin` to determine the resulting element type which is just a supertype of of all entries since they preserve the types of the entries. Other functions such as `cat` actually use `promote_type` since they promote all entries to a common type. Moreover, `promote_typejoin` was added (https://github.com/JuliaLang/julia/pull/25553) to preserve exactly `Union{T,Nothing/Missing}` and not to obtain `Any`. However, e.g.,
```julia
julia> Base.promote_typejoin(Real, Missing)
Any
```
whereas
```julia
julia> Base.promote_type(Real, Missing)
Union{Missing, Real}
```
Or for the example above:
```julia
julia> z = Any[0.5 1; 0.5 missing]
2×2 Array{Any,2}:
 0.5  1       
 0.5   missing

julia> mapreduce(typeof, Base.promote_typejoin, z)
Any

julia> mapreduce(typeof, promote_type, z)
Union{Missing, Float64}
```

So in case you actually want `Array{Union{Real,Missing}}` instead of `Array{Any}`, I think one workaround could be to evaluate `mapreduce(typeof, promote_type, z)` and depending on the result convert the array type. I'll try to change `concretize` accordingly.

devmotion:
With
```julia
function concretize(x::AbstractArray)
    if isconcretetype_recursive(typeof(x))
        return x
    else
        xnew = map(concretize, x)
        T = mapreduce(typeof, promote_type, xnew)
        if T <: eltype(xnew)
            return convert(AbstractArray{T}, xnew)
        else
            return xnew
        end
    end
end
```
I get
```julia
julia> z = Any[0.5 1; 0.5 missing]
2×2 Array{Any,2}:
 0.5  1       
 0.5   missing

julia> concretize(z)
2×2 Array{Union{Missing, Float64},2}:
 0.5  1.0     
 0.5   missing

julia> zz = [Any[0.5, 1, missing] for _ in 1:3]
3-element Array{Array{Any,1},1}:
 [0.5, 1, missing]
 [0.5, 1, missing]
 [0.5, 1, missing]

julia> concretize(zz)
3-element Array{Array{Union{Missing, Float64},1},1}:
 [0.5, 1.0, missing]
 [0.5, 1.0, missing]
 [0.5, 1.0, missing]

julia> zzz = Any[Any[0.5, 1, missing] for _ in 1:3]
3-element Array{Any,1}:
 Any[0.5, 1, missing]
 Any[0.5, 1, missing]
 Any[0.5, 1, missing]

julia> concretize(zzz)
3-element Array{Array{Union{Missing, Float64},1},1}:
 [0.5, 1.0, missing]
 [0.5, 1.0, missing]
 [0.5, 1.0, missing]
```

devmotion:
> In this part we drop the integer specification and convert it to a float, is this intended?

That's exactly what `promote_type` is intended for and used for in, e.g., `vcat`. I guess without any custom promotion rules checking if `promote_type` provides a more concrete type is the best one can do for obtaining a more concrete array type. Alternatively, one could start adding custom promotion rules (as suggested in the discourse discussion linked above); the drawback is that they will only fix the cases that we cover explicitly. More concretely, one could define
```julia
custom_promote_typejoin(::Type{S}, ::Type{T}) where {S,T} = Base.promote_typejoin(S, T)
custom_promote_typejoin(::Type{Missing}, ::Type{Real}) = Union{Real,Missing}
custom_promote_typejoin(::Type{Real}, ::Type{Missing}) = Union{Real,Missing}

function concretize(x::AbstractArray)
    if isconcretetype_recursive(typeof(x))
        return x
    else
        xnew = map(concretize, x)
        T = mapreduce(typeof, custom_promote_typejoin, xnew)
        if T <: eltype(xnew)
            return convert(AbstractArray{T}, xnew)
        else
            return xnew
        end
    end
end
```
which yields
```julia
julia> concretize(z)
2×2 Array{Union{Missing, Real},2}:
 0.5  1       
 0.5   missing

julia> typeof(concretize(z)[3])
Int64

julia> concretize(zz)
3-element Array{Array{Union{Missing, Real},1},1}:
 [0.5, 1, missing]
 [0.5, 1, missing]
 [0.5, 1, missing]

julia> concretize(zzz)
3-element Array{Array{Union{Missing, Real},1},1}:
 [0.5, 1, missing]
 [0.5, 1, missing]
 [0.5, 1, missing]
```

> I'm fine with doing this for now, but in general I think we need better support for mixed types. I started tooling around yesterday with a chain that uses NamedTuples on the backend, so types are freely allowed. It's quite nice so far.

I agree, IMO the proper way to deal with mixed types is to not merge them into a container with a common type in the first place. Alternatively, one could maybe just replace the AxisArray with a StructArray. In that case one could reuse a lot of existing functionality and does not have to come up with a custom implementation of the underlying container type.

cpfiffer:
I'd like to add this to MCMCChains and make a patch release. Currently, the most recent `concretize` function doesn't handle the use case in the tutorial for whatever reason. Here's a (not very minimum) working example:

```julia
using Revise
using Turing
using Turing.RandomMeasures
using Random

@model infiniteGMM(x) = begin
    
    # Hyper-parameters, i.e. concentration parameter and parameters of H.
    α = 1.0
    μ0 = 0.0
    σ0 = 1.0
    
    # Define random measure, e.g. Dirichlet process.
    rpm = DirichletProcess(α)
    
    # Define the base distribution, i.e. expected value of the Dirichlet process.
    H = Normal(μ0, σ0)
    
    # Latent assignment.
    z = tzeros(Int, length(x))
        
    # Locations of the infinitely many clusters.
    μ = tzeros(Float64, 0)
    
    for i in 1:length(x)
        
        # Number of clusters.
        K = maximum(z)
        nk = Vector{Int}(map(k -> sum(z .== k), 1:K))

        # Draw the latent assignment.
        z[i] ~ ChineseRestaurantProcess(rpm, nk)
        
        # Create a new cluster?
        if z[i] > K
            push!(μ, 0.0)

            # Draw location of new cluster.
            μ[z[i]] ~ H
        end
                
        # Draw observation.
        x[i] ~ Normal(μ[z[i]], 1.0)
    end
end

# Generate some test data.
Random.seed!(1)
data = vcat(randn(10), randn(10) .- 5, randn(10) .+ 10)
data .-= mean(data)
data /= std(data);

# MCMC sampling
Random.seed!(2)
iterations = 30
model_fun = infiniteGMM(data);
chain = sample(model_fun, SMC(), iterations);
```

I'm fine with using the promote type version (not the `custom_promote_type` version) that casts integers to floats -- are you alright if I open a PR upstream for this, or do you want to take another crack at allowing mixed integer/float types? I think it's probably not time well spent.

devmotion:
I made a PR and added the examples above as a test.

Issue #973, opened 2019-11-18
Test for RandomMeasures fail


trappmartin:
Tests in the distributions testset pass, however, all other testsets chinese restaurant processes, stick breaking, and size-based resampling error due to a KeyError when calling Libtask.consume(::Trace).

devmotion:
I noticed that sampling from these models works fine (without the KeyError) in the Julia REPL ([code](https://gist.github.com/devmotion/1fde42d174c6f53c9f7f749aaf4d6706)).

yebai:
Likely fixed by recent updates to `Libtask`. In addition, `TArray` will be removed after https://github.com/TuringLang/Turing.jl/pull/1858

PR #1180, opened 2020-03-25
Add concretize before generating Chains

cpfiffer:
Uses `MCMCChains.concretize` to determine type better. This fixes a bug in the infinite mixture tutorial.

I also switched us over to using `import MCMCChains` rather than using, since we only really use two things from that package (`Chains` and `concretize`).


Project.toml
```diff
@@ -45,7 +45,7 @@ EllipticalSliceSampling = "0.2"
 ForwardDiff = "0.10.3"
 Libtask = "0.3.1"
 LogDensityProblems = "^0.9, 0.10"
-MCMCChains = "3"
+MCMCChains = "^3.0.7"
```
devmotion: Since the caret specifier is the default, it's the same as
```suggestion
MCMCChains = "3.0.7"
```

cpfiffer:
I added the test. The `RandomMeasures.jl` tests weren't previously being run (see https://github.com/TuringLang/Turing.jl/issues/973), so I commented them out but added the infinite GMM model.

trappmartin:
Thanks!

PR #1197, opened 2020-04-03
Tutorials TOC links: more consistent naming

DilumAluthge:
Right now, the TOC links for the tutorials have inconsistent names: some of them use the word "Bayesian" and others do not.

Presumably, all of these tutorials are Bayesian, so it is redundant to include "Bayesian" in these TOC links.

### DIFF
--- a/docs/_data/toc.yml
+++ b/docs/_data/toc.yml
@@ -49,10 +49,10 @@
   - title: "Gaussian Mixture Models"
     url: "tutorials/1-gaussianmixturemodel"
 
-  - title: "Bayesian Logistic Regression"
+  - title: "Logistic Regression"
     url: "tutorials/2-logisticregression"
 
-  - title: "Bayesian Neural Networks"
+  - title: "Neural Networks"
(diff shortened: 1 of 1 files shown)


cpfiffer:
I'm fine with it. Anyone have any issues with the naming change?

yebai:
> I'm fine with it. Anyone have any issues with the naming change?

I think certain models are often associated with "Bayesian", e.g. Infinite Mixture Models. But some models, like logistic regression and neural networks, are not always the case. In the latter case, we do want to highlight that these tutorials are using Bayesian inference. I suggest we keep the name unless there is a stronger argument than consistency.

cpfiffer:
I suppose I would agree with that. The models that were changed are models that are very commonly non-Bayesian, so I think it's nice to point that out.

trappmartin:
I agree with Hong and would prefer to keep the prefix Bayesian if the model is not per se a Bayesian model.

yebai:
I'm closing this now. Please feel free to reopen if there is a need.

Issue #1199, opened 2020-04-03
CI tests fail on Julia 1.4 + master


devmotion:
I just noticed that the CI tests fail on Julia 1.4 + master (see e.g. https://travis-ci.com/github/TuringLang/Turing.jl/builds/158167819 which I reran on the current master branch of Turing) due to
```julia
julia: /buildworker/worker/package_linux64/build/src/codegen.cpp:4357: jl_cgval_t emit_expr(jl_codectx_t&, jl_value_t*, ssize_t): Assertion `token.V->getType()->isTokenTy()' failed.

signal (6): Aborted

in expression starting at /home/travis/build/TuringLang/Turing.jl/test/core/ad.jl:16
```
It seems to be AD related?

Googling revealed the following issue that was caused by the same failing assertion: https://github.com/JuliaLang/julia/issues/34247

devmotion:
Commenting out `include("core/ad.jl")` in test/runtests.jl fixes the issue, so it's definitely related to the AD tests.

devmotion:
OK, commenting out https://github.com/TuringLang/Turing.jl/blob/089b106f22a87f56312d525e99d395cbd339456f/test/core/ad.jl#L28 (and the subsequent lines that define and test `grad_Turing2`) fixes the issue. So there's some problem with the Zygote backend in Julia 1.4 + master.

devmotion:
@mohamed82008 Do you have any idea what's going on there?

devmotion:
The output related to Turing and DynamicPPL in the stack trace ends with
```julia
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2322
macro expansion at /home/david/.julia/packages/DynamicPPL/E1QSS/src/compiler.jl:348 [inlined]
macro expansion at /home/david/.julia/dev/Turing/test/test_utils/models.jl:12 [inlined]
##inner_function#577#43 at /home/david/.julia/packages/DynamicPPL/E1QSS/src/compiler.jl:493 [inlined]
_pullback at /home/david/.julia/packages/Zygote/KNUTW/src/compiler/interface2.jl:0
adjoint at /home/david/.julia/packages/Zygote/KNUTW/src/lib/lib.jl:167 [inlined] 
_pullback at /home/david/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
#_#5 at /home/david/.julia/packages/DynamicPPL/E1QSS/src/model.jl:24 [inlined]
[...]

PR #1202, opened 2020-04-04
Fix crashes in CI tests

devmotion:
This PR is a quick and dirty fix for https://github.com/TuringLang/Turing.jl/issues/1199 as long as the underlying issue is not resolved. Currently one Zygote AD test crashes Julia 1.4 and Julia master, which makes it impossible to evaluate any PRs on Julia 1.4.

Additionally it fixes a deprecation warning on Julia 1.4 (see https://github.com/TuringLang/Turing.jl/pull/1186#discussion_r398879717).


yebai:
Thanks, @devmotion!

PR #1212, opened 2020-04-08
Closes #1210 added working pdf for BinomialLogit and VecBinomialLogit…

### LINKED ISSUE #1210: StackOverflowError raised by the pdf of BinomialLogit
Hi, I raised this question in the Slack and was told to make a short PR for this bug:

```
pdf(BinomialLogit(1, 0.5), 1)
```
Results in:
```
ERROR: StackOverflowError:
Stacktrace:
 [1] pdf(::BinomialLogit{Float64,Int64}, ::Int64) at /Users/harrisonwilde/.julia/packages/Distributions/6HQ2U/src/univariates.jl:333 (repeats 80000 times)
```

The fix is to assign the following:
```
Distributions.pdf(d::BinomialLogit{<:Real}, k::Int) = exp(logpdf(d, k))
```

Where should this go in the source code? Additionally, I was wondering if it was worth defining `BernoulliLogit` as an explicit entity, to me it reads better than a `BinomialLogit` with `n=1` when doing logistic regression etc.

HarrisonWilde:
… to distributions.jl

Dealing with the bug mentioned in issue1210

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -107,7 +107,6 @@ export  @model,                 # modelling
         Flat,
         FlatPos,
         BinomialLogit,
-        VecBinomialLogit,
         OrderedLogistic,
         LogPoisson,
         NamedDist,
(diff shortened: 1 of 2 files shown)


HarrisonWilde:
Considering removing VecBinomialLogit functions as they are redundant, should I do here or make a separate PR?

devmotion:
I'd say it doesn't matter, whatever you prefer.

yebai:
Looks good to me, thanks @HarrisonWilde!

PR #1214, opened 2020-04-08
Closes #1213 added logpdf for BernoulliLogit

### LINKED ISSUE #1213: Add an explicit definition of the logit-parameterised Bernoulli distribution to Turing's distributions
Is there a need for this? I have added it to Turing for my own ease of use, will raise a PR, feel free to reject if not wanted.

HarrisonWilde:
…Logit to distributions.jl


devmotion:
I'm not sure if separate types are necessary here, wouldn't just another constructor be sufficient?

I also don't think one should define separate vector-valued distributions since one can define it easily as a product of univariate distributions (in fact, I think `VecBinomialLogit` should be removed).

HarrisonWilde:
That is fair enough, agreed on the vector distributions, I did think it unnecessary but thought I would follow the lead of BinomialLogit, should I remove all of them then?

And what do you mean by another constructor sorry?

devmotion:
> That is fair enough, agreed on the vector distributions, I did think it unnecessary but thought I would follow the lead of BinomialLogit, should I remove all of them then?

Just don't add VecBernoulliLogit, I'd say :smiley: 

> And what do you mean by another constructor sorry?

Sorry, I should have explained it better. I meant that maybe it would be sufficient to add the line
```julia
BernoulliLogit(logitp::Real) = BinomialLogit(1, logitp)
```
and export `BernoulliLogit`. Then users would be able to use `BernoulliLogit` but actually internally we would still use the `BinomialLogit` distribution, so it would just be an additional convenient constructor for `BinomialLogit`.

PR #1224, opened 2020-04-18
Update Turing to AbstractMCMC 1.0, AdvancedMH 0.5, and DynamicPPL 0.6

devmotion:
Make Turing compatible with all the new and [upcoming](https://github.com/TuringLang/DynamicPPL.jl/issues/73) releases.

Tests will fail until a new version of DynamicPPL is released.


devmotion:
I haven't updated the compiler docs yet, I thought we might want to do that separately in a more coordinated fashion (see https://github.com/TuringLang/DynamicPPL.jl/issues/65).

mohamed82008:
Let's also bump DistributionsAD to 0.5.

mohamed82008:
@torfjelde should we wait on this update until we sort out Bijectors then bump DAD and Bijectors versions as well? I think the new DAD-Bijectors versions will be technically breaking as the behavior of AD and bijectors AD will change.

mohamed82008:
By "wait", I mean not to release Turing yet. But we can merge the PR.

mohamed82008:
DPPL 0.6 is released.

devmotion:
So we update DistributionsAD in a separate PR?

mohamed82008:
> So we update DistributionsAD in a separate PR?

Yes

PR #1230, opened 2020-04-23
Add MLE/MAP functionality

cpfiffer:
This PR adds MLE and MAP functions to generate point estimates. I have not yet written tests, so this is still a WIP.

Usage:

```julia
@model function gdemo(x)
    s ~ InverseGamma(15,3)
    m ~ Normal(0, 1)

    for i in eachindex(x)
        x[i] ~ Normal(m, s)
    end

    return m, s
end

Random.seed!(100)
data = rand(Normal(0, 2), 10)
model = gdemo(data)

# Make point estimates
mle_gdemo = MLE(model)
map_gdemo = MAP(model)
```

MLE results:

```julia
2×3 Named Array{Float64,2}
A ╲ B │ parameter    std_err      tstat
──────┼────────────────────────────────
s     │   2.08011   0.465139    4.47202
m     │  0.596312   0.657789    0.90654
```

MAP results:

```julia
2×3 Named Array{Float64,2}
A ╲ B │ parameter    std_err      tstat
──────┼────────────────────────────────
s     │    1.3783    0.20177    6.83104
m     │  0.501096    0.40024    1.25199
```


mohamed82008:
I suppose that for many data points, `m` converges to 0 in MLE?

cpfiffer:
Yeah, the MLE estimate here is basically the mean.

devmotion:
Honestly, I had no clue what you mean with standard errors and tstat before I checked the code. I would have expected it to just return the mode, either as named tuple or VarInfo object, possibly together with the other results from the optimizer. Maybe one could have a separate function for these diagnostics?

If `mode_estimation` "just" performs the optimization, an alternative approach could also be to just provide the cost function and initial value to the user, similar to https://github.com/baggepinnen/FluxOptTools.jl, so users can just call Optim (or possibly other optimization packages) in whatever way without Turing having to worry about it.

Project.toml
```diff
@@ -17,6 +17,7 @@ Libtask = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
 MCMCChains = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
+Optim = "429524aa-4258-5aef-a3af-852621145aeb"
```
devmotion, line 20: Maybe make this an optional requirement?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
+module ModeEstimation
```
devmotion, line 1: I guess this is motivated by the fact that it might become a separate package at some point? IMO having nested modules is usually not very convenient for handling imports and exports (and debugging...), so I try to avoid it.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+A wrapper struct to store various results from a MAP or MLE estimation.
+
+Fields:
+
+- `values` is a vector with the resulting point estimates
+- `info_matrix` is the inverse Hessian
+- `optim_result` is the stored Optim.jl results
+- `summary_table` is a summary table with parameters, standard errors, and 
+  t-statistics computed from the information matrix.
+- `lp` is the final likelihood.
```
devmotion, line 31: I guess it would be nicer (and more standard) to add the documentation directly to the fields and use DocStringExtensions to include them in the docstring (see https://github.com/TuringLang/Turing.jl/pull/1131).

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+function Base.show(io::IO, m::ModeResult)
+    show(io, m.summary_table)
+end
+
+"""
+    make_logjoint(model::DynamicPPL.Model, ctx::DynamicPPL.AbstractContext)
+
+Constructs a log density function that accepts a vector `z` and returns 
+a tuple (-likelihood, `varinfo`). The model is run using the provided 
+context `ctx`.
```
devmotion, line 55: The [first one-line sentence in each docstring should use the imperative form](https://docs.julialang.org/en/latest/manual/documentation/):
```suggestion
Construct a log density function.

The `model` is run using the provided context `ctx`.
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+
+        unlinked && DynamicPPL.invlink!(varinfo_init, spl)
+        model(varinfo, spl, ctx)
+        unlinked && DynamicPPL.link!(varinfo_init, spl)
+
+        return -DynamicPPL.getlogp(varinfo)
+    end
+
+    return logπ
+end
```
devmotion, line 74: I know this pattern is used all over the place in Turing but IMO it would be cleaner to just have a proper callable struct, something like
```julia
struct OptimLogDensity{M<:Model,C<:Context,V<:VarInfo}
	model::M
	context::C
	init::V
end

function OptimLogDensity(model::Model, context::AbstractContext)
	init = VarInfo(model)
	DynamicPPL.link!(init, SampleFromPrior())
	return OptimLogDensity(model, context, init)
end

function (f::OptimLogDensity)(z)
	...
end
```

That would make it easier to access `model`, `context`, and the initial sample (which is created multiple times currently it seems) from the log density function.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+Optional arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
+
+"""
+function mode_estimation(
+    model::DynamicPPL.Model, 
+    lpf; 
+    optim_options=Optim.Options(),
```
devmotion, line 100: Maybe just forward `kwargs...` to Optim instead? Or just provide the cost function that can be optimized alternatively.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+
+    spl = DynamicPPL.SampleFromPrior()
+    vi = DynamicPPL.VarInfo(model)
+    init_params = model(vi, spl)
+    init_vals = vi[spl]
+
+    # Construct target function.
+    target(x) = lpf(x)
+    hess_target(x) = lpf(x; unlinked=true)
+
```
devmotion, line 115: There's no need to construct these functions. Instead of `target` one can just use `lpf`, and the second one can be avoided by using `do` syntax when computing the Hessian (which maybe should not be part of this function?).

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+Optional arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
+
+"""
+function mode_estimation(
+    model::DynamicPPL.Model, 
+    lpf;
```
devmotion, line 99: There is no need to provide `model` if `lpf` would be a callable struct, as mentioned above.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+    optim_options=Optim.Options(),
+    kwargs...
+)
+    # Do some initialization.
+    b = bijector(model)
+    binv = inv(b)
+
+    spl = DynamicPPL.SampleFromPrior()
+    vi = DynamicPPL.VarInfo(model)
+    init_params = model(vi, spl)
```
devmotion, line 109: I guess these model evaluations could be avoided by reusing the sample that was computed when constructing the log density function.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+    vals = binv(M.minimizer)
+
+    # Get the VarInfo at the MLE/MAP point, and run the model to ensure 
+    # correct dimensionality.
+    vi[spl] = vals
+    model(vi) # XXX: Is this a necessary step?
+
+    # Make one transition to get the parameter names.
+    ts = [Turing.Inference.Transition(DynamicPPL.tonamedtuple(vi), DynamicPPL.getlogp(vi))]
+    varnames, _ = Turing.Inference._params_to_array(ts)
```
devmotion, line 129: Can't you get the parameter names from `VarInfo` directly? If not, we should change that since it seems weird and inefficient to first convert it to a Transition and then extract the names.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+    # Optimize!
+    M = Optim.optimize(target, init_vals, optim_options)
+
+    # Retrieve the estimated values.
+    vals = binv(M.minimizer)
+
+    # Get the VarInfo at the MLE/MAP point, and run the model to ensure 
+    # correct dimensionality.
+    vi[spl] = vals
+    model(vi) # XXX: Is this a necessary step?
```
devmotion, line 125: Do you have an example for what could go wrong when the model is not rerun?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+
+"""
+    mode_estimation(
+        model::DynamicPPL.Model, 
+        lpf; 
+        optim_options=Optim.Options(),
+        kwargs...
+    )
+
+An internal function that handles the computation of a MLE or MAP estimate.
```
devmotion, line 84: Being picky about the imperative form again, maybe something like:
```suggestion
Estimate a mode, i.e., compute a MLE or MAP estimate.
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+Arguments: 
+
+- `model` is a `DynamicPPL.Model`.
+- `lpf` is a function returned by `make_logjoint`.
+
+Optional arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
+
```
devmotion, line 95: Argument lists [should be avoided if possible](https://docs.julialang.org/en/latest/manual/documentation/).

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+end
+
+model = f()
+mle_estimate = MLE(model)
+
+# Manually setting the optimizers settings.
+mle_estimate = MLE(model, optim_options=Optim.Options(iterations=500))
+```
+"""
+function MLE(model::DynamicPPL.Model; kwargs...)
```
devmotion, line 193: IMO this should be
```suggestion
function mle(model::DynamicPPL.Model; kwargs...)
```
since `MLE` suggests incorrectly that an object of type `MLE` is returned.

devmotion: Alternatively, one could use singleton types `MLE` and `MAP` as inputs to `mode_estimation` instead of the cost function.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+        @warn "Could not compute Hessian matrix" err
+        stable = NamedArrays.NamedArray([vals repeat([missing], length(vals)) repeat([missing], length(vals))], (varnames, ["parameter", "std_err", "tstat"]))
+        return ModeResult(vmat, missing, M, stable, M.minimum)
+    end
+end
+
+"""
+    MLE(model::DynamicPPL.Model; kwargs...)
+
+Returns a maximum likelihood estimate of the given `model`.
```
devmotion, line 164:
```suggestion
Compute a maximum likelihood estimate of the given `model`.
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+
+Arguments: 
+
+- `model` is a `DynamicPPL.Model`.
+
+Keyword arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
```
devmotion, line 173: Again, maybe avoid these lists and just forward regular keyword arguments to Optim?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+
+Usage:
```
devmotion, line 175: Code examples should be included in an [`# Examples` section](https://docs.julialang.org/en/latest/manual/documentation/):
```suggestion
# Examples
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+"""
+function MLE(model::DynamicPPL.Model; kwargs...)
+    lpf = make_logjoint(model, DynamicPPL.LikelihoodContext())
+    return mode_estimation(model, lpf; kwargs...)
+end
+
+"""
+    MAP(model::DynamicPPL.Model; kwargs...)
+
+Returns the maximum a posteriori estimate of the given `model`.
```
devmotion, line 201:
```suggestion
Compute the maximum a posteriori estimate of the given `model`.
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+Arguments: 
+
+- `model` is a `DynamicPPL.Model`.
+
+Keyword arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
+
+Usage:
```
devmotion, line 212:
```suggestion
# Examples
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+Returns the maximum a posteriori estimate of the given `model`.
+
+Arguments: 
+
+- `model` is a `DynamicPPL.Model`.
+
+Keyword arguments:
+
+- `optim_options` is a `Optim.Options` struct that allows you to change the number
+  of iterations run in an MLE estimate.
```
devmotion, line 210: Same as above, maybe try to avoid the explicit lists.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,235 @@
...
+end
+
+model = f()
+mle_estimate = MAP(model)
+
+# Manually setting the optimizers settings.
+mle_estimate = MAP(model, optim_options=Optim.Options(iterations=500))
+```
+"""
+function MAP(model::DynamicPPL.Model; kwargs...)
```
devmotion, line 230: This should be a function as well but unfortunately it can't be named `map`... Maybe use something more explicit?

cpfiffer:
> Honestly, I had no clue what you mean with standard errors and tstat before I checked the code. I would have expected it to just return the mode, either as named tuple or VarInfo object, possibly together with the other results from the optimizer. Maybe one could have a separate function for these diagnostics?

No, this is almost completely counter to how these analyses are supposed to work. Using a MLE/MAP estimate with no information on the dispersion of that estimator is next to useless, akin to running OLS without standard errors or using just the median or whatever from a full posterior.

I think if you are the kind of person who works in the MLE world, it's an absolute requirement to have statistics computed from the information matrix. Plus, you can use it to run hypothesis tests.

Granted, perhaps this should be separated out, such that you do something like

```julia
est = MLE(model)
info_matrix = information_matrix(est)
```

Or something equivalent. 

But I would prefer to hew to how OLS works, where the estimator covariance matrix is returned simultaneously with the point estimates.

src/modes/ModeEstimation.jl
cpfiffer: Well, we only export two things, and I'm hoping this is not going to live in Turing for very long, so I wanted to keep all of these things separate.

src/modes/ModeEstimation.jl
cpfiffer: Nice, I hadn't noticed that. I'll add it.

devmotion:
> Granted, perhaps this should be separated out, such that you do something like
>
> est = MLE(model)
> info_matrix = information_matrix(est)
>
> Or something equivalent.

Yes, that's what I had in mind (I just thought you might have to provide the `model` to `information_matrix` as well). IMO separating these computations seems cleaner, even if you are always interested in the information matrix, since it would allow to evaluate the information matrix also for estimates that are obtained in a different way. And, of course, one can still implement a function that just calls both functions and returns their results, if one wants to save one line in downstream code.

yebai:
A minor comment on the interface: maybe we can create a new `optimise(model, algorithm)` API on top of the `sample(model, algorithm)` method. This `optimise` function can support both point estimation and variational inference algorithms, e.g. 

```julia
# Make point estimates
mle_gdemo = optimise(model, MLE(...))
map_gdemo = optimise(model, MAP(...))
vi_gdemo = optimise(model, ADVI(...))
```

cc @torfjelde

cpfiffer:
One quibble with this is that `optimize` would return two different things -- the MLE/MAP return values are distinct from the ADVI stuff in that it is a vector of point estimates, which I think is a little counter to the interface method where we generally have a framework for the return value.

yebai:
> One quibble with this is that optimize would return two different things -- the MLE/MAP return values are distinct from the ADVI stuff in that it is a vector of point estimates, which I think is a little counter to the interface method where we generally have a framework for the return value.

That's true, but we can probably introduce some abstract types which can be shared by all deterministic inference algorithms -- point estimation returns a Dirac measure, which can be viewed as a special variational distribution.

src/modes/ModeEstimation.jl
cpfiffer: Not in a quick way, at least as far as I can tell. If someone's got a good fix other than this (reasonbly good) hack, I'm game.

src/modes/ModeEstimation.jl
cpfiffer: Not really. I pulled this out.

cpfiffer:
Okay, I've changed it quite a bit. There's now a callable struct `OptimLogDensity` which gets you the log joint or the log likelihood, as per David's suggestion.

I have also implemented the `Optim.optimize` call @yebai suggested, such that you now do

```julia
using Turing
using Optim

@model function gdemo(x)
    s ~ InverseGamma(15,3)
    m ~ Normal(0, 1)

    for i in eachindex(x)
        x[i] ~ Normal(m, s)
    end

    return m, s
end

Random.seed!(100)
data = rand(Normal(0, 2), 10)
model = gdemo(data)

mle = optimize(model, MLE())
mpa = optimize(model, MAP())
```

I also implemented a bunch of StatsBase.jl functions to handle the information matrix, covariance matrix, etc. It's much more palatable now, I think. 

Currently the MAP estimates are incorrect, and I can't figure out why -- they are biased downards by quite a bit. The current numbers for the default `gdemo` model are

```julia
A  │                                                                                                                           
───┼────────                                                                                                                   
s  │ 1.16668                                                                                                                   
m  │ 1.16668    
```

where we expect the results to be

```julia
A  │                                                                                                                           
───┼────────                                                                                                                   
s  │ 2.0416                                                                                                                   
m  │ 1.16666  
```

The fact that the mean is correct while the variance is not is making me suspect there's an issue with the bijection somewhere, which I'll have to look into later.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,205 @@
...
+function Optim.optimize(model::Model, ::MAP, args...; kwargs...)
+    return optimize(model, OptimLogDensity(model, DynamicPPL.DefaultContext()), args...; kwargs...)
+end
+
+"""
+    Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.BFGS(), args...; kwargs...)
+
+Estimate a mode, i.e., compute a MLE or MAP estimate.
+"""
+function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.BFGS(), args...; kwargs...)
```
trappmartin, line 177: The memory requirement of BFGS scales quadratic in the number of parameter you optimise. I would suggest to use a more memory friendly option as default.
```suggestion
function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim. LBFGS(), args...; kwargs...)
```
or 
```suggestion
function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.GradientDescent(), args...; kwargs...)
```

devmotion:
Could be interesting to use [GalacticOptim.jl](https://github.com/sciml/galacticoptim.jl) at some point.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,205 @@
...
+
+Evaluate the log joint (with `DefaultContext`) or log likelihood (with `LikelihoodContext`)
+at the array `z`. If `unlinked=true`, no change of variables will occur.
+"""
+function (f::OptimLogDensity)(z; unlinked::Bool = false)
+    spl = DynamicPPL.SampleFromPrior()
+
+    varinfo = DynamicPPL.VarInfo(f.init, spl, z)
+
+    unlinked && DynamicPPL.invlink!(f.init, spl)
```
trappmartin, line 55: Do I understand it correctly that this is done in each iteration. Is it really necessary to check this at each iteration, or could you pull out the code instead and do this once?

trappmartin:
This would be really cool! GalacticOptim.jl seems like the goto in the future.

devmotion:
Just throwing ideas around: a different approach to support optimization would be something like https://github.com/baggepinnen/FluxOptTools.jl, i.e., just providing a function that provides users with the functions that they plug into Optim (or other optimization packages).

cpfiffer:
If you want just the density to use somewhere, you can do that with

```julia
@model function f(x)
    m ~ Normal()
    x ~ Normal(m,1)
end

model = f(1.5)
target = Turing.OptimLogDensity(model, DynamicPPL.LikelihoodContext())
```

but perhaps there's a convenience function we could include as a first-class citizen in Turing -- maybe `logdensity(model)` and `loglikelihood(model)`.

trappmartin:
> but perhaps there's a convenience function we could include as a first-class citizen in Turing -- maybe `logdensity(model)` and `loglikelihood(model)`.

I like the idea. But maybe better: `logjoint(model)` and `loglikelihood(model)`, log density is not very descriptive as it doesn't say what the density is of.

cpfiffer:
Alright. I need some other eyes on this one. Something is wrong with the transformations, but I cannot figure out where -- MAP yields a result of

```julia
ModeResult with minimized lp of -4.60

2-element Named Array{Float64,1}
A  │ 
───┼────────
s  │  1.1667
m  │ 1.16667
```

when it should be 

```julia
2-element Named Array{Float64,1}
A  │ 
───┼────────
s  │  2.0416 
m  │ 1.16667
```

1.1667 is quite a bit lower than 2.0416, and I know it's because the transforms are wrong. I've tried checking to see if there's a missing inverse bijection (I don't think there is), if I'm using `link!` and `invlink!` correctly (not sure that I am, but couldn't figure out a different approach to fix it). 

Box minimization works fine, but I'd rather not add the computational overhead of figuring out bounds dynamically when transforms exist. 

Here's a MWE:

```julia
using Random
using Turing
using Optim

@model function gdemo()
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))

    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))

    return m, s
end

Random.seed!(100)
model = gdemo()

mle = optimize(model, MLE())
mpa = optimize(model, MAP())

mle |> display
mpa |> display
```

trappmartin:
I'll see if I can find the problem.

trappmartin:
@cpfiffer should the code you posted run on this PR? Seem like `NamedArrays` is missing.
```
ERROR: LoadError: LoadError: ArgumentError: Package Turing does not have NamedArrays in its dependencies:
- If you have Turing checked out for development and have
  added NamedArrays as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with Turing
```

Could you check if everything is committed.

cpfiffer:
Should be good to go with the recent fixes.

trappmartin:
Thx

cpfiffer:
Okay, I took another crack at trying to figure this out. I do _not_ have an answer, but I'm wondering if the evidence I've got might help a little. Here's my MWE (using this branch):

```julia
using Bijectors
using Turing
using Optim

@model function gdemo()
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))

    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))

    return m, s
end

model = gdemo()

bm = bijector(model)

globmax = [49/24, 7/6]

@info "Unlinked"
mpa_fun(globmax, linked=false) |> display

@info "Linked"
mpa_fun(bm(globmax), linked=true) |> display
```

The issue is basically that, even if I biject the true MAP estimate of `[49/24, 7/6]` and then have Turing link it internally, I get a result I'm not sure I understand. This is perhaps generally in line with my inability to grok bijectors. Here's the output:

```julia
[ Info: Unlinked

[2.0416666666666665, 1.1666666666666667]
(s = ([2.0416666666666665], ["s"]), m = ([1.1666666666666667], ["m"]))
Assume s: val: 2.0416666666666665 lp: -1.4134625810538644
Assume m: val: 1.1666666666666667 lp: -1.6091551004193465
Unlinked LP: 5.771540127209862
```

```julia
[ Info: Linked

[0.7137664677626809, 1.1666666666666667]
(s = ([2.0416666666666665], ["s"]), m = ([1.1666666666666667], ["m"]))
Assume s: val: 2.0416666666666665 lp: -0.6996961132911834
Assume m: val: 1.1666666666666667 lp: -1.6091551004193465
Linked LP: 5.057773659447182
```

My question: Why do I get different densities if I biject the true result first and let Turing handle the link internally, vs. using the unlinked variables with no transform? Shouldn't they be equivalent?

Maybe @torfjelde might have an idea -- I need an adult.

trappmartin:
The logpdf of the transformed density function is only for trivial cases the same as the untransformed density function, i.e. affine transformations. What is important is that both share the same maximum, i.e. the transformation has to be monotonic (I think). 

On a different note. 
I had a look at the PR a few days back and checked it works for the prior context, which it did. I haven't had the time to continue but it might be good to change the PR so that it uses the latest merged PR from DynamicPPL, which allow us to use `loglikelihood(model, vi)`. Thus, quite a bit of the code in this PR could be removed I believe. What do you think?

cpfiffer:
> The logpdf of the transformed density function is only for trivial cases the same as the untransformed density function, i.e. affine transformations. What is important is that both share the same maximum, i.e. the transformation has to be monotonic (I think).

They have different maxima. The untransformed MAP estimate is (correctly) 2.0416, whereas the transformed MAP estimate is 1.6667. So something is wrong with the transform?

> On a different note.
> I had a look at the PR a few days back and checked it works for the prior context, which it did. I haven't had the time to continue but it might be good to change the PR so that it uses the latest merged PR from DynamicPPL, which allow us to use `loglikelihood(model, vi)`. Thus, quite a bit of the code in this PR could be removed I believe. What do you think?

I don't think it would actually eliminate much code at all, since much of the callable struct functionality handles linking and contexts that those convenience functions don't.

trappmartin:
I just created a tiny gist, as I find it more convenient to implement this differently and try out stuff there. Actually, even the maximisation for the prior does not converge to the mode. No idea what the problem is atm.

https://gist.github.com/trappmartin/597b3de3d348006415f61c325f006657

cpfiffer:
Well, at the very least I'm happy that your method is also finding some strange behavior.

devmotion:
As far as I can see, the following is happening here when you do not convert the values back to the untransformed space:
- You update the `VarInfo` object (which has `trans` flag set to `true` for all variables since you called `link!` before) with the transformed values `xtrans` by calling `setval!`
- During model execution, the untransformed value `x` is retrieved when evaluating the assume statements: https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L134 and https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/varinfo.jl#L800-L801
- Then https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L141 evaluates `logpdf_with_trans(dist, x, true)` which just returns `logpdf(dist, x) - logabsdetjac(bijector(dist), x)`: https://github.com/TuringLang/Bijectors.jl/blob/1f3b581afe04f690bd93fba9edd88735cc1fc140/src/Bijectors.jl#L147-L153

However, if you convert the values manually to the untransformed space, the following happens:
- You update the `VarInfo` object (which has `trans` flag set to `false` for all variables since you did not call `link!` before) with the untransformed values `x` by calling `setval!`
- During model execution, the untransformed value `x` is retrieved when evaluating the assume statements: https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L134 and https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/varinfo.jl#L800-L801
- Then https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L141 evaluates `logpdf_with_trans(dist, x, false)` which just returns `logpdf(dist, x)`: https://github.com/TuringLang/Bijectors.jl/blob/1f3b581afe04f690bd93fba9edd88735cc1fc140/src/Bijectors.jl#L147-L153

So it is not surprising that you not obtain the same log densities in both cases. The optimization procedure should be unaffected by this difference only if the additional `logabsdetjac` term does not affect the location of the maximum (or minimum, depending on what you optimize). I don't know why that's not the case here and if bijectors are supposed to have this property in general at all.

However, since the values are converted to the untransformed space in DynamicPPL anyways during model evaluation, maybe one should just manually call `invlink!` after updating the values with `setval!` to get rid of the `logabsdetjac` term?

trappmartin:
I think all of the current steps should be fine. The transformation might effect the optimisation in terms of how well the optimiser finds the maximum but the maximum should not change given the transformation that is applied. I’ll look at it tomorrow again, it’s later over here. That said, @cpfiffer are actually comparing against the mode of the true posterior. I just did some calculations in my head (not reliable) and I think 2 sounds more like the mean of the posterior of the inverse gamma then the mode to me. I think the mode should be smaller. But again, it’s late over here and I’ll check again tomorrow.

devmotion:
> The transformation might effect the optimisation in terms of how well the optimiser finds the maximum but the maximum should not change given the transformation that is applied.

I just did the calculations on pen and paper for the prior example that you posted above, where optimization yielded an optimal value for `s` of around 1.2 whereas we would expect a value of 6/7, which is approx 0.86. Here's my calculation (constants denoted by C_1 and C_2):
1. We have `logpdf(IG(2, 3), s) = C_1 - (2 + 1) * log(s) - 3 / s = C_1 - 3 log(s) - 3 / s`
2. We have `logpdf(Normal(0, sqrt(s)), m) = C_2 - (log(s) + m^2 / s) / 2 = C_2 - log(s) / 2 - m^2 / (2 s)`
3. Thus in total we get `logpdf(IG(2, 3), s) + logpdf(Normal(0, sqrt(s)), m) = C_1 + C_2 - 7 log(s) / 2 - (6 + m^2) / (2 s)`
4. This expression is maximized for `m = 0` and `s = 6 / 7`.
5. The bijectors that are used in this case are the identity for `m` and `log` for `s`, and hence we have `logabsdetjac(bijector(IG(2, 3)), s) = - log(s)` (see https://github.com/TuringLang/Bijectors.jl/blob/d626f046b7f49160e22d597d2df6231413428180/src/bijectors/exp_log.jl#L42) and `logabsdetjac(bijector(Normal(0, sqrt(s)), m) = 0` (see https://github.com/TuringLang/Bijectors.jl/blob/a763e3ff922743d0cf2a0c6260e9ec5926d3adc9/src/interface.jl#L116).
6. Thus with the bijectors we obtain `logpdf(IG(2, 3), s) - logabsdetjac(bijector(IG(2, 3)), s) + logpdf(Normal(0, sqrt(s)), m) - logabsdetjac(bijector(Normal(0, sqrt(s))), m) = C_1 + C_2 - 7 log(s) / 2 - (6 + m^2) / (2 s) - (-log(s)) = C_1 + C_2 - 5 log(s) / 2 - (6 + m^2) / (2 s)`.
7. This expression is maximized for `m = 0` and `s = 6 / 5`.

Thus the optimization based on the transformed values (with the additional terms of `-logabsdetjac`) correctly returned a value close to 1.2. However, that maximum is different from the one you would obtain by performing optimization based on the `logpdf` of the untransformed values only.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,210 @@
...
+
+Evaluate the log joint (with `DefaultContext`) or log likelihood (with `LikelihoodContext`)
+at the array `z`. If `unlinked=true`, no change of variables will occur.
+"""
+function (f::OptimLogDensity)(z; unlinked::Bool = false)
+    spl = DynamicPPL.SampleFromPrior()
+
+    varinfo = DynamicPPL.VarInfo(f.init, spl, z)
+
+    if unlinked
```
mohamed82008, line 55: I think this makes more sense to be a field of `OptimLogDensity`. When finding the information matrix, we can construct a new `OptimLogDensity` with `unlinked = true` or `linked = false`.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,210 @@
...
+function Optim.optimize(model::Model, ::MAP, args...; kwargs...)
+    return optimize(model, OptimLogDensity(model, DynamicPPL.DefaultContext()), args...; kwargs...)
+end
+
+"""
+    Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+
+Estimate a mode, i.e., compute a MLE or MAP estimate.
+"""
+function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
```
mohamed82008, line 182: Could be nice to allow an initial solution here as a named tuple using PriorContext.

devmotion:
The calculation for the MAP estimate is very similar, we just get the additional terms `logpdf(Normal(m, sqrt(s)), 1.5)` and `logpdf(Normal(m, sqrt(s)), 2)` (since the bijector of the normal distribution is the identity, we can neglect it). Then without transformations we get the expression
`C_1 + C_2 - 7 log(s) / 2 - (6 + m^2) / (2 s) - log(s) / 2 - (m - 1.5)^2 / (2 s) - log(s) / 2 - (m - 2)^2 / (2 s) = C_1 + C_2 - 9 log(s) / 2 - (6 + m^2 + (m - 1.5)^2 + (m - 2)^2) / (2 s)`, which has the maximum `m = 7 / 6` and `s = 49 / 54`. With the transformations, and hence with the additional `logabsdetjac` term, we get the expression
`C_1 + C_2 - 5 log(s) / 2 - (6 + m^2) / (2 s) - log(s) / 2 - (m - 1.5)^2 / (2 s) - log(s) / 2 - (m - 2)^2 / (2 s) = C_1 + C_2 - 7 log(s) / 2 - (6 + m^2 + (m - 1.5)^2 + (m - 2)^2) / (2 s)`, which is maximized for `m = 7 / 6` and `s = 7 / 6`.

Thus the optimization gave the correct answer for the transformed values. It's just not the same maximum that you would get for the untransformed values.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,210 @@
...
+    # Do some initialization.
+    b = bijector(model)
+    binv = inv(b)
+
+    spl = DynamicPPL.SampleFromPrior()
+    vi = DynamicPPL.VarInfo(model)
```
mohamed82008, line 188: Can we use the VarInfo inside `f`?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,210 @@
...
+    init_params = model(vi, spl)
+    init_vals = vi[spl]
+
+    # Optimize!
+    M = Optim.optimize(f, init_vals, args...; kwargs...)
```
mohamed82008, line 193: `init_vals` here will be in the constrained domain. Using the `VarInfo` inside `f` should fix this since it is linked.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,210 @@
...
+end
+
+"""
+    Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+
+Estimate a mode, i.e., compute a MLE or MAP estimate.
+"""
+function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+    # Do some initialization.
+    b = bijector(model)
```
mohamed82008, line 184: I don't think we need this. We can just write the minimizer to `f.init`, `invlink!` it, get the solution in the constrained space, then `link!` it back.

mohamed82008:
I agree with @devmotion's analysis. I think we should have an `OptimizationContext` that always evaluates `logpdf` instead of `logpdf_with_trans` but still calls `invlink` in `vi[vn]`, i.e. `istrans(vi, vn)` is true but `logpdf` is called and not `logpdf_with_trans`.

devmotion:
BTW it's new to me but apparently it's a well-known fact that the MAP estimate are not invariant under reparameterization: https://stats.stackexchange.com/questions/330402/what-is-an-example-of-a-transformation-on-a-posterior-distribution-such-that-the

devmotion:
> I think we should have an OptimizationContext that always evaluates logpdf instead of logpdf_with_trans but still calls invlink in vi[vn], i.e. istrans(vi, vn) is true but logpdf is called and not logpdf_with_trans.

I'm not sure, isn't it sufficient to just call `invlink!` after having created the VarInfo object from the template `f.init`? Then the values will be untransformed and the `trans` flag is set to `false`, so we should only get the `logpdf` values.

mohamed82008:
>Then the values will be untransformed and the trans flag is set to false, so we should only get the logpdf values.

But the optimizer is unconstrained, so it may wander off into forbidden territory. We need the decision variables to be unconstrained but the objective to be the `logpdf` of the constrained parameters.

devmotion:
That's exactly what I suggested (or at least tried to) - you take the unconstrained values from the optimizer and transform them back to the original space before evaluating the model.

mohamed82008:
> and transform them back to the original space before evaluating the model.

If this is done in an out-of-place manner, then it should be enough yes. But doing it out-of-place is tricky.

devmotion:
I guess we could just have an alternative to https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/varinfo.jl#L119-L148 that transforms `x` back instead of just extracting the correct subsets?

devmotion:
So basically just changing https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/varinfo.jl#L133.

mohamed82008:
It's tricky because you need to get the transformation correct. If the distribution's support is stochastic, you are out of luck. We don't support stochastic support properly yet but if we do, the context will be the more efficient way to go because it delays the domain transformation until we have the distribution on the RHS of ~ in the `assume` functions.

mohamed82008:
Perhaps a different `AbstractSampler` would also do if it means less methods to define but a context makes more sense since optimization is not sampling.

devmotion:
> If the distribution's support is stochastic, you are out of luck.

That's a good point. Currently the transformation is independent of the RHS in the assume statement though: https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L134

mohamed82008:
> Currently the transformation is independent of the RHS in the assume statement though: https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L134

Yes which is a bug addressed by https://github.com/TuringLang/DynamicPPL.jl/pull/115. But I will get back to it after NeurIPS.

mohamed82008:
We can do what you say now and fix it later. But I think the context is easier anyways and will likely be faster, no need to `invlink` and `vcat`.

devmotion:
Actually, I think we should fix it correctly right away in this PR and define the `OptimizationContext` and the required `assume` and `observe` implementations, which can be very short and optimized I guess.

cpfiffer:
I feel like the quicker hack is just to define the following (more or less)

```julia
struct OptimizationSampler <: AbstractSampler end

function assume(
    spl::OptimizationSampler,
    dist::Distribution,
    vn::VarName,
    vi,
)
    if haskey(vi, vn)
        # Always overwrite the parameters with new ones for `SampleFromUniform`.
        if is_flagged(vi, vn, "del")
            unset_flag!(vi, vn, "del")
            r = init(dist, spl)
            vi[vn] = vectorize(dist, r)
            settrans!(vi, false, vn)
            setorder!(vi, vn, get_num_produce(vi))
        else
            r = vi[vn]
        end
    else
        r = init(dist, spl)
        push!(vi, vn, r, dist, spl)
        settrans!(vi, false, vn)
    end
    return r, Bijectors.logpdf_with_trans(dist, r, false)
end
```

cpfiffer:
I take that back. I see the benefit of the context, but I don't actually see how you would change this without rewriting `observe` and `assume`. What am I missing as far as implementation is concerned?

trappmartin:
> BTW it's new to me but apparently it's a well-known fact that the MAP estimate is not invariant under reparameterization: https://stats.stackexchange.com/questions/330402/what-is-an-example-of-a-transformation-on-a-posterior-distribution-such-that-the

Thanks, I wasn’t aware of that.

devmotion:
It would be nice to do something like
```julia
struct OptimizationContext{C}
	context::C
end

# assume
function tilde(ctx::OptimizationContext, sampler, right, vn::VarName, inds, vi)
    # transform variable to original space...
    return tilde(ctx.context, sampler, right, vn, vi)
end

# observe
function tilde(ctx::OptimizationContext, sampler, right, left, vi)
	# transform variable to original space...
    return tilde(ctx.context, sampler, right, left, vi)
end
```
since then one could support arbitrary nested contexts, but that leads to the same AD as transforming everything in the beginning I guess. Alternatively, we could just add special cases for the contexts that we want support by implementing
```
# assume
function tilde(ctx::OptimizationContext, sampler, dist, vn::VarName, inds, vi)
    r = vi[vn]
    return r, Distributions.logpdf(dist, r)
end

# observe
function tilde(ctx::OptimizationContext{<:PriorContext}, sampler, right, left, vi)
    return 0
end

function tilde(ctx::OptimizationContext, sampler, dist, value, vi)
    return Distributions.logpdf(dist, value)
end
```
and similar for `dot_tilde`. I'm not sure if we have to consider sampling at all, since optimization requires us to provide the full set of variables anyway I guess? Then one could just use user-provided initial values or sample them once with `SampleFromPrior()` (as currently done), and in the next steps just update these values with the ones provided by the optimization algorithm.

mohamed82008:
I like @devmotion's second suggestion. The `logpdf(NoDist(), ..)` can also be replaced by 0.

devmotion:
>  The logpdf(NoDist(..), ..) can also be replaced by 0.

I thought so as well initially, but it seems that it might actually return an array of zeros, depending on the underlying distribution and the provided value. Hence I kept `NoDist`.

mohamed82008:
>I thought so as well initially, but it seems that it might actually return an array of zeros, depending on the underlying distribution and the provided value. Hence I kept NoDist.

I think in all cases where an array is returned, e.g. `dot_assume`, we end up calling `sum` at the end. I think it is safe to assume (pun intended) that the second output of `tilde` and `dot_tilde` will always be a scalar. We call `acclogp!` on it after all.

devmotion:
Sounds reasonable, and should also be faster in the cases where `logpdf(::NoDist, x)` would return an array. Do we actually create the array of zeros and sum them up in some place? Sounds a bit inefficient :smile: 

I'll update the code above.

mohamed82008:
>Do we actually create the array of zeros and sum them up in some place? Sounds a bit inefficient 😄

Yes because I didn't want to special case `NoDist` everywhere, it's treated like any other dist. We can probably return a `FillArray`.

devmotion:
Maybe an alternative would be to actually work with `loglikelihood` instead of `logpdf` in cases where we always want a scalar, regardless if we have multiple inputs.

src/modes/ModeEstimation.jl
cpfiffer: Do you mean as a return value or a starting point?

cpfiffer:
Alright, I'm hitting the mark on @devmotion's math with the `PriorContext`. I'm going to open this up for formal review.

Project.toml
devmotion: These dependencies should be moved to the `[extras]` section, since the MAP/MLE functionality is optional.

src/Turing.jl
```diff
@@ -110,5 +115,8 @@ export  @model,                 # modelling
         LogPoisson,
         NamedDist,
         filldist,
-        arraydist
+        arraydist,
+
+        MLE,                    # mode estimation tools
+        MAP
```
devmotion, line 121: Can you move these exports to the `@require` block above?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+            settrans!(vi, false, vn)
+            setorder!(vi, vn, get_num_produce(vi))
+        else
+            r = vi[vn]
+        end
+    else
+        r = init(dist, spl)
+        push!(vi, vn, r, dist, spl)
+        settrans!(vi, false, vn)
+    end
```
devmotion, line 50: Is
```suggestion
    r = vi[vn]
```
not sufficient?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
```
devmotion, line 70: Same here, do we actually need the sampling functionality?
```suggestion
    r = vi[vn]
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:LikelihoodContext}, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = get_and_set_val!(vi, vns, dist, sampler)
+    return r, 0
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = get_and_set_val!(vi, vns, dist, sampler)
+    lp = sum(logpdf(dist, r))
```
devmotion, line 93: I guess this should be either
```suggestion
    lp = sum(logpdf.(dist, r))
```
or
```suggestion
    lp = loglikelihood(dist, r)
```
IIRC `logpdf(dist, r)` where `r` are multiple samples is deprecated in Distributions. I'd prefer the latter since it avoid the creation of an intermediate array if not needed (e.g., without AD) but one could still implement custom adjoints based on broadcasting. It seems, however, that we currently always use the first alternative in DynamicPPL, so I guess the custom adjoints are not available right now.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+end
+
+function DynamicPPL.tilde(ctx::OptimizationContext, sampler, dist, value, vi)
+    return Distributions.logpdf(dist, value)
+end
+
+# dot assume
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:LikelihoodContext}, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = get_and_set_val!(vi, vns, dist, sampler)
```
devmotion, line 86: Same here, if sampling is not needed, we should really think about implementing `get_val!` that just contains the logic from `get_and_set_val!` for extracting samples.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:PriorContext}, sampler, right, left, vi)
+    return 0
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext, sampler, right, left, vn, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = get_and_set_val!(vi, vns, dist, sampler)
+    lp = sum(logpdf(dist, r))
```
devmotion, line 109: Same here.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+
+Estimate a mode, i.e., compute a MLE or MAP estimate.
+"""
+function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+    # Do some initialization.
+    spl = DynamicPPL.SampleFromPrior()
+    init_params = model(f.vi, spl)
+    init_vals = f.vi[spl]
+
+    # Optimize!
```
trappmartin, line 301: Could we just use a closure here instead of the additional struct?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+"""
+struct OptimLogDensity{M<:Model,C<:AbstractContext,V<:VarInfo}
+    "A `DynamicPPL.Model` constructed either with the `@model` macro or manually."
+    model::M
+    "A `DynamicPPL.AbstractContext` used to evaluate the model. `LikelihoodContext` or `DefaultContext` are typical for MAP/MLE."
+    context::C
+    "A `DynamicPPL.VarInfo` struct that will be used to update model parameters."
+    vi::V
+    "Whether the to evaluate the model with transformed variables."
+    linked::Bool
```
devmotion, line 130: I don't think there should be a choice here? `OptimLogDensity` is only used internally, and hence we know that we always optimize in the unconstrained space.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+@model function f(x)
+    m ~ Normal(0, 1)
+    x ~ Normal(m, 1)
+end
+
+model = f(1.5)
+mle = optimize(model, MLE())
+
+# Use a different optimizer
+mle = optimize(model, MLE(), NelderMeade())
```
devmotion, line 257:
```suggestion
mle = optimize(model, MLE(), NelderMead())
```
I guess it would be nice to use a julia-repl code block here (see https://docs.julialang.org/en/latest/manual/documentation/, one could also write doctests).

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+@model function f(x)
+    m ~ Normal(0, 1)
+    x ~ Normal(m, 1)
+end
+
+model = f(1.5)
+map_est = optimize(model, MAP())
+
+# Use a different optimizer
+map_est = optimize(model, MAP(), NelderMeade())
```
devmotion, line 282: Same here:
```suggestion
map_est = optimize(model, MAP(), NelderMead())
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+"""
+    link(f::OptimLogDensity)
+
+Generate an linked (with variable transformions) version of an existing `OptimLogDensity`.
+"""
+function Bijectors.link(f::OptimLogDensity)
+    init = VarInfo(f.model, f.context)
+    DynamicPPL.link!(init, DynamicPPL.SampleFromPrior())
+    return OptimLogDensity(f.model, f.context, init, false)
+end
```
devmotion, line 177: I'm curious, why are `invlink` and `link` needed? BTW the docstring of `invlink` is incorrect.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+    "The final log likelihood or log joint, depending on whether `MAP` or `MLE` was run."
+    lp :: Float64
+    "The evaluation function used to calculate the output."
+    f :: M
+end
+#############################
+# Various StatsBase methods #
+#############################
+
+function Base.show(io::IO, m::ModeResult)
```
devmotion, line 207: Maybe change this to
```suggestion
function Base.show(io::IO, ::MIME"text/plain", m::ModeResult)
```
and implement another more compact `show` method with just one line (that would be used, e.g., when displaying arrays of ModeResult)? See https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing-1.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+    println(io, Printf.@sprintf("ModeResult with minimized lp of %.2f", m.lp), "\n")
```
devmotion, line 208:
```suggestion
    print(io, "ModeResult with minimized lp of ")
    Printf.@printf(io, "%.2f", m.lp)
    println(io)
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,318 @@
...
+    f.vi[spl] = M.minimizer
+    invlink!(f.vi, spl)
+    vals = f.vi[spl]
+
+    # Make one transition to get the parameter names.
+    ts = [Turing.Inference.Transition(DynamicPPL.tonamedtuple(f.vi), DynamicPPL.getlogp(f.vi))]
+    varnames, _ = Turing.Inference._params_to_array(ts)
+
+    # Store the parameters and their names in an array.
+    vmat = NamedArrays.NamedArray(vals, varnames)
```
devmotion, line 315: Do you know how it compares to https://github.com/SciML/LabelledArrays.jl?

src/modes/ModeEstimation.jl
mohamed82008: I think it's when computing the Hessian that `linked` needs to be false.

torfjelde:
Despite being mentioned TWICE, I've completely missed what this was all about. I saw it was a "Draft" and thought "Oh, Cameron is messing about, so I'll have a look with it becomes and actual PR". Did not realize there was a ongoing discussion about the implementation and that I was being pinged several times :flushed:  Sorry guys!

Anyways, first of I think I need to apologize to you @cpfiffer because I think I told you on Slack that for MAP you needed to scale with abs-det-jac but not for MLE, which is just not true (as nicely pointed out by @devmotion earlier!).

So if we want to optimize `p(x, θ)` wrt. `θ`, it does not matter how you arrive at your value `θ` as long it's the one that maximizes `p(x, θ)`. If you want to apply a sequence of transformations `f₁, …, fₙ` to some other variable `ϕ` and then let `θ := (fₙ ∘ ⋯ ∘ f₁)(ϕ)`, that's also fine (as long as you can "reach" the optimal value in `Ω`, it ain't no matter).

So just to clarify the `logabsdetjac` term comes up in either of these cases (mainly):
- in HMC we care about the *expectation* wrt. `p(x, θ) dθ`, i.e. and *integral* over `p(x, θ) dθ`, but after the transformation to ℝᵈ we're instead looking at `p(x, ϕ)` and so we need to rescale the term inside the integral accordingly (i.e. change-of-variables!)
- if we are interested in maximizing `p(x, θ)` where `θ = f(ϕ)` *but* we only have access to `p(x, ϕ)`, then we can derive `p(x, θ)` by scaling with abs-det-jacobian of the transformation `f` (this is the premise of normalizing flows)

Does that make more sense? I'll also have a proper look at the PR now:)

src/modes/ModeEstimation.jl
torfjelde: I haven't tried `NamedArrays` but remember `LabelledArrays` being quite slow when it came to _constructing_ the array, but other than it seemed good. So in this case, `LabelledArrays` should be fine to use :+1:

src/modes/ModeEstimation.jl
torfjelde: I'm also wondering about this. Is there a reason why we don't use something like `make_logjoint` already used for VI: https://github.com/TuringLang/Turing.jl/blob/cb56cd2de7b9f398d5619d78369cd7ee3a2510fb/src/variational/VariationalInference.jl#L230-L257

And for MLE we could just wrap a `LikelihoodCtx`. 

With that being said, the current impl looks nice but might be overly complex with out much gain? :confused:

torfjelde: With current Turing.jl you can do:
```julia
julia> @model function f(x)
           m ~ Normal(0, 1)
           x ~ Normal(m, 1)
       end
DynamicPPL.ModelGen{var"###generator#276",(:x,),(),Tuple{}}(##generator#276, NamedTuple())

julia> model = f(1.5)
DynamicPPL.Model{var"###evaluator#275",(:x,),Tuple{Float64},(),DynamicPPL.ModelGen{var"###generator#276",(:x,),(),Tuple{}}}(##evaluator#275, (x = 1.5,), DynamicPPL.ModelGen{var"###generator#276",(:x,),(),Tuple{}}(##generator#276, NamedTuple()))

julia> b = Turing.bijector(model);

julia> logπ = Turing.Variational.make_logjoint(model);

julia> objective = z -> - (logπ ∘ b)(z);

julia> map = optimize(objective, [0.], LBFGS());

julia> inv(b)(map.minimizer)
1-element Array{Float64,1}:
 0.7499999999941116

julia> map
 * Status: success

 * Candidate solution
    Minimizer: [7.50e-01]
    Minimum:   2.400377e+00

 * Found with
    Algorithm:     L-BFGS
    Initial Point: [1.00e+00]

 * Convergence measures
    |x - x'|               = 2.50e-01 ≰ 0.0e+00
    |x - x'|/|x'|          = 3.33e-01 ≰ 0.0e+00
    |f(x) - f(x')|         = 6.25e-02 ≰ 0.0e+00
    |f(x) - f(x')|/|f(x')| = 2.60e-02 ≰ 0.0e+00
    |g(x)|                 = 3.67e-11 ≤ 1.0e-08

 * Work counters
    Seconds run:   0  (vs limit Inf)
    Iterations:    1
    f(x) calls:    3
    ∇f(x) calls:   3
```

You can also get the corresponding ranges by
```julia
julia> b, sym2val = Turing.bijector(model; sym_to_ranges = Val(true));

julia> sym2val
(m = UnitRange{Int64}[1:1],)
```
which could be used internally to construct the `NamedArray`.

src/modes/ModeEstimation.jl
torfjelde: Potentially interesting though they don't have `LabelledArrays.jl`: https://github.com/mcabbott/AxisKeys.jl/blob/51550fe96f36937f7e4417adf77232f33f35657e/test/speed.jl

src/modes/ModeEstimation.jl
torfjelde: Could also use `Turing.gradient_logp` and pass this to `optimize` if there is some magic in there that we need.

devmotion: I suggested doing that and IMO we should change `make_logjoint` etc in the same way. It is much easier to hook into and analyze the internals if we don't use anonymous closures (see e.g. https://github.com/TuringLang/DynamicPPL.jl/issues/45).

torfjelde: That makes sense, and I agree that the closure-idea isn't particularly nice (unless the user does it themselves, making things more transparent) :+1: 

Also, I guess this `OptimizationContext` will cover most use-cases + be slightly more efficient than a closure over `DefaultContext`.

src/modes/ModeEstimation.jl
cpfiffer: I don't think it's needed anymore with `OptimizationContext`, I'll remove it.

src/modes/ModeEstimation.jl
cpfiffer: The main reason I picked NamedArrays is because they have good `show` methods. LabelledArrays is pretty damn ugly and not good for presenting results.

src/modes/ModeEstimation.jl
cpfiffer: I'd like to note here that this is more or less what I originally had -- @devmotion requested a callable struct in the first place ([here](https://github.com/TuringLang/Turing.jl/pull/1230#discussion_r413531662)).

I have no preference either way but I don't much care for going back and forth.

cpfiffer: Currently I like @devmotion's original suggestion and would prefer to keep the callable struct. In fact, I think later on it should be generalized outside of the `Optim` context so that everyone can use it for whatever purposes they have.

src/modes/ModeEstimation.jl
cpfiffer: Turns out we don't need to unlink it for the Hessian, since the OptimizationContext handles everything so nicely.

src/Turing.jl
devmotion: You forgot to remove the exports here.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+    return 0
+end
+
+function DynamicPPL.tilde(ctx::OptimizationContext, sampler, dist, value, vi)
+    return Distributions.logpdf(dist, value)
+end
+
+# dot assume
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:LikelihoodContext}, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
```
devmotion, line 53: `vns` and `dist` are not used, it seems?
```suggestion
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+
+"""
+    OptimizationContext{C<:AbstractContext} <: AbstractContext
+
+The `OptimizationContext` transforms variables to their constrained space, but
+does not use the density with respect to the transformation. This context is
+intended to allow an optimizer to sample in R^n freely.
+"""
+struct OptimizationContext{C<:AbstractContext} <: AbstractContext
+	context::C
```
devmotion, line 28: No tab space:
```suggestion
    context::C
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+end
+
+# assume
+function DynamicPPL.tilde(ctx::OptimizationContext{<:LikelihoodContext}, spl, dist, vn::VarName, inds, vi)
+    r = vi[vn]
+	return r, 0
```
devmotion, line 34:
```suggestion
    return r, 0
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+    return 0
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:PriorContext}, sampler, right, left, vi)
+    return 0
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext, sampler, right, left, vn, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = get_and_set_val!(vi, vns, dist, sampler)
```
devmotion, line 74: Why do we use `get_and_set_val!` here but used just `vi[vn]` in the other `dot_tilde` implementations? From looking at the implementation of `get_and_set_val!` (https://github.com/TuringLang/DynamicPPL.jl/blob/f74e039cb7fff30a10be882b6e12e9b686dbc77c/src/context_implementations.jl#L286-L347) I assumed that we need
```julia
function getval(
    vi,
    vns::AbstractVector{<:VarName},
)
    r = vi[vns]
    return r
end

function getval(
    vi,
    vns::AbstractArray{<:VarName},
)
    r = reshape(vi[vec(vns)], size(vns))
    return r
end
```
to just extract the values correctly, and just call `r = getval(vi, vns)` then here. It seems we could even drop the first definition.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+# dot assume
+function DynamicPPL.dot_tilde(ctx::OptimizationContext{<:LikelihoodContext}, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    return vi[vn], 0
+end
+
+function DynamicPPL.dot_tilde(ctx::OptimizationContext, sampler, right, left, vn::VarName, _, vi)
+    vns, dist = get_vns_and_dist(right, left, vn)
+    r = vi[vn]
+    return r, loglikelihood(dist, r)
```
devmotion, line 60: I think we should get rid of the `sum(logpdf.(dist, r))` calls but as I said I assume that just changing them to `loglikelihood` will break AD and be problematic if `dist` is an array of distributions. Maybe we should do that in a separate PR, I guess by introducing a separate `sum_logpdf` in DistributionsAD (to avoid having to commit type piracy when defining `loglikelihood(dist::AbstractArray{<:Distribution}, x)`) with optimized adjoints for AD and a fallback to `loglikelihood(dist::Distribution, x)`.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+
+"""
+    OptimLogDensity(model::Model, context::AbstractContext)
+
+Create a callable `OptimLogDensity` struct that evaluates a model using the given `context`.
+"""
+function OptimLogDensity(model::Model, context::AbstractContext)
+	init = VarInfo(model)
+	DynamicPPL.link!(init, DynamicPPL.SampleFromPrior())
+	return OptimLogDensity(model, context, init)
```
devmotion, line 104:
```suggestion
    init = VarInfo(model)
    DynamicPPL.link!(init, DynamicPPL.SampleFromPrior())
    return OptimLogDensity(model, context, init)
```

test/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,105 @@
...
+        @test isapprox(mle.values.array, true_beta)
+        @test isapprox(vcmat, vcmat_mle)
+    end
+
+    @testset "Dot tilde test" begin
+        @model function dot_gdemo(x)
+            s ~ InverseGamma(2,3)
+            m ~ Normal(0, sqrt(s))
+        
+            x .~ Normal(m, sqrt(s))
```
devmotion, line 91: `.~` can't be parsed on Julia 1.0

src/modes/ModeEstimation.jl
devmotion: I think that should be part of the `OptimLogDensity` constructor ideally and not be handled here. For convenience, one could then forward it from the `MAP` and `MLE` methods (or include it as a field?) when `OptimLogDensity` is constructed.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,267 @@
...
+
+"""
+    Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+
+Estimate a mode, i.e., compute a MLE or MAP estimate.
+"""
+function Optim.optimize(model::Model, f::OptimLogDensity, optimizer=Optim.LBFGS(), args...; kwargs...)
+    # Do some initialization.
+    spl = DynamicPPL.SampleFromPrior()
+    init_params = model(f.vi, spl)
```
devmotion, line 247: `init_params` is not used anywhere:
```suggestion
```

devmotion: (and `model(...)` just returns whatever the user returns in the model definition)

src/modes/ModeEstimation.jl
trappmartin: Make sense.

src/modes/ModeEstimation.jl
trappmartin: Cool.

src/modes/ModeEstimation.jl
trappmartin: Maybe this was supposed to be a `getall(f.vi)`?

src/modes/ModeEstimation.jl
cpfiffer: Agreed. I'm sure we'll see how this'll fit with AD as people test it out on more complex models.

src/modes/ModeEstimation.jl
cpfiffer: I've added these in for the moment. Seems to be working well.

cpfiffer:
Alright, tests are passing at the moment. Further comments or things you don't think have been addressed yet?

src/modes/ModeEstimation.jl
mohamed82008: >Do you mean as a return value or a starting point?

I mean as a starting point. We also need to make sure the varinfo here is linked. If the same objective is optimized twice, the varinfo will not be linked here I think.

cpfiffer: You're right on that -- I've linked it after the optimized values are extracted. Good catch.

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,284 @@
...
+
+    # Optimize!
+    M = Optim.optimize(f, init_vals, optimizer, args...; kwargs...)
+
+    # Get the VarInfo at the MLE/MAP point, and run the model to ensure 
+    # correct dimensionality.
+    f.vi[spl] = M.minimizer
+    invlink!(f.vi, spl)
+    vals = f.vi[spl]
+    link!(f.vi, spl)
```
mohamed82008, line 274: This means the hessian will not be wrt to the constrained parameters?

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,284 @@
...
+    estimates = m.values.array[:,1]
+    stderrors = StatsBase.stderror(m)
+    tstats = estimates ./ stderrors
+
+    StatsBase.CoefTable([estimates, stderrors, tstats], ["estimate", "stderror", "tstat"], terms)
+end
+
+function StatsBase.informationmatrix(m::ModeResult; hessian_function=ForwardDiff.hessian, kwargs...)
+    # Calculate Hessian and information matrix.
+    varnames = StatsBase.coefnames(m)
```
mohamed82008, line 191: Should probably check the vi is invlinked here. If not, then run `invlink!`.

cpfiffer: What's the quick, idiomatic way to check whether the whole VI is linked?

src/modes/ModeEstimation.jl
cpfiffer: Yup, you're right. This is what I get for relying on bad behavior elsewhere in the code.

src/modes/ModeEstimation.jl
mohamed82008: islinked(vi, SampleFromPrior())` should be good enough. It will only check the first varname of each symbol though.

cpfiffer: Okay, I made a quick work around to get this to work with `SampleFromPrior`, since `islinked` only works with `Sampler`. I opened a PR to fix this here: https://github.com/TuringLang/DynamicPPL.jl/pull/124

For right now, I think it's fine to use my quick workaround and replace it later if/when the DPPL PR goes through.

cpfiffer:
Aight, I think it's finally good to go. :fireworks:

src/Turing.jl
```diff
@@ -58,6 +58,11 @@ using .Variational
     end
 end
 
+@init @require Optim="429524aa-4258-5aef-a3af-852621145aeb" @eval begin
+    include("modes/ModeEstimation.jl")
+    export MAP, MLE, optimize
+end
```
devmotion, line 64: Just noticed: users will have problems if they load Optim but not NamedArrays. We should add a nested `@require` block that checks for NamedArrays if we want to keep it optional (which I guess we want). Otherwise one could think about returning a named tuple as default and just provide some optional way for converting it to a NamedArray.

src/Turing.jl
devmotion, line 121:
```suggestion
        arraydist
```

src/modes/ModeEstimation.jl
```diff
@@ -0,0 +1,301 @@
...
+using LinearAlgebra
+
+import ..AbstractMCMC: AbstractSampler
+import ..DynamicPPL
+import ..DynamicPPL: Model, AbstractContext, VarInfo, AbstractContext, VarName,
+    _getindex, getsym, getfield, settrans!,  setorder!,
+    get_and_set_val!, istrans, tilde, dot_tilde
+import Optim
+import Optim: optimize
+import NamedArrays
```
devmotion, line 12: Optional dependencies have to be loaded as 
```suggestion
import .Optim
import .Optim: optimize
import .NamedArrays
```
otherwise users will get a warning about them not being listed as dependencies of Turing.

mohamed82008:
So if `NamedArrays` is an optional dependency, why does it not have a require block?

cpfiffer:
I think I'd actually prefer to make `NamedArrays` a full dependency, just for the moment. I think all this stuff should probably be spun off into a different package so the deps can be managed more specifically.

torfjelde:
Also, the VI part would benefit from using `NamedArrays` because right now it's difficult for the user to figure out what the different elements in the vector refers to (you can use `bijector(model; symbol_to_ranges = Val(true))` but that's a "hack" that I've added in so that just have _some_ way of doing it). And the reason why I didn't make a proper solution was to reduce deps. So making `NamedArrays` a explicit dep is also useful for other parts in Turing:)

Issue #1233, opened 2018-11-10
Replace consume and Tasks with Julia iterator


mohamed82008:
I have scanned through the `consume`-`produce` pattern use in Turing.jl and I noticed the following:
1. A `Trace` aka `Particle` is nothing but a `model` wrapper that calls it with inputs `vi` and `spl` when "consumed".
2. A `ParticleContainer` is a collection of `Particle`s and their results when consumed, i.e. the `model` function is run once for each particle. When a `ParticleContainer` is consumed, it iterates over its `Particle`s and consumes them one by one.
3. There is no other place in Turing.jl where the `consume`-`produce` pattern is used.

Given the above, can't we just make `ParticleContainer` an iterator and instead of calling 
https://github.com/TuringLang/Turing.jl/blob/6e6197a6afe7e03ea91847f50804cea362736bee/src/samplers/pgibbs.jl#L70, 
we can do: 
```julia
for p in particles
    vi = p.vi
    ....
end
```
Notice that the above `for` loop will call `iterate(::ParticleContainer, state)` which we can define to do what `consume` https://github.com/TuringLang/Turing.jl/blob/master/src/core/container.jl#L76 currently does returning `p` which holds the result of the `model` call in `p.vi`. `state` can hold all the state variables used inside `consume(::ParticleContainer)` now. More conveniently, we can also define `consume` as a [ResumableFunction](https://github.com/BenLauwens/ResumableFunctions.jl) and still call it with the `for` loop syntax above. Did I miss anything?

PR #1237, opened 2020-04-25
Use `propagate!` instead of `Libtask.consume`

devmotion:
This PR removes Libtask in the propagation of `ParticleContainer`. Instead it introduces a function `propagate!` that propagates all particles one step and returns `true` if all particles are propagated to the final time step and `false` otherwise (maybe one would rather want to use more explicit symbols instead of booleans?).

I always had the same impression as @mohamed82008 (see https://github.com/TuringLang/Turing.jl/issues/1233) that in particular for the propagation of the particle container one does not have to use Libtask. However, since `collect(::ParticleContainer)` returns the particles IMO it would be surprising if iterating over a particle container would propagate it. Moreover, since it seems there are no states that have to be tracked by the container (I even removed `n_consume` and `logE` in the particle container), it felt more natural to add a method for propagation, similar to `resample!` for the resampling step, than using a dedicated iterator for `ParticleContainer`.

I didn't fully understand the existing implementation of the likelihood estimation (isn't the currently reported log evidence incorrect since `logZ` neglects the constant term `-log(num_particles)`?), so I used the algorithm in http://www.it.uu.se/research/systems_and_control/education/2019/smc/schedule/lecture9.pdf (pages 5, 6, and 7) as a template - basically, now the implementation for `SMC` and `PG` should perform exactly the same sequence of propagation and resampling steps as noted there using `propagate!` and `resample!`.

Aside from that, I'm slightly unhappy with the current implementation of `SMC` since it somehow misuses `spl.state.average_logevidence` to save the log likelihood estimate of all particles that is computed in `sample_init!`, then uses it in every iteration of `step!` in which the `ParticleTransition`s for every particle are created, and finally computes the "overall" log evidence in `bundle_samples!` by setting `spl.state.average_logevidence` to the mean of the log evidence values of these particles (i.e., the value that `spl.state.average_logevidence` was set to in `sample_init!`). It's also a bit annoying that for SMC an `iteration` keyword argument is needed (IMO that should be part of the state, i.e., transition and not be part of the generic algorithm in AbstractMCMC since it seems only/mostly SMC uses it).


test/core/container.jl
```diff
@@ -44,21 +42,38 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+        # Initial weights and likelihood.
+        weights = getweights(pc)
+        lz = logZ(pc)
+        @test weights == [1/3, 1/3, 1/3]
+        @test iszero(lz)
 
-        @test consume(pc) == log(1)
+        # Propagate particles.
+        propagate!(pc)
+        @test getweights(pc) == weights
```
yebai, line 53: I'm confused here. Shouldn't new weights be different from old weights, since no resampling is performed in `propagate!`?

yebai: OK, I forgot that there was no observation in the model. For the purpose of correctly testing `propagate!`, maybe consider adding some likelihood in the test?

yebai:
Thanks, @devmotion. It's a good idea to remove the dependency on `Libtask.comsume` since it does not add any extra functionality except for recycling a function name. I think we can further make `Particle`'s dependency on `Libtask` optional, by implementing an iterator interface mentioned by @mohamed82008 for `CTask` types. This way, the SMC implementation can support alternative approaches for resampling particles other than using `Libtask.CTask`, e.g. for less generic models that can be represented as Markov process. In such models, we can either manually handle particle duplication during the resampling step, or consider something like [`ResumableFunctions.jl`](https://github.com/BenLauwens/ResumableFunctions.jl). 


> I didn't fully understand the existing implementation of the likelihood estimation (isn't the currently reported log evidence incorrect since logZ neglects the constant term -log(num_particles)?)

I think we had a step for subtracting the constant term in previous versions of the code, but I can't recall where that was, and when that was removed.

devmotion:
>  I think we can further make Particle's dependency on Libtask optional, by implementing an iterator interface mentioned by @mohamed82008 for CTask types. This way, the SMC implementation can support alternative approaches for resampling particles other than using Libtask.CTask, e.g. for less generic models that can be represented as Markov processes. In such models, we can either manually handle particle duplication during the resampling step, or consider something like ResumableFunctions.jl.

I completely agree, probably it would be useful to implement and use an iterator interface for particles, so that one could use alternative particle implementations such as ordinary stateful iterators. IMO it's just not useful and unintuitive (due to the behaviour of `collect`) to implement an iterator interface for propagating a ParticleContainer. I thought the changes for Particles should better go into a separate PR (and as you said, one should first implement the interface for CTask).

devmotion:
I'll update the test in a separate PR.

PR #1242, opened 2020-04-27
Simplify ParticleContainer and add better tests for propagate!

devmotion:
This PR simplifies ParticleContainer a bit more by removing the `model` field which hasn't been used (and was actually misused in the tests by setting it to arbitrary functions). Moreover, it adds the additional tests promised in https://github.com/TuringLang/Turing.jl/pull/1237#issuecomment-620051358.

When I added the tests I was a bit surprised actually that the implementation of `propagate!` (and `consume` before, that code was not modified in the recent PR) increases the unnormalized logarithmic weights of the particles with the values that are `produce`d by the particles and not just sets them to these values (as I would have expected from the algorithm since, if I understand correctly, these values correspond to the log probability of the next "observation"). In the currently implemented samplers (`SMC` and `PG`) this does not matter though since the unnormalized weights are reset in the `resample!` step which is performed before every `propagate!` step. Nevertheless, I'm wondering what's the reason for the current implementation?

### DIFF
--- a/test/core/container.jl
+++ b/test/core/container.jl
@@ -1,7 +1,7 @@
 using Turing, Random
 using Turing: ParticleContainer, getweights, resample!,
-    effectiveSampleSize, Trace, current_trace, VarName,
-    Sampler, consume, produce, copy, fork
+    effectiveSampleSize, Trace, current_trace, VarName, VarInfo,
+    Sampler, consume, produce, copy, fork, getlogp
 using Turing.Core: logZ, propagate!
 using Test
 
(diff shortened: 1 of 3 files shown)


yebai:
> When I added the tests I was a bit surprised actually that the implementation of propagate! (and consume before, that code was not modified in the recent PR) increases the unnormalized logarithmic weights of the particles with the values that are produced by the particles and not just sets them to these values (as I would have expected from the algorithm since, if I understand correctly, these values correspond to the log probability of the next "observation"). In the currently implemented samplers (SMC and PG) this does not matter though since the unnormalized weights are reset in the resample! step which is performed before every propagate! step. Nevertheless, I'm wondering what's the reason for the current implementation?

IIRC, these are intended for supporting adaptive resampling, where the particle weights get accumulated for multiple time steps until the ESS falls below some threshold.

test/core/container.jl
```diff
@@ -10,76 +10,74 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+        sampler = Sampler(PG(5), empty_model())
 
-        # Initial weights and likelihood.
-        weights = getweights(pc)
-        lz = logZ(pc)
-        @test weights == [1/3, 1/3, 1/3]
-        @test iszero(lz)
+        # Create particle container.
+        logps = [0.0, -1.0, -2.0]
+        particles = [Trace(fpc(logp), empty_model(), sampler, VarInfo()) for logp in logps]
```
yebai, line 37: This looks good. On a side note, it's strange that we still have the following two `Trace` constructors:

```julia
function Trace(f::Function, m::Model, spl::AbstractSampler, vi::AbstractVarInfo)
function Trace(m::Model, spl::AbstractSampler, vi::AbstractVarInfo)
```

maybe we can consider merging these two in the future.

devmotion: Yep, that's a bit strange indeed.

PR #1243, opened 2020-04-29
Allow sampling from the prior

devmotion:
This PR is an alternative to https://github.com/TuringLang/Turing.jl/pull/1225. I tried to reduce the existing amount of type piracy and to clean the existing implementation by moving HMC-specific stuff to the implementation of the AbstractMCMC interface for HMC samplers. The bundle_samples implementations for SampleFromPrior are still type piracy, since neither the function nor the input types are owned by Turing. Also to avoid type piracy and too Turing-specific implementations of the interface for SampleFromPrior (Turing sets the default chain_type to `MCMCChains.Chains` which one can't do in a generic implementation in DynamicPPL), I added a `Prior` algorithm type that is owned by Turing and just samples using a `SampleFromPrior`. However, I'm not completely sure if that's a good idea.

The PR needs the implementation of `step!` for SampleFromPrior in the master branch of DynamicPPL.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -51,6 +51,7 @@ export  InferenceAlgorithm,
         SMC,
         CSMC,
         PG,
+        Prior,
         assume,
         dot_assume,
         observe,
@@ -69,6 +70,9 @@ abstract type AdaptiveHamiltonian{AD} <: Hamiltonian{AD} end
 
 getchunksize(::Type{<:Hamiltonian{AD}}) where AD = getchunksize(AD)
 getADbackend(::Hamiltonian{AD}) where AD = AD()
+
+# Algorithm for sampling from the prior
+struct Prior end
 
 """
     mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::Real)
@@ -138,7 +142,7 @@ const TURING_INTERNAL_VARS = (internals = [
 
 function AbstractMCMC.sample(
     model::AbstractModel,
-    alg::InferenceAlgorithm,
+    alg::Union{InferenceAlgorithm,Prior},
     N::Integer;
     kwargs...
 )
@@ -158,7 +162,7 @@ end
 function AbstractMCMC.sample(
     rng::AbstractRNG,
     model::AbstractModel,
-    sampler::Sampler,
+    sampler::Sampler{<:InferenceAlgorithm},
     N::Integer;
     chain_type=MCMCChains.Chains,
     resume_from=nothing,
@@ -174,8 +178,26 @@ function AbstractMCMC.sample(
 end
 
 function AbstractMCMC.sample(
+    rng::AbstractRNG,
     model::AbstractModel,
-    alg::InferenceAlgorithm,
+    alg::Prior,
+    N::Integer;
+    chain_type=MCMCChains.Chains,
+    resume_from=nothing,
+    progress=PROGRESS[],
+    kwargs...
+)
+    if resume_from === nothing
+        return AbstractMCMC.mcmcsample(rng, model, SampleFromPrior(), N;
+                                       chain_type=chain_type, progress=progress, kwargs...)
+    else
+        return resume(resume_from, N; chain_type=chain_type, progress=progress, kwargs...)
+    end
+end
+
+function AbstractMCMC.sample(
+    model::AbstractModel,
+    alg::Union{InferenceAlgorithm,Prior},
     parallel::AbstractMCMC.AbstractMCMCParallel,
     N::Integer,
     n_chains::Integer;
@@ -201,7 +223,7 @@ end
 function AbstractMCMC.sample(
     rng::AbstractRNG,
     model::AbstractModel,
-    sampler::Sampler,
+    sampler::Sampler{<:InferenceAlgorithm},
     parallel::AbstractMCMC.AbstractMCMCParallel,
     N::Integer,
     n_chains::Integer;
@@ -213,10 +235,25 @@ function AbstractMCMC.sample(
                                    chain_type=chain_type, progress=progress, kwargs...)
 end
 
+function AbstractMCMC.sample(
+    rng::AbstractRNG,
+    model::AbstractModel,
+    alg::Prior,
+    parallel::AbstractMCMC.AbstractMCMCParallel,
+    N::Integer,
+    n_chains::Integer;
+    chain_type=MCMCChains.Chains,
+    progress=PROGRESS[],
+    kwargs...
+)
+    return AbstractMCMC.sample(rng, model, SampleFromPrior(), parallel, N, n_chains;
+                               chain_type=chain_type, progress=progress, kwargs...)
+end
+
 function AbstractMCMC.sample_init!(
     ::AbstractRNG,
-    model::Model,
-    spl::Sampler,
+    model::AbstractModel,
+    spl::Sampler{<:InferenceAlgorithm},
     N::Integer;
     kwargs...
 )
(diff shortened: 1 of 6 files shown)


devmotion:
As I said, I'm not fully satisfied with the fact that we just introduce a `Prior` struct to be able to define some default `chain_type` and `progress` keyword arguments. So I thought a bit more about the current situation and these are my two cents.

One idea would be to move the `bundle_samples` for certain output types to their respective packages, e.g., for `MCMCChains.Chains` to MCMCChains where it would be implemented for all samplers, models, and samples as generically as possible by just converting the samples to an axis array (e.g., by calling `MCMCChains.samples2array(model, sampler, transitions)`), getting the name map (by, e.g., calling `MCMCChains.getnamemap(model, sampler, transitions)`), extracting the logevidence (by calling `MCMCChains.getlogevidence(model, sampler, transitions)`, and the info (by calling `MCMCChains.getinfo(model, sampler, transitions)`), and then using the results to build the `Chains` object. Then Turing (and other packages) could implement these methods for their combination of samplers, models, and transitions. However, that would not solve the type piracy problem for `SampleFromPrior` (as long as DynamicPPL does not depend on MCMCChains) and create a bunch of new methods that one would have to implement...

Another idea would be to just drop `chain_type` and to always return the actual samples (if one does not reimplement/modify the default `sample` logic in AbstractMCMC). If one would have an abstract interface in AbstractMCMC for accessing parameter values and additional information of samples, then users would still get some basic functionality and could retrieve, e.g., all samples of parameter `:p` or the log probabilities in an easy way. For some rudimentary use cases that might be enough, but additionally packages such as MCMCChains should then provide an implementation of how to construct a `Chains` object from any set of samples (and possibly the model and sampler) that implements this interface. Then it would still be fairly easy for users to construct a `Chains` object that gives them access to all the diagnostic tools in MCMCChains. And of course one could implement the same interface also for `VarInfo`, and hence even if `SampleFromPrior` returns a vector of `VarInfo`, users could extract samples etc. AbstractMCMC could even implement default methods for how to convert samples that implement this interface to types in Julia base such as `NamedTuple` or `Vector`. Not performing any transformations of the samples during or after the sampling procedure and just returning the "raw" samples would maybe also allow more advanced (and possibly more efficient?) use cases without hooking into the internals (too much).

src/inference/Inference.jl
```diff
@@ -70,6 +71,9 @@ abstract type AdaptiveHamiltonian{AD} <: Hamiltonian{AD} end
 getchunksize(::Type{<:Hamiltonian{AD}}) where AD = getchunksize(AD)
 getADbackend(::Hamiltonian{AD}) where AD = AD()
 
+# Algorithm for sampling from the prior
+struct Prior end
```
yebai, line 75: `Prior` can be viewed as the crudest way of performing approximate inference: 
```suggestion
struct Prior <: InferenceAlgorithm end
```

src/inference/Inference.jl
```diff
@@ -138,7 +142,7 @@ const TURING_INTERNAL_VARS = (internals = [
 
 function AbstractMCMC.sample(
     model::AbstractModel,
-    alg::InferenceAlgorithm,
+    alg::Union{InferenceAlgorithm,Prior},
```
yebai, line 145:
```suggestion
    alg::InferenceAlgorithm,
```

src/inference/Inference.jl
```diff
@@ -174,8 +178,26 @@ function AbstractMCMC.sample(
...
+        return AbstractMCMC.mcmcsample(rng, model, SampleFromPrior(), N;
+                                       chain_type=chain_type, progress=progress, kwargs...)
+    else
+        return resume(resume_from, N; chain_type=chain_type, progress=progress, kwargs...)
+    end
+end
+
+function AbstractMCMC.sample(
+    model::AbstractModel,
+    alg::Union{InferenceAlgorithm,Prior},
```
yebai, line 200:
```suggestion
    alg::InferenceAlgorithm,
```

src/inference/Inference.jl
```diff
@@ -268,19 +294,35 @@ end
...
+"""
+    getparams(t)
+
+Return a named tuple of parameters.
+"""
+getparams(t) = t.θ
+getparams(t::VarInfo) = tonamedtuple(TypedVarInfo(t))
+
+"""
+    getlp(t)
```
yebai, line 306: maybe `getlogp` for consistency with `VarInfo.logp` 
```suggestion
    getlogp(t)
```

src/inference/Inference.jl
```diff
@@ -300,7 +342,12 @@ function flatten_namedtuple(nt::NamedTuple)
     return [vn[1] for vn in names_vals], [vn[2] for vn in names_vals]
 end
 
-function get_transition_extras(ts::Vector)
+function get_transition_extras(ts::AbstractVector{<:VarInfo})
+    valmat = reshape([getlogp(t) for t in ts], :, 1)
+    return ["lp"], valmat
```
yebai, line 347:
```suggestion
    return ["logp"], valmat
```

src/inference/Inference.jl
```diff
@@ -340,50 +387,46 @@ function get_transition_extras(ts::Vector)
...
-    elseif :final_logevidence in fieldnames(typeof(spl.state))
-        le = getproperty(spl.state, :final_logevidence)
-    end
-
-    # Check whether to invlink! the varinfo
-    if islinked(spl.state.vi, spl)
-        invlink!(spl.state.vi, spl)
-    end
+    # Get the average or final log evidence, if it exists.
+    le = getlogevidence(spl)
```
yebai, line 425: nice!

yebai:
tests are failing with the following error:

```
  MethodError: no method matching step!(::MersenneTwister, ::DynamicPPL.Model{getfield(Main, Symbol("###evaluator#7567")),(),Tuple{},(),DynamicPPL.ModelGen{getfield(Main, Symbol("###generator#7568")),(),(),Tuple{}}}, ::SampleFromPrior, ::Int64, ::Nothing; iteration=1)
```

https://travis-ci.com/github/TuringLang/Turing.jl/jobs/325101686#L623

devmotion:
A new DynamicPPL release should fix the test errors, at least it worked for me with the master branch.

src/inference/Inference.jl
devmotion: I was a bit reluctant of overloading `DynamicPPL.getlogp` here. But maybe we should do that?

src/inference/Inference.jl
devmotion: I called it "lp" since then it would be consistent with all other inference algorithms (`Transition`, `ParticleTransition`, and `HamiltonianTransition` all report "lp"). Maybe we should rename all to "logp" (but probably in a separate PR) or just keep "lp" here?

yebai: Sounds good.

src/inference/Inference.jl
yebai: I think it should be fine to reuse `getlogp` here..

devmotion:
I'll update this PR and address the comments when https://github.com/TuringLang/Turing.jl/pull/1249 is merged.

src/inference/Inference.jl
devmotion: I removed `getlp` and instead defined `getlogp` for all transition types.

yebai:
The failing tests are related to the prior sampler tests. Seems the error bound for the prior sampler tests is too tight.

devmotion:
I set a seed and increased the number of samples. Locally tests pass now, hopefully this fixes CI tests as well :crossed_fingers:

yebai:
Looks like the tests are passing now - thanks, @devmotion!

PR #1246, opened 2020-05-01
Remove sidebar for news/teams

cpfiffer:
Removes the navigation sidebar and table of contents for the news and team pages, so we now have lovely white margins for things like blog posts.


yebai:
thanks, @cpfiffer!

PR #1248, opened 2020-05-02
Make Turing compatible with Libtask 0.4

devmotion:
Turing equivalent of https://github.com/TuringLang/DynamicPPL.jl/pull/95.


yebai:
thanks, @devmotion!

PR #1251, opened 2020-05-03
Update history

mohamed82008:
This PR prepares the history for the next release. We might need to update it before the release.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,7 +5,8 @@
 - There were some performance improvements in the automatic differentiation of functions in `DistributionsAD` and `Bijectors` leading to speeds closer to and sometimes faster than Stan's.
 - An HMC initialization bug was fixed. HMC initialization in Turing is now consistent with Stan's.
 - Sampling from the prior is now possible using `sample`.
-- Turing now uses AdvancedMH v0.5 and AbstractMCMC v1.0.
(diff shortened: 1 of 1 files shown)


HISTORY.md
```diff
@@ -1,3 +1,12 @@
+# Release 0.12.0
+
+- The interface for defining new distributions with a constrained support and making them compatible with `Turing` has changed. To make a custom distribution type `CustomDistribution` compatible with `Turing`, the user needs to define the method `bijector(d::CustomDistribution)` that returns an instance of type `Bijector` implementing the `Bijectors.Bijector` API.
+- `~` is now thread-safe when used for observations, but not assumptions yet.
+- There were some performance improvements in the automatic differentiation of functions in `DistributionsAD` and `Bijectors` leading to speeds closer to and sometimes faster than Stan's.
+- An HMC initialization bug was fixed. HMC initialization in Turing is now consistent with Stan's.
+- Sampling from the prior is now possible using `sample`.
+- Turing now uses AdvancedMH v0.5 and AbstractMCMC v1.0.
```
devmotion, line 8: Maybe it would be good to explain what is different with AbstractMCMC 1.0, namely that `psample` is removed and one instead has to specify `MCMCThreads()` or `MCMCDistributed()` as parallel sampling algorithms.

cpfiffer: For the HISTORY.md revisions I think it's better to just add a suggestion than to comment on the line -- everyone worked on different parts and should just directly add their updates.

cpfiffer, line 8:
```suggestion
- `psample` is now deprecated, in favor of `sample(model, sampler, parallel_method, n_samples, n_chains)` where `parallel_method` can be either `MCMCThreads()` or `MCMCDistributed()`. `MCMCThreads` will use your available threads to sample each chain (ensure that you have the environment variable `JULIA_NUM_THREADS` set to the number of threads you want to use) and `MCMCDistributed` will dispatch chain sampling to each available processes (you can add processes with `addprocs()`).
- Turing now uses AdvancedMH v0.5, which mostly provides behind-the -scenes restructuring.
```

cpfiffer: I added mine in a separate comment.

PR #1258, opened 2020-05-04
Fix HMC initialization and allow AD in SampleFromPrior

mohamed82008:
This PR fixes the HMC initialization. It also enables type specialization of type parameters even when using samplers other than `Sampler{<:Hamiltonian}` such as `SampleFromPrior`. This is useful for calling AD on the model function outside of Turing for benchmarking.


mohamed82008:
I haven't run all the tests locally, so let's see if they pass.

mohamed82008:
A new DPPL release is needed.

mohamed82008:
The following test fails with 2 threads.
```julia
# Test that chains with the same seed will sample identically.
chain1 = sample(Random.seed!(5), gdemo_default, HMC(0.1, 7), MCMCThreads(), 1000, 4)
chain2 = sample(Random.seed!(5), gdemo_default, HMC(0.1, 7), MCMCThreads(), 1000, 4)
@test all(chain1.value .== chain2.value)
check_gdemo(chain1)
```
Should it pass though?

mohamed82008:
Also it seems Travis tests don't use multiple threads so we need to fix that.

mohamed82008:
Once tests pass, this should be ready to merge.

mohamed82008:
The test failures seem like a seed issue. I think it is safe to merge and then fix the seeds and/or sample sizes in another PR.

src/contrib/inference/dynamichmc.jl
```diff
@@ -61,8 +61,15 @@ function AbstractMCMC.sample_init!(
     end
 
     model(spl.state.vi, SampleFromUniform())
+    link!(spl.state.vi, spl)
```
xukai92, line 64: why do we add `link!` here @mohamed82008

src/inference/hmc.jl
```diff
@@ -100,21 +100,52 @@ function HMC{AD}(
...
 )
-
     # Resume the sampler.
     set_resume!(spl; resume_from=resume_from, kwargs...)
 
     # Get `init_theta`
     initialize_parameters!(spl; verbose=verbose, kwargs...)
+    if init_theta !== nothing
+        # Doesn't support dynamic models
+        link!(spl.state.vi, spl)
```
xukai92, line 128: I would assume `init_theta` is in the constrained space. But here we move to unconstrained space?

src/inference/hmc.jl
```diff
@@ -100,21 +100,52 @@ function HMC{AD}(
...
+        model(spl.state.vi, spl)
+        theta = spl.state.vi[spl]
+        spl.state.z.θ .= theta
+    else
+        # Samples new values and sets trans to true, then computes the logp
+        model(empty!(spl.state.vi), SampleFromUniform())
```
xukai92, line 134: What does this line actually do? I'm confused because it follows by the `link!`call, but I would expect the unifromlly initalized values are already in the unconstrained space.

src/contrib/inference/dynamichmc.jl
mohamed82008: Because `SampleFromUniform` calls `invlink` https://github.com/TuringLang/DynamicPPL.jl/blob/master/src/utils.jl#L131 and sets `trans` to false. Whether we want that or not is a separate issue. But for now that's the case. I changed it for a brief moment in DPPL but it was changed back.

src/inference/hmc.jl
mohamed82008: It is in the constrained space so we transform it to the unconstrained. `initialize_parameters!` puts `init_theta` in `vi` so no we transform it here with `link!`. Here we assume the trans fields in `vi` are all false, but we probably need to check that before calling `initialize_parameters!`.

src/inference/hmc.jl
mohamed82008: It puts values in `vi` which are sampled from the uniform distribution then transformed with `invlink`. This happens in `inittrans` in https://github.com/TuringLang/DynamicPPL.jl/blob/master/src/utils.jl#L131.

mohamed82008: Looking at the model I don't update the phase point which I should to use the transformed logp and gradient. I will fix that. @yebai please don't release yet.

src/contrib/inference/dynamichmc.jl
xukai92: I see. Thanks for confirming.

src/inference/hmc.jl
```diff
@@ -100,21 +100,52 @@ function HMC{AD}(
...
     set_resume!(spl; resume_from=resume_from, kwargs...)
 
     # Get `init_theta`
     initialize_parameters!(spl; verbose=verbose, kwargs...)
+    if init_theta !== nothing
+        # Doesn't support dynamic models
+        link!(spl.state.vi, spl)
+        model(spl.state.vi, spl)
+        theta = spl.state.vi[spl]
+        spl.state.z.θ .= theta
```
xukai92, line 131: So here `z.θ` is still assumed to be in the constrained space, even though we have `link!` above?

src/contrib/inference/dynamichmc.jl
xukai92: Great thanks for confirming.

src/inference/hmc.jl
mohamed82008: No, `theta = spl.state.vi[spl]` is not in constrained space because we called `link!`. Also please check the new PR. This PR is buggy.

src/inference/hmc.jl
xukai92: >  I don't update the phase point which I should to use the transformed logp and gradient

I believe we had some similar discussion before when we visit this issue for the first time. IRRC, we had figure out a way to set the initial value in the very beginning s.t. we don't need to manually make the phase point updated, which is a bit unintuitive to ensure everywhere outside AHMC.

mohamed82008: I think we do to compute the logp and its gradient at the initial theta in the unconstrained space and check that it is finite.

xukai92: This is done while constructing the phase point in AHMC.

What I mean by my comment above is that, i think the correctly logic in Turing is the following

1. Do whatever initialization for a constrained value
2. Link it to unconstrained value
3. Initial a phase point
4. Rest...

If the order is so, there is no need to update a phase point.

xukai92: (surely we still need to do this in Gibbs as you pointed in the new PR)

PR #1261, opened 2020-05-05
Also test Julia 1.0

devmotion:
I just noticed that Julia 1.0 is not tested anymore. I assume it is due to https://travis-ci.community/t/travis-build-gets-not-triggered-for-single-julia-version/4655: if there exists only one job and one specifies `matrix.include`, then the single job is not triggered. It works in AbstractMCMC since there we test on MacOS as well, and hence there are two base jobs.


yebai:
@devmotion I didn't realise this PR is merging into Mohammad’s. Maybe open a new one for master?

PR #1263, opened 2020-05-05
Clean CI setup

devmotion:
This PR removes tests on Julia 1.2 and 1.3, updates CompatHelper (it has Julia installed by default, so there's no need anymore to download Julia), and enforces the use of the latest stable Julia version (to avoid updates when a new Julia version is released).

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -8,9 +8,6 @@ jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - uses: julia-actions/setup-julia@latest
-        with:
-          version: 1.4
       - name: Pkg.add("CompatHelper")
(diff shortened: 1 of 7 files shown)


devmotion:
It seems, with the modified HMC initialization (https://github.com/TuringLang/Turing.jl/pull/1258) we require more samples. For some tests the number of samples was increased in this PR, but this didn't fix all test failures. They are unrelated to multithreading or thread-local RNGs since the tests use only a single thread, as shown in the logs:
```julia
┌ Warning: Only a single thread available: MCMC chains are not sampled in parallel
```
Hence even though we currently don't respect user-provided RNGs in every `rand` call and don't use thread-local RNGs correctly, the results should be reproducible by setting, e.g., `Random.seed!(100)` when only using a single thread.

cpfiffer:
I'm going to merge this over since the test failure is unrelated.

PR #1266, opened 2020-05-06
Fix log evidence computation

devmotion:
This PR fixes https://github.com/TuringLang/DynamicPPL.jl/issues/104 and transfers the corresponding tests to Turing. The main problem is that the formula that I based the computation of the log evidence in https://github.com/TuringLang/Turing.jl/pull/1237 on is only valid if resampling is performed in every time step, or rather if the weights are reset to 1/N before every reweighting step. A reference for the more general formula (which was used before) is eq (14) in [Sequential Monte Carlo Samplers by P. Del Moral, A. Doucet and A. Jasra](https://www.stats.ox.ac.uk/~doucet/delmoral_doucet_jasra_sequentialmontecarlosamplersJRSSB.pdf). The fix is particularly important since by default we don't perform resampling in every step but only in an adaptive way based on the estimated ESS.

More formally, we save only the unnormalized logarithmic weights, and accumulate them until the next resampling is performed, in which case the are reset to 0. Let `logw_k^i` denote the unnormalized logarithmic weight of the `i`th particle in the `k`th step. Hence in the notation of Del Moral, Doucet, and Jasra we have `logw_k^i = logw_{k-1}^i + log(w_k^i)`, and hence by induction normalizing `logw_k^i` yields `exp(logw_k^i) / \sum_{j=1}^N exp(logw_k^j) = w_k^i * exp(logw_{k-1}^i) / \sum_{j=1}^N w_k^j * exp(logw_{k-1}^j) = w_k^i * W_{k-1}^i / \sum_{j=1}^N w_k^j W_{k-1}^j = W_k^i`, i.e., the unnormalized weights in algorithm 3.1.1, as desired. Hence from eq. (14) we can compute the increase of the log evidence by `log(Z_k) - log(Z_{k-1}) = log(Z_k / Z_{k-1}) = log(\sum_{i=1}^N W_{k-1}^i * w_{k-1}^i) = log(\sum_{i=1}^N exp(logw_{k-1}^i) * w_{k-1}^i) - log(\sum_{i=1}^N exp(logw_{k-1}^i)) = log(\sum_{i=1}^N \exp(logw_{k-1}^i + log(w_{k-1}^i))) - log(\sum_{i=1}^N exp(logw_{k-1}^i)) = log(\sum_{i=1}^N exp(logw_k^i)) - log(\sum_{i=1}^N exp(logw_{k-1}^i))`.

Thus only if `logw_{k-1}^i` are all 0 (such as in the initial step and after resampling) we obtain the formula `log(Z_k) - log(Z_{k-1}) = log(\sum_{i=1}^N exp(logw_k^i)) - log(N)` from the reference on which the linked PR was based on.

As a final remark, I'm a bit unsatisfied with the function name `logZ` since, as shown above, it does not compute `log(Z)` but the logarithm of the normalization factor of the current unnormalized weights.


devmotion:
The test error on Windows is still the HMC error that https://github.com/TuringLang/Turing.jl/pull/1264 is supposed to fix.

src/core/container.jl
```diff
@@ -212,11 +216,109 @@ function resample!(
...
+        # Here ``yᵢ`` are observations, ``xᵢ`` variables of the particle filter, and
+        # ``θᵢ`` are variables of other samplers.
+        score = Libtask.consume(p)
+
+        if score === nothing
+            numdone += 1
+        else
+            # Increase the unnormalized logarithmic weights, accounting for the variables
+            # of other samplers.
+            increase_logweight!(pc, i, score + getlogp(p.vi))
```
yebai, line 249: As a side note, `getlotp(p.vi)` will always return `0`, since the assume and observe functions for particle samplers does not modify `vi.logp` by default. This doesn't affect correctness, but worth to pay attention. 

See:
- https://github.com/TuringLang/Turing.jl/blob/v0.9.0/src/inference/AdvancedSMC.jl#L288
- https://github.com/TuringLang/Turing.jl/blob/v0.9.0/src/inference/AdvancedSMC.jl#L293

src/core/container.jl
```diff
@@ -212,11 +216,109 @@ function resample!(
...
+        logZ0 = logZ(pc)
+
+        # Reweight the particles by including the next observation ``yₜ``.
+        isdone = reweight!(pc)
+
+        # Compute the normalizing constant ``Z₁`` after reweighting.
+        logZ1 = logZ(pc)
+
+        # Compute the estimate of the log evidence ``\\log p(y₁, …, yₜ)``.
+        logevidence += logZ1 - logZ0
```
yebai, line 316: This looks correct to me. Sorry for missing this issue in the previous PR. I wrote these code a long time ago, and my memory must have got rusty I think.

src/inference/AdvancedSMC.jl
```diff
@@ -138,46 +132,75 @@ function AbstractMCMC.step!(
...
     pc = spl.state.particles
-    Ws = getweights(pc)
+    weight = getweight(pc, iteration)
 
     # update the master vi
     particle = pc.vals[iteration]
     params = tonamedtuple(particle.vi)
+
+    # This is pretty useless since we reset the log probability continuously in the
+    # particle sweep.
```
yebai, line 144: thanks for the note, also see my comment above about assume and observe functions.

src/core/container.jl
```diff
@@ -212,11 +216,109 @@ function resample!(
...
+end
+
+"""
+    sweep!(pc::ParticleContainer, resampler)
+
+Perform a particle sweep and return an unbiased estimate of the log evidence.
+
+The resampling steps use the given `resampler`.
+"""
+function sweep!(pc::ParticleContainer, resampler)
```
yebai, line 278: Consider add the Del Moral, Doucet, and Jasra paper as a reference here.

src/core/container.jl
devmotion: Sorry, it was my bad. I just checked the slides of the SMC course and there they only discussed the likelihood estimator for the bootstrap particle filter with resampling in every step, so I messed it up.

src/core/container.jl
devmotion: That's not completely true (or maybe I misunderstand you), in this line `getlogp` can actually return nonzero values due to https://github.com/TuringLang/Turing.jl/blob/35e3fe8e356212ba4e64265a870833e3a9b9f4e0/src/inference/AdvancedSMC.jl#L286. However, since we call `resetlogp!` in one of the following lines, this won't show up in the saved transitions.

yebai: I must have missed that line, thanks for the pointer!

PR #1272, opened 2020-05-08
Updated @code_warntype steps to match version

HarrisonWilde:
Following from https://julialang.slack.com/archives/CCYDC34A0/p1588954831274400

### DIFF
--- a/docs/src/using-turing/performancetips.md
+++ b/docs/src/using-turing/performancetips.md
@@ -98,7 +98,7 @@ model = tmodel(1.0)
 varinfo = Turing.VarInfo(model)
 spl = Turing.SampleFromPrior()
 
-@code_warntype model.f(model, varinfo, spl, Turing.DefaultContext())
+@code_warntype model(varinfo, spl)
 ```
 to inspect the type instabilities in the model.
 
@@ -136,4 +136,3 @@ Then define the `Turing` model using the new functions as such:
     x .~ Normal(c, d)
 end
 ```
-
--- a/docs/src/using-turing/performancetips.md
+++ b/docs/src/using-turing/performancetips.md
@@ -98,7 +98,7 @@ model = tmodel(1.0)
 varinfo = Turing.VarInfo(model)
 spl = Turing.SampleFromPrior()
 
-@code_warntype model.f(varinfo, spl, Turing.DefaultContext(), model)
+@code_warntype model.f(model, varinfo, spl, Turing.DefaultContext())
 ```
 to inspect the type instabilities in the model.
 
--- a/docs/src/using-turing/performancetips.md
+++ b/docs/src/using-turing/performancetips.md
@@ -98,7 +98,7 @@ model = tmodel(1.0)
 varinfo = Turing.VarInfo(model)
 spl = Turing.SampleFromPrior()
 
-@code_warntype model(varinfo, spl)
+@code_warntype model.f(model, varinfo, spl, Turing.DefaultContext())
 ```
 to inspect the type instabilities in the model.
 


docs/src/using-turing/performancetips.md
```diff
@@ -98,7 +98,7 @@ model = tmodel(1.0)
 varinfo = Turing.VarInfo(model)
 spl = Turing.SampleFromPrior()
 
-@code_warntype model.f(varinfo, spl, Turing.DefaultContext(), model)
+@code_warntype model.f(model, varinfo, spl, Turing.DefaultContext())
```
devmotion, line 101: Maybe it would be a bit more intuitive (or even more correct) to advise to check
```suggestion
@code_warntype model(varinfo, spl)
```
in particular now that `model(varinfo, spl)` does not necessarily call `model.f(model, varinfo, spl, Turing.DefaultContext())`?

cpfiffer:
Copying the Slack post referenced for posterity:

>Harrison Wilde  9:20 AM
I am trying to follow the example here on checking for type stability but it seems no matter which tmodel I define, @code_warntype doesn’t give any output, any ideas why this might be? https://turing.ml/dev/docs/using-turing/performancetips

docs/src/using-turing/performancetips.md
mohamed82008: I don't think this will "reveal" the model body which is the point of calling `model.f` instead of `model`.

devmotion: But why would you want to check `model.f` instead of `model`, in particular if it might never be called with these types?

mohamed82008: To reveal type instabilities in the user's code giving the user some hints where to look in the model to fix it.

mohamed82008: Calling `model` will just show some type instability in the output but no details.

HarrisonWilde: yeah I am with @mohamed82008 here, I am not sure on the underlying mechanisms but model.f is what I was after in checking types at least

HarrisonWilde: If you compare outputs for both I think model.f is more in line with what a user might expect as this I assume is what runs on each iteration during sampling?

mohamed82008: Yes that's kinda what runs during sampling plus AD.

devmotion: No, during sampling we call `model(varinfo, spl)` in every iteration, we never call `model.f` explicitly. That's why I assumed it might be better to check the output for `model` and not `model.f`. Additionally, internally we also use different types for `varinfo`, so there could be problems even if `model.f` looks fine. If you check `model` that's not the case.

HarrisonWilde: calling it on model only gives this though:

```
julia> @code_warntype model(varinfo, spl)Variables
  model::DynamicPPL.Model{var"###evaluator#516",(:x,),Tuple{Float64},(),DynamicPPL.Model
Gen{var"###generator#517",(:x,),(),Tuple{}}}
  vi::DynamicPPL.VarInfo{NamedTuple{(:p,),Tuple{DynamicPPL.Metadata{Dict{DynamicPPL.VarN
ame{:p,Tuple{Tuple{Int64}}},Int64},Array{Normal{Float64},1},Array{DynamicPPL.VarName{:p,
Tuple{Tuple{Int64}}},1},Array{Float64,1},Array{Set{DynamicPPL.Selector},1}}}},Float64}
  spl::Core.Compiler.Const(DynamicPPL.SampleFromPrior(), false)

Body::Float64
1 ─ %1 = DynamicPPL.DefaultContext()::Core.Compiler.Const(DynamicPPL.DefaultContext(), false)
│   %2 = (model)(vi, spl, %1)::Float64
└──      return %2
```

mohamed82008: Yes please revert the last change.

devmotion: I agree that it might be less obvious what causes the type instability if one exists (I guess going down one step and checking `model(varinfo, sampler, DefaultContext())` is not much more helpful). However, isn't it weird to mention/recommend checking a function in the documentation that is 1) never called explicitly by any sampling algorithm in Turing and 2) even more importantly might be called indirectly only with different input types? Maybe one could add a comment that `@code_warntype model(varinfo, sampler)` actually checks what is happening during sampling but checking the internals with `@code_warntype model.f(model, varinfo, sampler, DefaultContext())` might easier to interpret but does not provide any guarantees on the type stability?

devmotion: One could maybe even recommend to check type stability with `@code_warntype model(varinfo, sampler)` and in case there is one to use Cthulhu to figure out what's causing it.

PR #1275, opened 2020-05-12
Gibbs Conditionals (new PR)

phipsgabler:
**This is just the same as https://github.com/TuringLang/Turing.jl/pull/1172, but I did some bad things with Git there, so creating a new PR seemed easier**

---

Introduces a sampler `GibbsConditional` that can be used to manually specify a -- surprise! -- Gibbs conditional for a variable:

```
sample(model, Gibbs(GibbsConditional(:λ, cond_λ), GibbsConditional(:m, cond_m)), N)
```

See also [this discussion](https://github.com/phipsgabler/LittleGibbs.jl/issues/3).

where the `cond_v` are functions from a `NamedTuple` of the conditioned variables (`(λ,)` for `m` and vice versa in the above example) to a `Distribution` (or probably just a `Sampleable` is enough).  [Example code](https://gist.github.com/phipsgabler/dfac25deceedd30c6642b8b6333797a0).

Currently, only one variable per sampler object is handled (as opposed to some other samplers).  Of course extension is possible, but I believe it is better to leave it this way and let `Gibbs` handle the sweeps.

Since this seemed too simple, I'm quite sure there's work left.  At least:

- [x] There's a lot of stuff (esp. `assume`/`observe`.) simply copied from MH -- can this be simplified or at least abstracted?
- [x] Documentation!
- [ ] Tests, naturally
  - [x] Basic feasibility (gdemo, mixture of Gaussians)
  - [ ] Check whether indexed variables really work
  - [x] Mixing Gibbs conditionals with other samplers
- [ ] Doctests (discuss first)

Currently, the code is working on the tests when only the conditionals or a combination of conditioals and MH is used, but fails in combination with an HMC sampler due to some AD things I don't understand.


src/inference/gibbs.jl
```diff
@@ -166,27 +166,45 @@ function AbstractMCMC.step!(
...
 end
 
+
+"""
+    gibbs_step!(rng, model, spl, N, transition; kwargs...)
+
+Perform a conditional sampling step of the local sampler `spl` within a `Gibbs` sampler.  If the
+sampler is based on an `InferenceAlgorithm`, this just falls back to `AbstractMCMC.step!`.
+"""
+function gibbs_step!(
```
cpfiffer, line 196: I'm not clear on why we use `gibbs_step!` instead of just `step!`. What's the value add here?

phipsgabler: I honestely don't remember :) Presumably because `GibbsConditional` is not an `AbstractSampler`.  But removing it does not make it work not, so I did that.

cpfiffer: Perfect, thanks!

devmotion: It came up in https://github.com/TuringLang/Turing.jl/pull/1172#issuecomment-602189523. I don't remember exactly, but I think the idea was that you might want to perform some initial setup or configurations in the steps of the Gibbs sampler that you wouldn't perform in regular sampling. Currently that's all checked in the `step!` implementation (e.g., for HMC and ESS), but it would be cleaner to not mix these branches in the general `step!` implementation. The default implementation of `gibbs_step!` would just call `step!` (so not previous setup) but algorithms such as HMC could implement `gibbs_step!`, which would perform the setup and then call `step!`. Of course, `GibbsConditional` could then still just implement `step!` but I'm not sure if it makes sense to use GibbsConditional for anything else than Gibbs steps, so it might be a bit weird to keep it that general.

devmotion: Maybe it would actually make sense to keep `gibbs_step!`? :thinking:

cpfiffer: I feel like we should adhere to the interface as it currently is until we have a really good reason to deviate.

devmotion: Yes, maybe it's actually better to clean the Gibbs specific parts of `step!` in a separate PR.

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,134 @@
...
+    conditional::C
+
+    function GibbsConditional(sym::Symbol, conditional::C) where {C}
+        return new{sym, C}(conditional)
+    end
+end
+
+DynamicPPL.getspace(::GibbsConditional{S}) where {S} = (S,)
+alg_str(::GibbsConditional) = "GibbsConditional"
+isgibbscomponent(::GibbsConditional) = true
```
devmotion, line 60: I guess this should maybe also be part of another PR (since `GibbsComponent` is not removed yet)?
```suggestion
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,134 @@
...
+"""Takes care of removing the `tonamedtuple` indexing form."""
+extractparam(p::Tuple{Vector{<:Array{<:Real}}, Vector{String}}) = foldl(vcat, p[1])
+function extractparam(p::Tuple{Vector{<:Real}, Vector{String}})
+    values, strings = p
+    if length(values) == length(strings) == 1 && !occursin(r".\[.+\]$", strings[1])
+        # if m ~ MVNormal(1, 1), we could have have ([1], ["m[1]"])!
+        return values[1]
+    else
+        return values
+    end
```
devmotion, line 133: We should change the return type of `tonamedtuple` and not use strings (in a separate PR). In particular with the next MCMCChains release we would otherwise create strings in every step (when calling `tonamedtuple`) and then convert them back to symbols when creating the Chains object.

phipsgabler: Yeah, given that the new `VarName`s are used already, this is a bit crazy.

devmotion: I think we should leave that for a separate PR though, I assume it requires quite some changes that are otherwise completely unrelated to GibbsConditional.

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,138 @@
...
+            μ ~ arraydist(Normal.(fill(m, K), fill(λ, K)))
+            z ~ arraydist(Categorical.(fill(π, N)))
+            x ~ arraydist(Normal.(μ[z], σ))
+            return x
+        end
+
+        # see http://www.cs.columbia.edu/~blei/fogm/2015F/notes/mixtures-and-gibbs.pdf
+        function cond_z(c)
+            function mixtureweight(x)
+                p = π .* pdf.(Normal.(c.μ, σ), Ref(x))
```
mohamed82008, line 85: Ref here is not needed?

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,134 @@
...
+    N::Integer,
+    transition;
+    kwargs...
+) where {S}
+    if spl.selector.rerun # Recompute joint in logp
+        model(spl.state.vi)
+    end
+
+    condvals = conditioned(tonamedtuple(spl.state.vi), Val{S}())
+    conddist = spl.alg.conditional(condvals)
```
mohamed82008, line 85: This seems like a cool way to define a custom transition in general if we use all of the named tuple, not just the `condvals`. This would allow using the current value of the variable we are trying to sample to decide the next value, e.g. taking a gradient step. Without the MH criteria, this would not be exact MCMC but it can be useful to alternate between optimization and sampling for example. I would just pass the entire named tuple here.

bdeonovic:
Whats the ETA of this ending up in master branch?

phipsgabler:
@bdeonovic It's pretty much finished, and I'm bringing it up to current master right now.  There were always some strange things going on with AD, which is why it hasn't been merged yet (or rather, I stopped updating, since it basically was finished, but I'm not able to debug AD issues).

When I have it merged, can try it out on the current master, and I'd be happy to hear about any feedback.

src/inference/gibbs_conditional.jl
phipsgabler: Probably this will be done as part of what comes out from the discussion in https://github.com/TuringLang/DynamicPPL.jl/issues/119 over in DPPL.  Ideally, IMO, there would be `variable_values(vi) -> (x = ([1,2],), y = (3,))` and `variable_names(vi) ~> (x = (@varname(x[1]), @varname(x[2])), y = (@varname(y),))`, or something like this.

test/inference/gibbs_conditional.jl
trappmartin: I'd say it doesn't hurt to keep it.

src/inference/gibbs_conditional.jl
phipsgabler: Right, that's an interesting variant.  I did as you suggested.

phipsgabler:
@bdeonovic OK, should be good now.  Combining a conditional with HMC still fails (not so dramatically anymore, but the results are wrong).

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,144 @@
...
+        (α_0, θ_0) = (2.0, inv(3.0)),
+        λ_true = rand(Gamma(α_0, θ_0)),
+        σ_true = √(1 / λ_true),
+        m_true = rand(Normal(0, σ_true)),
+        x = rand(Normal(m_true, σ_true), N)
+        
+        @model inverse_gdemo(x) = begin
+            λ ~ Gamma(α_0, θ_0)
+            m ~ Normal(0, sqrt(1 / λ))
+            x .~ Normal(m, sqrt(1 / λ))
```
devmotion, line 18: You have to use, e.g.,
```suggestion
            @. x ~ $(Normal(m, sqrt(1 / λ)))
```
or
```suggestion
            (.~)(x, Normal(m, sqrt(1 / λ)))
```
since `.~` can't be parsed by Julia 1.0 (see, e.g., https://github.com/TuringLang/DynamicPPL.jl/blob/06e7fe0cd08a22e728306a9444cf45d0e3ede690/test/compiler.jl#L440-L447).

phipsgabler:
Back from zombie state :zombie: 

There's still the same error: the sampling of mu, in the GMM test case, fails with HMC using forwarddiff and reversediff.  No idea why.

Additionally, there needs to be a solution to make the test cases stable under permutation of cluster labels.  I'm not creative enough for that today... How is this usually done?

trappmartin:
Does the same happen when you use the gdemo model and sample one parameter with HMC and the other through Gibbs sampling?

phipsgabler:
@trappmartin Sometimes, but not consistently. And it also produces a huge pile of warnings:

```
┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfinite.((θ, r, ℓπ, ℓκ)) = (true, false, false, false)
└ @ AdvancedHMC ~/.julia/packages/AdvancedHMC/P9wqk/src/hamiltonian.jl:47
```

trappmartin:
If the tests run through, this can be merged.

trappmartin:
Looks like the test run through smoothly except for the multi threaded test, which terminates early due to time limit hit.

cc: @phipsgabler

phipsgabler:
:+1: Seems like this can be merged.

@yebai apparently the logp-recalculation was not necessary, but my test cases were in fact broken.

Shall we do one more test with fancier indexing, or are [the current ones](https://github.com/phipsgabler/Turing.jl/blob/phg/new_gibbs_conditionals/test/inference/gibbs_conditional.jl) enough?

devmotion:
Can you fix the Travis timeout?

The question is also if we want to merge this PR or https://github.com/TuringLang/Turing.jl/pull/1428 (updates Turing to AbstractMCMC 2 which has a different interface and hence requires different implementations of GibbsConditional) first. I have a deadline tomorrow but will try to finish the other PR (and the corresponding one to DynamicPPL) on Wednesday this week.

test/inference/gibbs.jl
```diff
@@ -114,7 +114,7 @@ include(dir*"/test/test_utils/AllUtils.jl")
         alg = Gibbs(MH(:s), HMC(0.2, 4, :m))
         sample(model, alg, 100; callback = callback)
     end
-
+
```
devmotion, line 117:
```suggestion

```

test/inference/gibbs.jl
```diff
@@ -138,5 +138,5 @@ include(dir*"/test/test_utils/AllUtils.jl")
         model = imm(randn(100), 1.0);
         sample(model, Gibbs(MH(10, :z), HMC(0.01, 4, :m)), 100);
         sample(model, Gibbs(PG(10, :z), HMC(0.01, 4, :m)), 100);
-    end
+    end
```
devmotion, line 141:
```suggestion
    end
```

cpfiffer:
> The question is also if we want to merge this PR or #1428 (updates Turing to AbstractMCMC 2 which has a different interface and hence requires different implementations of GibbsConditional) first. I have a deadline tomorrow but will try to finish the other PR (and the corresponding one to DynamicPPL) on Wednesday this week.

I would say this one first since it's "smaller" in that it touches less of the code base.

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+include(dir*"/test/test_utils/AllUtils.jl")
+
+
+@turing_testset "gibbs conditionals" begin
+    let N = 1000,
+        (α_0, θ_0) = (2.0, inv(3.0)),
+        λ_true = rand(Gamma(α_0, θ_0)),
+        σ_true = sqrt(1 / λ_true),
+        m_true = rand(Normal(0, σ_true)),
+        x = rand(Normal(m_true, σ_true), N)
```
devmotion, line 14: I'm curious, why a `let` block and not just regular assignments? Is it due to the other example below which uses variables of the same name? In this case two separate test sets might be an alternative.

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        
+        @model function inverse_gdemo(x)
+            λ ~ Gamma(α_0, θ_0)
+            m ~ Normal(0, sqrt(1 / λ))
+            @. x ~ $(Normal(m, sqrt(1 / λ)))
```
devmotion, line 19: A bit simpler would be
```suggestion
            σ2 ~ InverseGamma(α_0, θ_0)
            σ = sqrt(σ2)
            m ~ Normal(0, σ)
            @. x ~ $(Normal(m, σ))
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            x_bar = mean(x) # sample mean
+            s2 = var(x; mean=x_bar, corrected=false) # sample variance
+            return n, x_bar, s2
+        end
+
+        function cond_m(c)
+            n, x_bar, s2 = gdemo_statistics(x)
+            m_n = N * x_bar / (n + 1)
+            λ_n = c.λ * (n + 1)
+            σ_n = sqrt(1 / λ_n)
```
devmotion, line 34: Wouldn't it be easier to precompute `m_n` and a scaled variance outside of the function instead of closing over `x` and `N`? It is not clear if the compiler eliminates the redundant computations (it's also not possible for the square root).

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            return Normal(m_n, σ_n)
+        end
+
+        function cond_λ(c)
+            n, x_bar, s2 = gdemo_statistics(x)
```
devmotion, line 39: Same here, it seems one could precompute all required terms?

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            chain = sample(inverse_gdemo(x), alg, 10_000)
+            check_numerical(chain, [:m, :λ], [m_true, λ_true], atol=0.2)
+        end
+    end
+
+    Random.seed!(100)
+    let π = [0.5, 0.5],
+        K = length(π),
+        m = 0.5,
+        λ = 2.0,
```
devmotion, line 63: Again I'm curious why you use a `let` block here. It seems also a bit inconsistent/confusing to use `λ` for the precision above and for the standard deviation here.

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        σ = 0.1,
+        N = 20,
+        μ_true = [rand(Normal(m, λ)) for _ in π],
```
devmotion, line 66:
```suggestion
        μ_true = rand(Normal(m, λ), length(π)),
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        z_true = [rand(Categorical(π)) for _ in 1:N],
```
devmotion, line 67:
```suggestion
        z_true = rand(Categorical(π), N),
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        x = [rand(Normal(μ_true[z_n], σ)) for z_n in z_true]
```
devmotion, line 68:
```suggestion
        x = rand(MvNormal(μ_true[z_true], σ))
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        
+        @model function mixture(x)
+            μ ~ arraydist(Normal.(fill(m, K), fill(λ, K)))
```
devmotion, line 71:
```suggestion
            μ ~ MvNormal(Fill(m, K), λ)
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            z ~ arraydist(Categorical.(fill(π, N)))
```
devmotion, line 72:
```suggestion
            z ~ filldist(Categorical(π), N)
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            x ~ arraydist(Normal.(μ[z], σ))
```
devmotion, line 73:
```suggestion
            x ~ MvNormal(μ[z], σ)
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        function cond_μ(c)
+            z = c.z
+            n = [count(z .== k) for k = 1:K]
+
+            # If there were no observations assigned to center `k`, `n[k] == 0`, and
+            # we use the prior instead.
+            λ_hat = [(n[k] != 0) ? inv(n[k] / σ^2 + 1/λ^2) : λ for k = 1:K]
+            μ_hat = [(n[k] != 0) ? (sum(x[z .== k]) / σ^2) * λ_hat[k] : m for k = 1:K]
+
+            return arraydist(Normal.(μ_hat, λ_hat))
```
devmotion, line 95:
```suggestion
            return MvNormal(μ_hat, λ_hat)
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+        # Both variables sampled using the Gibbs conditional
+        Random.seed!(100)
+        for alg in (Gibbs(GibbsConditional(:z, cond_z), GibbsConditional(:μ, cond_μ)),
+                    Gibbs(GibbsConditional(:z, cond_z), MH(:μ)),
+                    Gibbs(GibbsConditional(:z, cond_z), HMC(0.01, 7, :μ)), )
+
+            chain = sample(mixture(x), alg, 10000)
+            
+            μ_hat = estimate(chain, :μ)
+            lμ_hat, uμ_hat = minimum(μ_hat), maximum(μ_hat)
```
devmotion, line 113:
```suggestion
            lμ_hat, uμ_hat = extrema(μ_hat)
```

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,122 @@
...
+            lμ_true, uμ_true = minimum(μ_true), maximum(μ_true)
```
devmotion, line 114: This could be precomputed outside of the loop it seems?
```suggestion
            lμ_true, uμ_true = extrema(μ_true)
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,135 @@
+"""
+    GibbsConditional(sym, conditional)
+
+A "pseudo-sampler" to manually provide analytical Gibbs conditionals to `Gibbs`.
+`GibbsConditional(:x, cond)` will sample the variable `x` according to the conditional `cond`, which
+must therefore be a function from a `NamedTuple` of the conditioned variables to a `Distribution`.
+
+
+The `NamedTuple` that is passed in contains all random variables from the model in an unspecified
+order, taken from the [`VarInfo`](@ref) object over which the model is run.  Scalars and vectors are
```
devmotion, line 10:
```suggestion
order, taken from the [`VarInfo`](@ref) object over which the model is run. Scalars and vectors are
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,135 @@
...
+stored as their respective types, but all as floats.  The tuple also contains the value of the
```
devmotion, line 11: Does the implementation require the limitation to floats?
```suggestion
stored as their respective types, but all as floats. The tuple also contains the value of the
```

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,135 @@
...
+t = (m = ([0.234, -1.23], ["m[1]", "m[2]"]), λ = ([1.233], ["λ"])
+```
+
+and this function implements the cleanup of indexing. `conditioned(t)` will therefore return
+
+```julia
+(λ = 1.233, m = [0.234, -1.23])
+```
+"""
+@generated function conditioned(θ::NamedTuple{names}) where {names}
```
devmotion, line 119: It seems this function (or maybe even rather a function that returns the same thing for a given VarInfo object directly without the intermediate `tonamedtuple`) could/should be added to DynamicPPL?

src/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,135 @@
...
+
+# Examples
+
+```julia
+α_0 = 2.0
+θ_0 = inv(3.0)
+
+x = [1.5, 2.0]
+
+function gdemo_statistics(x)
```
devmotion, line 24: I guess it would be reasonable to update the example as well?

test/inference/gibbs_conditional.jl
```diff
@@ -0,0 +1,121 @@
...
+        σ_true = sqrt(1 / λ_true)
+        m_true = rand(Normal(0, σ_true))
+        x = rand(Normal(m_true, σ_true), N)
+
+        # The conditionals and posterior can be formulated in terms of the following statistics:
+        x_bar = mean(x) # sample mean
+        s2 = var(x; mean=x_bar, corrected=false) # sample variance
+        m_n = N * x_bar / (N + 1)
+        
+        @model function inverse_gdemo(x)
```
devmotion, line 24: Is there a reason for not using the `gdemo` model directly (https://github.com/TuringLang/Turing.jl/blob/99ba1590dfbb121a38db55c8d9b5da8f2d4e22d4/test/test_utils/models.jl#L10-L16 - the main difference seems to be the number of observations?) or using an InverseGamma prior for the variance?

phipsgabler: Yes: because I found it easier to express (or primarily, derive) the conditional in the precision parametrization.  But if someone cares to write conditionals for `gdemo`, I don't mind.

src/inference/gibbs_conditional.jl
phipsgabler: Just to mention that: the convention of separating sencences with double spaces didn't come naturally to me (it's an American idea, I have heard) -- but it makes navigation in Emacs easier, so I kind of do it automatically now :D

src/inference/gibbs_conditional.jl
phipsgabler: I believed it's a limitation of `VarInfo`, at least the untyped one -- is that wrong?

src/inference/gibbs_conditional.jl
cpfiffer: Lol no, so many Americans hate this too. There was kind of a big culture war about this recently. My preference is for single space between sentences, and I'd rather not choose spacing rules on the margin of "what makes emacs people happy".

src/inference/gibbs_conditional.jl
devmotion: The untyped `VarInfo` saves everything in a `Vector{Real}`: https://github.com/TuringLang/DynamicPPL.jl/blob/9431df05af76460b0d5fea9572fcaaf946b4791d/src/varinfo.jl#L160

phipsgabler: OK, it probably is type-preserving with TypeVarInfo, but maybe not with the untyped one.  I removed the wording, it should rarely be a problem I guess.

phipsgabler:
Should be fine now; there's still some issue with variable subsumption, but @trappmartin agreed that I should merge already.  I'll make a separate issue.

test/inference/gibbs_conditional.jl
devmotion: Aren't they basically the same, just using and return the inverse?

devmotion:
You forgot to bump the version number - can you fix this in a new PR? :slightly_smiling_face:

PR #1277, opened 2020-05-13
Allow proposal matrix for MH

cpfiffer:
Allows you to specify a proposal matrix for MH and just use that instead of all the NamedTuple weirdness that happens by default:

```julia
using Turing

@model function gdemo()
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))

    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))

    return m, s
end

model = gdemo()

proposal_matrix = [1.5 -0.15; -0.15 1.5]

chain = sample(model, MH(proposal_matrix), 100000)
```

It's a little faster in some cases than the default NamedTuple (though not by much, the NamedTuple stuff is really fast), depending on your choice of proposal matrix. 

Tests are included, but for the most part this is a really minor change.

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -14,8 +14,11 @@ function MH(space...)
 
     for s in space
         if s isa Symbol
+            # If it's just a symbol, proceed as normal.
             push!(syms, s)
         elseif s isa Pair || s isa Tuple
+            # Check to see whether it's a pair that specifies a kernel
+            # or a specific proposal distribution.
             push!(prop_syms, s[1])
 
             if s[2] isa AMH.Proposal
@@ -25,6 +28,18 @@ function MH(space...)
             elseif s[2] isa Function
                 push!(props, AMH.StaticProposal(s[2]))
             end
+        elseif length(space) == 1
+            # If we hit this block, check to see if it's 
+            # a run-of-the-mill proposal or covariance
+            # matrix.
+            prop = if s isa AdvancedMH.Proposal
+                s
+            elseif typeof(s) <: Array{<:Real, 2}
+                AdvancedMH.RandomWalkProposal(MvNormal(s))
+            end
+
+            # Return early, we got a covariance matrix. 
+            return MH{(), typeof(prop)}(prop)
         end
     end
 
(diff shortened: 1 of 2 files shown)


src/inference/mh.jl
```diff
@@ -25,6 +28,18 @@ function MH(space...)
...
             end
+        elseif length(space) == 1
+            # If we hit this block, check to see if it's 
+            # a run-of-the-mill proposal or covariance
+            # matrix.
+            prop = if s isa AdvancedMH.Proposal
+                s
+            elseif typeof(s) <: Array{<:Real, 2}
+                AdvancedMH.RandomWalkProposal(MvNormal(s))
+            end
```
devmotion, line 39: Maybe it would be nicer and easier to extend if one defines
```julia
proposal(p) = p
proposal(cov::AbstractMatrix) = AdvancedMH.RandomWalkProposal(MvNormal(cov))
```
and then just uses
```suggestion
        elseif length(space) == 1
            # Extract or generate proposal.
            prop = proposal(s)
```

cpfiffer: Okay, I added this. Good comment.

src/inference/mh.jl
```diff
@@ -187,34 +201,77 @@ function AbstractMCMC.sample_init!(
 
     # Get `init_theta`
     initialize_parameters!(spl; verbose=verbose, kwargs...)
+
+    # If we're doing random walk with a covariance matrix,
+    # just link everything before sampling.
+    if spl.alg.proposals isa AdvancedMH.RandomWalkProposal{<:MvNormal}
+        link!(spl.state.vi, spl)
+    end
```
devmotion, line 209: Maybe also write this more generically as
```julia
maybe_link!(spl.state.vi, spl, spl.alg.proposals)
```
and define
```julia
maybe_link!(varinfo, sampler, proposal) = nothing
maybe_link!(varinfo, sampler, proposal::AdvancedMH.RandomWalkProposal) = link!(varinfo, sampler)
```
?

src/inference/mh.jl
```diff
@@ -187,34 +201,77 @@ function AbstractMCMC.sample_init!(
...
-function AbstractMCMC.step!(
+function AbstractMCMC.sample_end!(
     rng::AbstractRNG,
     model::Model,
     spl::Sampler{<:MH},
     N::Integer,
-    transition;
+    transitions;
+    verbose::Bool=true,
+    resume_from=nothing,
```
devmotion, line 219: These are not needed, I think?
```suggestion
```

src/inference/mh.jl
```diff
@@ -187,34 +201,77 @@ function AbstractMCMC.sample_init!(
...
     kwargs...
 )
+    # We are doing a random walk, so we unlink everything when we're done.
+    if spl.alg.proposals isa AdvancedMH.RandomWalkProposal{<:MvNormal}
+        invlink!(spl.state.vi, spl)
+    end
```
devmotion, line 225: Same here, maybe use something like `maybe_unlink!`?

src/inference/mh.jl
```diff
@@ -187,34 +201,77 @@ function AbstractMCMC.sample_init!(
...
-    densitymodel = AMH.DensityModel(MHLogDensityFunction(model, spl))
-    trans = AbstractMCMC.step!(rng, densitymodel, mh_sampler, 1, prev_trans)
-
-    # Update the values in the VarInfo.
-    set_namedtuple!(spl.state.vi, trans.params)
-    setlogp!(spl.state.vi, trans.lp)
+    # Cases:
+    # 1. A covariance proposal matrix
+    # 2. A bunch of NamedTuples that specify the proposal space
+    if spl.alg.proposals isa AdvancedMH.RandomWalkProposal{<:MvNormal}
```
devmotion, line 243: Also here I think it would be natural to call into a method that allows to dispatch on the type of proposal instead of hardcoding the check in the general implementation of `step!`.

src/inference/mh.jl
```diff
@@ -187,34 +201,77 @@ function AbstractMCMC.sample_init!(
...
+        # If this is the case, we can just draw directly from the proposal
+        # matrix.
+        vals = spl.state.vi[spl]
+
+        # Create a sampler and the previous transition.
+        mh_sampler = AMH.MetropolisHastings(spl.alg.proposals)
+        prev_trans = AMH.Transition(vals, getlogp(spl.state.vi))
+
+        # Make a new transition.
+        densitymodel = AMH.DensityModel(gen_logπ(spl.state.vi, spl, model))
```
devmotion, line 253: What's the reason for not using `MHLogDensityFunction` as in the branch below?

src/inference/mh.jl
cpfiffer: I think that's more complex than it needs to be -- I'd rather stick closely to convention (used elsewhere) so it's immediately visible when something is being linked.

src/inference/mh.jl
cpfiffer: `MHLogDensityFunction` is a specialized version of a density function that uses `setnamedtuple` in place of `vi[spl]`, and I don't need any of that functionality here.

src/inference/mh.jl
devmotion: I see that point, it's just that IMO hardcoding type checks in the general definitions of `sample_init!`, `step!` etc makes it more difficult to extend them and to achieve similar behaviour for different proposals. Maybe one could choose a name that's less focused on transforming variables and more something like a proposal-specific `sample_init!`, `step!`, and `sample_end!`, then it would be easy to define optimized behaviour depending on the type of proposal.

cpfiffer: Okay, I added `maybe_link!` and `maybe_invlink!`.

src/inference/mh.jl
cpfiffer: Okay, I added a `propose!` function that does this.

PR #1278, opened 2020-05-14
Updated section on custom distributions

torfjelde:
Updated the section on custom distributions to include defining `bijector` for a distribution.

### DIFF
--- a/docs/src/using-turing/advanced.md
+++ b/docs/src/using-turing/advanced.md
@@ -51,7 +51,7 @@ To transform from `[0, 1]` to `ℝ` we can use the `Logit` bijector:
 Bijectors.bijector(d::CustomUniform) = Logit(0., 1.)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/using-turing/advanced.md
```diff
@@ -46,17 +43,24 @@ In most cases, it may be required to define helper functions, such as the `minim
...
-
+To transform from `[0, 1]` to `ℝ` we can use the `Logit` bijector:
 
 ```julia
-Distributions.minimum(d::Flat) = -Inf
-Distributions.maximum(d::Flat) = +Inf
+Bijectors.bijector(d::CustomUniform) = Logit(0., 1.)
 ```
 
+You'd do the exact same thing for `MultivariateDistribution`, `MatrixDistribution`, etc. For example, `Wishart` defines a distribution over positive-definite matrices and so `bijector` returns a `PDBijector` when called with a `Wishart` distribution as an argument.
```
mohamed82008, line 54:
```suggestion
You'd do the exact same thing for `ContinuousMultivariateDistribution` and `ContinuousMatrixDistribution`. For example, `Wishart` defines a distribution over positive-definite matrices and so `bijector` returns a `PDBijector` when called with a `Wishart` distribution as an argument. For discrete distributions, there is no need to define a bijector; the `Identity` bijector is used by default.
```

Issue #1273, opened 2020-05-09
Export `acclogp!` and `getlogp`?


devmotion:
Just noticed in https://github.com/SciML/DiffEqBayes.jl/pull/158 that we don't export `acclogp!` and `getlogp`. Maybe we should do that, now that `@logpdf` was removed?

mohamed82008:
Sounds good to me.

PR #1284, opened 2020-05-17
Fix MAP code

karajan9:
added DynamicPPL for `getlogp`


devmotion:
Thanks for the PR! I'm not sure if we should expose DynamicPPL though. Personally I would prefer to either fix https://github.com/TuringLang/Turing.jl/issues/1273 or use `Turing.getlogp` instead of `getlogp`.

karajan9:
Sure, it's much better this way. I feel like I'm going crazy, I wasn't able to find that method in Turing before.

yebai:
thanks @karajan9!

PR #1286, opened 2020-05-20
Adds `predict` method to make it easier to work with the predictive posterior

torfjelde:
With this method you can now do:
```julia
julia> using Turing

julia> @model function linear_reg(x, y, σ = 0.1)
           β ~ Normal(0, 1)

           for i ∈ eachindex(y)
               y[i] ~ Normal(β * x[i], σ)
           end
       end
DynamicPPL.ModelGen{var"###generator#384",(:x, :y, :σ),(:σ,),Tuple{Float64}}(##generator#384, (σ = 0.1,))

julia> f(x) = 2 * x + 0.1 * randn()

julia> Δ = 0.1
0.1

julia> xs_train = 0:Δ:10; ys_train = f.(xs_train);

julia> xs_test = [10 + Δ, 10 + 2 * Δ]; ys_test = f.(xs_test);

julia> # Infer
       m_lin_reg = linear_reg(xs_train, ys_train);

julia> chain_lin_reg = sample(m_lin_reg, NUTS(100, 0.65), 200);
└   ϵ = 0.003125

julia> m_lin_reg_test = linear_reg(xs_test, Vector{Union{Missing, Float64}}(undef, length(ys_test)));

julia> predictions = Turing.Inference.predict(m_lin_reg_test, chain_lin_reg)
Object of type Chains, with data of type 100×2×1 Array{Float64,3}

Iterations        = 1:100
Thinning interval = 1
Chains            = 1
Samples per chain = 100
parameters        = y[1], y[2]

Summary Statistics
  parameters     mean     std  naive_se     mcse       ess   r_hat
  ──────────  ───────  ──────  ────────  ───────  ────────  ──────
        y[1]  20.1978  0.0921    0.0092  missing  224.2130  0.9904
        y[2]  20.3739  0.1056    0.0106  missing  273.7488  0.9919

Quantiles
  parameters     2.5%    25.0%    50.0%    75.0%    97.5%
  ──────────  ───────  ───────  ───────  ───────  ───────
        y[1]  20.0378  20.1254  20.1863  20.2668  20.3704
        y[2]  20.1650  20.3117  20.3800  20.4430  20.5696


julia> ys_pred = collect(vec(mean(predictions[:y].value; dims = 1)))
2-element Array{Float64,1}:
 20.19778013967426
 20.373873670853182

julia> sum(abs2, ys_test - ys_pred)
0.0016295979352728104
```

**One question:** should this method be exported? I'm currently overloading `StatsBase.predict` because Turing will often used together with StatsBase and so I figured overloading rather than defining our own is preferable to avoid warnings. This is also up for discussion; I'm fine either way.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -30,6 +30,7 @@ import DynamicPPL: get_matching_type,
 import EllipticalSliceSampling
 import Random
 import MCMCChains
+import StatsBase
 
 export  InferenceAlgorithm,
         Hamiltonian,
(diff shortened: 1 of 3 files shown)


devmotion:
>  I'm currently overloading StatsBase.predict because Turing will often used together with StatsBase and so I figured overloading rather than defining our own is preferable to avoid warnings.

IMO we should definitely overload `StatsBase.predict` and not define our own method, but that's orthogonal to the question if we should (re-)export it. Maybe it's more convenient for users if we reexport it, but I don't have a strong opinion on that question.

src/inference/Inference.jl
```diff
@@ -593,4 +594,105 @@ end
...
+        y[2]  0.1948  0.0950    0.0095  missing  131.6113  0.9969
+
+Quantiles
+  parameters     2.5%    25.0%   50.0%   75.0%   97.5%
+  ──────────  ───────  ───────  ──────  ──────  ──────
+        y[1]  -0.1581  -0.0567  0.0114  0.0550  0.1758
+        y[2]   0.0200   0.1198  0.2105  0.2667  0.3773
+
+```
+"""
```
devmotion, line 656: That's a nice and detailed documentation but could you reformat the docstring according to the [official guidelines](https://docs.julialang.org/en/latest/manual/documentation/)? E.g., use a single one-line sentence for describing the function in the imperative form and use a `# Examples` section with code put into a `julia-repl` or `jldoctest` code block? You could also use DocStringExtensions if you want to automate the display of the signature etc.

src/inference/Inference.jl
```diff
@@ -593,4 +594,105 @@ end
...
+            end
+        end
+        # Execute `model` on the parameters set in `vi` and sample thouse with `"del"` flag using `spl`
+        model(vi, spl)
+
+        # Convert `VarInfo` into `NamedTuple` and save
+        theta = DynamicPPL.tonamedtuple(vi)
+        lp = Turing.getlogp(vi)
+        transitions[i] = Transition(theta, lp)
+    end
```
devmotion, line 686: Can we separate this logic into a separate function so that we can just map over the rows of the chain? Then one would not have to define the array `transitions` with abstract element type a priori.

src/inference/Inference.jl
```diff
@@ -593,4 +594,105 @@ end
...
+
+Quantiles
+  parameters     2.5%    25.0%   50.0%   75.0%   97.5%
+  ──────────  ───────  ───────  ──────  ──────  ──────
+        y[1]  -0.1581  -0.0567  0.0114  0.0550  0.1758
+        y[2]   0.0200   0.1198  0.2105  0.2667  0.3773
+
+```
+"""
+function StatsBase.predict(model::Turing.Model, chain::MCMCChains.Chains; include_all = false)
```
devmotion, line 657: Can we implement this as
```suggestion
function StatsBase.predict(model::Model, chain::MCMCChains.Chains; kwargs...)
    return StatsBase.predict(Random.GLOBAL_RNG, chain; kwargs...)
end

function StatsBase.predict(
    rng::Random.AbstractRNG,
    model::Model,
    chain::MCMCChains.Chains;
    include_all = false
)
```
instead, to prepare for the upcoming DynamicPPL release that supports RNGs everywhere?

src/inference/Inference.jl
torfjelde: Yeah of course; thanks for the comment! I didn't do `julia-repl` because the example takes up so much space already, but now I agree with you (+ the example in this PR above is a better one).

src/inference/Inference.jl
torfjelde: But are we guaranteed to release the DynamicPPL release before we put this into a release? Though I agree, it seems like we should add this once we're releasing the new DynamicPPL. I would suggest getting this out before any breaking changes (unless there are already some in #master)

devmotion: The additional argument doesn't hurt, I guess, we just don't use it right now (apart from forwarding it to `bundle_samples` I guess). It would be good to keep in mind that we should update `predict` if we upgrade Turing to the upcoming DynamicPPL release (which should be out soon).

src/inference/Inference.jl
torfjelde: Have now attempted to do that:)

src/inference/Inference.jl
torfjelde: Yeah, it's more that if you see a method taking in `rng` you'd expect it to respect it. So I'd prefer to leave it until DynamicPPL release PR (this way we can also make a release of Turing with these smaller new features before we introduce breaking changes).

Issue #1260, opened 2020-05-04
Taking multi-threaded random number generation a bit more seriously


mohamed82008:
Currently, a thread-local random number generator is used in certain places even when a random number generator is passed by the user to `sample`. This should be fixed by passing the input `rng` through to any function that might call `rand`.

devmotion:
As on Slack, I want to note that this problem is not specific to multithreading but exists for singlethreading as well. The AbstractMCMC interface correctly provides Turing (and any other downstream package) with (deep copies of) reproducible random number generators. However, since not all sampling steps and model evaluations use the provided RNG any sampling with non-standard RNGs in singlethreading and any RNG in multithreading will be non reproducible.

cpfiffer:
I think this is a problem endemic almost everywhere. I'm guilty of this almost everywhere in AdvancedMH, too.

devmotion:
Not fixed in Turing yet.

xukai92:
> The AbstractMCMC interface correctly provides Turing (and any other downstream package) with (deep copies of) reproducible random number generators. However, since not all sampling steps and model evaluations use the provided RNG any sampling with non-standard RNGs in singlethreading and any RNG in multithreading will be non reproducible.

I guess for HMC type alg. the only randomness on Turing side is initialisation. As long as RNGs are passed to AHMC correctly, it should be reproducibile - there are reproducibility test on the AHMC side.

For SMC type alg., we need seperate reproducibility test on model evaluation in DynamicPPL, and reproducibility test for e.g. AdvancedSMC module.

We can then look what's missing if things still can't be reproduced.

devmotion:
The fixes for Turing are in https://github.com/devmotion/Turing.jl/tree/rng (was linked in the DynamicPPL PR), but they require a new DynamicPPL release. I added tests for HMC, MH, IS, PG, and ESS that all pass.

PR #1288, opened 2020-05-22
Use non-global RNGs

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1260.


test/inference/Inference.jl
```diff
@@ -12,13 +12,25 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+
+                samplers = (HMC(0.1, 7),
+                            PG(10),
+                            IS(),
+                            MH(),
+                            Gibbs(PG(3, :s), HMC(0.4, 8, :m)),
+                            Gibbs(HMC(0.1, 5, :s), ESS(:m)))
+                for sampler in samplers
+                    Random.seed!(5)
+                    chain1 = sample(model, sampler, MCMCThreads(), 1000, 4)
```
yebai, line 26: How threads rng are initialised? Are they sharing the same random seed?

devmotion: That is handled in AbstractMCMC: https://github.com/TuringLang/AbstractMCMC.jl/blob/509b5cf3fe24c6a8809ba6ca98f42ae8165e12e2/src/sample.jl#L215-L216
We sample separate seeds for every chain, using the provided RNG, and create separate copies of the RNG for every thread. When sampling the `i`th chain, the RNG copy of the active thread is seeded with the `i`th seed. Hence the results can be reproduced even with different numbers of threads. The same happens when sampling with multiple processes.

yebai: thanks for the clarification!

yebai:
Ready to merge once the tests pass.

torfjelde:
Should we add `predict(rng::AbstractRNG, ...)` as we spoke about @devmotion in this PR or should I make another one after this has gotten merged?:)

devmotion:
The PR is based on Turing master before `predict` and `MLE` and `MAP` were added, so if we want to fix `predict` we would have to merge master first (or rebase). I don't mind if it's addressed in this PR or in a separate one, but I guess we should also make sure that VI uses user-provided RNGs.

cpfiffer:
It'll be cleaner if done separately.

Issue #1287, opened 2020-05-22
Question: Mistakes in Custom Distribution documentation? 


marty0801:
Question on: https://github.com/TuringLang/Turing.jl/edit/master/docs/_docs/using-turing/advanced.md

I'm concerned about mistakes in the CustomUniform example. Should `Flat` be replaced by `CustomUniform` in the below? 
```
Distributions.rand(rng::AbstractRNG, d::Flat) = rand(rng) # sample in [0, 1]
Distributions.logpdf(d::Flat, x::Real) = zero(x)          # p(x) = 1 → logp(x) = 0
```
 Also, it is unclear how/whether the "rand", and "logpdf" functions mentioned under "3. Define Helper Functions" is related to the functions defined just before.

torfjelde:
> I'm concerned about mistakes in the CustomUniform example. Should Flat be replaced by CustomUniform in the below?

Yes it should:)

> Also, it is unclear how/whether the "rand", and "logpdf" functions mentioned under "3. Define Helper Functions" is related to the functions defined just before.

Yeah that's a typo; `rand` and `logpdf` shouldn't be mentioned in section 3. Also, should probably make it clear that Turing.jl will perform the transformation internally when necessary using the defined `bijector`?

marty0801:
Thanks!

PR #1289, opened 2020-05-22
Updated doc on custom distribution

torfjelde:
See #1287

### DIFF
--- a/docs/src/using-turing/advanced.md
+++ b/docs/src/using-turing/advanced.md
@@ -10,7 +10,7 @@ title: Advanced Usage
 Turing.jl supports the use of distributions from the Distributions.jl package. By extension it also supports the use of customized distributions, by defining them as subtypes of `Distribution` type of the Distributions.jl package, as well as corresponding functions.
 
 
-Below shows a workflow of how to define a customized distribution, using a flat prior as a simple example.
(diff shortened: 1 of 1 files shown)


docs/src/using-turing/advanced.md
```diff
@@ -60,8 +59,16 @@ Distributions.minimum(d::CustomUniform) = 0.
...
 
-and `Bijectors.jl` will ensure that everything just works.
+and `Bijectors.jl` will return a default `Bijector` called `TruncatedBijector` which makes use of `minimum` and `maximum` derive the correct transformation.
+
+Internally, Turing basically does the following when it needs to convert a constrained distribution to an unconstrained distribution, e.g. when sampling using `HMC`:
+```julia
+b = bijector(dist)
+transformed_dist = transformed(dist, b) # results in distribution with transformed support + correction for logpdf
+```
+and then we can call `rand(transformed_dist)` and `logpdf(transformed_dist, y)` where `y` is any real number, as usual.
```
yebai, line 69: Consider adding a sentence to clarify whether

1. `rand` returns a sample in the unconstrained real space, and
2.  `logpdf(transformed_dist, y)` returns the log density for the original distribution, but with y living in the unconstrained space. 

At the moment, these points are a bit implicit, and it might help to be a bit more explicit.

torfjelde: Done:)

Issue #1293, opened 2020-05-24
Update docs for MAP/MLE estimation


torfjelde:
Need to update https://turing.ml/dev/docs/using-turing/advanced#maximum-a-posteriori-estimation to instead use the new MAP/MLE functionality.

Maybe even make an own section for it, rather than putting it in the "Advanced" section?

PR #1296, opened 2020-05-26
Update MLE/MAP and parallel sampling docs

cpfiffer:
Fixes https://github.com/TuringLang/Turing.jl/issues/1293 by adding documentation for MLE/MAP. I also consolidated and added full examples for parallel sampling so people can see the differences between the distributed/threaded workflow.

I adjusted spacing in examples to four spaces to match our style guide, and I converted the model declarations from 

```julia
@model f() = begin
    . . . 
end
```

to the more modern

```julia
@model function f()
    . . . 
end
```


docs/src/using-turing/guide.md
```diff
@@ -231,7 +279,7 @@ Inputs to the model that have a value `missing` are treated as parameters, aka r
 
 
 ```julia
-@model gdemo(x, ::Type{T} = Float64) where {T} = begin
+@model function gdemo(x, ::Type{T} = Float64) where {T}
```
yebai, line 282: Can we add some doc on the new feature `sample(model, Prior(), N)`? I can't find it in the docs at the moment, but it might be I don't know where it is located.

cpfiffer: Good catch, I've added this.

PR #1297, opened 2020-05-26
Affine invariant ensemble sampler

cpfiffer:
An attempt to link Turing up to the AdvancedMH Emcee sampler (https://github.com/TuringLang/AdvancedMH.jl/pull/21).

Currently it is very rough, so don't bother doing a style/tidy review yet -- there's lots of stuff that needs to be cleaned before this is ready to go.

I'm currently running into some numerical issues (detailed below) that I might like some comments on.

Gibbs is not supported yet, and it appears that Emcee is really good at finding modes. Not so much the whole posterior.

Usage:

```julia
using Turing

@model gdemo(x, y) = begin
    s ~ InverseGamma(2,3)
    m ~ Normal(0, sqrt(s))

    x ~ Normal(m, sqrt(s))
    y ~ Normal(m, sqrt(s))
end

n_samples = 100
n_walkers = 1000

spl = Turing.Inference.Emcee(n_walkers, MvNormal(2, 100), 2.0)
model = gdemo(1.5, 2.0)

chain1 = sample(model, spl, n_samples)
```

Summary stats using `chain[50:100]`:

```julia

Summary Statistics
  parameters    mean     std  naive_se    mcse         ess   r_hat
  ──────────  ──────  ──────  ────────  ──────  ──────────  ──────
           m  1.1584  0.3622    0.0016  0.0016  34787.3740  0.9999
           s  1.3484  0.4785    0.0021  0.0024  24231.6361  1.0032

Quantiles
  parameters    2.5%   25.0%   50.0%   75.0%   97.5%
  ──────────  ──────  ──────  ──────  ──────  ──────
           m  0.4173  0.9504  1.1564  1.3624  1.9099
           s  0.6990  1.0622  1.2711  1.5270  2.4930
```

Compare Emcee to NUTS:

ensemble-reject

There is notable bias in `s`, but the modes have been heavily overrepresented. The mean is substantially lower than what we'd expect to see in the full posterior.

Let's also look at the transformed/untransformed results. Here's a closeup of `s`. I took the second half of iterations for all walkers and threw them on a scatter plot, with the parameter value on the x-axis and the log joint on the y-axis. This graph rejects samples that fall outside the support of `s`:

s-scatter

If I change this to sample from the transformed space (and not use support-based rejection), you get a somewhat better exploration:

s-scatter-transform

Any thoughts on how exactly to interpret these results? Particularly in the fact that the mean for the posterior is too low, but the shape appears right if you use the transformed space?


devmotion:
I played around with it (but didn't do any detailed analysis), but instead I implemented the model as a `AdvancedMH.DensityModel` (both the constrained and unconstrained model (with `log s`)) to avoid any potential issues or complications introduced by Turing. I could observe the same condensed modes without Turing. Maybe it would be good to fix potential issues in AdvancedMH first (and add some tests, I noticed they are missing) before integrating it with Turing.

I'm not familiar with the literature, so I don't know if people would rather perform the estimation in constrained or unconstrained space, and if there are any recommendations or differences to be expected. To me intuitively a stretch move makes most sense if you can actually perform these vector operations without additional constraints :man_shrugging:

devmotion:
Would also be interesting to see what one gets with emcee (the Python package) for the same model and hyperparameters.

devmotion:
I implemented both the constrained and unconstrained version of the model with emcee, both implementations seem to work fine and produce correct mean values and densities in line with Turing's NUTS sampler: https://nbviewer.jupyter.org/gist/devmotion/6e33af3f36a93448663a690177ee05e0

cpfiffer:
Strange. I've gone through my implementation and I can't find where this issue comes from. It's probably some incredibly small issue somewhere. 

I'll go through it again soon.

cpfiffer:
Thanks to @devmotion, this now works:

```julia
Summary Statistics
  parameters    mean     std  naive_se    mcse        ess   r_hat
  ──────────  ──────  ──────  ────────  ──────  ─────────  ──────
           m  1.0702  1.3908    0.0028  0.0233  1004.0161  1.8152
           s  2.3557  3.4325    0.0069  0.0562  1004.0161  1.7763
```

It is currently functional, though it is not set up to work with Gibbs yet. I still am not sure what that would look like, exactly, or how it would even fit into our current sampling system. Would draws from other samplers be updated across all walkers?

devmotion:
Related: https://github.com/dfm/emcee/issues/134

devmotion:
Also an interesting read (linked in the issue above): https://statmodeling.stat.columbia.edu/2017/03/15/ensemble-methods-doomed-fail-high-dimensions/

devmotion:
From reading through the issue linked above, I get the impression that just updating draws from other samplers across all walkers won't work since, e.g., it is unclear how these samplers should be updated (samples of which walker should be used?) and it introduces additional correlations between the walkers. Could one perform Gibbs steps from other samplers for each walker independently instead, similar to parallel sampling but with combining them in the proposal step?

yebai:
> Also an interesting read (linked in the issue above): https://statmodeling.stat.columbia.edu/2017/03/15/ensemble-methods-doomed-fail-high-dimensions/

yes, it's known that affine invariant MH methods won't scale well with high dimensions (see below for an example). But it is still an interesting algorithm for problems within a few hundred dimensions. Besides, it is possible to improve Emcee by replacing the MH adjusted stretch move with a slice sampling-based stretch move, which I think could be an interesting future direction. I mentioned this idea a while ago, maybe we can have some discussions over the summer. 


Huijser, D., Goodman, J., & Brewer, B. J. (2015). Properties of the Affine Invariant Ensemble Sampler in high dimensions. In arXiv [stat.CO]. arXiv. http://arxiv.org/abs/1509.02230

yebai:
> It is currently functional, though it is not set up to work with Gibbs yet. I still am not sure what that would look like, exactly, or how it would even fit into our current sampling system. Would draws from other samplers be updated across all walkers?

I think we can treat Emcee as a special (top-level) algorithm that doesn't fit into Gibbs for now. Users should only use Emcee for static models with continuous variables and only as a standalone algorithm.

devmotion:
> I mentioned this idea a while ago, maybe we can have some discussions over the summer.

Yes, I remember that you mentioned this a while ago but it was not completely clear to me what exactly you had in mind. I would be interested to discuss it more after the NeurIPS deadline.

cpfiffer:
Cool. I'm happy to let this sit as a non-Gibbs method for the moment. It's ready for a review.

src/inference/emcee.jl
```diff
@@ -0,0 +1,243 @@
...
+
+struct Emcee{space, E<:AMH.Ensemble} <: InferenceAlgorithm 
+    ensemble::E
+end
+
+function Emcee(n_walkers::Int, stretch_length=2.0)
+    # Note that the proposal distribution here is just a Normal(0,1)
+    # because we do not need AdvancedMH to know the proposal for
+    # ensemble sampling.
+    prop = AMH.StretchProposal(Normal(0, 1), stretch_length)
```
devmotion, line 13: I guess type restrictions in AdvancedMH don't allow to use
```suggestion
    prop = AMH.StretchProposal(nothing, stretch_length)
```
to make it clear that `Normal(0, 1)` should never be used by any method?

cpfiffer: Actually, I don't think there's any such restriction. I think your proposal should work.

src/inference/emcee.jl
```diff
@@ -0,0 +1,243 @@
...
+
+    ss = DynamicPPL.SampleFromPrior()
+    walkers = Vector{AMH.Transition}(undef, spl.alg.ensemble.n_walkers)
+
+    for (i, t) in enumerate(transition)
+        DynamicPPL.invlink!(spl.state.vi, spl)
+        spl.state.vi[spl] = identity.(t[ss])
+        DynamicPPL.link!(spl.state.vi, spl)
+        walkers[i] = AMH.Transition(spl.state.vi[spl], getlogp(spl.state.vi))
+    end
```
devmotion, line 88: Is it possible to not preallocate `walkers`? It results in a vector with abstract element type which we could avoid by using, e.g., `map`, I guess. Also I guess it would be nice if we could get rid of the use of `sample`, `invlink!`, and `link!` here. Can't we just run the model multiple times (using the default `SampleFromPrior`) and create `AMH.Transition`s with the resulting variables?

src/inference/emcee.jl
```diff
@@ -0,0 +1,243 @@
...
+    # Chain construction.
+    return MCMCChains.Chains(
+        parray,
+        string.(nms),
+        deepcopy(TURING_INTERNAL_VARS);
+        evidence=le,
+        info=info,
+        sorted=true
+    )
+end
```
devmotion, line 177: IMO it would be nice if we don't have to reimplement (most parts of) the logic for bundling samples in the end, and I guess it would be nice (and consistent with AdvancedMH and emcee) to return a Chains object with `nwalkers` numbers of chains instead of concatenating everything. The advantage would also be that the `iteration` column would not be needed.

Maybe it would be sufficient to implement `_params_to_array` for `Vector{<:Vector{<:AMH.Transition}}` (maybe one has to add a `sampler` input argument for linking and implement some other functions called in the default implementation of `bundle_samples` as well...)?

src/inference/emcee.jl
```diff
@@ -0,0 +1,243 @@
...
+
+function DynamicPPL.dot_observe(
+    spl::Sampler{<:Emcee},
+    ds::Union{Distribution, AbstractArray{<:Distribution}},
+    value::AbstractArray,
+    vi,
+)
+    return DynamicPPL.dot_observe(SampleFromPrior(), ds, value, vi)
+end
```
devmotion, line 243: I guess this might not be needed at all if we never run the `model` with `Sampler{<:EMCEE}` - I thought it should be sufficient to just always use `SampleFromPrior`? I guess one mainly has to change the construction of the `AMH.DensityModel` and pass `SampleFromPrior` instead of `spl` there.

src/inference/emcee.jl
cpfiffer: This function does that -- it parses all the walkers out into separate chains. But I suppose the `_params_to_array` fix is a little more succinct. Although, in this case, one of the big time sucks is running `transform_transitions` to convert the walkers to the constrained space. Perhaps that should be in `sample_end!`, though.

src/inference/emcee.jl
cpfiffer: Yeah, I can do that.

src/inference/Inference.jl
```diff
@@ -138,6 +139,8 @@ const TURING_INTERNAL_VARS = (internals = [
     "nom_step_size",
     "tree_depth",
     "is_adapt",
+    "walker_id",
+    "iteration"
```
devmotion, line 143: I don't think these are needed if the samples of each walker end up in a separate chain in a `Chains` object? Due to these internal parameter names I assumed you're concatenating samples in a different way, but judging from your comment below you want to handle it in the same way as in AdvancedMH.

devmotion:
```suggestion
```

src/inference/emcee.jl
devmotion: > Although, in this case, one of the big time sucks is running transform_transitions to convert the walkers to the constrained space. Perhaps that should be in sample_end!, though.

I'm not sure, maybe you could do that in `_params_to_array` but on the other you probably always want to get the untransformed samples back, no matter if you output a `Chains` or a vector of named tuples.

I was just surprised that you would need to implement `bundle_samples` from scratch, I would have assumed `_params_to_array` for vectors of vectors of `AMH.Transition`s (and maybe `get_logevidence`) would be sufficient since as soon as you have the transitions as a 3d array everything else should be the same.

src/inference/Inference.jl
cpfiffer: Yeah, this was leftover code when I was spitting everything out into one chain.

test/inference/emcee.jl
```diff
@@ -0,0 +1,20 @@
...
+    @testset "gdemo" begin
+        Random.seed!(9876)
+
+        n_samples = 1000
+        n_walkers = 250
+        
+        spl = Emcee(n_walkers, 2.0)
+        chain = sample(gdemo_default, spl, n_samples)
+
+        check_gdemo(chain[div(n_samples, 2):end])
```
devmotion, line 18: Maybe just run it for a bit longer or adjust the tolerance instead?

cpfiffer:
I settled for writing a wrapper around `bundle_samples`, because _`params_to_arry` isn't aware of the sampler or the `VarInfo` and can't do the transforms. It does however make some very minor, general changes, and then uses our internal default `bundle_samples`, which I think is more elegant that what I had before.

The walkers are also initialized a lot better from the prior.

test/inference/emcee.jl
cpfiffer: Well, the burn-in on Emcee is quite extreme, and I think it's probably good modelling practice to remove the bulk of your early samples.

src/inference/emcee.jl
```diff
@@ -63,6 +63,15 @@ function AbstractMCMC.sample_end!(
     invlink!(spl.state.vi, spl)
 end
 
+function _typed_draw(model, vi::VarInfo)
+    spl = DynamicPPL.SampleFromPrior()
+    empty!(vi)
+    model(vi, spl)
+    DynamicPPL.link!(vi, spl)
```
devmotion, line 70: I'm wondering why `link!` is not imported (I think?) even though `invlink!` is. Or maybe we should always use qualified names for internal stuff from DynamicPPL?

test/inference/emcee.jl
devmotion: I agree in practice, I was just a bit surprised about doing it in the tests since the tests in AdvancedMH are fine without discarding the warm-up phase.

src/inference/emcee.jl
cpfiffer: I'm not sure. Part of the reason all the DynamicPPL stuff is unqualified is because it was in Turing before they were separated, so I'm not sure what our go-forward style should be for this.

cpfiffer:
I'd rather prefer to just go back to my original formulation, where I overloaded `bundle_samples` and generated the 3d array there. I prefer to do this because `Emcee` is such a bizarre case that does not fit with the rest of Turing's ecosystem -- `bundle_samples` was meant to be overloaded in this exact case.

devmotion:
> bundle_samples was meant to be overloaded in this exact case.

I know I'm repeating myself, but I still think one should not implement `bundle_samples` at all here (and I don't think it's meant to be implemented either). The underlying logic is exactly the same as in the current implementation of `bundle_samples` - get an array of samplers, the transition extras (such as lp), the log evidence and combine everything in a `Chains` object, together with the sampler if the state is supposed to be saved. Hence it should be sufficient to just implement `_params_to_array`, `get_transition_extras`, and `get_logevidence`. If there are any missing arguments such as `sampler` they should just be added to the default implementation instead of completely rewriting `bundle_samples` (and in fact, `sampler` was forwarded to `_params_to_array` up until some weeks ago but I removed it since it was not used).

cpfiffer:
In general I would agree that we should be doing this for any other sampler method, when or if they needed to access `spl` from `_params_to_array`. But extending three methods to fit a single edge case instead of neatly assembling everything in one place seems over-engineered and wasteful. 

I very much prefer to keep these all separate because of how distinct `Emcee` is from everything else -- it has a unique control flow, a unique chain assembly mechanism, and it is not integrated with `Gibbs` or other methods. We shouldn't adapt the rest of the ecosystem by gluing on extra arguments to a larger subset of methods when it is perfectly sufficient to contain everything needed in one uncomplicated function.

devmotion:
I see that point, but I just thought it could be generalized and used even more generally :smile: 

I had the impression that the main logic in `bundle_samples` is so general that it could even be defined for `AbstractModel` and `AbstractSampler` and moved to `MCMCChains` since given your transitions, model, and sampler you always just
- put all samples in a 3d-array and get the names
- get the internal stuff
- extract the log evidence
- get the additional information you want to save
and then put that into a `Chains`, if the chain type is `Chains`. Since these things correspond exactly to the fields of `Chains`, I assume that this will be the procedure in 99% of the cases (and for the remaining special cases you could still just overload `bundle_samples` in a different way).

So in my impression, although Emcee is quite different from all other algorithms, when you want to construct the `Chains` object the main algorithm should be the same as for any other `AbstractSampler`.

cpfiffer:
I'm quite wary of making the interface methods "too general". Interfaces in non-Julia languages is that they are very specific to the class or object in question, but in Julia's case, this is a lot less clear a dichotomy -- you can have both general and specific interface methods, and it's a lot fuzzier from a design perspective where you should stop generalizing. `bundle_samples` is one of those that I anticipated would be overloaded far more than it currently is, partly because we added stuff like `get_transition_extras` and such. I think actually that was a very good design pattern because all of the samplers that hit the default `bundle_samples` are more or less identical in terms of structure, since they only ever return a single chain.

In the case of `Emcee`, it has divulged from the traditional control flow, and should be treated as such. Overloading `bundle_samples` seems to me like a good way to handle divergences from the generic program flow we've got everywhere else, because it's the top-most level at which those changes occur. I don't want the default `bundle_samples` to sometimes handle 2d arrays and sometimes 3d arrays because those are fundamentally different things.

I think generalizing too much obviates the benefit of the interface methods in AbstractMCMC -- we can have a general method in Turing to handle all the samplers unless we have exceptions, and it seems like this is a good use case for that kind of thing.

test/inference/emcee.jl
cpfiffer: Turns out not to matter so much. I've adjusted the test.

cpfiffer:
Tests are passing here.

cpfiffer:
Thanks for the help, @devmotion!

PR #1299, opened 2020-05-27
Allow users to provide a starting point for MLE/MAP

### LINKED ISSUE #1298: MLE/MAP with start values gives ambiguity
Using MLE/MAP with a start values (to help conversion) results in the following ambiguity error:
```
julia> optimize(m, MLE(), [178.0, 25.0])
ERROR: MethodError: optimize(::DynamicPPL.Model{var"###evaluator#316",(:heights,),Tuple{Array{Float64,1}},(),DynamicPPL.ModelGen{var"###generator#317",(:heights,),(),Tuple{}}}, ::MLE, ::Array{Float64,1}) is ambiguous. Candidates:
  optimize(model::DynamicPPL.Model, ::MLE, args...; kwargs...) in Turing at C:\Users\Jens.Adam\.julia\packages\Turing\GMBTf\src\modes\ModeEstimation.jl:258
  optimize(f, g, initial_x::AbstractArray; inplace, autodiff, kwargs...) in Optim at C:\Users\Jens.Adam\.julia\packages\Optim\L5T76\src\multivariate\optimize\interface.jl:68
Possible fix, define
  optimize(::DynamicPPL.Model, ::MLE, ::AbstractArray)
```
I hope this is the right place for this issue.

cpfiffer:
Fixes #1298 by allowing users to provide a combination of initial values and optimizers.


devmotion:
Looking at all the `Optim.optimize` functions that are needed here, maybe we should think about providing an interface for constructing the objective function instead of working with `Optim.optimize` directly (and have a separate interface for getting the stats and untransformed parameters from the optimization result). Then users could use all Optim-specific features without us having to worry about how to support them. It would also be trivial for users to wrap the objective function inside a custom function, e.g., to combine it with another objective function or to add regularization.

cpfiffer:
I think it's fine to support just a function, but that's a distinct issue, and one I think we should address later.

For the moment, it's really not very expensive for us to add these signatures. Another reason for providing the full `Optim` framework is that the internal `_optimize` function does a lot of things that would be mentally costly for and end-user to learn about, such as getting parameter names, linking, etc.

cpfiffer:
Tests are passing here.

devmotion:
A bit sad that it requires adding so many lines to the code base just to avoid method ambiguities (I guess?).

cpfiffer:
There's probably a "dispatch-foo" way to do this in far fewer lines, but my reading of the Optim signatures tilts towards them having a very complex interface structure. I think this is sufficient for a fix, anyhow.

PR #1307, opened 2020-06-04
Make GibbsComponent a simple trait (fix #1306)

### LINKED ISSUE #1306: Make Gibbs components extensible (somewhat RFC...)
We have discussed about this somewhere before... currently, anything you want to use in `Gibbs` has to be a subtype of

```julia
const GibbsComponent = Union{Hamiltonian,MH,ESS,PG,GibbsConditional}
```

so you can't just implement a compatible sampler in a separate package (which is what I would like to do these days). 

@devmotion mentioned that we should use some kind of trait, but then of course we can't have

```julia
struct Gibbs{space, A<:Tuple{Vararg{GibbsComponent}}}
```

anymore (or maybe only with some heavy trait dispatch machinery that we'd like to avoid).

I'd really like to write a quick fix for this now, but of course having a permanent improvement would be cool.

So, what about (re)introducing `gibbs_step` as [here](https://github.com/TuringLang/Turing.jl/pull/1275/commits/e9f95138fe441c32fa5863978d10f4c31fb64cab), but with the implementation

```julia
struct IsGibbs{S} end

isgibbs(::Type{S}) where {S<:Hamiltoniam} = IsGibbs{S}
# etc. for other samplers

gibbs_step!(rng::AbstractRNG, model::Model, spl::S, N, transition; kwargs...) where {S<:Sampler} = 
    gibbs_step!(isgibbs(S), rng, model, spl, N, transition; kwargs)

function gibbs_step!(
    ::IsGibbs{S},
    rng::AbstractRNG,
    model::Model,
    spl::S,
    N,
    transition;
    kwargs...) where {S<:Sampler}

    return AbstractMCMC.step!(rng, model, spl, N, transition; kwargs...)
end
```

Just a hand-written Holy trait -- I hope I got that right.  

Maybe with a macro `@allowgibbs MySampler` expanding to `DynamicPPL.isgibbs(::Type{S}) where {S<:MySampler} = IsGibbs{S}`.

phipsgabler:


src/inference/gibbs.jl
```diff
@@ -2,7 +2,16 @@
...
-const GibbsComponent = Union{Hamiltonian,MH,ESS,PG}
+
+"""
+    isgibbscomponent(spl::A)
+    isgibbscomponent(::Type{A})
+
+Marks the type `S` to be allowed withing a `Gibbs` sampler.
+"""
+isgibbscomponent(::Type{<:InferenceAlgorithm}) = false
+isgibbscomponent(alg::InferenceAlgorithm) = isgibbscomponent(typeof(alg))
```
devmotion, line 13: It is recommended to define methods only for either instances or types, and recommended to define them for instances if possible (see https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type-1). I would suggest to define only
```suggestion
isgibbscomponent(alg::InferenceAlgorithm) = alg
```
since it seems currently it is sufficient to be able to reason about instances.

src/inference/gibbs.jl
```diff
@@ -30,11 +39,16 @@ Tips:
 methods like Particle Gibbs. You can increase the effectiveness of particle sampling by including
 more particles in the particle sampler.
 """
-struct Gibbs{space, A<:Tuple{Vararg{GibbsComponent}}} <: InferenceAlgorithm
+struct Gibbs{space, A<:Tuple} <: InferenceAlgorithm
     algs::A   # component sampling algorithms
+
+    function Gibbs{space, A}(algs::A) where {space, A<:Tuple}
+        @assert all(isgibbscomponent, algs)
```
devmotion, line 46: `@assert` can be removed at various optimization levels (see https://docs.julialang.org/en/v1/base/base/#Base.@assert), so it is safer to use something like
```suggestion
        all(isgibbscomponent, algs) || error("all algorithms have to support Gibbs sampling")
```

src/inference/gibbs.jl
phipsgabler: I know that phrase in the docs well, but every single time I write a function like this I think "oh, in this case it would surely be nice to make an exception" :P

phipsgabler: Another thing I was just wondering about: restricting to `InferenceAlgorithm` makes a narrower choice than the old `GibbsComponent`.  Is is even a good idea?

phipsgabler: I'm assuming it might not be, since there was this discussion that `GibbsConditional` should not be an `InferenceAlgorithm`, for example.

devmotion: Before it was fixed to `PG`, `Hamiltonian`, `MH`, and `ESS`, so it's not actually narrower, is it? And everyone could just implement `isgibbscomponent` for her custom algorithm, so it's not even generally restricted to `InferenceAlgorithm`. Whereas before it was fixed to the predefined algorithms.

phipsgabler: It was restricted to the end user, so to say, but it wasn't if you actually extended it internally.  The `GibbsConditional` in my other PR is not an `InferenceAlgorithm`, but still was added to `GibbsComponent`.  

The reason I'd at least add a non-restricted fallback is: if you have a new component that is not `<: InferenceAlgorithm`, and don't add the trait, you get a `MethodError`, instead of the error message from the constructor.  (Just happened to me :))

devmotion: > The reason I'd at least add a non-restricted fallback is: if you have a new component that is not <: InferenceAlgorithm, and don't add the trait, you get a MethodError, instead of the error message from the constructor. (Just happened to me :))

Yep, that's a good reason, I agree :+1:

> (the only way you could implement an algorithm was by changin Turing, anyway)

Is that true? Ideally you should be able to just define a new subtype of `InferenceAlgorithm`, implement the `AbstractMCMC` interface, and get access to all predefined dispatches (for `sample` etc).

I assumed the only/main problem with such a custom implementation would be that you can't hook into the Gibbs sampler (currently) due to `GibbsComponent` being defined internally.

devmotion: > I know that phrase in the docs well, but every single time I write a function like this I think "oh, in this case it would surely be nice to make an exception" :P

BTW same here :smile: I'm always inclined to implement both but then try to stick to the recommendations.

phipsgabler: > I assumed the only/main problem with such a custom implementation would be that you can't hook into the Gibbs sampler (currently) due to `GibbsComponent` being defined internally.

Yeah, that's what I meant.  As soon as you went so far as to change `GibbsComponent`, which you had to, you could put in any type.  But you're right, now that it works by dispatch it's open anyway.

devmotion:
Maybe add an upper bound for Distributions for now until the problem with Distributions 0.23.3 is fixed (0.23.2 should work fine).

phipsgabler:
Do you mean the Laplacian `eltype` thing in that other issue?  Is it good to fix that in this same PR?

And if so: please write me the exact thing to put in the compat field, I never remember how version bounds work exactly :D

devmotion:
No, that's not what I had in mind. In Distributions 0.23.3 the field `c0` of `InverseWishart` was renamed to `logc0` (and I guess the internal implementation changed accordingly). This is why currently tests fail on Turing master, and it has to be fixed properly in DistributionsAD. Adding a bound here would just be a temporary solution to avoid the problem and make sure that tests pass.

So basically I'm just suggesting to change
https://github.com/TuringLang/Turing.jl/blob/85c772d0d79438a6189d43cf950ef90320efa5d4/Project.toml#L37
to
```julia
Distributions = "0.22, = 0.23.0, = 0.23.1, = 0.23.2"
```
(I think that's the only way to restrict it to versions [0.22, 0.23.3) according to https://julialang.github.io/Pkg.jl/v1/compatibility/, since usually patch versions shouldn't break anything - but on the other hand changes of field names are usually considered non-breaking AFAIK).

devmotion:
I'm sorry, I forgot to mention that I already fixed it in another PR yesterday evening 🙈 That's why you get the merge conflicts now.

phipsgabler:
No worries, that was an easy one.

PR #1310, opened 2020-06-04
Make Bijectors.jl lower-bound 0.7 instead of 0.6.4

torfjelde:
Thanks to @xukai92 we know that the following fails if you're using `Bijectors@0.6`:
```julia
using Turing
@model model(x, q=nothing) = begin
    if isnothing(q)
        s ~ InverseGamma(2, 3)
        m ~ Normal(0.0, sqrt(s))
    else
        θ ~ q
        s, m = θ
    end
    for i = 1:length(x)
        x[i] ~ Normal(m, sqrt(s))
    end
end
x = randn(200)
q1 = vi(model(x),     ADVI(10, 1_000))
q2 = vi(model(x, q1), ADVI(10, 1_000))
```
Because calling `bijector` on `q1` results in a `DistributionBijector` which has been removed from Bijectors.jl (https://github.com/TuringLang/Bijectors.jl/pull/104).

On `Bijectors@0.7` this correctly results in the following error:
```julia
julia> q2 = vi(model(x, q1), ADVI(10, 1_000))
ERROR: MethodError: no method matching bijector(::Bijectors.TransformedDistribution{DistributionsAD.TuringDiagMvNormal{Array{Float64,1},Array{Float64,1}},Bijectors.Stacked{Tuple{Bijectors.Exp{0},Bijectors.Identity{0}},2},Multivariate})
```
since we cannot know what `bijector(d::TransformedDistribution)` should be in general. But for the above example, we just need to add the line:
```julia
Turing.bijector(d::typeof(q1)) = inv(d.transform)
```
to make it work.

Also added re-export for Bijectors.jl.


yebai:
thanks, @torfjelde!

Issue #1317, opened 2020-06-09
`no method matching reconstruct` with MH on a vector of `Dirichlet` vectors


phipsgabler:
```julia
julia> K = 2
2

julia> @model function test()
           T = Vector{Vector}(undef, K)
           for i = 1:K
               T[i] ~ Dirichlet(K, 1.0)
           end
       end
DynamicPPL.ModelGen{var"###generator#433",(),(),Tuple{}}(##generator#433, NamedTuple())

julia> sample(test(), Gibbs(MH(:T)), 2)
Sampling 100%|███████████████████████████████████████████████████████████████| Time: 0:00:01
ERROR: MethodError: no method matching reconstruct(::Array{Dirichlet{Float64},1}, ::Array{Float64,1})
Closest candidates are:
  reconstruct(::Distribution, ::AbstractArray{T,1} where T) at /home/philipp/.julia/dev/Turing/src/inference/mh.jl:138
  reconstruct(::AbstractArray{#s168,1} where #s168<:(Distribution{Univariate,S} where S<:ValueSupport), ::AbstractArray{T,1} where T) at /home/philipp/.julia/dev/Turing/src/inference/mh.jl:143
Stacktrace:
 [1] _val_tuple(::DynamicPPL.VarInfo{NamedTuple{(:T,),Tuple{DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:T,Tuple{Tuple{Int64}}},Int64},Array{Dirichlet{Float64},1},Array{DynamicPPL.VarName{:T,Tuple{Tuple{Int64}}},1},Array{Float64,1},Array{Set{DynamicPPL.Selector},1}}}},Float64}, ::NamedTuple{(:T,),Tuple{Array{DynamicPPL.VarName{:T,Tuple{Tuple{Int64}}},1}}}) at /home/philipp/.julia/dev/Turing/src/inference/mh.jl:0
[...]

PR #1318, opened 2020-06-09
Fix MH heuristics for vectors of multivariate distributions

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1317.


src/inference/mh.jl
```diff
@@ -90,7 +90,11 @@ function set_namedtuple!(vi::VarInfo, nt::NamedTuple)
             vi[vns[1]] = vals
         elseif v_isarr && n_vals == n_vns > 1
             for (vn, val) in zip(vns, vals)
-                vi[vn] = [val]
+                if val isa AbstractArray
+                    vi[vn] = val
+                else
+                    vi[vn] = [val]
+                end
```
phipsgabler, line 97: I always write that as vi[vn] = [val;]. But it's annoying that you have to do it at all...

yebai:
thanks @devmotion!

PR #1323, opened 2020-06-11
Fix argument name

harryscholes:
Also fixed a bullet point line break

### DIFF
--- a/src/inference/hmc.jl
+++ b/src/inference/hmc.jl
@@ -61,8 +61,7 @@ HMC(0.05, 10)
 
 Tips:
 
-- If you are receiving gradient errors when using `HMC`, try reducing the
-`step_size` parameter, e.g.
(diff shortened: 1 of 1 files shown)


src/inference/hmc.jl
```diff
@@ -61,8 +61,7 @@ HMC(0.05, 10)
 
 Tips:
 
-- If you are receiving gradient errors when using `HMC`, try reducing the
-`step_size` parameter, e.g.
+- If you are receiving gradient errors when using `HMC`, try reducing the `ϵ` parameter, e.g.
```
yebai, line 64:
```suggestion
- If you are receiving gradient errors when using `HMC`, try reducing the leapfrog step-size `ϵ`, e.g.
```

yebai:
thanks, @harryscholes -- looks good, I added a minor comment.

harryscholes:
Cool :) I just removed the hyphens in 'step size' for consistency

PR #1333, opened 2020-06-21
[WIP] Implement nested sampling

SaranjeetKaur:


src/inference/ns.jl
```diff
@@ -0,0 +1,354 @@
...
+##    prior_transform(X) = quantile.(priors, X)
+##    return NestedModel(loglike, prior_transform)
+##end
+
+## revisit the Nested sampler definition here: https://github.com/TuringLang/NestedSamplers.jl/blob/master/src/staticsampler.jl
+
+## comment on each step of the code and also include docstrings for sections of the code
+## both NestedModel and Nested are to be utilized in this code
+
+struct NS{space, T, P <: AbstractProposal, B <: AbstractBoundingSpace{T}} <: InferenceAlgorithm  ## refer the comment on line 39
```
cpfiffer, line 44: No need to duplicate all these fields, just make `NS` wrap around `Nested` from [here](https://github.com/TuringLang/NestedSamplers.jl/blob/19f5273b25a5eb3a6b1278f98a1dacf5b302acb0/src/staticsampler.jl#L3).

src/inference/ns.jl
```diff
@@ -0,0 +1,354 @@
...
+bound(b::NestedSamplers.Bounds) = b
+
+NS() = NS{()}()
+NS(space::Symbol) = NS{(space,)}()
+
+isgibbscomponent(::NS) = true # this states that NS alg is allowed as a Gibbs component
+
+## Define a general `funtion NS` here ?
+
+mutable struct NSState{V<:VarInfo} <: AbstractSamplerState   ## where is this being used? in the function Sampler ?
```
cpfiffer, line 74: To answer your comment: The sampler state is where you store state information for the sampler, whatever that could mean. In this case it looks like you only need `vi`, which is provided by the default sampler state here: https://github.com/TuringLang/Turing.jl/blob/614f593424f54f458f992ebe81454c29215f28fc/src/inference/Inference.jl#L528

You can remove this `NSState` stuff and just use `SamplerState`.

src/inference/ns.jl
```diff
@@ -0,0 +1,354 @@
...
+       info = Dict{Symbol, Any}()
+    
+       # set up state struct
+       state = NSState(vi)
+    
+       # generate a sampler
+       return Sampler(alg, info, s, state)
+end
+
+function AbstractMCMC.sample_init!(
```
cpfiffer, line 98: You probably need almost none of this code -- it duplicates everything in `NestedSamplers`. You should redo this to just do something like

```julia
function AbstractMCMC.sample_init!(    
	
Attach files by dragging & dropping, selecting or pasting them.
    rng::AbstractRNG,
    model::Model,   ## check `NestedModel`?
    spl::Sampler{<:NS},    ## `Nested{T,B}` ## s == spl ## or `Sampler{<:NS{space, T, P, B}}`
    N::Integer,
    verbose::Bool=true,
    resume_from=nothing,
    debug::Bool = false;
    kwargs...
) where {T, P, B}
    # You need to make a `NestedModel` somehow
    # spl.nested is what I imagine you'd call `Nested` if you placed it in `NS` and just wrapped around everything in 
    # NestedSamplers.jl
    AbstractMCMC.sample_init(rng, nested_model, spl.nested, N)
end
```

cpfiffer: This way you forward everything to `NestedSamplers.jl` to do all the work.

src/inference/ns.jl
cpfiffer: Scratch that, you should probably put the `Nested` struct in `NSState`, since `Nested` is mutable and seems to be modified pretty heavily.

src/inference/ns.jl
cpfiffer: See comment below -- just make a `Nested` struct and stick it as a field in `NSState`. Don't duplicate all this code. Rely on `NestedSamplers.jl` code as much as possible.

src/inference/ns.jl
```diff
@@ -0,0 +1,354 @@
...
+        # get weights
+        wsum = mapreduce(t->exp(t.log_wt - spl.logz), +, transitions)
+
+        # check with h
+        err = spl.h ≠ 0 ? 3sqrt(spl.h / spl.nactive) : 1e-3
+        isapprox(wsum, 1, atol = err) || @warn "Weights sum to $wsum instead of 1; possible bug"
+    end
+
+    return vals
+end
```
cpfiffer, line 328: None of this is necessary. If you are correctly returning a `Transition`, the rest of Turing will handle this for you.

src/inference/ns.jl
```diff
@@ -0,0 +1,354 @@
...
+    spl.active_points[:, idx] = v
+    spl.active_logl[idx] = logl
+    spl.ndecl = log_wt < prev.log_wt ? spl.ndecl + 1 : 0
+    spl.ncall += ncall
+    spl.since_update += 1
+
+    # Shrink interval
+    spl.log_vol -=  1 / spl.nactive
+
+    return NestedTransition(draw, logL, log_wt)
```
cpfiffer, line 212: Each `step!`, you should be extracting whatever the values are inside `NestedTransition` and putting them inside `spl.state.vi`, and then returning `vi(spl)`.

yebai:
Closing this - @ParadaCarleton please feel free to open a new PR for the nested sampling integration.

PR #1334, opened 2020-06-21
[WIP] Approximate Bayesian Computation (ABC)

torfjelde:
This is an initial draft of implementing ABC in Turing.

I've implemented the following samplers:
1. `GetData <: InferenceAlgorithm`: a hacky way of extracting the observations by running the model once before inference. We need the observations to compute the reference statistics at each `step!` for any ABC method, and this allows us to get the observations once in the construction of the state. This also means that I'm currently assuming the model to be static, i.e. observations doesn't change between each execution of the model.
2. `AbstractABC <: InferenceAlgorithm`: allows some code-reuse between different ABC methods.
3. `ABC <: AbstractABC`: Essentially makes use of `AbstractMH.jl` to make a simple random-walk ABC method. Doesn't have any additional dependencies beyond Turing and it's dependencies.
4. `GPABC <: AbstractABC`: Attempts to make use of the `GpABC.jl` package under the hood. After looking at a couple of packages, this seemed like pretty good one that is being actively maintained. But none of these frameworks are really designed with an interface such as `AbstractMCMC.jl` in mind, though algorithmically they should fit the bill perfectly.

There are a couple of awkward hacks to make this work in Turing:
1. `GetData` is a hack.
2. Currently only works for univariate observations. I bet we have some functionality somewhere in the VI impls which can deal with flattening all sorts of things, so I didn't bother with that just now.
3. Generally the code is quick-and-dirty, so def needs some work in general.

Other relevant packages:
1. [`GpABC.jl`](https://github.com/tanhevg/GpABC.jl): contains ABC-rejection and ABC-SMC methods, and has additional emulation-approaches using GPs. Looks like it's active and maintained, but hasn't reached v1.0 yet (or at least isn't on available via `Pkg.install`).
2. [`ApproximateBayesianComputing.jl`](https://github.com/eford/ApproximateBayesianComputing.jl): looks like it's a bit more in the earlier stages of development.
3. [`ApproxBayes.jl`](https://github.com/marcjwilliams1/ApproxBayes.jl): contains ABC-rejection and ABC-SMC methods. Looks a bit more "mature" than others maybe, but their README points to `GpABC.jl` for a more feature-complete implementation.

IMHO, `GpABC.jl` seems like the thing to interface with if we want some more feature-complete impls.

# Examples
## Setup
```julia
julia> using LinearAlgebra

julia> using Turing

julia> import Turing.Inference: ABC, GPABC

julia> import AdvancedMH

julia> const AMH = AdvancedMH
AdvancedMH

julia> Turing.turnprogress(false)
[ Info: [Turing]: progress logging is disabled globally
false

julia> @model function demo(x)
           m ~ Normal(0, 0.1)

           for i ∈ eachindex(x)
               x[i] ~ Normal(m, 1)
           end
       end;

julia> xs = randn(10_000);

julia> m = demo(xs);
```
## `AdvancedMH.jl` based
```julia
julia> f(x, y) = norm(x - y) / length(x) # mean squared error

julia> proposal = (m = AMH.RandomWalkProposal(Normal(0.0, 0.1)), );

julia> alg = ABC(proposal, f; epsilon = 0.05);

julia> chain = sample(m, alg, 1000);

julia> mean(chain)
Mean
  parameters    mean
  ──────────  ──────
           m  0.0026
```

## `GpABC.jl` interop
```julia
julia> # GpABC.jl integration
       using GpABC

julia> threshold = 0.05 * length(xs);

julia> n_particles = 1;

julia> max_iter = 10 * n_particles;

julia> priors = [Normal()];

julia> summary_statistic = "keep_all";

julia> reference_data = reshape(xs, (:, 1));

julia> distance_function = GpABC.Distances.euclidean;

julia> summary_statistic = GpABC.build_summary_statistic(summary_statistic)
keep_all_summary_statistic (generic function with 1 method)

julia> alg = Turing.Inference.GPABC(summary_statistic, distance_function, threshold, n_particles, max_iter)
GPABC{typeof(GpABC.keep_all_summary_statistic),typeof(Distances.euclidean),Float64}(GpABC.keep_all_summary_statistic, Distances.euclidean, 500.0, 1, 10)

julia> samples_gpabc = sample(m, alg, 1000);

julia> mean(samples_gpabc)
Mean
  parameters     mean
  ──────────  ───────
           m  -0.0014
```


Project.toml
```diff
@@ -13,6 +13,7 @@ DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 DynamicPPL = "366bfd00-2699-11ea-058f-f148b4cae6d8"
 EllipticalSliceSampling = "cad2338a-1db2-11e9-3401-43bc07c9ede2"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+GpABC = "e850a1a4-d859-11e8-3d54-a195e6d045d3"
```
yebai, line 16: Let's make this an optional dependency via using `Requires.jl`.

src/inference/abc.jl
```diff
@@ -0,0 +1,279 @@
...
+    r = rand(rng, dist)
+    push!(vi, vn, r, dist, spl)
+    return r, 0
+end
+
+function DynamicPPL.observe(spl::Sampler{<:AbstractABC}, dist::Distribution, value, vi)
+    # 1. Sample from likelihood
+    proposed = rand(dist)
+
+    # HACK: 2. Store proposal and observation (`value`) in sampler state so that we can later compute distance
```
yebai, line 55: I can't find where observation is stored in sampler state. Am I missing anything?

src/inference/abc.jl
```diff
@@ -0,0 +1,279 @@
...
+            # Return old sample
+            return transition
+        else
+            return Transition(spl)
+        end
+    end
+end
+
+
+#############
```
yebai, line 179: Let's move this `GPABC` section  into `src/contrib/gpabc.jl` to indicate it is a third party inference support.

src/inference/abc.jl
```diff
@@ -0,0 +1,279 @@
...
+    y = zeros(n_design_points)
+    for i in 1:n_design_points
+        model_output = simulator_function(parameters[i,:])
+        y[i] = distance_metric(
+            summary_statistic(model_output),
+            reference_summary_statistic
+        )
+    end
+    return y
+end
```
yebai, line 279: Let's try to add a simple test on the `gdemo` model, or something similar.

jbrea:
This looks cool! With @francescoalemanno I started the [JuliaApproxInference org](https://github.com/JuliaApproxInference) which aims at a coherent interface for ABC and approximate point estimation. We have already some methods implemented in [LikelihoodfreeInference.jl](https://github.com/JuliaApproxInference/LikelihoodfreeInference.jl) and [KissABC.jl](https://github.com/francescoalemanno/KissABC.jl) but we intend to refactor these packages such that we can have a common API. Some initial ideas are written up [here](https://github.com/JuliaApproxInference/ApproxInferenceBase.jl/issues/1). Would be great, if you would chime in such that we arrive at something that works smoothly with Turing.jl!

torfjelde:
@yebai Thanks for the comments! But when I made this it was only as a demonstration of how we could approach this in Turing. The current implementations only works with a subset of possible models, e.g. univariate continuous observations. It's also the issue of how to extract the data from a `Model`; the current approach of using a `Sampler` to do so seems _very_ hacky imo. 

The same with the dependency on `GpABC.jl`; this should not be forced (as you mentioned, e.g. using requires), but I've added it at the moment just so people can try the branch out.

My hope here is to get a discussion going of how to implement ABC in Turing :+1:

src/inference/abc.jl
torfjelde: This is a left-over comment from before; I've removed this now :+1:

src/inference/abc.jl
torfjelde: :+1:

torfjelde:
@jbrea I'm very happy to join the conversation so we can make the integration with Turing as smooth as possible!:) I'll follow and try to chime in over at your repo from now on :+1:

Project.toml
torfjelde: Done :+1:

mcreel:
I made a few scripts that do one version of ABC, based on summary statistics and the asymptotic Gaussian distribution of the statistics, using Turing and AdvancedMH. They are just an example. They might be helpful for more ambitious examples: https://github.com/mcreel/ABC-MSM-Turing

yebai:
Probably better to do this via a separate ABC package that supports models specified via `DynamicPPL`.

Issue #1337, opened 2020-06-28
Unable to run "Using DynamicHMC" example


jerome-white:
From the [Using DynamicHMC documentation](https://turing.ml/stable/docs/using-turing/dynamichmc):
```julia
# Import Turing and DynamicHMC.
using LogDensityProblems, DynamicHMC, Turing

# Model definition.
@model gdemo(x, y) = begin
  s ~ InverseGamma(2, 3)
  m ~ Normal(0, sqrt(s))
  x ~ Normal(m, sqrt(s))
  y ~ Normal(m, sqrt(s))
end

# Pull 2,000 samples using DynamicNUTS.
chn = sample(gdemo(1.5, 2.0), DynamicNUTS(), 2000)
```
I get the following error:
```
┌ Error: Exception while generating log record in module Turing.Inference at /Users/abc/.julia/packages/Turing/GMBTf/src/contrib/inference/dynamichmc.jl:130
│   exception =
│    MethodError: no method matching alg_str(::DynamicNUTS{Turing.Core.ForwardDiffAD{40},()})
│    Closest candidates are:
│      alg_str(!Matched::DynamicPPL.Sampler{#s104,S} where S<:DynamicPPL.AbstractSamplerState where #s104<:Turing.Inference.Hamiltonian) at /Users/abc/.julia/packages/Turing/GMBTf/src/inference/hmc.jl:80
│      alg_str(!Matched::DynamicPPL.Sampler{#s106,S} where S<:DynamicPPL.AbstractSamplerState where #s106<:MH) at /Users/abc/.julia/packages/Turing/GMBTf/src/inference/mh.jl:65
│    Stacktrace:
│     [1] macro expansion at ./logging.jl:322 [inlined]
│     [2] sample(::Random._GLOBAL_RNG, ::DynamicPPL.Model{var"###evaluator#279",(:x, :y),Tuple{Float64,Float64},(),DynamicPPL.ModelGen{var"###generator#280",(:x, :y),(),Tuple{}}}, ::DynamicNUTS{Turing.Core.ForwardDiffAD{40},()}, ::Int64; chain_type::Type{T} where T, resume_from::Nothing, progress::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/abc/.julia/packages/Turing/GMBTf/src/contrib/inference/dynamichmc.jl:130
[...]

PR #1338, opened 2020-06-28
Use and test `DynamicPPL.alg_str`

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1337.


yebai:
thanks, @devmotion!

PR #1342, opened 2020-06-30
Remove DEBUG

devmotion:
Same as https://github.com/TuringLang/DynamicPPL.jl/pull/148.


devmotion:
I'm unsure if it interferes with progress logging which is disabled in the tests, see https://github.com/TuringLang/DynamicPPL.jl/pull/148#issuecomment-651711363.

devmotion:
Couldn't spot any obvious performance regression in the CI tests. I also performed a simple benchmark with progress logging:
```julia
using Turing, BenchmarkTools

@model function gauss_unknown(y)
    N = length(y)
    m ~ Normal(0, 1)
    s ~ truncated(Cauchy(0, 5), 0, Inf)
    y ~ filldist(Normal(m, s), N)
end
```

The results with this PR and the corresponding one to DynamicPPL are
```julia
@btime sample($(gauss_unknown(randn(100))), $(NUTS(0.65)), 10_000); # 1.357 s (5360686 allocations: 511.90 MiB)
@btime sample($(gauss_unknown(randn(100))), $(MH()), 10_000); # 955.633 ms (2759622 allocations: 220.54 MiB)
```
whereas on Turing master + DynamicPPL dev branch (for a fair comparison) I get
```julia
@btime sample($(gauss_unknown(randn(100))), $(NUTS(0.65)), 10_000); # 1.371 s (5396111 allocations: 517.47 MiB)
@btime sample($(gauss_unknown(randn(100))), $(MH()), 10_000); # 998.275 ms (2759622 allocations: 220.54 MiB)
```
i.e., at least for this model and these samplers I can't observe any performance regressions.

BTW on Turing master + the latest release of DynamicPPL I get
```julia
@btime sample($(gauss_unknown(randn(100))), $(NUTS(0.65)), 10_000); # 1.614 s (5435581 allocations: 522.35 MiB)
@btime sample($(gauss_unknown(randn(100))), $(MH()), 10_000); # 937.352 ms (2759622 allocations: 220.54 MiB)
```

xukai92:
Does `@code_warntype` complain type stability without `DEBUG`?

devmotion:
Which function are you concerned about? `sample` with `NUTS`?

xukai92:
Yes. But that would be about removing `DEBUG` in AHMC then.

yebai:
@xukai92 Can you confirm that this PR doesn't affect type stability? If so we'll remove all `DEBUG` gradually.

xukai92:
> @xukai92 Can you confirm that this PR doesn't affect type stability? If so we'll remove all `DEBUG` gradually.

It's hard to tell it _for Turing_ because it seems even for master branch (9049c5e2), 
the `sample` function, as a whole, is not type stable

```julia
julia> using Turing
julia> @model function gauss_unknown(y)
           N = length(y)
           m ~ Normal(0, 1)
           s ~ truncated(Cauchy(0, 5), 0, Inf)
           y ~ filldist(Normal(m, s), N)
       end
julia> ydata = randn(100)
```
image

But still

- As from David's numbers this PR doesn't seem to change the performance for users.
- As this PR only touches `DEBUG` _in Turing_, it won't cause any type stability issue to other packages, e.g. AHMC.

yebai:
Can we construct a simple MWE (without depending on Turing) to verify whether `@debug` is type stable overall?

xukai92:
```julia
julia> function f(x)
           @debug "x is $x"
           x + x
       end
```

image

It seems that the error itself is not. But I don't think this has any effect on performance.

mohamed82008:
I think one of the reasons for adding DEBUG was that the code generated by `@debug` was not eliminated when debugging is off. This is especially annoying calling `@code_warntype` because it creates lots of noise.

devmotion:
There shouldn't be any performance regressions due to `@debug` since one of its design goals is to be performant (as explained in the docs), and I haven't experienced any regressions with this branch. IMO the verbose output of `@code_warntype` doesn't justify a custom logging solution in Turing (maybe the printing could be improved upstream in base Julia though). Additionally, I'm not sure how often users would want to inspect `@code_warntype` for the affected methods (actually, I wouldn't expect them to do it all since it should be sufficient if they inspect the type stability of their own models which shouldn't be affected by these changes - and I guess `@code_warntype` is intended to be used by more advanced users mainly).

yebai:
It makes sense to fix the `verbose output` from `@debug` upstream. In the meantime, does it make sense define a customised `Turing.@debug` to replace `Turing.DEBUG && @debug`? 

 I think it might be possible to easily switch between `Turing.@debug` and Julia's default `@debug` via some tricks.

yebai:
Any objections for merging this?

cpfiffer:
None from me.

PR #1345, opened 2020-07-01
residual sampling weird weighting

francescoalemanno:
this PR fixes:
residual sampling is using unmodified weights on the stochastic correction, this will introduce an unwanted bias on the mode.

### DIFF
--- a/src/inference/AdvancedSMC.jl
+++ b/src/inference/AdvancedSMC.jl
@@ -388,37 +388,27 @@ function resample_residual(w::AbstractVector{<:Real}, num_particles::Integer)
 
     # "Repetition counts" (plus the random part, later on):
     M_w = M .* w
-    Ns = floor(Int, M_w)
+    residuals = similar(w)
+    indices = similar(Int, w)
+    # deterministic assignment
+    i = 1
+    m = length(w)
     
-    # The "remainder" or "residual" count
-    R = sum(Ns)
-    
-    # The number of particles which will be drawn stochastically:
-    M_rdn = num_particles - R
-
-    # The modified weights:
-    Ws = @. (M_w - Ns) / M_rdn
-
-    # The "remainder" or "residual" count:
-    R = Int(sum(Ns))
-
-    # The number of particles which will be drawn stocastically:
-    M_rdn = num_particles - R
-
-    # The modified weights:
-    Ws = (M .* w - floor.(M .* w)) / M_rdn
-
-    # Draw the deterministic part:
-    indx1, i = Array{Int}(undef, R), 1
-    for j in 1:M
-        for k in 1:Ns[j]
-            indx1[i] = j
+    @inbounds for j in 1:m
+        x = m * w[i]
+        floor_x = floor(Int, x)
+        for k in 1:floor_x
+            indices[i] = j
             i += 1
         end
+        residuals[j] = x - floor_x
     end
-
-    # And now draw the stocastic (Multinomial) part:
-    return append!(indx1, rand(Categorical(Ws), M_rdn))
+    
+    # sampling from residuals
+    residuals ./= sum(residuals)
+    rand!(Categorical(residuals), view(indices, i:m))
+    
+    return indices
 end
 
 """
--- a/src/inference/AdvancedSMC.jl
+++ b/src/inference/AdvancedSMC.jl
@@ -383,16 +383,13 @@ function resample_multinomial(w::AbstractVector{<:Real}, num_particles::Integer)
 end
 
 function resample_residual(w::AbstractVector{<:Real}, num_particles::Integer)
-
-    M = length(w)
-
+    m = length(w)
+    @assert m==num_particles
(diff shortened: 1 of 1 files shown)


src/inference/AdvancedSMC.jl
```diff
@@ -387,7 +387,7 @@ function resample_residual(w::AbstractVector{<:Real}, num_particles::Integer)
     M = length(w)
 
     # "Repetition counts" (plus the random part, later on):
-    Ns = floor.(length(w) .* w)
+    Ns = floor.(M .* w)
```
devmotion, line 390: Actually now that I look at the implementation I think we might want to change it even more and make it a bit more efficient.
I'd suggest the following (replacing also the lines below):
```suggestion
    M_w = M .* w
    Ns = floor(Int, M_w)
    
    # The "remainder" or "residual" count
    R = sum(Ns)
    
    # The number of particles which will be drawn stochastically:
    M_rdn = num_particles - R

    # The modified weights:
    Ws = @. (M_w - Ns) / M_rdn
```
Currently `Ns` is an array of floating point numbers, which is not needed and makes all computations involving `Ns` such as `sum(Ns)` and `1:Ns[j]` operate on floating point numbers instead of integers. Furthermore, the computation of `M .* w` is repeated three times currently.

devmotion: Maybe even more efficiently would be something like
```suggestion
    residuals = similar(w)
    indices = similar(Int, w)

    # deterministic assignment
    i = 1
    m = length(w)
    @inbounds for j in 1:m
        x = m * w[i]
        floor_x = floor(Int, x)
        for k in 1:floor_x
            indices[i] = j
            i += 1
        end
        residuals[j] = x - floor_x
    end
    
    # sampling from residuals
    residuals ./= sum(residuals)
    rand!(Categorical(residuals), view(indices, i:m))
    
    return indices
end
```

That avoids the creation of all the intermediate arrays, the computation of how many particles are assigned deterministically, and in particular avoids the possibly costly call of `append!` in the end.

francescoalemanno:
I applied the suggestions blindly, sorry for the many commits, not it should be good

src/inference/AdvancedSMC.jl
```diff
@@ -383,34 +383,32 @@ function resample_multinomial(w::AbstractVector{<:Real}, num_particles::Integer)
...
-    # The modified weights:
-    Ws = (M .* w - floor.(M .* w)) / M_rdn
-
-    # Draw the deterministic part:
-    indx1, i = Array{Int}(undef, R), 1
-    for j in 1:M
-        for k in 1:Ns[j]
-            indx1[i] = j
+    residuals = similar(w)
+    indices = similar(w,Int)
```
devmotion, line 390:
```suggestion
    indices = Vector{Int}(undef, num_particles)
```

src/inference/AdvancedSMC.jl
```diff
@@ -383,34 +383,32 @@ function resample_multinomial(w::AbstractVector{<:Real}, num_particles::Integer)
...
+    # deterministic assignment
+    i = 1
+    
+    @inbounds for j in 1:m
+        x = m * w[i]
```
devmotion, line 395:
```suggestion
    @inbounds for j in 1:length(w)
        x = num_particles * w[i]
```

src/inference/AdvancedSMC.jl
```diff
@@ -383,34 +383,32 @@ function resample_multinomial(w::AbstractVector{<:Real}, num_particles::Integer)
...
         end
+        residuals[j] = x - floor_x
     end
-
-    # And now draw the stocastic (Multinomial) part:
-    return append!(indx1, rand(Distributions.sampler(Categorical(w)), M_rdn))
+    
+    # sampling from residuals
+    residuals ./= sum(residuals)
+    rand!(Categorical(residuals), view(indices, i:m))
```
devmotion, line 406:
```suggestion
    # sampling from residuals
    if i <= num_particles
        residuals ./= sum(residuals)
        rand!(Categorical(residuals), view(indices, i:num_particles))
    end
```

src/inference/AdvancedSMC.jl
```diff
@@ -383,34 +383,32 @@ function resample_multinomial(w::AbstractVector{<:Real}, num_particles::Integer)
 end
 
 function resample_residual(w::AbstractVector{<:Real}, num_particles::Integer)
-
-    M = length(w)
-
+    m = length(w)
+    @assert m==num_particles
```
devmotion, line 387: This doesn't have to be the case (and that's the reason why some of the code below has to be changed)
```suggestion
```
BTW I believe this was an additional bug in the previous implementation since it used `M = length(w)` where `num_particles` would have been more appropriate.

francescoalemanno:
@devmotion i'm having a hard time figuring which tests are failing, can you help me out?

devmotion:
Thanks @francescoalemanno!

PR #1348, opened 2020-07-03
add ColPrac guide

willtebbutt:
Adds the new collaborative practices guide. This seems to have buy in from several members of the team, but would require everyone to be on board.

We already do pretty much everything, the only thing that we _don't_ do is release a patch version after every merged PR. The reason to like this is that
1. it makes identifying where a bug was introduced really easy (if we accidentally introduce a bug with a release)
2. it means that patches get to users instantly, rather than having to wait for minor releases that we only do every so often

It doesn't really have any downsides because there's plenty of tooling in place to handle everything.

@yebai @xukai92 @devmotion @cpfiffer and anyone else that I've missed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 [![Build Status](https://github.com/TuringLang/Turing.jl/workflows/Turing-CI/badge.svg)](https://github.com/TuringLang/Turing.jl/actions?query=workflow%3ATuring-CI+branch%3Amaster)
 [![Coverage Status](https://coveralls.io/repos/github/yebai/Turing.jl/badge.svg?branch=master)](https://coveralls.io/github/yebai/Turing.jl?branch=master)
 [![Documentation](https://img.shields.io/badge/doc-latest-blue.svg)](https://turing.ml/dev/docs/using-turing/)
+[![ColPrac: Contributor's Guide on Collaborative Practices for Community Packages](https://img.shields.io/badge/ColPrac-Contributor's%20Guide-blueviolet)](https://github.com/SciML/ColPrac)
 
 **Turing.jl** is a Julia library for general-purpose [probabilistic programming](https://en.wikipedia.org/wiki/Probabilistic_programming_language). Turing allows the user to write models using standard Julia syntax, and provides a wide range of sampling-based inference methods for solving problems across probabilistic machine learning, Bayesian statistics, and data science. Compared to other probabilistic programming languages, Turing has a special focus on modularity, and decouples the modelling language (i.e. the compiler) and inference methods. This modular design, together with the use of a high-level numerical language Julia, makes Turing particularly extensible: new model families and inference methods can be easily added.
 


devmotion:
> We already do pretty much everything, the only thing that we don't do is release a patch version after every merged PR. 

I think that would help a lot since currently we accumulate quite many changes in some upstream packages which then "suddenly" require fixes in, e.g., Turing when they are released, although some of them could have been integrated much earlier and weren't breaking. I'm wondering if these guidelines indicate that one should not work with a separate dev branch (apart from maybe if one wants to split work on some new features that belong together in separate PRs?)?

willtebbutt:
@yebai just bumping this. I agree with @devmotion that currently Turing tends to accumulate quite a few non-breaking changes between releases, which is bad because bug fixes / new but non-breaking take longer to make it to users.

If it _is_ the case that you want to be able to make multiple breaking changes between releases, it might make sense to move to a (very slightly) more sophisticated set up with git whereby

- master contains the most recent release at any point in time. Any non-breaking changes (bug fixes etc) get merged directly into master and a new patch version is released immediately
- a separate `dev` branch contains all breaking changes, and is merged into master when a minor version release happens.

For example, suppose we're currently on version 0.13.5.
- if someone produces a bug fix, it gets merged directly into master and bumps the version to 13.6. This change is also merged `dev` so that it remains up-to-date with master
- if someone is working on a new feature that _isn't_ breaking (performance-related, fancy new syntax that is backwards-compatible etc), the same happens
- new breaking changes get merged into `dev` until a release is ready to go, at which point `dev` is merged into `master` and version 0.14 is released.

cpfiffer:
I think one of the issues I've noticed with the `dev` workflow is making sure the minor changes that go directly into `master` are reflected in some way in the `dev` branch -- does anyone know a way to automate that kind of thing?

xukai92:
@willtebbutt's proposal sounds great.

> I think one of the issues I've noticed with the `dev` workflow is making sure the minor changes that go directly into `master` are reflected in some way in the `dev` branch -- does anyone know a way to automate that kind of thing?

I think we will have to choose which branch (`master` or `dev`) to merge when crating the PR.
Automatically making PRs that go in to `master` also go to `dev` should be possible via GitHub Actions but I don't know exactly how.

yebai:
This looks like a good plan. It would be nice to have an automatic mechanism for keeping dev and master branches synced for bug-fixing PRs. For example, if a (non-dev) PR is directly merged into against master, then automatically open a PR for the dev branch.

devmotion:
> It would be nice to have an automatic mechanism for keeping dev and master branches synced for bug-fixing PRs. For example, if a (non-dev) PR is directly merged into master, then automatically open a PR for the dev branch.

Something like https://github.com/TuringLang/DynamicPPL.jl/blob/master/.github/workflows/pullrequest.yml could maybe work. At least just opened the first PR to the dev branch.

PR #1357, opened 2020-07-17
Fix for issue #1352

### LINKED ISSUE #1352: Wrong prediction results on multivariate params
The `predict` function outputs wrong results depending on how the multivariate parameters are constructed. I present below a simple linear regression problem which always converge fine, but whose `predict` results depend on how the `coef` parameter is constructed.

```julia
using Turing, Plots, StatsPlots
using Turing.Inference: predict

@model function simple_linear(x, y)
    intercept ~ Normal(0,1)

    ## this corrupts `predict` output
    coef ~ MvNormal(2, 1)

    ## this alternative also
    # coef ~ filldist(Normal(0,1), 2)

    ## but this version works fine
    # coef = Vector(undef, 2)
    # for i in axes(coef, 1)
    #     coef[i] ~ Normal(0,1)
    # end

    ## this works too
    # coef1 ~ Normal(0,1)
    # coef2 ~ Normal(0,1)
    # coef = [coef1, coef2]

    coef = reshape(coef, 1, size(x,1))

    mu = intercept .+ coef * x |> vec

    error ~ truncated(Normal(0,1), 0, Inf)

    y ~ MvNormal(mu, error)
end


# simple linear transformation
x = randn(2, 100)
y = [1 + 2 * a + 3 * b for (a,b) in eachcol(x)]

chain = sample(simple_linear(x, y), NUTS(), 1000)

# model converges fine
plot(chain) |> display
@show chain

p = predict(simple_linear(x, missing), chain)

# prediction correctness depends on how multivariate params were constructed
@show y[1]
@show p["y[1]"].value.data |> mean # should be close to y[1] above
@show p["y[1]"].value.data |> std # sould be close to 0.0
```

I'm trying this on Julia 1.4.1 with Turing 0.13.0.

torfjelde:
Fixes #1352, i.e. makes `transitions_from_chain` and thus `predict` compatible with `MCMCChains@0.4` and higher.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -753,9 +753,17 @@ function transitions_from_chain(
         md = vi.metadata
         for v in keys(md)
             for vn in md[v].vns
-                vn_symbol = Symbol(vn)
-                if vn_symbol ∈ c.name_map.parameters
-                    val = c[vn_symbol]
+                vn_sym = Symbol(vn)
+
+                # This returns `()` if `vn` not present in `c`
+                # Otherwise it returns `(a = ..., )` even if
+                # `a` represents non-univariate.
+                res = get(c, vn_sym; flatten = false)
+                if !isempty(res)
+                    # FIXME: this does not handle the cases where
+                    # only a subset of the indices are set, e.g.
+                    # if `a[1]` is in `chain` but `a[2]` is not.
+                    val = copy.(vec(c[vn_sym].value))
                     DynamicPPL.setval!(vi, val, vn)
                     DynamicPPL.settrans!(vi, false, vn)
                 else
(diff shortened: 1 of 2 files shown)


cpfiffer:
Does the test [here](https://github.com/TuringLang/Turing.jl/blob/master/test/inference/utilities.jl) cover the issue in #1352? Do we need to expand the tests with their specific problem?

torfjelde:
Good point! Nah, that doesn't catch it because it only deals with univariate variables. I'll make it so that it also includes multivariate and matrixvariates :+1:

torfjelde:
Done :+1:

torfjelde:
Classic. We got AD-issues completely unrelated to the changes introduced in this PR again!

Bleh. Do we just merge and open an issue related to the AD-stuff?

torfjelde:
Seems to be an issue for Tracker.jl; unrelated to Turing.

yebai:
I am merging now; let's fix the AD issue in a separate PR.

Issue #1358, opened 2020-07-17
Tests failing due to the ambiguity error for `\` with `PDMat` and `TrackedArray`


torfjelde:
See logs: https://github.com/TuringLang/Turing.jl/pull/1357/checks?check_run_id=883458568.

Seems to be an issue with `\` being a bit "too" specialized in PDMats.jl. Uncertain where to raise the issue, so I've just raised it here and we can move it or bother someone else about it if that's the right course of action.

mohamed82008:
I think changing `TrackedArray` in Tracker's method to `TrackedVecOrMat` should fix this.

yebai:
Could we fix this from the Turing side first to get #1357 merged?

torfjelde:
> I think changing TrackedArray in Tracker's method to TrackedVecOrMat should fix this.

Unfortunately it doesn't seem to do the trick :confused: 

```julia
PDMatDistribution AD: Error During Test at /home/tor/.julia/dev/Turing/test/core/ad.jl:284
  Got exception outside of a @test
  MethodError: \(::PDMat{Float64,Array{Float64,2}}, ::TrackedArray{…,Array{Float64,2}}) is ambiguous. Candidates:
    \(a::PDMat, x::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T) in PDMats at /home/tor/.julia/packages/PDMats/8bndI/src/pdmat.jl:51
    \(A::AbstractArray, B::Union{TrackedArray{T,1,A}, TrackedArray{T,2,A}} where A where T) in Tracker at /home/tor/.julia/dev/Tracker/src/lib/array.jl:299
    \(A::AbstractArray{T,2} where T, B::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T) in LinearAlgebra at /opt/julia-1.4.0/share/julia/stdlib/v1.4/LinearAlgebra/src/generic.jl:1103
  Possible fix, define
    \(::PDMat, ::Union{TrackedArray{T,1,A}, TrackedArray{T,2,A}} where A where T)
```
seems like the issue is that `A` is not specified. It also seems like a bug that is "recognized" (or at least they're aware that such issues could arise): https://github.com/FluxML/Tracker.jl/blob/b460dcf5199c7d64ec5ea9e2392fa131f7369b2e/src/lib/array.jl#L297

devmotion:
[...]

PR #1362, opened 2020-07-27
Use AdvancedVI.jl for variational inference

torfjelde:
Most of the code related to variational inference has been moved over to https://github.com/TuringLang/AdvancedVI.jl and this PR makes Turing depend on AdvancedVI.jl for most of its VI-related functionality.


Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.13.0"
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AdvancedHMC = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
 AdvancedMH = "5b7e9947-ddc0-4b3f-9b55-0d8042f74170"
+AdvancedVI = "b5ca4192-6429-45e5-a2d9-87aec30a685c"
```
devmotion, line 9: We should add a compat entry for AdvancedVI.

yebai:
It eems the failing tests are caused by #1358.

torfjelde:
Yeah; essentially this is the PR that made me aware of that bug. Up to you whether or not we should merge. I tried the VI-tests locally on my laptop and they passed without issue. The only thing that _could_ (but really shouldn't) cause issues globally is the fact that now `Turing.setadbackend` also does it for `Bijectors.jl`, but there are no bijectors currently used which makes use of the AD-backend in Bijectors so shouldn't matter.

Issue #1365, opened 2020-07-29
Apply AD gradient if `optimizer` is a first-order one


wupeifan:
In estimating `MLE` and `MAP`, the routines from `Optim.jl` are called. However, even for gradient-based methods, Optim.jl only support `ForwardDiff` as an AD backend, and apply finite differences otherwise. Therefore, it makes sense to use the AD backend in Turing (`Turing.setadbackend`) to define the gradient function and feed it to the optimizer.

Basically, we replace https://github.com/TuringLang/Turing.jl/blob/master/src/modes/ModeEstimation.jl#L383 by a structure like
``` Julia
if optimizer isa Optim.FirstOrderOptimizer
   ...
else
   ...
end
```
In the first branch, you will need to pass the original function and another that computes the function value and gradient-based on `Turing.gradient_logp`. 
An example is https://github.com/JuliaNLSolvers/Optim.jl/blob/7b660484724755ee2b306dd3ceed13d6633067ae/test/multivariate/optimize/interface.jl#L54

pkofod:
Let me know if you need help here. Is it possible to calculate the gradient and get the value simultaneously with your setup? That would help performance.

wupeifan:
I think in this setting the gradient is `Turing.gradient_logp`. Therefore, you can wrap the value and the gradient together. @mohamed82008 also pointed out this https://github.com/JuliaNLSolvers/NLSolversBase.jl/blob/9e89526817d5932489d8fdfead5f6e537c1291a1/src/objective_types/oncedifferentiable.jl#L172 in Slack discussion thread.

devmotion:
I guess the branch is not needed? It seems one could just define
```julia
function (f::OptimLogDensity)(F, G, H, x)
    if G !== nothing
        ...
    end
    if H !== nothing
        ...
    end
    if F !== nothing
        return ...
    end
    nothing
end
```
and then call `Optim.optimize(Optim.only_fgh!(f), ...)`, similar to https://github.com/JuliaNLSolvers/Optim.jl/blob/7b660484724755ee2b306dd3ceed13d6633067ae/test/multivariate/optimize/interface.jl#L73-L78.

wupeifan:
Yeah, probably. Thanks!
I'm swamped with more urgent stuff so I might not be able to make a PR myself for this one... Maybe wait for Cameron after he finishes his qualifier, or maybe I'll revisit this in a couple of days...

cpfiffer:
[...]

PR #1369, opened 2020-08-04
Provide AD gradient for MLE/MAP

cpfiffer:
Currently, MLE/MAP is not using the AD gradients calculated with `gradient_logp`, as noted in #1365. This PR modifies the MLE/MAP code to use the AD-generated gradient.

I incremented the version number to 0.13.1, but this might have to be 0.14.0 since we are technically dropping support for second-order methods with this PR. I intend to follow up in a separate PR with code that adds a `hessian_logp` method so people can use `Optim.Newton` or whatever -- it'll also make the information matrix calculations much better to use the AD-Hessian.


src/modes/ModeEstimation.jl
```diff
@@ -147,6 +147,32 @@ function (f::OptimLogDensity)(z)
...
+    # Calculate log joint and the gradient
+    l, g = gradient_logp(
+        z, 
+        DynamicPPL.VarInfo(f.vi, spl, z), 
+        f.model, 
+        DynamicPPL.SampleFromPrior(),
+        f.context
+    )
+
+    if G !== nothing
```
devmotion, line 162: If `G === nothing` it seems wasteful to compute the gradient at all. I guess in that case we should only evalute the log joint probability.

src/modes/ModeEstimation.jl
```diff
@@ -369,6 +395,11 @@ function _optimize(
     args...; 
     kwargs...
 )
+    # Throw an error if we received a second-order optimizer.
```
devmotion, line 398: Do we have to do that? Doesn't Optim just use ForwardDiff (or FD?) to compute the Hessian in this case? If that's the case, then we shouldn't throw an error IMO. It might not be the most efficient approach and would not adhere to the user-provided AD settings but as long as it works we could only print a warning.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.13.0"
+version = "0.13.1"
```
devmotion, line 3: We should make sure that we haven't introduced any breaking changes since 0.13.0. (IMO we should adopt the ColPrac practice of making patch releases for every PR).

src/modes/ModeEstimation.jl
wupeifan: In FD yes; but (for example in the project that I am working on) it could be that users only define custom adjoints for the gradients but not the Hessian. Therefore even the user provides an AD backend, it might not be a great idea if it by default take that for Hessian function.

devmotion: You mean we shouldn't even print a warning? Would be fine with me as well.

wupeifan: Oh, I think throwing an error when some Hessian-required optimizer is received is a great idea, just like what Cameron did here.

devmotion: If we want to throw an error if the Hessian is evaluated, I suggest using `only_fgh!(f)` and implementing `f(F, G, H, x)` that contains the check
```julia
if H !== nothing
    error("second order methods are not supported at the moment")
end
```
In general, this approach is more flexible, avoids baking in a hardcoded check for a special type of a different package in our implementation, and avoids incorrect and unexpected behaviour for second-order optimization algorithms that don't subtype this specific type (since multiple inheritance is not possible in Julia, that's not an impossible scenario per se).

cpfiffer: I had this, but using `only_fgh!` when the optimizer is first order throws the error

```julia
ERROR: LoadError: MethodError: objects of type Nothing are not callable
Stacktrace:
 [1] (::NLSolversBase.var"#61#62"{NLSolversBase.InplaceObjective{Nothing,Nothing,Turing.OptimLogDensity{DynamicPPL.Model{var"###evaluator#331",(),Tuple{},(),DynamicPPL.ModelGen{var"###generator#332",(),(),Tuple{}}},Turing.OptimizationContext{DynamicPPL.LikelihoodContext{Nothing}},DynamicPPL.VarInfo{NamedTuple{(:s, :m),Tuple{DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:s,Tuple{}},Int64},Array{InverseGamma{Float64},1},Array{DynamicPPL.VarName{:s,Tuple{}},1},Array{Float64,1},Array{Set{DynamicPPL.Selector},1}},DynamicPPL.Metadata{Dict{DynamicPPL.VarName{:m,Tuple{}},Int64},Array{Normal{Float64},1},Array{DynamicPPL.VarName{:m,Tuple{}},1},Array{Float64,1},Array{Set{DynamicPPL.Selector},1}}}},Float64}},Nothing,Nothing},Float64})(::Array{Float64,1}, ::Array{Float64,1}) at /home/cameron/.julia/packages/NLSolversBase/mGaJg/src/objective_types/incomplete.jl:45
 [2] value_gradient!!(::OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}) at /home/cameron/.julia/packages/NLSolversBase/mGaJg/src/interface.jl:82
 [3] initial_state(::LBFGS{Nothing,LineSearches.InitialStatic{Float64},LineSearches.HagerZhang{Float64,Base.RefValue{Bool}},Optim.var"#15#17"}
, ::Optim.Options{Float64,Nothing}, ::OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}) at /home/cameron/.julia/packages/Optim/TNmSw/src/multivariate/solvers/first_order/l_bfgs.jl:158
 [4] optimize(::OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}, ::LBFGS{Nothing,LineSearches.InitialStatic{
Float64},LineSearches.HagerZhang{Float64,Base.RefValue{Bool}},Optim.var"#15#17"}, ::Optim.Options{Float64,Nothing}) at /home/cameron/.julia/packages/Optim/TNmSw/src/multivariate/optimize/optimize.jl:33
```

@pkofod do you know if this is supposed to happen?

cpfiffer: At the very least maybe the fix is 
```julia
    M = if optimizer isa Optim.SecondOrderOptimizer
        Optim.optimize(Optim.only_fgh!(f), init_vals, optimizer, options, args...; kwargs...)
    else
        Optim.optimize(Optim.only_fg!(f), init_vals, optimizer, options, args...; kwargs...)
    end
```

cpfiffer: In addition to throwing the error @devmotion added above, I mean. It seems to be working well on my local run.

src/modes/ModeEstimation.jl
cpfiffer: Fixed.

Project.toml
cpfiffer: I'll just bump it up to 0.14.0. Honestly at this point we should consider moving to 1.0 as well.

wupeifan: More importantly, current 0.13.0 is by default failing so `]  add Turing` and `using Turing` will fail. I think maybe you guys want to bump up the version really soon...

devmotion: We first have to fix the bug introduced by the changes in PDMats 0.10 on master before releasing 0.14.0. What was your package setup that failed, i.e. can you post the output of `] st`?

wupeifan: Oh, I have mine on Turing#master, but there are more than one guys on slack that faces an issue: `] add Turing` installs an older version and `using Turing` somehow fails.

devmotion: Usually these problems are caused by unbounded compatibilities of old Turing versions (there are many closed issues in the repo here). These issues should be fixed by running `] add Turing@0.13.0` and possibly adjusting conflicting packages (by users) and adding correct bounds in the registry (by us). I fixed some bounds a while ago, but it seems the old version are still missing some compatibilty bounds.

src/modes/ModeEstimation.jl
pkofod: It's a bug. I think there might be an issue for it.

devmotion: I just found https://github.com/JuliaNLSolvers/Optim.jl/issues/718, I guess that's the related issue.

pkofod: Yeah, I have a fix. Sorry to cossio for waiting a year and a half :grimacing:

pkofod: I mean, I will tag a fix in an hour or so, so please don't special case with a branch.

pkofod: 19141

src/modes/ModeEstimation.jl
```diff
@@ -147,6 +147,50 @@ function (f::OptimLogDensity)(z)
     return -DynamicPPL.getlogp(varinfo)
 end
 
+function (f::OptimLogDensity)(F, G, z)
```
devmotion, line 150: I'm not sure if it's useful to keep this separate definition? It seems we only need `f(F, G, H, z)`, so the implementation could just be included there directly.

wupeifan:
It seems that there is nothing blocking this PR, I think?

devmotion:
Tests fail currently.

cpfiffer:
I'll take a look when I'm done teaching today, looks like I've done something strange to the methods.

devmotion (approved):
LGTM.

PR #1373, opened 2020-08-12
Add some missing RNGs

devmotion:
This might have caused some issues with parallel sampling. However, with AbstractMCMC 2 this should be changed (and hopefully improved) quite a bit (see https://github.com/TuringLang/DynamicPPL.jl/pull/150).


yebai:
Thanks, @devmotion!

PR #1375, opened 2020-08-16
Add "Turing is fast" to splash page

cpfiffer:
Adds a little section to the feature ribbon about Turing's performance. I'd like to hyperlink to some performance information there -- what's the central spot I should link to?


yebai:
We don't have a nice page form that yet. In the meantime, perhaps link to the following arxiv paper?

https://arxiv.org/abs/2002.02702

cpfiffer:
Cool, added. I also changed the splash page so hyperlinks are underlined -- previously they were hard to identify as hyperlinks.

PR #1378, opened 2020-08-17
Require Julia 1.3

devmotion:
This PR drops support of Julia < 1.3 since other packages such as Tracker and NNlib only support Julia >= 1.3 which would require us to backport fixes if we want to support Julia < 1.3 properly.

### DIFF
--- a/.github/workflows/DynamicHMC.yml
+++ b/.github/workflows/DynamicHMC.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         version:
+          - '1.3'
           - '1'
         os:
           - ubuntu-latest
--- a/.github/workflows/Numerical.yml
+++ b/.github/workflows/Numerical.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         version:
+          - '1.3'
           - '1'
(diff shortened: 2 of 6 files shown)


devmotion:
Test fail due to the error fixed in https://github.com/TuringLang/Turing.jl/pull/1377.

devmotion:
There seems to be a problem with Julia 1.5 on Windows 32bit. There is no clear error message though, and tests seem to be fine with Julia 1.3 on Windows 32bit, so it might be a problem upstream or in some other package. Should we disable tests on Windows 32bit completely? Or just with Julia 1.5?

yebai:
It’s helpful to test Win32 on at least one Julia release. It allows us to sport syntactical and numerical errors for some algorithms, e.g. HMC.

devmotion:
Wouldn't it be sufficient to test on Linux 32bit to spot numerical errors? It seems we can test both Julia versions there.

yebai:
Yea, that's enough. I did not realise we are testing 32 bit on Linux.

devmotion:
All remaining test errors are due to https://github.com/TuringLang/Turing.jl/pull/1377.

PR #1387, opened 2020-08-23
Use separate test environment

devmotion:
Since we dropped Julia < 1.3, we can use the [new way of specifying test dependencies on Julia >= 1.2](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1).


.gitignore
```diff
@@ -24,7 +24,7 @@ docs/.jekyll-cache
 
 .vscode
 .DS_Store
-Manifest.toml
-test/Project.toml
+/Manifest.toml
```
yebai, line 27: why switching to root level directory?

yebai (approved):

.gitignore
devmotion: True, we don't have to do that here. It's mostly useful for packages where the documentation is built with a specific reproducible set of packages which requires to add a `docs/Manifest.toml` file.

yebai: I see - thanks. Maybe keep both?

devmotion: Do you mean both `Manifest.toml` and `/Manifest.toml`? `test/Project.toml` is needed for specifying the test environment.

yebai: yes

PR #1392, opened 2020-08-24
Update DynamicPPL

devmotion:
This PR updates Turing to DynamicPPL 0.9.

Should be straightforward, though I'm wondering if we should export `@addlogprob!`?


yebai:
> Should be straightforward, though I'm wondering if we should export @addlogprob!?

Let's keep these internal unless users start to complain about it.

yebai (approved):

PR #1396, opened 2020-08-27
ModeResult description inconsistency

wupeifan:
Currently, it says "minimized at ..." but it actually returns the log-density instead of the minus log-density used in optimization, hence inconsistent. Better to change it to "maximized", or report the minus log-density.

In this PR I changed the word to "maximized", but feel free to change it if you think reporting the original "minus log-density" is a better idea. That line is at the end of this file.


cpfiffer:
I incremented the version number so we can tag a release after. I'll merge this after the tests pass, but they shouldn't fail since the changes are cosmetic. Thanks @wupeifan!

Issue #1400, opened 2020-09-06
Precompilation fails while using ReverseDiff backend


BlackWingedKing:
I have created an MWE [here](https://github.com/BlackWingedKing/precompile_mwe). The description is in README.md of the repository.  
[This reply](https://julialang.slack.com/archives/CCYDC34A0/p1598865070158200?thread_ts=1598863706.158100&cid=CCYDC34A0) in the slack thread might be relevant

devmotion:
As explained on Slack, `Turing.setadbackend(:reverse_diff)` is incorrect and should not be used - it does not work even if you run it in a script.

On the other hand `Turing.setadbackend(:reversediff)` (or `Turing.setadbackend(Val(:reversediff))`) which should be used is only defined inside of a `@require` block if ReverseDiff is loaded. Reading through https://github.com/JuliaLang/Pkg.jl/issues/1285#issue-476496324, it seems that one of the bad things about optional dependencies with Requires is that the optionally loaded code does not end up in the precompile files. I suspect this fundamental problem with Requires leads to the precompilation issues - however, a MWE without Turing would be needed to confirm this hypothesis.

BlackWingedKing:
Hi @devmotion 
Thanks for your reply. I understand the issue. I have created this [fork](https://github.com/BlackWingedKing/Turing.jl) which merges this [PR](https://github.com/TuringLang/Turing.jl/pull/1401) with master and adds a few things on top of it.

* Add ReverseDiff and Memoization as a dependency
* Remove the @requires macro for both ReverseDiff and Memoization

But I still get a similar error with `:reversediff` backend
This [README](https://github.com/BlackWingedKing/precompile_mwe/blob/master/README2.md) provides steps for reproducing the over the [MWE](https://github.com/BlackWingedKing/precompile_mwe)

devmotion:
Ah nice, so it seems my hypothesis was correct and the precompilation problem is caused by Requires. The error messages that you get now with your Turing fork without Requires are caused by AdvancedVI - and there exactly the same problem exists: the ReverseDiff support is optional, hidden in a `@require` block. I assume if you would make ReverseDiff a proper dependency of both AdvancedVI and Bijectors (the same problem should be present there as well) as well, then precompilation should work.

BlackWingedKing:
Thank you @devmotion. I will make the corresponding changes. 
Also, do you guys have any plans to add ReverseDIff and Memoization as a dependency for Turing?

[...]

PR #1401, opened 2020-09-06
Make `setadbackend` more consistent + use ZygoteRules

devmotion:
This PR fixes an inconsistency observed in https://github.com/TuringLang/Turing.jl/issues/1400 which leads to different AD settings depending on whether `setadbackend(::Symbol)` or `setadbackend(::Val)` is called. Moreover, it separates the deprecations into a separate file so that (hopefully) we don't forget to remove them in the next breaking release. Additionally, ZygoteRules is used instead of Zygote inside of a `@require` block, hopefully eliminating potential precompilation and loading time issues for Zygote similar to those observed for ReverseDiff in https://github.com/TuringLang/Turing.jl/issues/1400.


devmotion:
PkgServer lags behind (again)... The new DynamicPPL release is not available in CI yet. Just have to retry until it works.

cpfiffer (approved):
LGTM, thanks!

PR #1408, opened 2020-09-15
Fix MLE/MAP with Zygote and ReverseDiff

cpfiffer:
MLE/MAP was not being tested with ReverseDiff or Zygote, so I missed the fact that the sampling context was not being passed to `gradient_logp` for those two backends. Thanks to @wupeifan for flagging this.

I added tests for this as well to make sure this doesn't happen in the future.

### DIFF
--- a/test/modes/ModeEstimation.jl
+++ b/test/modes/ModeEstimation.jl
@@ -42,31 +42,6 @@ include(dir*"/test/test_utils/AllUtils.jl")
         @test all(isapprox.(m4.values.array - true_value, 0.0, atol=0.01))
     end
 
-    @testset "AD backends" begin
-        Random.seed!(222)
-        true_value = [0.0625, 1.75]
-        
-        Turing.setadbackend(:forwarddiff)
-        m1 = optimize(gdemo_default, MLE())
-        
-        Turing.setadbackend(:reversediff)
-        m2 = optimize(gdemo_default, MLE())
-
-        Turing.setadbackend(:tracker)
-        m3 = optimize(gdemo_default, MLE())
-
-        Turing.setadbackend(:zygote)
-        m4 = optimize(gdemo_default, MLE())
-
-        # Go back to normal forwarddiff for the rest of the tests
-        Turing.setadbackend(:forwarddiff)
-
-        @test all(isapprox.(m1.values.array - true_value, 0.0, atol=0.01))
-        @test all(isapprox.(m2.values.array - true_value, 0.0, atol=0.01))
-        @test all(isapprox.(m3.values.array - true_value, 0.0, atol=0.01))
-        @test all(isapprox.(m4.values.array - true_value, 0.0, atol=0.01))
-    end
-
     @testset "StatsBase integration" begin
         Random.seed!(54321)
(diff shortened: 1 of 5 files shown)


cpfiffer:
For anyone stumbling on this issue, any MLE/MAP estimates run with Zygote or ReverseDiff as the AD backends in Turing 14.0, 14.1, or 14.2 are incorrect and should be re-run.

test/modes/ModeEstimation.jl
```diff
@@ -41,6 +42,31 @@ include(dir*"/test/test_utils/AllUtils.jl")
...
+
+        # Go back to normal forwarddiff for the rest of the tests
+        Turing.setadbackend(:forwarddiff)
+
+        @test all(isapprox.(m1.values.array - true_value, 0.0, atol=0.01))
+        @test all(isapprox.(m2.values.array - true_value, 0.0, atol=0.01))
+        @test all(isapprox.(m3.values.array - true_value, 0.0, atol=0.01))
+        @test all(isapprox.(m4.values.array - true_value, 0.0, atol=0.01))
+    end
+
```
devmotion, line 69: Maybe it would be better to move the mode tests to the AD block in runtests.jl, to more easily comment out tests of different AD backends?

(In general, IMO in a different PR we might want to restructure the tests similar to DistributionsAD or Bijectors such that we can test specific AD backends more easily using environment variables. This would also allow to split the tests in CI tests for each backend.)

cpfiffer: I'd prefer to keep it here because it falls under the optional dependency Optim -- there's really no great place to put these, so I think it's better to keep all of these "feature" tests under one banner.

devmotion: Oh maybe I didn't explain clearly what I mean. I just thought that instead of switching between AD backends here, you could move the `include(...)` line for the mode tests in runtests.jl inside of the block that cycles through the AD backends.

devmotion: Then you would not have to add any new tests at all, the mode tests would run with every AD backend automatically.

cpfiffer: Oh, yeah, that's way better. Done.

devmotion (approved):
LGTM

cpfiffer:
Cool, I'll merge this and flag a release once the tests pass.

wupeifan:
The loop of AD backends only includes Forward, Tracker and Reverse. There's no Zygote I think.

devmotion:
Oh good catch. I think we should split the CI tests and use environment variables to determine which AD backend to run in runtests.jl (a bit similar to DistributionsAD).

wupeifan:
@devmotion Yeah, I think maybe you guys want to do a similar exercise as in DistributionsAD... But I'm convinced that the previous issue on MLE/MAP with Zygote is resolved. Thanks!
It would be great if a new version is tagged soon and then I can use it...

devmotion:
The new version should already be available, @cpfiffer released it right after merging the PR: https://github.com/TuringLang/Turing.jl/commit/1d87a1d4084299de74c2d53956895741a6e0001b

Issue #1412, opened 2020-09-23
Multi-threaded sampling with reversediff backend and rdcache gives bad samples


BlackWingedKing:
The following are the posterior plots for the [coin flip](https://turing.ml/dev/tutorials/0-introduction/) turing example 
![image](https://user-images.githubusercontent.com/28155507/94046078-d15f9480-fded-11ea-9bb0-d2151f8e6102.png)
As we can see the posterior from multi-threaded sampling with rdcache is not at all close to others
Code to reproduce this
```
# Load the modules
using Turing, MCMCChains, Distributions, StatsPlots, Random
using ReverseDiff, Memoization, Zygote

println("loaded modules")
# Set the true probability of heads in a coin.
p_true = 0.5

# Iterate from having seen 0 observations to 100 observations.
Ns = 0:100
Random.seed!(12)
data = rand(Bernoulli(p_true), last(Ns))
# define the model
@model coinflip(y) = begin
    p ~ Beta(1, 1)
    N = length(y)
    for n in 1:N
        y[n] ~ Bernoulli(p)
    end
end
# parameters
num_chains = 10
iterations = 100
model_coin = coinflip(data)

# sampling
chain_serial = mapreduce(c -> sample(model_coin, NUTS(0.65), iterations), chainscat, 1:num_chains)
chain_multithread = sample(model_coin, NUTS(0.65), MCMCThreads(), iterations, num_chains)

# now enable Zygote backend
Turing.setadbackend(:zygote)
chain_zymultithread = sample(model_coin, NUTS(0.65), MCMCThreads(), iterations, num_chains)

# now enable ReverseDiff backend without rdcache
Turing.setadbackend(:reversediff)
Turing.setrdcache(false)
chain_rdmultithread = sample(model_coin, NUTS(0.65), MCMCThreads(), iterations, num_chains)

# with rdcache
Turing.setrdcache(true)
chain_rdcachemultithread = sample(model_coin, NUTS(0.65), MCMCThreads(), iterations, num_chains)

# plot the combined density of all those chains
density(chain_serial[:p][:], label="serial", legend=:topleft)
density!(chain_multithread[:p][:], label="multi-thread-normal", legend=:topleft)
density!(chain_zymultithread[:p][:], label="multi-thread-zygote", legend=:topleft)
density!(chain_rdmultithread[:p][:], label="multi-thread-reversediff", legend=:topleft)
[...]

Issue #1393, opened 2020-08-26
Cannot reuse model when using ReverseDiff and Memoization


hstrey:
Here is a little program that illustrates the problem:

```
using Turing, ReverseDiff, Memoization
using Statistics
using Random
Turing.setadbackend(:reversediff)
turing.setrdcache(true)
Turing.turnprogress(false)
using Distributions
for i in 1:10
    N = 1000
    d = Normal(0.0, i)
    data=rand(d,N)
    @model g(rn,T) = begin
        sigma ~ Uniform(0.0,20.0)
        rn ~ MvNormal(zeros(T),sigma)
    end
    chn = sample(g(data,N),NUTS(0.65),1000)
    println("Std of chain: ",mean(Array(chn[:sigma])),"Std of data: ",std(data))
end

```
I am creating a dataset that samples from a Normal distributions and our model is estimating the sigma of the distribution.  When you run the code you can see that the sampler samples only from the first dataset and therefore the estimated mean for sigma is always around 1.  I tried to run Turing.emptyrdcache() after sampling but that did not solve the problem.

Here is my Project.toml:
Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
Memoization = "6fafb56a-5788-4b4e-91ca-c0cea6611c73"
Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
Turing = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"

mohamed82008:
This issue is a bit weird. If emptying the cache is not helping then it is probably not emptying it properly. Maybe something changed with Memoization.jl. The second unusual thing here is that the same tape should be good enough for all the iterations and it should give the correct gradient. So I am suspecting something maybe wrong with ReverseDiff here.

PR #1414, opened 2020-09-24
Fix memoization issue

devmotion:
This seems to fix https://github.com/TuringLang/Turing.jl/issues/1412 for me. Seems I somehow messed up the git history, it is based on the old reversediff PR.

@mohamed82008 @BlackWingedKing can you have a look at the PR and check if it works and makes sense?

Edit: Fixes https://github.com/TuringLang/Turing.jl/issues/1393 and https://github.com/TuringLang/Turing.jl/issues/1412.

### DIFF
--- a/src/core/compat/reversediff.jl
+++ b/src/core/compat/reversediff.jl
@@ -46,12 +46,10 @@ end
 @require Memoization = "6fafb56a-5788-4b4e-91ca-c0cea6611c73" @eval begin
     setrdcache(::Val{true}) = RDCache[] = true
     function emptyrdcache()
-        for k in keys(Memoization.caches)
-            if k[1] === typeof(memoized_taperesult)
-                pop!(Memoization.caches, k)
-            end
-        end
+        Memoization.empty_cache!(memoized_taperesult)
+        return
     end
+
     function gradient_logp(
         backend::ReverseDiffAD{true},
         θ::AbstractVector{<:Real},
@@ -81,15 +79,13 @@ end
         f::F
         x::Tx
     end
-    function Memoization._get!(f::Union{Function, Type}, d::IdDict, keys::Tuple{Tuple{RDTapeKey}, Any})
+    function Memoization._get!(f, d::IdDict, keys::Tuple{Tuple{RDTapeKey}, Any})
         key = keys[1][1]
-        return Memoization._get!(f, d, (typeof(key.f), typeof(key.x), size(key.x)))
+        return Memoization._get!(f, d, (key.f, typeof(key.x), size(key.x)))
     end
     memoized_taperesult(f, x) = memoized_taperesult(RDTapeKey(f, x))
     Memoization.@memoize function memoized_taperesult(k::RDTapeKey)
         return compiledtape(k.f, k.x), GradientResult(k.x)
     end
-    memoized_tape(f, x) = memoized_tape(RDTapeKey(f, x))
-    Memoization.@memoize memoized_tape(k::RDTapeKey) = compiledtape(k.f, k.x)
     compiledtape(f, x) = compile(GradientTape(f, x))
 end
--- a/src/core/compat/reversediff.jl
+++ b/src/core/compat/reversediff.jl
@@ -23,7 +23,7 @@ function gradient_logp(
     context::DynamicPPL.AbstractContext = DynamicPPL.DefaultContext()
 )
     T = typeof(getlogp(vi))
-    
+
     # Specify objective function.
     function f(θ)
         new_vi = VarInfo(vi, sampler, θ)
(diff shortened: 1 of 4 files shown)


src/core/compat/reversediff.jl
```diff
@@ -46,12 +46,10 @@ end
 @require Memoization = "6fafb56a-5788-4b4e-91ca-c0cea6611c73" @eval begin
     setrdcache(::Val{true}) = RDCache[] = true
     function emptyrdcache()
-        for k in keys(Memoization.caches)
-            if k[1] === typeof(memoized_taperesult)
-                pop!(Memoization.caches, k)
-            end
-        end
+        Memoization.empty_cache!(memoized_taperesult)
```
mohamed82008, line 49: This will empty any user-memoized stuff as well. The old implementation only removed the relevant parts of the memo.

src/core/compat/reversediff.jl
```diff
@@ -81,15 +79,13 @@ end
         f::F
         x::Tx
     end
-    function Memoization._get!(f::Union{Function, Type}, d::IdDict, keys::Tuple{Tuple{RDTapeKey}, Any})
+    function Memoization._get!(f, d::IdDict, keys::Tuple{Tuple{RDTapeKey}, Any})
         key = keys[1][1]
-        return Memoization._get!(f, d, (typeof(key.f), typeof(key.x), size(key.x)))
+        return Memoization._get!(f, d, (key.f, typeof(key.x), size(key.x), Threads.threadid()))
```
mohamed82008, line 84: We need the type here otherwise the memoization will not work when the functions are different but not identical. I think we construct the closure in every `gradient_logp` call so the function differentiated is different every time but its type is the same. This change will actually break tape memoization.

BlackWingedKing:
@devmotion yes this fixes #1412. Here is the plot I get by running the mwe
image

mohamed82008:
@BlackWingedKing check the running times. I think the cached tape may not even be used here.

mohamed82008:
I think this needs a rebase. There are a lot of unrelated changes.

src/core/compat/reversediff.jl
devmotion: This should only remove memoized calls of `memoized_taperesult`, shouldn't it?

src/core/compat/reversediff.jl
devmotion: This change is taken from the other PR (on which the changes here are based), which tries to fix https://github.com/TuringLang/Turing.jl/issues/1393. Memoization leads to incorrect results if it is only based on the type here.

BlackWingedKing:
> @BlackWingedKing check the running times. I think the cached tape may not even be used here.

@mohamed82008 These are the running times
reversediff without rdcache - 426ms
reversediff with rdcache - 423ms
I have used `@btime` macro (`@btime sample(...)` ) with the code in #1412

mohamed82008:
Yes the cached tape here is not actually being used.

src/core/compat/reversediff.jl
mohamed82008: Let me see.

src/core/compat/reversediff.jl
devmotion: BTW this fixed the cache emptying issues in https://github.com/TuringLang/Turing.jl/issues/1393 IIRC

devmotion:
> Yes the cached tape here is not actually being used.

The problem (or one of the problems at least) was that it didn't actually used the custom `_get!` function since I forgot to change the dictionary type in the arguments - hence everything was based on value of `x` as well and of course no caching was possible then. With the latest commit I get
- ~ 40ms (cached)
- ~ 355ms (not cached)

for the MWE using BenchmarkTools.

@BlackWingedKing can you rerun your benchmarks also with the latest changes in this PR?

BlackWingedKing:
> @BlackWingedKing can you rerun your benchmarks also with the latest changes in this PR?

@devmotion with the latest commit I get
* ~44ms (cached)
* ~404ms (not cached) 
:smiley:

src/core/compat/reversediff.jl
devmotion: Memoization seems to work still: https://github.com/TuringLang/Turing.jl/pull/1414#issuecomment-698215278

src/core/compat/reversediff.jl
devmotion: I also added tests that check if the cache is emptied.

mohamed82008: Cool, thanks!

devmotion:
@mohamed82008 Tests pass and I think this PR is ready for another round of review :slightly_smiling_face:

mohamed82008 (approved):
Looks good to me. Great PR @devmotion!

PR #1420, opened 2020-09-28
Re-export `elementwise_loglikelihoods` and `generated_quantities`

torfjelde:
I think it would be a good idea to re-export `elementwise_loglikelihoods` and `generated_quantities` from DynamicPPL. v0.9.2 since these methods are mainly meant to be used by the end-user, rather than just internally. 

It also bumps the compat-entry of DynamicPPL from 0.9.2, on which these are available (though this is the only difference between 0.9.1 and 0.9.2).


src/Turing.jl
```diff
@@ -16,7 +16,7 @@ using Libtask
 using Tracker: Tracker
 
 import AdvancedVI
-import DynamicPPL: getspace, NoDist, NamedDist
+import DynamicPPL: getspace, NoDist, NamedDist, elementwise_loglikelihoods, generated_quantities
```
devmotion, line 19: Maybe use `using` to avoid accidentally overloading it and move the `using` and export statements to the re-export section (IIRC there was a comment about reexports somewhere in the file)?

Project.toml
```diff
@@ -39,7 +39,7 @@ Bijectors = "0.8"
 Distributions = "0.23.3"
 DistributionsAD = "0.6"
 DocStringExtensions = "0.8"
-DynamicPPL = "0.9.1"
+DynamicPPL = "0.9.2"
```
devmotion, line 42: Should bump the Turing version as well :slightly_smiling_face:

src/Turing.jl
torfjelde: Yeah! Uhh, am I blind though? I can't find any comment on reexports in the file...

src/Turing.jl
```diff
@@ -66,6 +66,8 @@ end
 ###########
 # Exports #
 ###########
+# `using` statements for stuff to re-export
+using DynamicPPL: elementwise_loglikelihoods, generated_quantities
```
torfjelde, line 70: How's this @devmotion ?

src/Turing.jl
devmotion: Hmm seems I misremembered, it was in DynamicPPL: https://github.com/TuringLang/DynamicPPL.jl/blob/161f8200c5d608c481b58e056c2687a380f1c0d5/src/DynamicPPL.jl#L98-L100
Here we only use `@reexport` it seems. The question is - should we just `@reexport` DynamicPPL as well?

src/Turing.jl
devmotion: Good - if we don't want to reexport everything :slightly_smiling_face:

torfjelde: Hmm, maybe. Only reason not to IMO is that a lot of the functionality in DynamicPPL is meant for developers rather than end-users (people who run models, rather than developing samplers). And I think for end-users it's nice to just have the "outwards-facing" functionality, and not clutter the namespace with `VarInfo`, etc.

BUT maybe we  also should at least re-export `logpior` and siblings?

torfjelde: Any other methods?

src/Turing.jl
torfjelde: See other comments

src/Turing.jl
trappmartin: Resxporting `logprior` and co. sounds like a good idea to me.

trappmartin:
Is there going to be some documentation on turing.ml regarding both methods?

src/Turing.jl
devmotion: Sounds reasonable, so maybe just add `logjoint` and `logprior` as well for now (`loglikelihood` is already exported since Distributions is reexported).

torfjelde:
> Is there going to be some documentation on turing.ml regarding both methods?

IMO that should be included into a larger effort put into a tutorial about diagnostics + we should really have DynamicPPL and MCMCChains also rendered under the API section IMO.

src/Turing.jl
torfjelde: Done diddit!

torfjelde:
Ready to go?

trappmartin (approved):

devmotion:
> Is there going to be some documentation on turing.ml regarding both methods?

For now there are at least docstrings, but I guess it would help if we could prepare and update the documentation in DynamicPPL (or e.g. MCMCChains) more independently.

devmotion (approved):

PR #1421, opened 2020-09-28
`predict` implemented for multiple chains

torfjelde:
Now `predict` should work even for multiple chains.

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -66,7 +66,8 @@ end
 ###########
 # Exports #
 ###########
-
+# `using` statements for stuff to re-export
+using DynamicPPL: elementwise_loglikelihoods, generated_quantities, logprior, logjoint
 using StatsBase: predict
 
 # Turing essentials - modelling macros and inference algorithms
@@ -120,5 +121,9 @@ export  @model,                 # modelling
         filldist,
         arraydist,
 
-        predict
+        predict,
+        elementwise_loglikelihoods,
+        genereated_quantities,
+        logprior,
+        logjoint
 end
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -67,6 +67,8 @@ end
 # Exports #
 ###########
 
+using StatsBase: predict
+
 # Turing essentials - modelling macros and inference algorithms
 export  @model,                 # modelling
         @varname,
@@ -116,5 +118,7 @@ export  @model,                 # modelling
         LogPoisson,
         NamedDist,
         filldist,
-        arraydist
+        arraydist,
+
+        predict
 end
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -683,7 +683,7 @@ function predict(rng::AbstractRNG, model::Model, chain::MCMCChains.Chains; inclu
     chain_result = reduce(
         MCMCChains.chainscat, [
             AbstractMCMC.bundle_samples(
-                Distributions.GLOBAL_RNG,
+                rng,
                 model,
                 spl,
                 length(chain),
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -637,7 +637,7 @@ julia> chain_lin_reg = sample(m_train, NUTS(100, 0.65), 200);
 
 julia> m_test = linear_reg(xs_test, Vector{Union{Missing, Float64}}(undef, length(ys_test)), σ);
 
-julia> predictions = Turing.Inference.predict(m_test, chain_lin_reg)
+julia> predictions = predict(m_test, chain_lin_reg)
 Object of type Chains, with data of type 100×2×1 Array{Float64,3}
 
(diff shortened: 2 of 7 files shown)


src/Turing.jl
```diff
@@ -67,6 +67,8 @@ end
 # Exports #
 ###########
 
+using StatsBase: predict
```
devmotion, line 70: Maybe move this to the re-export section when the other PR is merged?

src/inference/Inference.jl
```diff
@@ -667,21 +667,24 @@ julia> sum(abs2, ys_test - ys_pred) ≤ 0.1
...
-        transitions,
-        MCMCChains.Chains
-    )
+    chain_result = mapreduce(MCMCChains.chainscat, MCMCChains.chains(chain)) do chain_idx
+        transitions = transitions_from_chain(
+            model, chain[:, :, chain_idx];
+            sampler = spl
+        )
+        AbstractMCMC.bundle_samples(
+            Distributions.GLOBAL_RNG,
```
devmotion, line 680: We should use
```suggestion
            Random.GLOBAL_RNG,
```
Even better would be to add a `rng` argument as first argument to `predict` which would be set to `Random.GLOBAL_RNG` if omitted.

devmotion:
Would be good to add some tests that show that it actually works now :slightly_smiling_face:

torfjelde:
> Would be good to add some tests that show that it actually works now slightly_smiling_face

I was like, "what is he talking about?! I literally added that 5mins ago!", and then I realized I'd pushed the updates to my github rather the `TuringLang` :sweat_smile:

src/inference/Inference.jl
torfjelde: Btw, is the `rng` actually used for anything in MCMCChains?

torfjelde: Actually, one thing we've forgot is that`transitions_from_chain` ought to have `rng`, right? Do you pass it to `evaluate` for the `Model`?

devmotion: Every time we run the model we use an RNG (if not specified, it's the default one). The current approach is to omit it when (we assume that) all samples are provided. However, maybe we should always use one explicitly since users could also make use of in Turing-external code (when they call an SDE solver e.g.)

torfjelde: > The current approach is to omit it when (we assume that) all samples are provided.

Not sure what you mean here :confused: I mean to ask if `model = ...; model(rng, ...)` was the way to go (which it seems like it is). One thing though, it doesn't seem like calling `model(rng)` works but `model(rng, varinfo)` does. Is that an issue or am I missing something?

torfjelde: > However, maybe we should always use one explicitly since users could also make use of in Turing-external code (when they call an SDE solver e.g.)

Inside the model you mean? If so, I agree!

devmotion: I mean in multiple implementions (e.g., when computing the elementwise loglikelihoods) we do not pass an explicit RNG since we assume that there is no randomness if all variables are provided.

> Is that an issue or am I missing something?

Yes, the dispatches in https://github.com/TuringLang/DynamicPPL.jl/blob/161f8200c5d608c481b58e056c2687a380f1c0d5/src/model.jl#L79-L92 should be changed (the issue can be fixed by just defining `(model::Model)(rng, varinfo = VarInfo(), ...)` instead, i.e., adding the default value to the method definition).

devmotion: > Inside the model you mean? If so, I agree!

Yep, inside the model.

torfjelde:
Cool now?

src/inference/Inference.jl
```diff
@@ -667,20 +667,30 @@ julia> sum(abs2, ys_test - ys_pred) ≤ 0.1
...
-        Distributions.GLOBAL_RNG,
-        model,
-        spl,
-        length(chain),
-        transitions,
-        MCMCChains.Chains
+    chain_result = reduce(
+        MCMCChains.chainscat, [
+            AbstractMCMC.bundle_samples(
+                Distributions.GLOBAL_RNG,
```
devmotion, line 686:
```suggestion
                rng,
```

torfjelde: Watt, how did the tests pass?

devmotion (approved):
LGTM, just bump the Turing version.

src/inference/Inference.jl
torfjelde: Ah, nvm I see why :+1:

torfjelde:
Es bueno :+1:

torfjelde:
I guess we don't have bors here.. That was awkward..

torfjelde:
> I guess we don't have bors here.. That was awkward..

I guess we _do_; doubly awkward!

devmotion:
Hehe I guess bors is enabled for TuringLang but the config file in Turing.jl is missing.

PR #1428, opened 2020-10-06
Update to AbstractMCMC 2

devmotion:
This is the Turing counterpart to https://github.com/TuringLang/DynamicPPL.jl/pull/150.

CI tests will fail as long as there is no new release of DynamicPPL with the changes from the linked PR, but tests pass locally with the branch in DynamicPPL.


cpfiffer (approved):

src/contrib/inference/dynamichmc.jl
```diff
@@ -41,82 +41,83 @@ function DynamicNUTS{AD}(space::Symbol...) where AD
...
-mutable struct DynamicNUTSState{V<:VarInfo, D} <: AbstractSamplerState
+struct DynamicNUTSState{V<:AbstractVarInfo,D}
     vi::V
     draws::Vector{D}
 end
 
 DynamicPPL.getspace(::DynamicNUTS{<:Any, space}) where {space} = space
 
-function AbstractMCMC.sample_init!(
+# initial step: general interface for resuming and
```
cpfiffer, line 51: Looks like this comment wasn't finished.

src/inference/Inference.jl
```diff
@@ -439,95 +352,54 @@ function AbstractMCMC.bundle_samples(
...
-        for v in values(params)
-            push!(vs, v[1])
-        end
-
-        push!(k, :lp)
-
-        nts[i] = NamedTuple{tuple(k...)}(tuple(vs..., getlogp(t)))
+    return map(ts) do t
+        params = map(first, getparams(t))
+        return merge(params, (lp = getlogp(t),))
```
cpfiffer, line 372: Couldn't we use `metadata(t)` here?

cpfiffer (approved):
Crap, I reviewed the test files in the DynamicPPL PR instead. I replicated my comments from there over to here -- sorry for the noise.

I've only got two minor comments.

src/inference/Inference.jl
devmotion: Sure, we can. I guess I didn't want to change the current behaviour but maybe it is actually better (and more consistent with the Chains output) to add the full metadata?

src/contrib/inference/dynamichmc.jl
devmotion:
```suggestion
```

src/inference/Inference.jl
cpfiffer: Yeah, it'll prevent future errors I think if we start to lean more heavily on `metadata` (which is correct, IMO).

```suggestion
        return merge(params, metadata(t))
```

torfjelde (commented):
From a brief look, this looks identical to the Turing-part in DynamicPPL, right?

devmotion:
> From a brief look, this looks identical to the Turing-part in DynamicPPL, right?

Yes, it is the same apart from the commits that I added yesterday and today (i.e. basically your suggestions). I'll update the PR when a new version of DynamicPPL is released.

torfjelde:
Cool, cool; I'll approve this once we have DPPL released then :+1:

devmotion:
One thing that should be addressed is the documentation of the sampling interface. Both https://turing.ml/dev/docs/for-developers/interface and https://turing.ml/dev/docs/for-developers/how_turing_implements_abstractmcmc are outdated. Moreover, they feel a bit redundant and do not belong to Turing - the sampling interface is defined in AbstractMCMC and explained in the README there, and the special implementation is mostly DynamicPPL-specific (and not Turing).

Actually I think it might be more appropriate to move these longer explanations to DynamicPPL and AbstractMCMC (of course, after updating them) and maybe linking to them from the Turing docs. It would also be easier to update them, to reduce redundancy, and would avoid the impression that all these things only work or are designed for Turing. I also don't think that anyone who studies the DynamicPPL or AbstractMCMC repos currently will notice these more detailed explanations.

devmotion:
Hmm seems there are some numerical inaccuracies on some platforms regarding the normalization of probability vectors of the Categorical distribution. Tests passed locally (Linux 64bit, Julia 1.5.3).

torfjelde:
> Actually I think it might be more appropriate to move these longer explanations to DynamicPPL and AbstractMCMC (of course, after updating them) and maybe linking to them from the Turing docs. It would also be easier to update them, to reduce redundancy, and would avoid the impression that all these things only work or are designed for Turing. I also don't think that anyone who studies the DynamicPPL or AbstractMCMC repos currently will notice these more detailed explanations.

Given how many packages we have at this point, it's really a hassle to have one big centralized documentation like we do now. IMO we ought to have separate project documentation for each package, with the possible exception of Turing.jl which is then hosted on turing.ml, and then we simply link (or if we really want to, re-render them) on Turing.jl. This significantly lowers the barrier to actually document the packages we're working on. But yeah, this is a larger issue and should be addressed outside of this PR of course :sweat_smile:

torfjelde (approved):
Given that this is the same as for DynamicPPL, with the exception of the last commit improving the testing of `GibbsConditional` (which I looked at, and it looks good), this looks good to me:)

PR #1431, opened 2020-10-07
Improve documentation and implementation of custom distributions

devmotion:
This PR adds proper docstrings to the included distributions, improves the implementation by implementing `sampler` and fixing some type instability issues, and prepares the code for Distributions 0.24 (which fixes the `logpdf`/`pdf` weirdness for univariate distributions).

I noticed that we currently can't support Distributions 0.24 since it requires https://github.com/TuringLang/DynamicPPL.jl/pull/150 (both AdvancedMH and DynamicPPL block the update, and the latest versions of AdvancedMH are only compatible with AbstractMCMC 2, so it has to be supported by DynamicPPL and Turing as well).

Probably similarly the random measures could be improved.

### DIFF
--- a/src/stdlib/distributions.jl
+++ b/src/stdlib/distributions.jl
@@ -1,138 +1,249 @@
-import Random: AbstractRNG
-
-# No info
 """
-    Flat <: ContinuousUnivariateDistribution
+    Flat()
 
-A distribution with support and density of one
-everywhere.
+The *flat distribution* is the improper distribution of real numbers that has the improper
+probability density function
+
+```math
+f(x) = 1.
+```
 """
 struct Flat <: ContinuousUnivariateDistribution end
 
-Distributions.rand(rng::AbstractRNG, d::Flat) = rand(rng)
-Distributions.logpdf(d::Flat, x::Real) = zero(x)
-Distributions.minimum(d::Flat) = -Inf
-Distributions.maximum(d::Flat) = +Inf
+Base.minimum(::Flat) = -Inf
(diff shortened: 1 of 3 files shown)


src/stdlib/distributions.jl
```diff
@@ -1,139 +1,250 @@
...
+```math
+P(X = k) = \\frac{e^{k \\cdot logλ}{k!} e^{-e^{logλ}}, \\quad \\text{ for } k = 0,1,2,\\ldots.
+```
+
+See also: [`Poisson`](@ref)
 """
-struct LogPoisson{T<:Real} <: DiscreteUnivariateDistribution
+struct LogPoisson{T<:Real,S} <: DiscreteUnivariateDistribution
     logλ::T
+    λ::S
```
cpfiffer, line 219: Is it worth adding an extra field here to remove the `exp` call later?

cpfiffer (approved):

src/stdlib/distributions.jl
devmotion: I don't have a strong opinion on this one. It will be (slightly) more efficient if you evaluate the logpdf at least twice or sample twice, and it won't be less efficient if you evaluate or sample at least once. I noticed that it is possible to cache it after having updated the BinomialLogit distribution (in this case the cached computations are definitely more expensive though).

cpfiffer: Alright, then I'm happy. Let's stick with this for now.

PR #1439, opened 2020-10-11
Clear RDCache for MLE and MAP estimation

### LINKED ISSUE #1418: rdcache not cleared while using MLE,MAP estimation
I was using MLE, MAP estimation with `:reversediff` backend and `rdcache`
The following is the MWE
```
using Turing,Distributions, Random
using ReverseDiff, Memoization, Optim
println("loaded modules")
Turing.setadbackend(:reversediff)
Turing.setrdcache(true)

@model function mwe(x)
    n = length(x)
    sigma ~ filldist(Uniform(0.5,1.5), n)
    TT = eltype(sigma)
    u ~ arraydist(truncated.(Normal.(0, sigma), TT(-4), TT(4)))
    x ~ MvNormal(u,1)
end

# generate some fake data
a = rand(2000)
model = mwe(a)

niter = 10
println("starting estimation.....")
for i in 1:niter
    optimize(model, MLE())
    # Turing.emptyrdcache()
    println("estimation done for $(i)")
end
wait()
```
I could see a significant rise in ram usage of my system per iteration (> 0.5GB)
I'm not sure if this is expected behavior. I have done serial sampling with `rdcache` enabled and I didn't notice any growth per every iteration.

An easy fix for RAM increase is to clear the `rdcache` every time using `Turing.emptyrdcache`

cpfiffer:
Fixes #1418. The memoization cache is now cleared after each MLE/MAP `optimize` call only if ReverseDiff & Memoization are loaded -- I used a dispatched function called `clearcache` that does nothing if the two named packages aren't loaded, but clears the cache if they are both loaded.

Not sure what the tests here should be?

### DIFF
--- a/src/core/ad.jl
+++ b/src/core/ad.jl
@@ -198,3 +198,7 @@ for F in (:link, :invlink)
         end
     end
 end
+
+# This function will call emptyrdcache only if ReverseDiff & Memoization are loaded
+# See Turing issue #1418
+clearcache(x=nothing) = ()
--- a/src/core/compat/reversediff.jl
+++ b/src/core/compat/reversediff.jl
@@ -50,6 +50,9 @@ end
         return
     end
 
+    # This function will call emptyrdcache only if ReverseDiff & Memoization are loaded
+    clearcache() = emptyrdcache()
+
     function gradient_logp(
         backend::ReverseDiffAD{true},
         θ::AbstractVector{<:Real},
--- a/src/modes/ModeEstimation.jl
+++ b/src/modes/ModeEstimation.jl
@@ -419,6 +419,9 @@ function _optimize(
     # Optimize!
     M = Optim.optimize(Optim.only_fgh!(f), init_vals, optimizer, options, args...; kwargs...)
 
+    # Clear the cache if Memoization & ReverseDiff are loaded.
+    Turing.Core.clearcache()
+
     # Warn the user if the optimization did not converge.
     if !Optim.converged(M)
         @warn "Optimization did not converge! You may need to correct your model or adjust the Optim parameters."


src/core/ad.jl
```diff
@@ -198,3 +198,7 @@ for F in (:link, :invlink)
         end
     end
 end
+
+# This function will call emptyrdcache only if ReverseDiff & Memoization are loaded
+# See Turing issue #1418
+clearcache(x=nothing) = ()
```
mohamed82008, line 204: why do we need an argument here?

cpfiffer: This was just a stupid trick to get the compiler to dispatch on the version I overwrite it with when we load ReverseDiff and Memoization. There's probably a better way to do this.

devmotion: In general, I don't think it's good to overwrite method definitions. IMO `clearcache` is not needed, instead one could add a check before calling `emptyrdcache()` such as
```julia
if isdefined(Turing, :emptyrdcache)
    emptyrdcache()
end
```

devmotion (commented):
IMO this is not an optimal solution since it will clear everything, i.e., also caches of other sampling procedures that a user might want to keep.

In general, it would be good to know why this is happening. I assume the problem is that a new `OptimLogDensity` is created every time a user calls `optimize`. It would be good to test if the problem is fixed by creating the `OptimLogDensity` object once and providing it explicitly every time.

If this is the case, it would be an additional argument for returning the objective function or an optimization problem (for GalacticOptim) instead of hacking into the `Optim.optimize` dispatches.

cpfiffer:
Or, perhaps we should just export `emptyrdcache` for users to track their own cache (not necessarily ideal). I'm wary of requiring people to create an `OptimLogDensity` themselves because they should be able to do whatever work they need with an instantiated model, rather than constructing miscellaneous intermediate types.

devmotion:
> I'm wary of requiring people to create an OptimLogDensity themselves because they should be able to do whatever work they need with an instantiated model, rather than constructing miscellaneous intermediate types.

Oh, I don't think users should deal with creating an OptimLogDensity directly. I think instead of calling `optimize` they should call a function that gives them something they can just throw at (Galactic)Optim - either an objective function or an optimization problem that takes care of all conversions from parameter space to VarInfo, contexts etc.

cpfiffer:
How would we handle wrapping it up for people? I.e., how could we serve the optimal values in an interperetable way, so that they don't have to deal with a vector of parameters themselves?

devmotion:
By returning an `OptimizationProblem`: https://github.com/sciml/galacticoptim.jl

devmotion:
By providing both the optimization function/problem and a function for mapping the results back to the intuitive/interpretable formulation, similar to e.g. [`functor`](https://github.com/FluxML/Functors.jl).

yebai:
@cpfiffer Could you give this PR another shot? It seems like low-hanging fruit and we should be able to merge it quickly once the comments from @devmotion are addressed.

cpfiffer:
I'm not really convinced that this PR is the fix -- I think I agree with David that we should export a more generally optimization function and allow users to handle their own cache clearing. Maybe we should close this PR?

yebai:
Closing it sounds good to me.

PR #1440, opened 2020-10-11
Fix documentation formatting

hsm207:
Fix indentation so that each element in the numbered list appears in a new line for readability


cpfiffer (approved):
Awesome, thanks!

PR #1448, opened 2020-10-28
A more general MAP/MLE optimisation interface

jonasmac16:
Starting this PR after a discussion with David and Cameron.

A first attempt in adding functions to create optimisation problems from a Turing model both in constrained and unconstrained space which can be passed to any Julia optimisation library such as `BlackBoxOptim.jl`. Also added a `GalacticOptim`-specific function which creates a `GalacticOptim.OptimisationFunction`. Added a series  of test for both implementation and the interaction with `GalacticOptim`.


src/Turing.jl
```diff
@@ -68,7 +68,12 @@ end
 
 @init @require Optim="429524aa-4258-5aef-a3af-852621145aeb" @eval begin
     include("modes/ModeEstimation.jl")
-    export MAP, MLE, optimize
+    export MAP, MLE, optimize, constrained, unconstrained, instantiate_optimisation_problem
+end
+
+@init @require GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577" @eval begin
+    include("modes/GalacticBridge.jl")
+    export instantiate_galacticoptim_function
```
devmotion, line 76: These optional dependencies should all be removed.

GalacticOptim supports Optim (by default), so there is no need for handling Optim separately anymore. Moreover, both OptimizationProblem and OptimizationFunction (of GalacticOptim) are actually defined in DiffEqBase, so we should just make DiffEqBase a proper dependency (and never load Optim or GalacticOptim).

IMO the names `constrained` and `unconstrained` are too general (and should use uppercase characters, following the usual convention in Julia). Maybe it should just be a type parameter of `MAP` and `MLE`.

src/modes/GalacticBridge.jl
```diff
@@ -0,0 +1,37 @@
...
+  end
+  
+  function instantiate_galacticoptim_function(model::DynamicPPL.Model, ::MLE , ::constrained)
+    obj, init, t = instantiate_optimisation_problem(model, MLE() , constrained())
+    
+    l(x,p) = obj(x)
+    f = GalacticOptim.OptimizationFunction(l, grad = (G,x,p) -> obj(G,x))
+  
+    return (f=f, init=init, transform = t)
+  end
```
devmotion, line 37:
```suggestion
  function instantiate_galacticoptim_function(model::DynamicPPL.Model, optimtype, constrained)
    obj, init, t = instantiate_optimisation_problem(model, optimtype, constrained)
    
    l(x,p) = obj(x)
    f = GalacticOptim.OptimizationFunction(l, grad = (G,x,p) -> obj(G,x))
  
    return (f=f, init=init, transform = t)
  end
```

src/modes/ModeEstimation.jl
```diff
@@ -17,6 +17,8 @@ import Printf
 struct MLE end
 struct MAP end
 
+struct constrained end
+struct unconstrained end
```
devmotion, line 21: Maybe rather make it a type parameter:
```suggestion
struct MLE{constrained} end
struct MAP{constrained} end

MLE() = MLE{false}()
MAP() = MAP{false}()
```
BTW I'm not sure if the default should actually be to optimize in the constrained space (so that models don't have to deal with linking etc.).

src/modes/ModeEstimation.jl
```diff
@@ -440,3 +442,190 @@ function _optimize(
...
+
+function transform!(f::OptimLogDensity)
+  spl = DynamicPPL.SampleFromPrior()
+
+  ## Check link status of vi in OptimLogDensity
+  linked = DynamicPPL.islinked(f.vi, spl) 
+
+  ## transform into constrained or unconstrained space depending on current state of vi
+  if !linked
+    f.vi[spl] = f.vi[DynamicPPL.SampleFromPrior()]
```
devmotion, line 479: This seems redundant:
```suggestion
```

src/modes/ModeEstimation.jl
```diff
@@ -440,3 +442,190 @@ function _optimize(
...
+    DynamicPPL.link!(f.vi, spl)
+  else
+    DynamicPPL.invlink!(f.vi, spl)
+  end
+
+  return nothing
+end
+
+
+function transform2constrained(par::AbstractArray, vi::DynamicPPL.VarInfo)
```
devmotion, line 489:
```suggestion
function transform2constrained(vi::DynamicPPL.VarInfo, par::AbstractArray)
```

src/modes/ModeEstimation.jl
```diff
@@ -440,3 +442,190 @@ function _optimize(
...
+    if array
+      return par_sor[tmp_idx]
+    else
+      return (; zip(keys(par),par_sor[par_idx])...)
+    end
+  end
+end
+
+
+function transform2constrained(res::Optim.MultivariateOptimizationResults, vi::DynamicPPL.VarInfo)
```
devmotion, line 576: I don't think we should implement any special methods for Optim result types. Users should just use GalacticOptim and transform the resulting optimum if they operated in the unconstrained space.

src/modes/ModeEstimation.jl
```diff
@@ -440,3 +442,190 @@ function _optimize(
...
+end
+
+
+
+function instantiate_optimisation_problem(model::DynamicPPL.Model, ::MAP , ::unconstrained)
+  obj = OptimLogDensity(model, OptimizationContext(DynamicPPL.DefaultContext()))
+
+  transform!(obj)
+
+  init(init_vals) = transform2unconstrained(init_vals, obj.vi)
```
devmotion, line 600: IMO the initialization should be kept separate from defining the cost function and the transformation.

src/modes/ModeEstimation.jl
```diff
@@ -440,3 +442,190 @@ function _optimize(
...
+    end
+  else
+    if array
+      return par_sor[tmp_idx]
+    else
+      return (; zip(keys(par),par_sor[par_idx])...)
+    end
+  end
+end
+
```
devmotion, line 574: IMO we should not implement this here (at least not for now). It feels like we should add better support for using NamedTuples in DynamicPPL instead (also for specifying initial values, outputs, etc.).
```suggestion
```

test/Project.toml
```diff
@@ -2,13 +2,15 @@
...
 AdvancedVI = "b5ca4192-6429-45e5-a2d9-87aec30a685c"
+BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
 CmdStan = "593b3428-ca2f-500c-ae53-031589ec8ddd"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DistributionsAD = "ced4e74d-a319-5a8a-b0ac-84af2272839c"
 DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
 DynamicPPL = "366bfd00-2699-11ea-058f-f148b4cae6d8"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
```
devmotion, line 13: Optim should be removed from the explicit dependencies.

test/modes/GalacticBridge.jl
```diff
@@ -0,0 +1,111 @@
+using Turing
+using Optim
```
devmotion, line 2: As mentioned above, I don't think Optim should be supported explicitly anymore. Instead users should just use GalacticOptim (whose default backend is Optim).
```suggestion
```

test/modes/ModeEstimation.jl
```diff
@@ -15,32 +15,90 @@ include(dir*"/test/test_utils/AllUtils.jl")
     @testset "MLE" begin
         Random.seed!(222)
         true_value = [0.0625, 1.75]
-        
+        true_value_named = (s=true_value[1],m=true_value[2])
```
devmotion, line 18: This whole test file should just be switched to GalacticOptim.

devmotion (commented):
Thanks for the PR and all your efforts!

I added some comments.

src/modes/ModeEstimation.jl
jonasmac16: How come you want to change the order of arguments? I had put it this way so an inplace ! version which mutating par follows naturally

src/modes/ModeEstimation.jl
jonasmac16: Makes sense. I will remove this

src/modes/ModeEstimation.jl
jonasmac16: It doesn't so much initialise but provides the user with transformation function (with the correct `vi`) together with the objective to transform the initial value into the correct space of the objective

src/modes/ModeEstimation.jl
devmotion: Ah my bad, I somehow thought it is called `transform2constrained!` here as well (and would mutate `vi`). BTW I don't think the in-place versions are particularly useful since they do not operate in-place actually - they still create an intermediate new array (with the out-of-place version) which is then copied. The usual approach when you want to have both is to implement an in-place version without any intermediate allocations, and then create a new separate output array first for the out-of-place method before calling the in-place implementation.

A different criticism here is that one has to provide `vi` at all. I imagined something like
```julia
struct TransformFunction{V}
  vi::V
end

function (t::TransformFunction)(p::AbstractArray)
    vi = t.vi
	...
end 
```
Then you also don't have to define the new closures in the instantiate_.... functions.

src/modes/ModeEstimation.jl
cpfiffer: For now we should set the default to the unconstrained space, and then we can change it to use the constrained space when it's a little easier to get bounds.

test/modes/GalacticBridge.jl
cpfiffer: Hm, I don't want to do this shift quite yet -- lots of people use the MLE/MAP functionality as it currently exists, so I don't want to break everyone's code without offering a nice way to use GalacticOptim instead. Maybe we could add some convenience functions for `optimize` that map to GalacticOptim?

jonasmac16: I agree, I have kept the `optimize` function in a separate file in the subsequent commit, which is conditionally loaded when `Optim.jl` is loaded.

src/modes/ModeEstimation.jl
jonasmac16: That sounds sensible. Will work that in.

cpfiffer:
I'm pleased in general with the format here. My only comment would be whether or not we could find a shorter name for `instantiate_galacticoptim_function` and `instantiate_optimisation_problem`? Maybe `galacticoptim_function` and `optim_problem`, respectively. I'd also prefer not to use the full word for optimize (optim instead) because the spelling varies regionally.

Also, the tests seem to be failing. It looks like there's an issue with the transformation because initial values are falling outside the parameter bounds.

Project.toml
```diff
@@ -8,6 +8,7 @@ AdvancedHMC = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
 AdvancedMH = "5b7e9947-ddc0-4b3f-9b55-0d8042f74170"
 AdvancedVI = "b5ca4192-6429-45e5-a2d9-87aec30a685c"
 Bijectors = "76274a88-744f-5084-9051-94815aaf08c4"
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
```
devmotion, line 11: This needs a compat bound

src/Turing.jl
```diff
@@ -66,11 +66,15 @@ using .Variational
...
 end
 
+include("modes/ModeEstimation.jl")
+using .ModeEstimation
+
 @init @require Optim="429524aa-4258-5aef-a3af-852621145aeb" @eval begin
-    include("modes/ModeEstimation.jl")
-    export MAP, MLE, optimize
+    include("modes/OptimInterface.jl")
+    export  optimize
```
devmotion, line 74:
```suggestion
    export optimize
```

src/Turing.jl
```diff
@@ -66,11 +66,15 @@ using .Variational
...
 end
 
+
```
devmotion, line 77:
```suggestion
```

test/Project.toml
```diff
@@ -2,13 +2,15 @@
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AdvancedMH = "5b7e9947-ddc0-4b3f-9b55-0d8042f74170"
 AdvancedVI = "b5ca4192-6429-45e5-a2d9-87aec30a685c"
+BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
```
devmotion, line 5: This requires a compat bound

test/Project.toml
devmotion: This also needs a compat bound.

jonasmac16:
@devmotion What exactly are Compat bounds? I will research myself over the coming days but can you give me a pointer. @cpfiffer Regarding the names, I agree they should be shorter and only use optim to avoid ambiguity of spelling. Also should there be a way to calculate the summary stats of a MAP/MLE (similiar to the current `optimize` implementation) obtained from any kind of optimiser interface, e.g. another entry of the tuple which is created when calling `instantiate_optimisation_problem`. Also, at the moment the function which transforms an initial value array only returns "one" sample from the prior but I think it might good if it sample a new initial value when you call `obj.init()`. What do you think? What is the best way achivieng with the curreent `Turing` and `DynamicPPL` interface?

devmotion:
Compat bounds specify which package versions are compatible with Turing: https://julialang.github.io/Pkg.jl/v1/compatibility/

yebai:
We're happy to review and merge this PR asap. @jonasmac16  Could you take a look at review comments from @devmotion and address them where possible?

devmotion:
Before merging this PR, DiffEqBase should be replaced with SciMLBase. We just moved the more general SciML stuff (such as for optimization problems) to the more lightweight package SciMLBase some days ago, DiffEqBase is a heavier dependency with common stuff for the DiffEq packages such as OrdinaryDiffEq and StochasticDiffEq.

jonasmac16:
> Before merging this PR, DiffEqBase should be replaced with SciMLBase. We just moved the more general SciML stuff (such as for optimization problems) to the more lightweight package SciMLBase some days ago, DiffEqBase is a heavier dependency with common stuff for the DiffEq packages such as OrdinaryDiffEq and StochasticDiffEq.

I am incorporating all the comments but the switch to `SciMLBase` as dependency seems to cause issues as `GalacticOptim` still depends on `DiffEqBase` and doesn't work with the `SciMLBase.OptimizationFunction`

devmotion:
Yeah, the change really just happened a day ago, so it is not propagated to the downstream packages yet. Currently DiffEqBase just reexports SciML such that downstream packages are not broken suddenly. But it should be fixed very soon, and so we should wait for it and not introduce a very heavy dependency on DiffEqBase here.

jonasmac16:
Made the changes except for the change to `SciMLBase` and will amend it once `GalacticOptim` is updated. Also, should there be some function which calculates the summary stats of a given point estimate similar to the current `Optim` specific implementation but optimization package agnostic?

jonasmac16:
I have finally gotten around to make the changes suggested eg SciMLBase. I have also made the functions name more general and concise. Furthermore, I have added some additional functions to create a `OptimizattionProblem` directly in Turing. Also, each functions which creates the optimisation problem in the constrained space automatically tries to retrieve the parameter bounds from the model.

jonasmac16:
@devmotion would you mind having a look if you are happy with the changes I have made. Sorry, I don't mean to push, if you saw the commits already.

jonasmac16:
Do you want me to change, remove or amend anything to/from the PR.

devmotion (changes requested):
Sorry, I completely missed the review request. I had a quick look at the first few files and added some comments.

However, my general impression is that the design is too complicated. I think it should be as simple as
```julia
prob, reconstruct = optimization_problem(model::Model, u0=nothing; constrained=Val(true), ad=Turing.ADBackend[], kwargs...)
```
where `u0` is sampled from the prior if not specified. Intuitively, I'd say users should provide all quantities such as initial values or bounds in the original space since it seems much more intuitive. I am not completely sure what would be the best representation that `reconstruct` would transform the optimized values to - either just values in the original space (i.e., `identity` for constrained sampling and a transformation otherwise) or a more intuitive representation such as a NamedTuple (or a VarInfo object that one obtains by rerunning the model with these values - but this could the users compute anyway?).

devmotion:
Regarding constrained/unconstrained values and optimization: maybe we should use https://github.com/invenia/ParameterHandling.jl and derive a set of variables with the deferred transformations from our model?

jonasmac16:
I have addressed hopefully all your requests except for the transformation of constrained/unconstrained values. The transformation is handled within `Turing` at the moment. What kind of workflow did you imaging using `ParameterHandling`. Something like `value(x, model)`?

jonasmac16:
@devmotion Would you be able to have a look at the recent changes? I hope I have addressed all your suggestions.

jonasmac16:
@devmotion Is this PR still of interest or are there maybe different efforts to link Turing to GalacticOptim which I might be able to help with

cpfiffer:
I think David's out on vacation, I'll take this one over and we'll see what we can do.

cpfiffer:
Handled some of the merge conflict, let's see what happens with the tests.

src/modes/ModeEstimation.jl
```diff
@@ -1,22 +1,30 @@
...
+using DynamicPPL: Model, AbstractContext, VarInfo, VarName,
     _getindex, getsym, getfield, settrans!,  setorder!,
-    get_and_set_val!, istrans
-import .Optim
-import .Optim: optimize
-import ..ForwardDiff
-import NamedArrays
-import StatsBase
-import Printf
+    get_and_set_val!, istrans, dot_tilde, get_vns_and_dist
```
cpfiffer, line 10:
```suggestion
    get_and_set_val!, istrans, get_vns_and_dist
```

src/modes/ModeEstimation.jl
```diff
@@ -1,22 +1,30 @@
...
+using DynamicPPL: Model, AbstractContext, VarInfo, VarName,
     _getindex, getsym, getfield, settrans!,  setorder!,
-    get_and_set_val!, istrans
-import .Optim
-import .Optim: optimize
-import ..ForwardDiff
-import NamedArrays
-import StatsBase
-import Printf
+    get_and_set_val!, istrans, get_vns_and_dist
```
cpfiffer, line 10:
```suggestion
    get_and_set_val!, istrans
```

cpfiffer:
Ugh now we're getting some weird CUDA error. I'll have to poke around this later.

jonasmac16:
Thank you for reviewing. Are the CUDA errors maybe linked to `GalacticOptim `.

cpfiffer:
Perhaps, though the stacktraces seem to suggest it's a Flux-type thing.

jonasmac16:
Hi, do you still want to go ahead with this PR? There seem to be some changes ModeEstimation.jl code.

yebai:
@jonasmac16 yes, I think there are some broken tests previously. Once fixed, we can fix the merge conflicts and get the PR merged.

jonasmac16:
I think I removed the culprit lines of code. There was second call to include "test_utils/AllUtils.jl", which meant that the constant `gdemo_models` would be redefined.

yebai:
Many thanks @jonasmac16 @cpfiffer and @devmotion!

PR #1449, opened 2020-11-09
Add missing Distributions import in quick-start.md

mrandri19:


devmotion:
Did you get an error without importing Distributions?

It should not be necessary to load Distributions since it is reexported by Turing in https://github.com/TuringLang/Turing.jl/blob/99ba1590dfbb121a38db55c8d9b5da8f2d4e22d4/src/Turing.jl#L15, i.e., it is loaded automatically when you load Turing.

mrandri19:
I think something is wrong with my dependencies as now I'm getting `@model` is not defined. After restarting the notebook kernel I get, but I have Turing installed
```
ERROR: LoadError: ArgumentError: Package MCMCChains [c7f686f2-ff18-58e9-bc7b-31028e88f75d] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```
 I'll close it for now, sorry for the trouble

PR #1487, opened 2020-12-03
Remove non-Turing specific AD tests

devmotion:
I noticed that some of the AD tests error on master. Maybe this is caused by some recent changes in FiniteDifferences, in combination with the specific seeds and tolerances used here. Regardless of what exactly caused it and if it could be fixed by adjusting the tolerances: In my opinion, the AD tests that do not use Turing at all should be removed.

The tests in DistributionsAD are designed to catch these errors, they allow to test each AD backend separately, and they allow to test a larger set of distributions.


cpfiffer (approved):
Works for me -- I have no problem with removing these tests. They should definitely be caught elsewhere.

PR #1492, opened 2020-12-17
Update EllipticalSliceSampling

devmotion:
This PR updates EllipticalSliceSampling to 0.4. The new version requires fewer user definitions (e.g., no need to define how to compute the proposal) and improves the cache handling (see https://github.com/TuringLang/EllipticalSliceSampling.jl/pull/13 for details).

In particular, it is not necessary anymore to define how to compute the proposal and more naturally we just define the prior and loglikelihood function for the Turing model and pass it to EllipticalSliceSampling.

### DIFF
--- a/src/inference/ess.jl
+++ b/src/inference/ess.jl
@@ -64,67 +64,82 @@ function AbstractMCMC.step(
     end
 
     # define previous sampler state
-    oldstate = EllipticalSliceSampling.ESSState(f, getlogp(vi))
+    # (do not use cache to avoid in-place sampling from prior)
+    oldstate = EllipticalSliceSampling.ESSState(f, getlogp(vi), nothing)
 
     # compute next state
-    _, state = AbstractMCMC.step(rng, ESSModel(model, spl, vi),
-                                 EllipticalSliceSampling.ESS(), oldstate)
+    sample, state = AbstractMCMC.step(
+        rng,
+        EllipticalSliceSampling.ESSModel(
+            ESSPrior(model, spl, vi), ESSLogLikelihood(model, spl, vi),
+        ),
+        EllipticalSliceSampling.ESS(),
+        oldstate,
+    )
 
     # update sample and log-likelihood
-    vi[spl] = state.sample
+    vi[spl] = sample
     setlogp!(vi, state.loglikelihood)
 
     return Transition(vi), vi
 end
 
-struct ESSModel{M<:Model,S<:Sampler{<:ESS},V<:AbstractVarInfo,T} <: AbstractMCMC.AbstractModel
+# Prior distribution of considered random variable
+struct ESSPrior{M<:Model,S<:Sampler{<:ESS},V<:AbstractVarInfo,T}
     model::M
-    spl::S
-    vi::V
+    sampler::S
+    varinfo::V
     μ::T
-end
-
-function ESSModel(model::Model, spl::Sampler{<:ESS}, vi::AbstractVarInfo)
-    vns = _getvns(vi, spl)
-    μ = mapreduce(vcat, vns[1]) do vn
-        dist = getdist(vi, vn)
-        vectorize(dist, mean(dist))
+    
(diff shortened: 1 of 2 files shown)


src/inference/ess.jl
```diff
@@ -64,67 +64,82 @@ function AbstractMCMC.step(
...
-    vi = model.vi
-    vi[spl] = f
-    model.model(vi, spl)
-    getlogp(vi)
+function (ℓ::ESSLogLikelihood)(f)
+    sampler = ℓ.sampler
+    varinfo = ℓ.varinfo
+    varinfo[sampler] = f
+    ℓ.model(varinfo, sampler)
+    return getlogp(varinfo)
```
yebai, line 142: It's slightly unclear: Does this include the `logprior` terms for model parameters?

src/inference/ess.jl
```diff
@@ -64,67 +64,82 @@ function AbstractMCMC.step(
...
 
-# sample from the prior
-function EllipticalSliceSampling.sample_prior(rng::Random.AbstractRNG, model::ESSModel)
-    spl = model.spl
-    vi = model.vi
-    vns = _getvns(vi, spl)
-    set_flag!(vi, vns[1][1], "del")
-    model.model(rng, vi, spl)
-    return vi[spl]
+function ESSPrior(model::Model, sampler::Sampler{<:ESS}, varinfo::AbstractVarInfo)
```
yebai, line 108: `ESSPrior`, `ESSLoglikdlihood`, and potentially `LogDensity` are probably useful for other samplers. Maybe we can consider to generalise them and put them in `AbstractPPL` in future PRs.

src/inference/ess.jl
```diff
@@ -64,67 +64,82 @@ function AbstractMCMC.step(
...
-        vectorize(dist, mean(dist))
+    
+    function ESSPrior{M,S,V}(model::M, sampler::S, varinfo::V) where {
+        M<:Model,S<:Sampler{<:ESS},V<:AbstractVarInfo
+    }
+        vns = _getvns(varinfo, sampler)
+        μ = mapreduce(vcat, vns[1]) do vn
+            dist = getdist(varinfo, vn)
+            EllipticalSliceSampling.isgaussian(typeof(dist)) ||
+                error("[ESS] only supports Gaussian prior distributions")
```
yebai, line 101: looks good!

src/inference/ess.jl
devmotion: The prior is a bit special since we want to check the distributions, have to define `EllipticalSliceSampling.isgaussian`, and have to extract the mean in the constructor. I don't think this can be generalized.

The loglikelihood functions and logdensity are implemented for other samplers as well, and also there exists `logprior`, `logjoint`, and `loglikelihood` implementations already in DynamicPPL (e.g., https://github.com/TuringLang/DynamicPPL.jl/blob/9d4280cf4b0a31e812c87acaad8a27d57b84db0e/src/model.jl#L212). However, in contrast to the implementations in DynamicPPL here we want to evaluate the log likelihood for a vectorized sample instead of a given `VarInfo` object. One could just use
```
loglik = let model=model, sampler=sampler, varinfo=varinfo
    x -> begin
        varinfo[sampler] = x
        return logjoint(model, sampler, varinfo)
    end
end
```
but it felt clearer to use a callable struct.

BTW we have to use `logjoint` here since we do not want to evaluate the loglikelihood for the ESS-sampled variable only but also require the prior terms for the other model paramters. The dispatches of `tilde`, `dot_tile`, etc. handle this scenario below.

src/inference/ess.jl
devmotion: See my comment below: we need a mix of `logjoint` and `loglikelihood` since we want the log joint probability minus the prior terms for the ESS-sampled variables. Regular `DefaultContext` and `LikelihoodContext` implementations are not sufficient here. This is handled by the `tilde` and `dot_tilde` implementations below.

yebai (approved):

PR #1497, opened 2020-12-22
Use iteration interface of DynamicHMC

devmotion:
This PR replaces https://github.com/TuringLang/Turing.jl/pull/1186. It makes use of the (inofficial) iteration interface of DynamicHMC to exploit more features of AbstractMCMC and enable Gibbs sampling.


src/contrib/inference/dynamichmc.jl
```diff
@@ -55,44 +64,41 @@ function DynamicPPL.initialstep(
...
-        link!(vi, spl)
-        l, dl = _lp(vi[spl])
-    end
-
-    if spl.selector.tag == :default && !islinked(vi, spl)
-        link!(vi, spl)
-        model(vi, spl)
+    # Ensure that initial sample is in unconstrained space.
+    if !DynamicPPL.islinked(vi, spl)
+        @debug "X -> R..."
```
cpfiffer, line 71: I feel like we should maybe add this to the definition of `link!` so that we don't have random idiosyncratic debug calls for some samplers and not others.

src/contrib/inference/dynamichmc.jl
```diff
@@ -55,44 +64,41 @@ function DynamicPPL.initialstep(
...
+    steps = DynamicHMC.mcmc_steps(results.sampling_logdensity, results.final_warmup_state)
+    Q, _ = DynamicHMC.mcmc_next_step(steps, results.final_warmup_state.Q)
+
+    # Update the variables.
+    vi[spl] = Q.q
+    DynamicPPL.setlogp!(vi, Q.ℓq)
+
+    # If a Gibbs component, transform the values back to the constrained space.
+    if spl.selector.tag !== :default
+        @debug "R -> X..."
```
cpfiffer, line 93: Same as above, maybe `invlink!` should have this debug call in it.

cpfiffer (approved):
Looks good to me. I left two comments that are more cosmetic, feel free to address them or not before merge.

src/contrib/inference/dynamichmc.jl
```diff
@@ -55,44 +64,41 @@ function DynamicPPL.initialstep(
...
-        model(vi, SampleFromUniform())
-        link!(vi, spl)
-        l, dl = _lp(vi[spl])
-    end
-
-    if spl.selector.tag == :default && !islinked(vi, spl)
-        link!(vi, spl)
-        model(vi, spl)
+    # Ensure that initial sample is in unconstrained space.
+    if !DynamicPPL.islinked(vi, spl)
```
torfjelde, line 70: AFAIK `link!` won't actually do anything if the variables are already transformed.

`islinked` and `link!` both perform the same check:
- `link!`: https://github.com/TuringLang/DynamicPPL.jl/blob/6ac3922bf09cf8e0d124b448353dd18b7afee00d/src/varinfo.jl#L701
- `islinked`: https://github.com/TuringLang/DynamicPPL.jl/blob/6ac3922bf09cf8e0d124b448353dd18b7afee00d/src/varinfo.jl#L797

I.e. `islinked` is redundant.

EDIT: Nvm, I see that there's a warning beeing logged if we're linking an already linked `vi` :+1:

src/contrib/inference/dynamichmc.jl
torfjelde: https://github.com/TuringLang/DynamicPPL.jl/pull/206#issue-544639845

torfjelde (approved):
Muy bueno :+1:

Issue #526, opened 2018-09-13
Debug statements etc


willtebbutt:
We have quite a lot of `@debug` statements floating around the code base, and I wonder how much additional value a lot of them provide. Debug messages like `@debug "run model..."` in a function called `runmodel` really don't provide any additional information over and above what a stack trace already provides, while cluttering the code.

I would be keen to understand whether
1. anyone feels that these debug messages provide value over and above what we get for free from stack traces and
2. there would be any objections to removing a lot of the debug messages.

yebai:
@willtebbutt That's a good point - many of these code are written when Julia's debugging environment isn't so mature.  Indeed, some `@debug` messages are a bit redundant and can be safely removed.

willtebbutt:
Great. I'll do this once #513 is merged.

PR #1499, opened 2020-12-23
Remove some `@debug` statements

devmotion:
As a follow-up to https://github.com/TuringLang/DynamicPPL.jl/pull/206, this PR removes some `@debug` statements that are redundant or do not provide much information. Basically, only the `@debug` statement that reports the current stepsize of HMC is kept.

Fixes https://github.com/TuringLang/Turing.jl/issues/526.


cpfiffer (approved):
I approve, all of these were totally useless debug calls.

PR #1500, opened 2020-12-23
Separate Gibbs handling

devmotion:
This PR is my take on a long-standing issue/discussion around Gibbs sampling. In my opinion, the `Gibbs` sampler should provide an interface for the updates of states, Hamiltonians etc. of the components such that sampler can easily opt-in and enable Gibbs sampling but do not have to adapt the implementation of their sampling steps.

So with this change, every sampler can expect that the provided state is accurate and uptodate. For Gibbs sampling, updates of the states should be performed in `gibbs_state`. The interface currently does not include a way to avoid recomputing the log probabilities required by the components. For some samplers these are the log joint probabilities but e.g. ESS uses neither the standard log joint probability nor the log likelihood and DynamicHMC also requires the gradient of the log probability, so probably it is not applicable for every sampler. I imagine the best way would be to check if `varinfo.logp` should be recomputed before updating the state in, e.g., a separate function such as `gibbs_rerun(prev_sampler, sampler)` which would return `true` by default. Then one would have to implement `gibbs_rerun` explicitly for all combinations of samplers for which this computation is not needed.


src/inference/gibbs.jl
```diff
@@ -83,11 +83,24 @@ metadata(t::GibbsTransition) = (lp = t.lp,)
 DynamicPPL.getlogp(t::GibbsTransition) = t.lp
 
 # extract varinfo object from state
-getvarinfo(state) = state.vi
-getvarinfo(state::AbstractVarInfo) = state
+"""
+    gibbs_varinfo(model, sampler, state)
+
+Return the variables corresponding to the current `state` of the Gibbs component `sampler`.
```
torfjelde, line 89: What is meant by "variables" in the above? Is it supposed to be `VarInfo`?

src/inference/gibbs.jl
```diff
@@ -83,11 +83,24 @@ metadata(t::GibbsTransition) = (lp = t.lp,)
...
 
-# update state with new varinfo object
-gibbs_update_state(state::AbstractVarInfo, varinfo::AbstractVarInfo) = varinfo
+"""
+    gibbs_state(model, sampler, state, varinfo)
+
+Return the state of the Gibbs component with updated variables `varinfo`.
+
+The Gibbs component samples from the `model` using the provided `sampler` and its current
+state is `state`.
```
torfjelde, line 101:
```suggestion
The Gibbs component samples from the `model` using the provided `sampler` with the returned state.
```
?

torfjelde (commented):
Not certain I file comfortable with my own familiarity with the impl of the Gibbs sampler to approve, so for now I've just left some comments.

src/inference/gibbs.jl
```diff
@@ -162,13 +175,13 @@ function AbstractMCMC.step(
...
         # Update state of current sampler with updated `VarInfo` object.
-        current_state = gibbs_update_state(_state, vi)
+        current_state = gibbs_state(model, _sampler, _state, vi)
 
         # Step through the local sampler.
         _, newstate = AbstractMCMC.step(rng, model, _sampler, current_state; kwargs...)
 
         # Update `VarInfo` object.
-        vi = getvarinfo(newstate)
+        vi = gibbs_varinfo(model, _sampler, _state)
```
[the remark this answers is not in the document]
devmotion: It should be `newstate`, of course.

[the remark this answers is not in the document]
devmotion: 1. 
```suggestion
        vi = gibbs_varinfo(model, _sampler, newstate)
```

src/inference/hmc.jl
```diff
@@ -18,8 +18,25 @@ struct HMCState{
 end
 
 # TODO: Include recompute Hamiltonian here?
-function gibbs_update_state(state::HMCState, varinfo::AbstractVarInfo)
-    return HMCState(varinfo, state.i, state.traj, state.hamiltonian, state.z, state.adaptor)
+function gibbs_state(
```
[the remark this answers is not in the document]
devmotion: Hmm, why would this be needed? To me it seemed all `link!`/`invlink!` statements should be removed. No sampler should modify any variables sampled by another sampler, so it should not be needed to transform them. The only thing that is (sometimes) needed is to reevaluate `vi.logp`, isn't it?

src/inference/gibbs.jl
devmotion: I wanted to say that the current state of the sampler, as computed in the previous iteration, is `state`. However, the function updates it by taking into account the new variables sampled by the other samplers.

src/inference/gibbs.jl
devmotion: Yes, maybe that should be made more explicit.

src/inference/hmc.jl
```diff
@@ -18,8 +18,25 @@ struct HMCState{
 end
 
 # TODO: Include recompute Hamiltonian here?
```
devmotion, line 20:
```suggestion
# Update state in Gibbs sampling
```

src/inference/hmc.jl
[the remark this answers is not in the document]
torfjelde: > No sampler should modify any variables sampled by another sampler

Ah, yeah that makes sense :+1:

src/inference/gibbs.jl
torfjelde: Hmm, how about:
```julia
Computes the new state, taking into account the new variables sampled by other Gibbs components.

# Arguments
- `model`: model targeted by the samplers.
- `sampler`: the sampler for this Gibbs component.
- `state`: the current state of `sampler`, as computed in the previous iteration.
- `varinfo`: `VarInfo` for this sampler.

```

src/inference/gibbs.jl
```diff
@@ -83,11 +88,69 @@ metadata(t::GibbsTransition) = (lp = t.lp,)
...
+varinfo(state) = state.vi
+varinfo(state::AbstractVarInfo) = state
+
+"""
+    gibbs_state(model, sampler, state, varinfo)
+
+Return the state of the Gibbs component with updated variables `varinfo`.
+
+The Gibbs component samples from the `model` using the provided `sampler` and its current
+state is `state`.
```
devmotion, line 106: @torfjelde I slightly modified your suggestion:
```suggestion
Return an updated state, taking into account the variables sampled by other Gibbs components.

# Arguments
- `model`: model targeted by the Gibbs sampler.
- `sampler`: the sampler for this Gibbs component.
- `state`: the state of `sampler` computed in the previous iteration.
- `varinfo`: the variables, including the ones sampled by other Gibbs components.
```
What do you think?

src/inference/gibbs.jl
```diff
@@ -10,6 +10,11 @@ Determine whether algorithm `alg` is allowed as a Gibbs component.
 """
 isgibbscomponent(alg) = false
 
+isgibbscomponent(::ESS) = true
+isgibbscomponent(::GibbsConditional) = true
+isgibbscomponent(::Hamiltonian) = true
```
devmotion, line 15: I am wondering if we should be more restrictive here, to avoid any surprises for someone who implements another `Hamiltonian` algorithm but does not define `gibbs_state`, i.e., whose implementation is not "Gibbs-ready".

phipsgabler: I agree.  And beyond that: since the "Gibbs-readyness" is a trait implementation, why move most of the code away from the individual samplers, and into this file?  Especially since we are moving away from priviledged implementations here.

IMHO things are clearer to look up when the Gibbs-parts are also near them.  E.g., from the perspective of someone looking up how to implement a Gibbs component.

devmotion: > why move most of the code away from the individual samplers, and into this file?

I agree that this would be better, and in fact it was the case initially in this PR. However, I noticed that it does not work for `gibbs_rerun` - Turing can't be compiled anymore if `gibbs_rerun` is defined next to the corresponding sampler due to the dependency on two samplers, e.g., `gibbs_rerun(::MH, ::Hamiltonian)` requires `MH` to be defined before Hamiltonian and similarly `gibbs_rerun(::Hamiltonian, ::MH)` requires `Hamiltonian` to be defined before `MH`. Therefore I decided it is more consistent to move all Gibbs parts for the Turing-internal samplers to the `Gibbs` implementation - otherwise some traits would be implemented in `gibbs.jl` and some in the files of the corresponding samplers. Only DynamicNUTS is an exception here and treated like a user-defined sampler that has to bundle also the Gibbs parts since it is loaded optionally if DynamicHMC is available.

phipsgabler: Uh, I see.  An unfortunate situation.  But how does the two-argument `gibbs_rerun` generalize between user-defined samplers?  

In a way this makes me more sympathetic to the old way.  Was the meaning of `selector.rerun` exactly "the logp has not been calculated", or was there more logic to it?

devmotion: There's much more to it: `vi.logp` means different things for different samplers (sometimes it is the joint log probability, sometimes it is the likelihood, sometimes it is something in between, and sometimes it is something completely different), some samplers just don't care about the value of `vi.logp` in their sampling step or reset everything to zero, etc.

Therefore I think the only reasonable way to define `rerun` (regardless how it is saved or implemented) is to compare two concrete samplers. The default setting of `gibbs_rerun(prev_sampler, sampler) = true` makes sure that every user implementation is always mathematically correct, and then users have to explicitly define if one can save one computation for a specific previous sampler by implementing `gibbs_rerun` for this specific sampler. The current implementation is just wrong, to be honest, since it ignores all these facts - however, I assume it does not lead to any problems since `vi.logp` is basically always recomputed, apart from `MH` with a previous sampler that is neither `ESS`, `PG`, nor `GibbsConditional`. :shrug:

src/inference/gibbs.jl
cpfiffer: I like it.

cpfiffer (approved):
I see that there's some outstanding discussions on this one, but in general I'm quite fond of this PR and would like to see it merged over. I think many of the choices David made here are reasonable and in some cases massively preferred (`gibbs_rerun`, for example).

I'll approve this for now, but I wonder if @torfjelde or @phipsgabler have updates to their comments.

phipsgabler (approved):

phipsgabler:
I still think that `gibbs_rerun` could impact extensibility sometime, but I cannot think of anything better than what David argued.  It only concerns an optimization, anyway.

src/inference/gibbs.jl
torfjelde: Sorry, this went under my radar: that looks good!

torfjelde (approved):
Other than the just committing that docstring suggestion by @devmotion, I'm happy!

yebai:
I'll review this PR soon - sorry for the delay!

devmotion:
Bump :slightly_smiling_face:

src/inference/ess.jl
```diff
@@ -58,11 +56,6 @@ function AbstractMCMC.step(
     # obtain previous sample
     f = vi[spl]
 
-    # recompute log-likelihood in logp
-    if spl.selector.tag !== :default
-        model(rng, vi, spl)
-    end
```
yebai, line 64: The `spl.selector` field was intended to store the assignment of model parameters to samplers in the Gibbs framework. It was meant to be more efficient than just using the `space` field of Gibbs component samplers since it avoids the model parameters name matching step. It might have other undocumented uses (e.g. determining whether parameters are transformed), so removing this mechanism requires some caution. Is there any reason why we want to remove this mechanism? 

cc @xukai92

src/inference/gibbs.jl
```diff
@@ -83,11 +88,72 @@ metadata(t::GibbsTransition) = (lp = t.lp,)
...
 # extract varinfo object from state
-getvarinfo(state) = state.vi
-getvarinfo(state::AbstractVarInfo) = state
+"""
+    gibbs_varinfo(model, sampler, state)
+
+Return the variables corresponding to the current `state` of the Gibbs component `sampler`.
+"""
+gibbs_varinfo(model, sampler, state) = varinfo(state)
+varinfo(state) = state.vi
```
yebai, line 97: Maybe consider introducing an `AbstractMCMCState` type to avoid potential misuses of `varinfo`? It also improves clarity I think. In simple cases, samplers can use a generic `MCMCState`, wherein more complicated ones, a sampler-specific `MCMCState` such as `HMCState` can be defined.

src/inference/gibbs.jl
```diff
@@ -99,41 +165,24 @@ function DynamicPPL.initialstep(
...
-        else
-            prev_alg = algs[i-1]
-        end
-        rerun = !isa(alg, MH) || prev_alg isa PG || prev_alg isa ESS ||
-            prev_alg isa GibbsConditional
-        selector = Selector(Symbol(typeof(alg)), rerun)
-        Sampler(alg, model, selector)
-    end
-
-    # Add Gibbs to gids for all variables.
```
yebai, line 116: See also my comment related to `spl.selector`.

yebai (commented):
Hi @devmotion, sorry for the delay - I was busy after Christmas for some teaching responsibilities.  

Overall, it makes sense to introduce a more rigorous mechanism for deciding when to re-run a model. However, I wonder whether it is necessary to be as strict as implemented here. For example, I would expect the previous sampler to fully determine whether a re-run a model in most cases. If that's the case, maybe we can consider storing the re-run flag as part of MCMC states (e.g. `HMCState`)? Then we would avoid defining `gibbs_rerun` methods for all combinations of samplers, and also avoid the issue to moving Gibbs related code into a single file mentioned in [here](https://github.com/TuringLang/Turing.jl/pull/1500/files#r549742861).

Another potential (internal) breaking change is the removal of `spl.selector`. It was introduced during the previous round of refactoring, but didn't get fully supported by various samplers. However, it was meant to be a more flexible and efficient mechanism than matching model parameters using the `space` field of samplers.

src/inference/ess.jl
devmotion: As far as I remember, currently `spl.selector` is only used to determine if we perform Gibbs sampling or not and rerun the model, i.e., repopulate `vi.logp`, if we do. Since this PR "cleans" the sampling state and moves the Gibbs-specific logic to the Gibbs sampler and `gibbs_rerun`, there is no need for it anymore.

devmotion:
> Overall, it makes sense to introduce a more rigorous mechanism for deciding when to re-run a model. However, I wonder whether it is necessary to be as strict as implemented here. For example, I would expect the previous sampler to fully determine whether to re-run a model in most cases.

As far as I can tell, usually we have to know both the previous and the current sampler to determine if it is necessary to rerun the model, i.e., to repopulate `vi.logp`. E.g., some samplers such as PG always reset `vi.logp`, regardless of the previous sampler, and hence the model does not have to be rerun - however, for other samplers (and the same previous sampler) one has to repopulate `vi.logp`.

@phipsgabler also asked about the two-argument `gibbs_rerun` and why it is necessary, I tried to  while ago (https://github.com/TuringLang/Turing.jl/pull/1500#discussion_r549788223) explains it in more detail:

> vi.logp means different things for different samplers (sometimes it is the joint log probability, sometimes it is the likelihood, sometimes it is something in between, and sometimes it is something completely different), some samplers just don't care about the value of vi.logp in their sampling step or reset everything to zero, etc.
>
> Therefore I think the only reasonable way to define rerun (regardless how it is saved or implemented) is to compare two concrete samplers. The default setting of gibbs_rerun(prev_sampler, sampler) = true makes sure that every user implementation is always mathematically correct, and then users have to explicitly define if one can save one computation for a specific previous sampler by implementing gibbs_rerun for this specific sampler. The current implementation is just wrong, to be honest, since it ignores all these facts - however, I assume it does not lead to any problems since vi.logp is basically always recomputed, apart from MH with a previous sampler that is neither ESS, PG, nor GibbsConditional. shrug

src/inference/ess.jl
devmotion: So it's not the main purpose of this PR to remove `spl.selector`, rather it disappears automatically when the more fine-tuned (and correct) two-argument logic of `gibbs_rerun` is introduced.

yebai:
thanks, @devmotion for the clarifications. It's now clear to me why we need the two-argument implementation of `gibbs_rerun`. I am happy to merge this PR as is except that 

- Consider keeping `updategid!` in this PR, and refactor/remove it in separate PRs. I think it's sensible to consider refactoring the variable assignment in Gibbs samplers, but feels it's better to do it in a separate effort to avoid undesirable effects. 

Regarding `gibbs_rerun(prev_sampler, sampler)`, I think it is possible to consider to make it a bit more modular (in future PRs), e.g. by introducing an additional sampler-specific attribute: 

```julia
gibbs_attribute (sampler) = one of a few possible attributes 
gibbs_rerun(prev_sampler, sampler) = gibbs_rerun(gibbs_attribute (prev_sampler), gibbs_attribute (sampler))
```

Such designs allow us to reuse some `gibbs_rerun` definitions and potentially avoiding the need to gathering sampler specific implementation for Gibbs into a single file.

devmotion:
I reverted the `gid` related changes, these parts of the PR could go into a future PR if desired.

The more modular approach sounds very reasonable. Unfortunately, it seems it would still require to define dispatches for all existing attributes when an algorithm does not fit into any existing attribute categories and has to define its own new attribute. One could still group together e.g. `MH`, `HMC`, and `NUTS` it seems :slightly_smiling_face:

devmotion:
Tests on master, and in the same way in this PR, fail terribly currently :open_mouth: It seems not a single AdvancedHMC-related test passes and the logs (e.g. https://pipelines.actions.githubusercontent.com/N6hMANmENxs4asRtCYxRx2T5dwtGFqB7REAP98WHa7yPx1CJ2q/_apis/pipelines/1/runs/13034/signedlogcontent/4?urlExpires=2021-01-28T02%3A04%3A15.5098764Z&urlSigningMethod=HMACV1&urlSignature=vljLf7fo15GLGBuVX4H6SoFlm2E%2FpDlq8ilNYVpd%2BRQ%3D) contain just 6 hours of repeated warnings of AdvancedHMC about numerical errors and rejected proposals. I have no idea why this happens.

yebai (approved):

yebai:
I'm running the tests locally ATM - will merge if it passes.

devmotion:
The test errors are real and consistent. You can already see them in the tests of the PRs that you just merged into the master branch.

yebai:
> The test errors are real and consistent. You can already see them in the tests of the PRs that you just merged into the master branch.

It's a bit strange - I just checked `AdvancedHMC`, we don't have any new release since 2 Dec 2020.

devmotion:
I wonder if it is related to the test errors that suddenly appeared in DistributionsAD. I still haven't figured out what package causes them. After hours of down- and upgrading packages I finally managed to reproduce the successful run in https://github.com/TuringLang/DistributionsAD.jl/runs/1659751199. But I still don't know what upstream change broke it.

yebai:
It seems all AD backends have test errors including `ForwardDiff`, `Tracker` and `ReverseDiff`, which hints the error might come from a shared dependency of all AD implementations. 

```julia
Test Summary:                          | Pass  Error  Total
Turing                                 |    4     13     17
  core                                 |    4      1      5
    ad.jl                              |    4             4
  inference: forwarddiff               |           1      1
    samplers                           |           1      1
  variational algorithms : forwarddiff |           1      1
  modes                                |           1      1
  inference: tracker                   |           1      1
    samplers                           |           1      1
  variational algorithms : tracker     |           1      1
  modes                                |           1      1
  inference: reversediff               |           1      1
    samplers                           |           1      1
  variational algorithms : reversediff |           1      1
  modes                                |           1      1
  variational optimisers               |           1      1
  stdlib                               |           1      1
  utilities                            |           1      1
ERROR: LoadError: Some tests did not pass: 4 passed, 0 failed, 13 errored, 0 broken.


```

devmotion:
Yeah one of my hypotheses currently is that there's a problem with the latest ForwardDiff release (which is used e.g. by Tracker and Zygote for broadcasting) but I don't know why it would cause any problems... It's just that I don't have tried so many other things already.

devmotion:
Hmm... But now I can reproduce the test failure also by just upgrading Distributions from 0.24.10 to 0.24.11...

yebai:
> Hmm... But now I can reproduce the test failure also by just upgrading Distributions from 0.24.10 to 0.24.11...

It might make sense to add `DistributionsAD` as a downstream/reverse test for `Distributions` given how closely interconnected they are.

devmotion:
I am confused how the Dirichlet changes in Distributions can break DistributionsAD though since we should always use DistributionsAD.TuringDirichlet instead. Something seems wrong.

devmotion:
https://github.com/TuringLang/DistributionsAD.jl/pull/151

yebai:
Excellent work, many thanks @devmotion!

PR #1502, opened 2020-12-24
Update SGHMC and SGLD

devmotion:


torfjelde:
I'll do a proper review soon, but a quick qesh while we wait: would it be possible to make proper stochastic algorithms, i.e. do data-subsampling? 

Either by introducing a way to filter `VarName` in DynamicPPL using a context, or somehow extracting the observations (maybe something similar to what we do with the `GetLikelihoodContext` or whatever it's called) and then just update in place. Thoughts? Or should we maybe make a separate issue over at DynamicPPL and do that work in a separate PR?

devmotion:
>  would it be possible to make proper stochastic algorithms, i.e. do data-subsampling?

I thought about this and was surprised that it was implemented but I didn't include it here since it was not part of the old implementation and it requires some additional thought and discussion which I assumed would be better to have in a separate PR or issue.

Mainly I thought about two approaches, both based on a special context and some special `observe` implementations to handle multiple observations in one `~` expression. One could try to extract the number of observations by running the fully sampled model once and then use the context to only sample one index (or a batch of indices) in every run and only include the corresponding observations. However, I assume that stochastic control flow would break this approach. As an alternative, one could instead specify the probability of an observation to be included (similar to dropout) and just decide randomly in each `observe` call if a (subset of) observation(s) will be included. This does not require a fixed number of observations and should work for all kind of models but there is a non-zero probability that all observations would be dropped - in this case it is not possible to normalize the (gradient of the) log probability with the number of considered observations.

torfjelde:
I made a separate issue for this discussion: https://github.com/TuringLang/DynamicPPL.jl/issues/208

cpfiffer (approved):
I love it! Thanks for updating these.

devmotion:
OK, let's address the subsampling issue in a separate PR!

PR #1527, opened 2021-01-26
Fix dev docs for VarName

phipsgabler:
Parts of this seem to have been written before `indexing` started using tuples.

### DIFF
--- a/docs/src/for-developers/compiler.md
+++ b/docs/src/for-developers/compiler.md
@@ -184,10 +184,13 @@ without any default values. Finally, in the new function body a `model::Model` w
 
 # `VarName`
 
-In order to track random variables in the sampling process, `Turing` uses the struct `VarName{sym}` which acts as a random variable identifier generated at runtime. The `VarName` of a random variable is generated from the expression on the LHS of a `~` statement when the symbol on the LHS is in `P`. Every `vn::VarName{sym}` has a symbol `sym` which is the symbol of the Julia variable in the model that the random variable belongs to. For example, `x[1] ~ Normal()` will generate an instance of `VarName{:x}` assuming `x` is in `P`. Every `vn::VarName` also has a field `indexing` which stores the indices requires to access the random variable from the Julia variable indicated by `sym`. For example, `x[1] ~ Normal()` will generate a `vn::VarName{:x}` with `vn.indexing == "[1]"`. `VarName` also supports hierarchical arrays and range indexing. Some more examples:
-- `x[1] ~ Normal()` will generate a `VarName{:x}` with `indexing == "[1]"`.
-- `x[:,1] ~ MvNormal(zeros(2))` will generate a `VarName{:x}` with `indexing == "[Colon(),1]"`.
-- `x[:,1][2] ~ Normal()` will generate a `VarName{:x}` with `indexing == "[Colon(),1][2]"`.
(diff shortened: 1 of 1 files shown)


docs/src/for-developers/compiler.md
```diff
@@ -184,10 +184,13 @@ without any default values. Finally, in the new function body a `model::Model` w
...
-- `x[1] ~ Normal()` will generate a `VarName{:x}` with `indexing == "[1]"`.
-- `x[:,1] ~ MvNormal(zeros(2))` will generate a `VarName{:x}` with `indexing == "[Colon(),1]"`.
-- `x[:,1][2] ~ Normal()` will generate a `VarName{:x}` with `indexing == "[Colon(),1][2]"`.
+In order to track random variables in the sampling process, `Turing` uses the struct `VarName{sym}` which acts as a random variable identifier generated at runtime. The `VarName` of a random variable is generated from the expression on the LHS of a `~` statement when the symbol on the LHS is in `P`. Every `vn::VarName{sym}` has a symbol `sym` which is the symbol of the Julia variable in the model that the random variable belongs to. For example, `x[1] ~ Normal()` will generate an instance of `VarName{:x}` assuming `x` is in `P`. Every `vn::VarName` also has a field `indexing`, which stores the indices requires to access the random variable from the Julia variable indicated by `sym` as a tuple of tuples.  Each element of the tuple thereby contains the indices of one indexing operation (`VarName` also supports hierarchical arrays and range indexing). Some examples:
```
devmotion, line 187: Minor suggestions, mainly since the changes to `VarName` also introduced an additional type parameter (hence `VarName{:x}` is an abstract supertype rather than the type of a `VarName` instance):

```suggestion
In order to track random variables in the sampling process, `Turing` uses the `VarName` struct which acts as a random variable identifier generated at runtime. The `VarName` of a random variable is generated from the expression on the LHS of a `~` statement when the symbol on the LHS is in the set `P` of unobserved random variables. Every `VarName` instance has a type parameter `sym` which is the symbol of the Julia variable in the model that the random variable belongs to. For example, `x[1] ~ Normal()` will generate an instance of `VarName{:x}` assuming `x` is an unobserved random variable. Every `VarName` also has a field `indexing`, which stores the indices required to access the random variable from the Julia variable indicated by `sym` as a tuple of tuples.  Each element of the tuple thereby contains the indices of one indexing operation (`VarName` also supports hierarchical arrays and range indexing). Some examples:
```

docs/src/for-developers/compiler.md
```diff
@@ -184,10 +184,13 @@ without any default values. Finally, in the new function body a `model::Model` w
...
+- `x ~ Normal()` will generate a `VarName{:x}` with `indexing == ()`.
+- `x[1] ~ Normal()` will generate a `VarName{:x}` with `indexing == ((1,),)`.
+- `x[:,1] ~ MvNormal(zeros(2))` will generate a `VarName{:x}` with `indexing == ((Colon(), 1),)`.
+- `x[:,1][2] ~ Normal()` will generate a `VarName{:x}` with `indexing == ((Colon(), 1), (2,))`.
```
devmotion, line 191: I'm wondering if this should be changed to account for the fact that `VarName{:x}` is an abstract supertype rather than the type of the generated `VarName` object.

devmotion: Maybe one should explain `VarName` more with the constructor `VarName{:x}(indexing)` than the type and the `indexing` field?

phipsgabler: :+1:  And I'll do the same thing in the docstring as well.

devmotion: I was a bit surprised but I just noticed that the constructor is actually `VarName(:x, indexing)` (or `VarName{:x,typeof(indexing)}(indexing)`).

phipsgabler: Yep, same here.  I'll try to find all the mistaken instances.

yebai (approved):

yebai:
Looks good - thanks @phipsgabler!

devmotion:
@phipsgabler bors is enabled for Turing.jl but not configured. So far we have only used it in DynamicPPL, in all other repos PRs are just merged manually.

yebai:
I'm merging this PR now since it's a pure doc improvement - the failed CI tests seems not caused by this PR.

PR #1545, opened 2021-02-15
`vi` should not work with Matrixvariate distributions

torfjelde:
## Issue
https://github.com/TuringLang/Bijectors.jl/issues/169

This happens because `Bijectors.Stacked` is only being compatible with 0- and 1-dimensional inputs. To "fix" this I've added a `Vec` bijector which wraps any higher-than-1-dimensional bijector with a "flattening" operation. 

## Now works
```julia
julia> using Turing

julia> @model function demo(xs)
           Σ ~ LKJ(size(xs, 1), 1)

           for i = 1:size(xs, 2)
               xs[:, i] ~ MvNormal(Σ)
           end

           return Σ
       end;

julia> m = demo(randn(2, 100));

julia> # This now works
       q = Turing.Variational.meanfield(m);

julia> rand(q)
4-element Array{Float64,1}:
  1.0
 -0.43345701408941034
 -0.43345701408941034
  1.0000000000000002

julia> advi = ADVI(1, 1000)
ADVI{AdvancedVI.ForwardDiffAD{40}}(1, 1000)

julia> q = vi(m, advi);
┌ Info: [ADVI] Should only be seen once: optimizer created for θ
└   objectid(θ) = 0x8016f7ba5cf582c2
[ADVI] Optimizing...100% Time: 0:00:03

julia> rand(q)
4-element Array{Float64,1}:
 1.0
 0.07902135820833789
 0.07902135820833789
 1.0
```

## Questions
- [ ] I'm a bit split on the `Vec` bijector as it's a bit "hacky". Ideally we'd be able to specify both input- and output-dimensionality (in full generality, ideally _spaces_ rather than just dimensionality, but this seems non-trivial), but we did not do this from the get-go due to additional complexity for both user and implementer. So should we keep `Vec` here as a way of only fixing this particular issue _or_ should I move it to Bijectors.jl so it's available there to? I'm slightly in favour of the last option.


yebai:
@torfjelde is this related to the more general `batching`-issue?

yebai (approved):

yebai:
CI is passing after using the most recent Bijector release. Let's merge this PR for now since it's taking quite a bit of time already; the code only affects `advi`. Let's make further improvements in new PRs.

Issue #1556, opened 2021-03-11
RWMH inside Gibbs sampling


julianjohs:
Dear all, 

I wrote a model which would use multiple RWMH samplers inside a Gibbs sampler. See the example below. 
It seems like the MH() function inside the Gibbs sampler would not use RWMH even though I supplied a covariance matrix. 
Am I making any mistake? Thanks in advance. 

In the picture, I used a very small covariance matrix to see if the chain would actually only make small steps. That seems to be the case for chn2 (see below), but not chn. 

```
# reproducible example

using Random, Distributions, Turing
using Plots, LinearAlgebra

# generate data
x = rand(Normal(5, 10), 200)
y = rand(LogNormal(-3, 2), 200)

# Turing model
@model function twomeans(x, y)
    # Set Priors
    μ ~ MvNormal(2, 3)
    σ ~ filldist(Exponential(1), 2)

    # Distributions of supplied data
    x .~ Normal(μ[1], σ[1])
    y .~ LogNormal(μ[2], σ[2])

end
mod = twomeans(x, y)

# generate covariance matrix for RWMH
# with small-valued VC matrix to check if we only see very small steps
vc_μ = convert(Array, 1e-2*I(2))
vc_σ = convert(Array, 1e-2*I(2))

# Sample via Gibbs: RWMH doesn't work
chn = sample(mod,
    Gibbs(
        MH(:μ, vc_μ),
        MH(:σ, vc_σ),
        ), 3_000 # draws
    )
#save values to df
chn_df = DataFrame(chn)

# This works
# define new VC matrix
vc_total = convert(Array, 1e-2*I(4))
# and sample from only one RWMH
chn2 = sample(mod, MH(vc_total), 3_000)

chn2_df = DataFrame(chn2)

# see that the steps from chn are very large - can't be the result of the supplied VC matrix!
plot(chn_df.iteration, hcat(chn_df."μ[1]", chn2_df."μ[1]"), label=["Gibbs-MH" "RWMH over all parameters"])

```

Thanks in advance!

cpfiffer:
Yeah, this is a bug here:
https://github.com/TuringLang/Turing.jl/blob/a4d277855b760d63f9cbbb27889f11f5f20cd25e/src/inference/mh.jl#L158

It just sees a symbol in the first argument for `MH` and says "ah whatever time for static MH". I can fix that.

cpfiffer:
[...]

PR #1557, opened 2021-03-11
Make MH proposal generation more sane

cpfiffer:
Makes it easier to do the kind of thing requested in #1556. Now we handle `proposal` generation more consistently when constructing `MH`, and there's even a helpful error when a proposal type is not supported.

No real functional change here, just a slight bug fix & consistency improvements.

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -7,7 +7,10 @@ struct MH{space, P} <: InferenceAlgorithm
 end
 
 proposal(p::AdvancedMH.Proposal) = p
+proposal(f::Function) = AdvancedMH.StaticProposal(f)
+proposal(d::Distribution) = AdvancedMH.StaticProposal(d)
 proposal(cov::AbstractMatrix) = AdvancedMH.RandomWalkProposal(MvNormal(cov))
+proposal(x) = error("Proposal type $x not supported")
 
 """
     MH(space...)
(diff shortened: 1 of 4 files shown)


src/inference/mh.jl
```diff
@@ -7,7 +7,10 @@ struct MH{space, P} <: InferenceAlgorithm
 end
 
 proposal(p::AdvancedMH.Proposal) = p
+proposal(f::Function) = AdvancedMH.StaticProposal(f)
+proposal(d::Distribution) = AdvancedMH.StaticProposal(d)
 proposal(cov::AbstractMatrix) = AdvancedMH.RandomWalkProposal(MvNormal(cov))
+proposal(x) = error("Proposal type $x not supported")
```
devmotion, line 13:
```suggestion
proposal(x) = error("proposals of type ", typeof(x), " are not supported")
```

devmotion (commented):
Looks good, I just added a minor suggestion. Maybe also add some tests?

test/inference/mh.jl
```diff
@@ -106,6 +106,48 @@
...
+                MH((:σ, vc_σ)),
+                ), 3_000 # draws
+            )
+        
+            
+        chn2 = sample(mod, MH(), 3_000)
+
+        # Test that the small variance version is actually smaller.
+        v1 = var(diff(chn["μ[1]"].data, dims=1))
+        v2 = var(diff(chn2["μ[1]"].data, dims=1))
```
devmotion, line 146: Is it possible to use something like
```suggestion
        v1 = var(diff(Array(chn["μ[1]"]), dims=1))
        v2 = var(diff(Array(chn2["μ[1]"]), dims=1))
```
instead of the internals here?

devmotion (approved):

PR #1564, opened 2021-03-30
Update "Contributing" link

tjburch:
Very minor change - was looking for information about contributing and noticed that the link from the README to the contributing section was outdated. Updated it to the current location in the documentation.


torfjelde (approved):
Nice catch! Thank you:)

Issue #1554, opened 2021-03-08
Segmentation fault when running tests on Julia v1.6.0-rc1


lmshk:
On a fresh installation of Julia `v1.6.0-rc1` with Turing `v0.15.1`, running `pkg> test Turing` always causes segmentation faults on my machine (`x86_64 GNU/Linux 5.11.1-arch1-1`).

The tracebacks vary, but the segfaults always seem to happen during GC. I have attached a couple of specimens:
- [Turing1.txt](https://github.com/TuringLang/Turing.jl/files/6103179/Turing1.txt)
- [Turing2.txt](https://github.com/TuringLang/Turing.jl/files/6103181/Turing2.txt)
- [Turing3.txt](https://github.com/TuringLang/Turing.jl/files/6103183/Turing3.txt)

When playing around with various Turing tutorials this problem has not occured yet, so this is not currently blocking my work, but I am willing to investigate if somebody points me in the right direction. (I have never used Turing before.)

devmotion:
This is expected and caused by Libtask. Its binary dependency Libtask_jll is not compatible with Julia 1.6 yet - it has to be linked against libjulia but the version for Julia 1.6 is not available yet (since Julia 1.6 is not released). However, there exists already a PR that will fix this issue once the new version of libjulia is available: https://github.com/JuliaPackaging/Yggdrasil/pull/2176

Therefore there's nothing that can be done here and it will be fixed (more or less) automatically when Julia 1.6 is released. As long as you don't use Libtask-dependent features (i.e., if you do not use particle filters), Turing should work fine on Julia 1.6 already now (but due to the Libtask_jll incompatibility an older version will be installed by default IIRC).

lmshk:
Thanks for the quick and detailed reply!

> Turing should work fine on Julia 1.6 already now (but due to the Libtask_jll incompatibility an older version will be installed by default IIRC)

I am a little confused about this: Doing `pkg> add Turing` installed the (older) `v0.15.1`, which has the tests failing as described above. So it also has the Libtask_jll incompatibility, right?

(Do I understand correctly that this issue would also affect Turing on a Julia compiled from `master`?)

lmshk:
All clear, thank you.

PR #1567, opened 2021-04-04
Fix for #1352

### LINKED ISSUE #1352: Wrong prediction results on multivariate params

torfjelde:
Fixes #1352 and increases the lower-bound of DynamicPPL (the recent version includes bug-fixes for `predict` and `generated_quantities`).


src/inference/Inference.jl
```diff
@@ -603,44 +603,23 @@ function transitions_from_chain(
 )
     return transitions_from_chain(Random.GLOBAL_RNG, model, chain; kwargs...)
 end
+
+
```
devmotion, line 607:
```suggestion
```

Project.toml
```diff
@@ -42,7 +42,7 @@ Bijectors = "0.8"
 Distributions = "0.23.3, 0.24"
 DistributionsAD = "0.6"
 DocStringExtensions = "0.8"
-DynamicPPL = "0.10.2"
+DynamicPPL = "0.10.9"
```
devmotion, line 45: I guess we should update the bounds in test/Project.toml as well.

src/inference/Inference.jl
```diff
@@ -649,4 +628,5 @@ function transitions_from_chain(
     return transitions
 end
 
+
```
devmotion, line 631:
```suggestion
```

Project.toml
torfjelde: Done :+1:

devmotion:
The tests fail since Libtask_jll does not support Julia 1.6.0 yet - it will be fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/2176 but the PR depends on https://github.com/JuliaPackaging/Yggdrasil/pull/1987 (see https://github.com/TuringLang/Turing.jl/issues/1554 for more details).

devmotion:
Maybe we should run Turing tests with Julia 1.5, here and in DynamicPPL?

torfjelde:
> Maybe we should run Turing tests with Julia 1.5, here and in DynamicPPL?

Sounds like a good idea to me :+1:

torfjelde:
Just for reference: this is now only waiting on packages to get up-to-date with Bijectors.jl and running #1568

devmotion:
The test error is strange (negative standard deviation in a Normal distribution), do you know what's going on there?

torfjelde:
> The test error is strange (negative standard deviation in a Normal distribution), do you know what's going on there?

Very strange indeed. And no, I don't. It shouldn't have anything to do with this particular PR though :confused:

torfjelde:
It's all the Gibbs-sampler btw

torfjelde:
This should finally be good to go now, once tests pass :+1:

src/inference/Inference.jl
```diff
@@ -517,24 +517,24 @@ function predict(model::Model, chain::MCMCChains.Chains; kwargs...)
     return predict(Random.GLOBAL_RNG, model, chain; kwargs...)
 end
 function predict(rng::AbstractRNG, model::Model, chain::MCMCChains.Chains; include_all = false)
+    # Don't need all the diagnostics
+    chain = MCMCChains.get_sections(chain, :parameters)
```
devmotion, line 521: Can we maybe use different variable names here? It can be a bit tricky, both for the compiler and reader, if different objects of possibly different type share the same name.

devmotion (commented):
Looks good, just had a minor suggestion. It seems you also have to update the version number 🙂

src/inference/Inference.jl
torfjelde: Good point :+1:

devmotion (approved):

torfjelde:
What in the world. Gibbs is failing again

torfjelde:
I'm very confused by that test. Nothing that I did in the fix for MH and linking changed the behavior of `MH()`, which is NOT linked and uses `StaticProposal`. But now it results in `v2` in the test at https://github.com/TuringLang/Turing.jl/blob/master/test/inference/mh.jl#L149-L152 vanishing, which makes the test fail o.O

Is it okay if I just disable that particular comparison for now and make an issue for it? *Really* want to get the PR merged, and the only thing holding it back is tests unrelated to this PR failing due to random seed :sweat_smile:

yebai:
> Is it okay if I just disable that particular comparison for now and make an issue for it? Really want to get the PR merged, and the only thing holding it back is tests unrelated to this PR failing due to random seed 😅

Happy to merge this PR as-is and open an issue for the failed test.

PR #1577, opened 2021-04-08
Run tests on Julia 1.6

devmotion:
Reverts https://github.com/TuringLang/Turing.jl/pull/1568 since now a version of Libtask_jll is available that is compatible with Julia 1.6 (segfaults caused by multithreading on Julia 1.5.4 require a different version of Libtask_jll which hopefully is available soon).

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Julia
         uses: julia-actions/setup-julia@v1
         with:
-          version: '1.5'
+          version: '1'
       - name: Set up Ruby 2.6
         uses: actions/setup-ruby@v1
         with:
--- a/.github/workflows/DynamicHMC.yml
+++ b/.github/workflows/DynamicHMC.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         version:
           - '1.3'
-          - '1.5'
+          - '1'
         os:
           - ubuntu-latest
         arch:
--- a/.github/workflows/Numerical.yml
+++ b/.github/workflows/Numerical.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         version:
           - '1.3'
-          - '1.5'
+          - '1'
         os:
           - ubuntu-latest
         arch:
--- a/.github/workflows/StanCI.yml
+++ b/.github/workflows/StanCI.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         version:
           - '1.3'
-          - '1.5'
+          - '1'
         os:
           - ubuntu-latest
         arch:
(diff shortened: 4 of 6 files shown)


torfjelde (approved):
LGTM (as long as tests pass)!

Btw, you're the best @devmotion :heart:

devmotion:
I get the impression that some tests are take significantly more time to complete than what I am used to (e.g. 75 minutes for Turing CI whereas I thought it usually was 35-45 minutes). Also the integration tests in DynamicPPL timed out earlier today. Could it be that the compiler change in DynamicPPL introduced these regressions @torfjelde?

Also tests on Julia 1.3 seem slower, which would indicate that it is not caused by Julia 1.6.

torfjelde:
> I get the impression that some tests are take significantly more time to complete than what I am used to (e.g. 75 minutes for Turing CI whereas I thought it usually was 35-45 minutes). Also the integration tests in DynamicPPL timed out earlier today. Could it be that the compiler change in DynamicPPL introduced these regressions @torfjelde?
> 
> Also tests on Julia 1.3 seem slower, which would indicate that it is not caused by Julia 1.6.

Hmm, maybe that change + APPL change really did have an impact, since now that check will occur at every assume-statement rather than just the `observe` statement *and* be compiled away (if possible).

How would be go about checking this properly? Revert the changes and run tests for that?

It does kind of suck if we were to do that though, as it would not make it possible to take the submodel-approach we discussed, since now `args` need to be known at compile-time :confused:

torfjelde:
And you're right btw, it used to take 35mins a couple of PRs back :confused:

devmotion:
> How would be go about checking this properly? Revert the changes and run tests for that?

I think the best approach would be to create a new branch from the PR in DynamicPPL (the one that switches tests to Julia 1.6) and revert the changes in this branch. And then run `bors try` in the branch and compare it with the PR that timed out twice now.

devmotion:
I don't think the APPL change causes these regressions, it should rather improve type inference.

torfjelde:
> I don't think the APPL change causes these regressions, it should rather improve type inference.

But improving type-inference will put more pressure on the compiler, right? When it doesn't know the type, it will leave all the checks for `observe` to be dealt with at runtime rather than trying to compile them away.

Even so, I am worried (and think you're right here) that it's the fact that we're performing this check at every `~` now which causes the insane slow-down. I'll make a PR as you said and try.

torfjelde:
#1578

torfjelde:
Still waiting for 1.6 over at #1578, but notice that 1.3 tests over there are only a couple of minutes slower.

Also, why is 1.6 *slower* than 1.3? o.O

devmotion:
> Also, why is 1.6 *slower* than 1.3? o.O 

Maybe Libtask became slower? Or there are some regressions in 1.6? Did you observe any differences in your experiments?

devmotion:
Or some AD backend is slower on 1.6?

PR #1578, opened 2021-04-08
Upper-bound DPPL to remove most recent compiler change

torfjelde:


torfjelde:
@devmotion Tests are running like 10mins faster here compared to #1577 , so we're looking at a ~15% slowdown?

devmotion:
Is this relative increase consistent with the timings in your experiments?

devmotion:
Is this relative increase consistent with the timings in your experiments?

torfjelde:
Was difficult to see a significant slowdown in those experiments. We observed a ~15% slowdown for `model2` (bernoulli coin-flip), so maybe?

devmotion:
I am a bit surprised though since my impression was that it was mainly compile time regressions. So I assumed it would have a smaller impact in the tests since we use the same model repeatedly.

torfjelde:
I guess we didn't check AD-related stuff :/ E.g. maybe Zygote is significantly slower?
We need to recompile the model for different types, so maybe that's why we're seeing such an increase in testing time too?

Anyways, it's probably a good idea to run the benchmarks present in https://github.com/TuringLang/TuringExamples for the different versions and see if we can find something clearcut.
I'll get on it, but uncertain if I'll have it available today.

In the meantime, can we try to get the outstanding PRs merged? E.g. bump AbstractMCMC versions across the board and merge #1567 ?

devmotion:
> In the meantime, can we try to get the outstanding PRs merged? E.g. bump AbstractMCMC versions across the board and merge #1567 ?

Sure, let's finish these other PRs.

PR #1579, opened 2021-04-09
Bump AbstractMCMC major version

torfjelde:
Supersedes #1574 and #1575

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -32,7 +32,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
-AbstractMCMC = "2.1"
+AbstractMCMC = "2.1, 3"
 AdvancedHMC = "0.2.24"
 AdvancedMH = "0.5.2"
 AdvancedPS = "0.1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.15.14"
+version = "0.15.15"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -28,7 +28,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
-AbstractMCMC = "2.1"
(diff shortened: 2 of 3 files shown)


devmotion (approved):
Looks good, just have to make sure that it's actually tested and needs a version bump :+1:

devmotion:
The same strange errors in the Gibbs MH test. I'm even more surprised that the errors occur on Julia 1.3 :thinking:

devmotion:
The failing tests were added recently in https://github.com/TuringLang/Turing.jl/pull/1557.

devmotion:
@cpfiffer Maybe we should just change the test in https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/test/inference/mh.jl#L137? To me it seems, it is likely to fail randomly (even with a small step size) since we use a random walk proposal for the standard deviation that *has* to be positive (otherwise the model errors)?

torfjelde:
But it should be linking, no?

torfjelde:
At least that's the intention: https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L404

torfjelde:
I think I see the issue. https://github.com/TuringLang/Turing.jl/blob/master/src/inference/mh.jl#L404 passes `spl.alg.proposals`, but in this case the proposals are actually `NamedTuple{sym, RandomWalkProposal}` and so they hit the `maybe_link!` which does nothing.

Project.toml
```diff
@@ -32,7 +32,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
-AbstractMCMC = "2.1"
+AbstractMCMC = "2.1, 3"
```
cpfiffer, line 35:
```suggestion
AbstractMCMC = "3"
```

This should cut out AbstractMCMC 2 because the internal API changes.

test/Project.toml
```diff
@@ -28,7 +28,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
-AbstractMCMC = "2.1"
+AbstractMCMC = "2.1, 3"
```
cpfiffer, line 31:
```suggestion
AbstractMCMC = "3"
```

Same here.

Project.toml
devmotion: But this is only relevant for users that use custom callbacks? I.e., for Turing it should be fine to support both versions since we do not depend on these internals, do we?

cpfiffer: Hm, I suppose not. IGNORE ME

torfjelde: Yeah, leaving it as is:)

PR #1582, opened 2021-04-10
Improve `maybe_link!` for MH sampler

torfjelde:
Fixes issues as discussed in #1579 regarding failing tests by making `maybe_link!` also link if `proposal` is a `NamedTuple` of proposals which *all* should be linked. If only one of the proposals does not require linking, we don't link.

It might be better to not dispatch here, but instead check at runtime and warn the user that there's a mix between proposals that are by default linked but don't link any?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.15.13"
+version = "0.15.14"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -338,6 +338,13 @@ maybe_link!(varinfo, sampler, proposal) = nothing
 function maybe_link!(varinfo, sampler, proposal::AdvancedMH.RandomWalkProposal)
     link!(varinfo, sampler)
 end
+function maybe_link!(
+    varinfo,
+    sampler,
+    proposal::NamedTuple{names, vals}
+) where {names, vals<:NTuple{<:Any, <:AdvancedMH.RandomWalkProposal}}
+    return link!(varinfo, sampler)
+end
 
 # Make a proposal if we don't have a covariance proposal matrix (the default).
 function propose!(
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -334,16 +334,25 @@ end
 end
 
 # Utility functions to link
-maybe_link!(varinfo, sampler, proposal) = nothing
-function maybe_link!(varinfo, sampler, proposal::AdvancedMH.RandomWalkProposal)
-    link!(varinfo, sampler)
+function should_link(varinfo, sampler, proposal::NamedTuple{(), Tuple{}})
+    # If it's an empty `NamedTuple`, we're using the priors as proposals
+    # in which case we shouldn't link.
+    return false
 end
-function maybe_link!(
+function should_link(varinfo, sampler, proposal::AdvancedMH.RandomWalkProposal)
+    return true
+end
+function should_link(
     varinfo,
     sampler,
     proposal::NamedTuple{names, vals}
 ) where {names, vals<:NTuple{<:Any, <:AdvancedMH.RandomWalkProposal}}
-    return link!(varinfo, sampler)
+    return true
+end
+
+function maybe_link!(varinfo, sampler, proposal)
+    should_link(varinfo, sampler, proposal) && link!(varinfo, sampler)
+    return varinfo
 end
 
 # Make a proposal if we don't have a covariance proposal matrix (the default).
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -334,6 +334,7 @@ end
 end
 
 # Utility functions to link
+should_link(varinfo, sampler, proposal) = false
 function should_link(varinfo, sampler, proposal::NamedTuple{(), Tuple{}})
     # If it's an empty `NamedTuple`, we're using the priors as proposals
     # in which case we shouldn't link.
--- a/test/inference/mh.jl
+++ b/test/inference/mh.jl
@@ -130,13 +130,17 @@
         # with small-valued VC matrix to check if we only see very small steps
         vc_μ = convert(Array, 1e-4*I(2))
         vc_σ = convert(Array, 1e-4*I(2))
-        
-        chn = sample(mod,
-            Gibbs(
-                MH((:μ, vc_μ)),
-                MH((:σ, vc_σ)),
-                ), 3_000 # draws
-            )
+
+        alg = Gibbs(
+            MH((:μ, vc_μ)),
+            MH((:σ, vc_σ)),
+        )
+
+        chn = sample(
+            mod,
+            alg,
+            3_000 # draws
+        )
         
             
         chn2 = sample(mod, MH(), 3_000)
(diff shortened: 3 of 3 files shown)


devmotion:
Somehow this feels a bit wrong - wouldn't it be better to extract the different proposals from the NamedTuple and link only the variables that should be linked?

torfjelde:
> Somehow this feels a bit wrong - wouldn't it be better to extract the different proposals from the NamedTuple and link only the variables that should be linked?

Yep. But isn't this when you should be using `Gibbs`?

torfjelde:
Also tests are now failiing o.O

torfjelde:
> Also tests are now failiing o.O

Ah crap. My definition is too general and will also catch `proposal == NamedTuple()`. Though why would "unnecessarily" linking cause issues?

devmotion:
> > Somehow this feels a bit wrong - wouldn't it be better to extract the different proposals from the NamedTuple and link only the variables that should be linked?
> 
> Yep. But isn't this when you should be using `Gibbs`? 

But the docstrings explicitly explain how to set different proposals for different variables without Gibbs, so it should work it seems. I still don't understand what the underlying problem is here but it seems wrong to fail or emit warning here, and it seems suboptimal to link everything.

devmotion:
Ideally, the linking would be more transparent and less confusing. There are so many bugs and hours of debugging...

IIRC when moving to AbstractMCMC 2 we discussed that in general it would be good if one could avoid to link the samples only in the beginning and at the end of sampling since it leads to surprising results when one uses the iteration interface.

torfjelde:
> But the docstrings explicitly explain how to set different proposals for different variables without Gibbs, so it should work it seems. I still don't understand what the underlying problem is here but it seems wrong to fail or emit warning here, and it seems suboptimal to link everything.

I agree! But not clear to me how to go about doing that without additional implementations in DynamicPPL :confused: 

> Ideally, the linking would be more transparent and less confusing. There are so many bugs and hours of debugging...

100% agree.

> IIRC when moving to AbstractMCMC 2 we discussed that in general it would be good if one could avoid to link the samples only in the beginning and at the end of sampling since it leads to surprising results when one uses the iteration interface.

Wait, you want to NOT only link at beginning and end of sampling?

torfjelde:
> > Also tests are now failiing o.O
> 
> Ah crap. My definition is too general and will also catch `proposal == NamedTuple()`. Though why would "unnecessarily" linking cause issues?

Ah, so when we do `MH()` we use the priors as proposals and thus linking causes issues.

torfjelde:
Btw, one simple way of allowing the desired behavior is to add a `DynamicPPL.link!(vi, spl, space)` rather than forcing `link!` to work on the "entire space" of `spl`. E.g. 
```julia
function link!(vi::TypedVarInfo, spl::AbstractSampler)
    return link!(vi, spl, Val(getspace(spl)))
end
function link!(vi::TypedVarInfo, spl::AbstractSampler, spaceval::Val)
    vns = _getvns(vi, spl)
    return _link!(vi.metadata, vi, vns, spaceval)
end
```

devmotion:
> Wait, you want to NOT only link at beginning and end of sampling?

Yep, maybe it would be better to just apply the bijective mapping in `set_namedtuple!` (called in https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L251-L263 and https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L362) and `dist_val_tuple` (in https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L351), and in https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L378 and in https://github.com/TuringLang/Turing.jl/blob/2eab753d1ad744cd3020a807966f21c2659d1d6c/src/inference/mh.jl#L389. Then one could perform whatever transformation is required for a specific variable and proposal.

As I said above, I am always confused by the linking. In general, IMO it is problematic if the samples that are returned in the intermediate steps return the transformed samples. I am not completely sure if this happens here but it is one reason for why I think the current design is a bit suboptimal. Also, if I understand correctly, the linking only makes it easier to set and retrieve the transformed samples (which could be moved to the other functions mentioned above) but when the model is executed the variables still have to be converted to the original anyway, so I wonder if and to what extent it actually improves performance. Maybe generally it would be cleaner to always store the samples in the original space in the VarInfo object and to map them on the fly if needed (or at least not modify the VarInfo object in-place since this e.g. makes it impossible to implement bijections from a submanifold in a higher dimensional space to some lower dimensional space where the dimensions and potentially the type changes).

torfjelde:
> Yep, maybe it would be better to just apply the bijective mapping in set_namedtuple!

I disagree with this right now, simply because it diverges from the current approach of the rest of the samplers.

BUT I do agree that the current approach is suboptimal; just feels like we should address that head on rather than approach it differently for different samplers.

>  In general, IMO it is problematic if the samples that are returned in the intermediate steps return the transformed samples. I am not completely sure if this happens here but it is one reason for why I think the current design is a bit suboptimal.

I see what you're saying and I remember the conversation we had about potential issues this causes when using the iterator interface for AbstractMCMC.jl.

And yes, it does happen here. We only `link!` at the beginning of the sampling process.

> Also, if I understand correctly, the linking only makes it easier to set and retrieve the transformed samples (which could be moved to the other functions mentioned above) but when the model is executed the variables still have to be converted to the original anyway, so I wonder if and to what extent it actually improves performance. 

Currently we do:
```julia
for i = 1:num_samples
    # Propose in unconstrained space
    ϕ_new = step(ϕ)

    # Transform to constrained
    θ_new = f(ϕ)

    # Evaluate logpdf
    logpdf(model, θ_new)

    # Update sample
    ϕ = θ_new
end
```
but if we made it so that it should be transformed on the fly, we'd get
```julia
for i = 1:num_samples
    # ADDITIONAL STEP
    # Transform previous sample to unconstrained
    ϕ = f⁻¹(θ)
    
    # Propose in unconstrained space
    ϕ_new = step(ϕ)

    # Transform to constrained
    θ_new = f(ϕ)

    # Evaluate logpdf
    logpdf(model, θ_new)

    # Update sample
    ϕ = θ_new
end
```

So it def helps performance-wise in the case where the linking is actual a significant part of the computation.

> Maybe generally it would be cleaner to always store the samples in the original space in the VarInfo object and to map them on the fly if needed (or at least not modify the VarInfo object in-place since this e.g. makes it impossible to implement bijections from a submanifold in a higher dimensional space to some lower dimensional space where the dimensions and potentially the type changes).

It does have a performance cost though, as mentioned above. But I do think it's an important point, and for certain bijectors we could possibly even gain some performance by doing what you mention.

With all this being said, I do agree that the implicit transformation is both:
1. Very confusing at times.
2. Often not even worth it.

I'm also of the opinion that we should exploit Bijectors.jl more.
These days, using Bijectors.jl we easily construct a bijector which handles the entire `VarInfo` without any additional overhead (for static models). We can even optimize the structure of the transformation, e.g. by combining contiguous ranges into a single vectorized bijector.
So as a "step" along the way, we could for example make something similar to `TransformedDistribution` for `VarInfo`, which is a bijector + `VarInfo`, and then we overload the necessary operations, e.g. `setindex!`, to take in unconstrained variables, convert them to constrained, and set the underlying `VarInfo`. 

This would also make it possible to construct completely arbitrary transformations, use different bijectors than default if that's wanted, etc., by simply calling `b = bijector(varinfo)` and then replacing parts of `b` as desired.

torfjelde:
Also, we should take this discussion to an issue over at DPPL :+1:

In addition, open an issue regarding how MH is a bit disingenuous when it comes to what it claims and what it can actually do wrt. unconstrained variables.

devmotion:
> but if we made it so that it should be transformed on the fly, we'd get

The unconstrained sample could just be part of the state.

But whatever, I agree that the general design should be discussed (and possibly changed) separately and not in this PR :+1:

devmotion (commented):
Looks good but it seems it doesn't handle the case where we use not only random walk proposals in the NamedTuple, does it?

And I wonder if we should add specific tests or it's sufficient that the tests don't fail anymore?

torfjelde:
> Looks good but it seems it doesn't handle the case where we use not only random walk proposals in the NamedTuple, does it?

Correct; IMO we should fix the general case through https://github.com/TuringLang/DynamicPPL.jl/pull/231. 
I'll open an issue here that we need to handle the general case though.

> And I wonder if we should add specific tests or it's sufficient that the tests don't fail anymore?

Well, technically the tests to catch this were already present :sweat_smile: Issue was that they only did so every now and then, depending on the `rng`. But good point, I'll add some tests.

devmotion (approved):
Looks good to me, given that tests pass. Can you also update the version number?

PR #1585, opened 2021-04-11
Make `_params_to_array` faster, fixes #1584

### LINKED ISSUE #1584: _params_to_array is very slow
I am using Turing on a model with ~2 million parameters, I found out that doing just one sample takes 20 minutes. The time is almost exclusively taken by the `_params_to_array` function. In particular, line 252 to 256 of Inference.jl is equivalent to the following function
```
function foo(nms)
    names = Vector{Symbol}()
    for nm in nms
        if !(nm in names)
            push!(names, nm)
        end
    end
    return names
end
```
This is inefficient because testing for membership of a `Vector{Symbol}` is not efficient. We can modify this implementation to the following

```
function bar(nms)
    names = Vector{Symbol}()
    append!(names, nms)
    unique!(names)
    return names
end
```

Giving an argument of `nms = [Symbol("param_", i) for i in 1:2000000]`,  `foo(nms)` takes ~20 minutes to run, while `bar(nms)` takes 0.5 seconds.

Can we change the implementation of `_params_to_array`  in the same fashion?
```
function _params_to_array(ts::Vector)
    names = Vector{Symbol}()
    # Extract the parameter names and values from each transition.
    dicts = map(ts) do t
        nms, vs = flatten_namedtuple(getparams(t))
        append!(names, nms)
        # Convert the names and values to a single dictionary.
        return Dict(nms[j] => vs[j] for j in 1:length(vs))
    end
    unique!(names)
    vals = [get(dicts[i], key, missing) for i in eachindex(dicts), 
        (j, key) in enumerate(names)]

    return names, vals
end
```

haozhangphd:


src/inference/Inference.jl
```diff
@@ -249,15 +249,11 @@ function _params_to_array(ts::Vector)
...
-            if !(nm in names)
-                push!(names, nm)
-            end
-        end
+        append!(names, nms)
         # Convert the names and values to a single dictionary.
         return Dict(nms[j] => vs[j] for j in 1:length(vs))
     end
-    # names = collect(names_set)
+    unique!(names)
```
devmotion, line 256: Probably we should just use a `Set` (actually, I thought we operate on `Set`s but maybe this was lost in some refactoring).

cpfiffer:
Could you increment the version number in Project.toml to 0.15.16?

Location: https://github.com/TuringLang/Turing.jl/blob/be40a19e7024f56c6a5024e937b47da42dc9e800/Project.toml#L3

Also, I'd be curious to see some kind of benchmarking here -- how big is the improvement?

src/inference/Inference.jl
cpfiffer: I think I changed it away from using a `Set` because the order was getting messed up, though I cannot recall the exact reason.

cpfiffer: It may not be relevant anymore, though.

cpfiffer: I wonder if an [`OrderedSet`](https://juliacollections.github.io/DataStructures.jl/v0.8/ordered_containers.html) would be faster.

torfjelde: `OrderedSet` seems perfect 👍

haozhangphd:
For my particular model with ~2 million parameters, the time for 10 samples went down from more than 2 hours to less than 1 minute. I did not do a very precise benchmark for the version in 0.15.15, because it is very long. I think the effect of this change can be more precisely benchmarked with code snippets listed in #1584, because the appending behavior of `names` is the only thing that's changed.  Changing from a loop-based implementation to a `unique` based implementation reduces the running time from 20 minutes to 0.5 seconds, using an input of 2-million-element vector of Symbols.

devmotion:
A quick benchmark:
```julia
julia> using BenchmarkTools, DataStructures

julia> function bar(nms)
           names = Vector{Symbol}()
           append!(names, nms)
           unique!(names)
           return names
       end

julia> function bar_set(nms)
           names = Set{Symbol}()
           for nm in nms
               push!(names, nm)
           end
           return collect(names)
       end

julia> function bar_orderedset(nms)
           names = OrderedSet{Symbol}()
           for nm in nms
               push!(names, nm)
           end
           return collect(names)
       end

julia> nms = [Symbol("param_", i) for i in 1:2000000]

julia> @btime bar($nms);
  448.749 ms (55 allocations: 85.76 MiB)

julia> @btime bar_set($nms);
  428.193 ms (54 allocations: 85.76 MiB)

julia> @btime bar_orderedset($nms);
  387.644 ms (65 allocations: 63.59 MiB)
```

cpfiffer:
Excellent! @haozhangphd for an additional speedup, could you change the `Vector{Symbol}` at the start to `OrderedSet{Symbol}`? I think you'll need to add `DataStructures` to the `Project.toml` and throw in a `using DataStructures` somewhere.

devmotion:
So IMO we should use `OrderedSet` if we want to not mess with the insertion order (and it seems to be the fastest as well, even though I don't know why it would be faster than `Set`).

haozhangphd:
Thanks for suggesting to use `OrderedSet`!

src/inference/Inference.jl
```diff
@@ -245,19 +246,17 @@ getparams(t) = t.θ
...
-    names = Vector{Symbol}()
+    names_set = OrderedSet{Symbol}()
     # Extract the parameter names and values from each transition.
     dicts = map(ts) do t
         nms, vs = flatten_namedtuple(getparams(t))
         for nm in nms
-            if !(nm in names)
-                push!(names, nm)
-            end
+               push!(names_set, nm)
```
torfjelde, line 254: Is indentation a bit different here?

haozhangphd: Sorry I messed up...

torfjelde: No worries; nice work!

cpfiffer (approved):
Alright, I'm happy to merge when the tests pass.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.15.15"
+version = "0.15.16"
```
devmotion, line 3: Since 0.15.16 was just released this morning I'll update it to
```suggestion
version = "0.15.17"
```

devmotion:
Thanks for the PR @haozhangphd!

PR #1586, opened 2021-04-12
Make SGHMC more flexible

haozhangphd:
Currently the `velocity` field of `SGHMCState` is hard-coded to `Vector{Float64}`. This is not very flexible.

For example, when training a Bayesian neural network on CUDA, the parameters are usually assumed to `Float32`. The current setup does not allow the priors to generate `Float32` parameters. And mixing `Float32` and `Float64` types can give difficult-to-debug errors during backpropagation, because many of the Zygote adjoint rules assumes all the arguments have the same numerical type (for example [here](https://github.com/FluxML/NNlib.jl/blob/2c3bdb8e45fa912282ed71823bce13fc2f45c1a0/src/pooling.jl#L119)).

I would like to make the type of `SGHMCState` automatically consistent with the type of `VarInfo`, and the prior distribution is no longer restricted to `Float64` type.


torfjelde (approved):
This is nice:)
Thanks!

torfjelde:
Oh one thing: could you bump the version?

haozhangphd:
Thanks! The `Project.toml` file is updated based on the latest version in `Master`

ma-fei:
Is it possible to get this PR merged soon? I just realized a final project for my school work actually depend on this PR, otherwise my project implementation throw errors. I really appreciate it!

devmotion (approved):

devmotion:
Thanks for the PR!

devmotion:
The new version will be available soon: https://github.com/JuliaRegistries/General/pull/34799

ma-fei:
@devmotion Thank you very much for your very quick response!

Issue #1598, opened 2021-05-01
PG - not retaining reference particle?


m9ko:
Hello,

I've recently updated Turing from v0.13.0 to v0.15.18 and noticed that `PG()` does not work the same way as before, specifically when it comes to retaining reference particles (I'm using Julia 1.4.2 for both). Using an example from Turing Guide: https://turing.ml/dev/docs/using-turing/guide

```julia
@model function gdemo(x, y)
    s ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s))
    x ~ Normal(m, sqrt(s))
    y ~ Normal(m, sqrt(s))
end

# Use only one particle for PG
c2 = sample(gdemo(1.5, 2), PG(1), 1000)
```

I tested it out with only one particle, where PG should retain its original particle throughout the iterations. This is the case for v0.13.0, but not for v0.15.18 (plot of `m` shown below). I'm wondering if this is due to resetting VarInfo before creating the reference particle (line 282 onward in Turing.jl/src/inference/AdvancedSMC.jl). Please let me know if this is intended or I'm mistaken somewhere, as I'm just a Turing/Julia beginner :')

Old Turing ver:

![image](https://user-images.githubusercontent.com/53410259/116788512-fd27b480-aa77-11eb-8cea-ab25a28d2704.png)

New Turing ver:

![image](https://user-images.githubusercontent.com/53410259/116788547-33653400-aa78-11eb-8094-d183fa98c97e.png)

devmotion:
Thanks for opening the issue and providing this clear and simple example! Recently, we started to extract the particle sampling logic to a separate package AdvancedPS.jl such that it can be used even without Turing. Currently, this is not possible (in a nice way) and both packages still not clearly separated. I assume that during the refactoring we (or probably rather I :smile:) broke something accidentally.

In principle, https://github.com/TuringLang/Turing.jl/blob/46c05895d7e51028211a856c78b1cca60838c5d4/src/inference/AdvancedSMC.jl#L294 should ensure that the reference particle is retained but it seems this is not the case anymore. Maybe because the VarInfo object is reset before. I'll have a closer look at the example and check why it fails.

devmotion:
[...]

PR #1599, opened 2021-05-01
Retain reference particle properly

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1598.

### DIFF
--- a/src/inference/AdvancedSMC.jl
+++ b/src/inference/AdvancedSMC.jl
@@ -279,6 +279,9 @@ function AbstractMCMC.step(
     vi::AbstractVarInfo;
     kwargs...
 )
+    # Create reference particle that will be retained
+    reference = AdvancedPS.forkr(AdvancedPS.Trace(model, spl, vi))
+
     # Reset the VarInfo before new sweep.
     reset_num_produce!(vi)
     set_retained_vns_del_by_spl!(vi, spl)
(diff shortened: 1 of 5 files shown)


devmotion:
Hmm it seems it breaks some other tests, even though it fixes an existing bug. I have to check why e.g. https://github.com/TuringLang/Turing.jl/pull/1599/checks?check_run_id=2483639336#step:6:493 fails now. Additionally, there seems to be some precompilation error caused by LoopVectorization: https://github.com/TuringLang/Turing.jl/pull/1599/checks?check_run_id=2483639317#step:6:257

devmotion:
I opened an issue regarding the LoopVectorization errors: https://github.com/JuliaSIMD/LoopVectorization.jl/issues/258

devmotion:
OK, now everything but the LoopVectorization issue should be fixed.

src/inference/AdvancedSMC.jl
```diff
@@ -281,17 +281,21 @@ function AbstractMCMC.step(
 )
     # Reset the VarInfo before new sweep.
     reset_num_produce!(vi)
-    set_retained_vns_del_by_spl!(vi, spl)
     resetlogp!(vi)
 
+    # Create reference particle for which the samples will be retained.
+    reference = AdvancedPS.forkr(AdvancedPS.Trace(model, spl, vi))
```
yebai, line 287: nice catch - maybe add a comment referencing the Github issue so that we won't incidentally re-introduce the same bug? 

EDIT: I see there is a comment in the tests.

yebai (approved):
thanks, @devmotion - great work as always!

devmotion:
The LoopVectorization issue was fixed upstream. One numerical test is a bit unstable, I'll check if increasing the number of samples or tuning the parameters of the samplers helps.

PR #1605, opened 2021-05-04
Use correct statistical notation in Hello World

palday:
The notation here is problematic. `σ` should represent the standard deviation, not the variance. Similarly `σ²` represents the variance in standard statistical notation. If you want to put a prior on the variance, then call it  `σ²`, so that the standard deviation is  `sqrt(σ²)` as expected.

 `sqrt(σ)` would be the square root of the standard deviation, and that's definitely not a common quantity nor a common way to parameterize the normal distribution.

### DIFF
--- a/docs/pages/index.html
+++ b/docs/pages/index.html
@@ -20,11 +20,11 @@ code-sample:
   snippet: |
     @model gdemo(x, y) = begin
       # Assumptions
-      σ ~ InverseGamma(2,3)
-      μ ~ Normal(0,sqrt(σ))
+      σ² ~ InverseGamma(2,3)
+      μ ~ Normal(0,sqrt(σ²))
       # Observations
-      x ~ Normal(μ, sqrt(σ))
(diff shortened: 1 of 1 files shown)


palday:
This problem seems to occur elsewhere in the documentation with `s` instead of `σ`.

yebai:
Many thanks, @palday - there was still some early-stage sloppy code around. Hopefully, we will iron them out in the next doc/tutorial refactoring. In the meantime, please feel free to open PRs for the ones you noticed.

yebai (approved):

docs/pages/index.html
```diff
@@ -20,11 +20,11 @@
   snippet: |
     @model gdemo(x, y) = begin
```
devmotion, line 21: We could use the opportunity and change the example to the syntax that is more common in the implementation andd documentation nowadays:
```suggestion
    @model function gdemo(x, y)
```

PR #1627, opened 2021-05-28
Don't `sort` chain by default

torfjelde:
This drops the default behavior of sorting the chain in `sample`. Sorting can be turned on again using the kwarg `sort_chain=true`.

IMO we should get https://github.com/TuringLang/DynamicPPL.jl/pull/251 merged first, so that uses of `setval!` and others are working correctly. This requires upping the compat-entry for DPPL, and so I figured we might as well also add re-export of `@submodel` in this PR.


src/Turing.jl
```diff
@@ -82,6 +82,7 @@ using StatsBase: predict
 # Turing essentials - modelling macros and inference algorithms
 export  @model,                 # modelling
         @varname,
+        @submodel,
```
devmotion, line 85: Is this intended? I'm fine with it but it seems a bit unrelated here :slightly_smiling_face:

src/inference/Inference.jl
```diff
@@ -356,13 +357,15 @@ function AbstractMCMC.bundle_samples(
...
+    chain = MCMCChains.Chains(
         parray,
         nms,
         (internals = extra_params,);
         evidence=le,
         info=info,
-    ) |> sort
+    )
+
+    return sort_chain ? sort(chain) : chain
```
devmotion, line 368: Ah I think I remember why it was removed - doesn't this create a type instability? IIRC `sample` can't be inferred due to limitations of Julia's logging system anyway, so I guess it doesn't matter :shrug:

src/Turing.jl
cpfiffer: Yeah I saw someone on Slack noted that it wasn't exported. I think it's fine to bundle in here.

cpfiffer:
We should probably go up a minor version since this is a big change from the UI perspective.

src/Turing.jl
torfjelde: Yep:) From above:

> This requires upping the compat-entry for DPPL, and so I figured we might as well also add re-export of @submodel in this PR.

devmotion: Ah missed your comment :+1:

src/inference/Inference.jl
torfjelde: Ah, but do we care? :confused:

devmotion: As I said, I am pretty sure it's unstable either way, so one should always use a function barrier if it's important. Hence I'm fine with the change :slightly_smiling_face:

devmotion (commented):
Looks good, maybe add a test as well?

torfjelde:
> Looks good, maybe add a test as well?

Will do :+1: Just waiting for the DPPL release and I'll add tests and get this through.

test/inference/Inference.jl
```diff
@@ -131,4 +131,12 @@
         @test mean(x[:s][1] for x in chains) ≈ 3 atol=0.1
         @test mean(x[:m][1] for x in chains) ≈ 0 atol=0.1
     end
+
+    @testset "chain ordering" begin
+        chain_sorted = sample(gdemo_default, Prior(), 1, sort_chain=true)
+        @test names(MCMCChains.get_sections(chain_sorted, :parameters)) == [:m, :s]
+
+        chain_unsorted = sample(gdemo_default, Prior(), 1, sort_chain=false)
+        @test names(MCMCChains.get_sections(chain_unsorted, :parameters)) == [:m, :s]
```
devmotion, line 140: Not completely surprising, this test fails. I guess it should be
```suggestion
        @test names(MCMCChains.get_sections(chain_unsorted, :parameters)) == [:s, :m]
```
?

devmotion: BTW it might be good to check the output without keyword argument as well, to check that the default value of `sort_chains` is set as expected.

devmotion: One additional thought: maybe `sort_parameters` would be more explicit since the chain is a 3D array and hence in principle could be sorted with respect to different axes?

torfjelde: Ehh yes:)

yebai:
nice work - thanks, @torfjelde and @devmotion!

PR #1628, opened 2021-05-31
Supporting MCMCTempering in Turing

HarrisonWilde:
Adds a dependency to Turing on MCMCTempering, and in return allows for all of its samplers to be wrapped with `Tempered` to apply the Parallel Tempering algorithm during execution. To do this I have added a file to the inference submodule that implements a minimal set of API requirements exposed by MCMCTempering and then exported the relevant functions.


Project.toml
```diff
@@ -20,6 +20,7 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Libtask = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MCMCChains = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
+MCMCTempering = "ce233488-44ea-4441-b732-192676ce2298"
```
devmotion, line 23: Can you add a `[compat]` entry as well and bump the Turing version?

src/Turing.jl
```diff
@@ -138,5 +138,10 @@ export  @model,                 # modelling
         elementwise_loglikelihoods,
         generated_quantities,
         logprior,
-        logjoint
+        logjoint,
+
+        get_densities_and_θs,     # Tempering 
+        make_tempered_model,
+        make_tempered_logπ,
+        get_θ
```
devmotion, line 146: Is it necessary to export these functions?

devmotion: In particular, I am not happy about `get_θ` and `get_densities_and_θs`: the API should not rely on unicode characters (so there should be at least an ASCII alternative) and also the names seem quite general and not limited to tempering.

src/inference/Inference.jl
```diff
@@ -61,8 +62,12 @@ export  InferenceAlgorithm,
         dot_observe,
         resume,
         predict,
-        isgibbscomponent
-
+        isgibbscomponent,
+        get_densities_and_θs,     # Tempering 
+        make_tempered_model,
+        make_tempered_logπ,
+        get_θ
```
devmotion, line 69: Same here, is this needed?

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+    sampler = MCMCTempering.TemperedSampler(
+        Sampler(sampler.internal_sampler, model),
+        sampler.Δ,
+        sampler.Δ_init,
+        sampler.N_swap,
+        sampler.swap_strategy
+    )
+
+    if resume_from === nothing
+        return AbstractMCMC.mcmcsample(rng, model, sampler, N;
```
devmotion, line 29: I guess it would be more natural to call `sample` (which might fall back to `mcmcsample` but this should be defined by MCMCTempering):
```suggestion
        return AbstractMCMC.sample(rng, model, sampler, N;
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+)
+    sampler = MCMCTempering.TemperedSampler(
+        Sampler(sampler.internal_sampler, model),
+        sampler.Δ,
+        sampler.Δ_init,
+        sampler.N_swap,
+        sampler.swap_strategy
+    )
+
+    return AbstractMCMC.mcmcsample(rng, model, sampler, ensemble, N, n_chains;
```
devmotion, line 56: Same here:
```suggestion
    return AbstractMCMC.sample(rng, model, sampler, ensemble, N, n_chains;
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+
+function AbstractMCMC.bundle_samples(
+    ts::Vector,
+    model::DynamicPPL.Model,
+    spl::MCMCTempering.TemperedSampler,
+    state::MCMCTempering.TemperedState,
+    chain_type::Union{Type{MCMCChains.Chains},Type{Vector{NamedTuple}}};
+    kwargs...
+)
+    return AbstractMCMC.bundle_samples(ts, model, spl.internal_sampler, state, chain_type; kwargs...)
```
devmotion, line 69: Maybe just define this in MCMCTempering for `model::AbstractMCMC.AbstractModel` and arbitrary chain type?

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+
+
+###########################
+#   MODEL
+###########################
+
+struct TemperedEval{T<:AbstractFloat}
```
devmotion, line 78: This seems a bit restrictive, could it be changed to
```suggestion
struct TemperedEval{T<:Real}
```
?

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+
+function (te::TemperedEval)(
+    rng::Random.AbstractRNG,
+    model::DynamicPPL.Model,
+    varinfo::DynamicPPL.AbstractVarInfo,
+    sampler::DynamicPPL.AbstractSampler,
+    context::DynamicPPL.AbstractContext,
+    args...
+)
+    context = DynamicPPL.MiniBatchContext(
```
devmotion, line 91: A bit clearer if one would use a different name here:
```suggestion
    minibatchcontext = DynamicPPL.MiniBatchContext(
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+
+
+###########################
+#   MODEL
+###########################
+
+struct TemperedEval{T<:AbstractFloat}
+    model :: DynamicPPL.Model
```
devmotion, line 79: I wanted to suggest to add a type parameter for `model` (currently it's a field with abstract type which should be avoided if possible) but it seems it is not actually used in the definition below (see my comment below for why accessing `model.f` is dangerous)?
```suggestion
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+    varinfo::DynamicPPL.AbstractVarInfo,
+    sampler::DynamicPPL.AbstractSampler,
+    context::DynamicPPL.AbstractContext,
+    args...
+)
+    context = DynamicPPL.MiniBatchContext(
+        context,
+        te.β
+    )
+    te.model.f(rng, model, varinfo, sampler, context, args...)
```
devmotion, line 95: I think this is dangerous - `model.f` is a very internal thing and we even avoid calling it explicitly in DynamicPPL in most places. We explicitly wrap it in https://github.com/TuringLang/DynamicPPL.jl/blob/5ffbbd235a1638b4b231bd1741e55c75e83215cc/src/model.jl#L150-L152 to ensure that the types of the inputs are correct, and even this function is not called directly but rather we use `model(...)` which ensures that evaluation is performed in a threadsafe way: https://github.com/TuringLang/DynamicPPL.jl/blob/5ffbbd235a1638b4b231bd1741e55c75e83215cc/src/model.jl#L85-L89

devmotion: In any case, it seems surprising to use `te.model.f` instead of `model.f`?

src/inference/tempering.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+"""
+    MCMCTempering.make_tempered_model
+
+Uses a `TemperedEval` to temper a `model`'s `f` by using a `DynamicPPL.MiniBatchContext` to inject `β` as a multiplier on evaluation
+"""
+function MCMCTempering.make_tempered_model(model::DynamicPPL.Model, β::T) where {T<:AbstractFloat}
```
devmotion, line 103: The type parameter is not needed:
```suggestion
function MCMCTempering.make_tempered_model(model::DynamicPPL.Model, β::Real)
```

src/inference/tempering.jl
HarrisonWilde: Yep my mistake, meant to remove this!

src/inference/tempering.jl
HarrisonWilde: Yes, this was our main concern with the approach to inject a context for Turing samplers, is there an obvious alternative here to achieve the same thing that would be safer? We concluded that as it was at least some external functionality it might be ok to have it work as it does.

As for your second question, I believe the call to te.model.f is required to avoid an infinite loop, as in this case the `Model`'s `F` is just the TemperedEval  call itself. We define a new model object with this wrapper but store the original model within it to be able to just alter the context and then call the original `model.f` with a new context via `te.model.f`, as to call  just `model.f` would be to precisely call on the `TemperedEval` object, hopefully that makes sense?

HarrisonWilde: i.e. the `model` in the args of `function (te::TemperedEval)` is the output of `make_tempered_model`, not the original model with functional `f` that we want to evaluate

src/Turing.jl
HarrisonWilde: Fair enough on unicode, wasn't aware of that design parameter, I will just rename to `get_params` and `get_tempered_densities_and_params`? As for the functionality, I agree its very general, and the former `get_params` function is certainly not limited to usage for tempering but is at least first implemented for this, is there somewhere else we could put it? Perhaps in the inference sub-module given it should act on any individual sampler offered by Turing?

HarrisonWilde: Then there is also a question of where to put `get_vi` which is similarly general but first required here it seems

src/inference/Inference.jl
HarrisonWilde: It should be yes, or MCMCTempering's functions would be unable to call them no?

src/inference/tempering.jl
HarrisonWilde: The problem I see here is that if you look at `Inference.jl`, the `Turing` specific AMCMC sample calls expect a `Sampler` subtyping `InferenceAlgorithm`, so we cannot call `sample` with a `TemperedSampler` and still pass through anything Turing specific here, hence why I just rewrote the final step of the chain so to speak but injecting the `Sampler` into the `internal_sampler` of the `TemperedSampler`.

src/Turing.jl
devmotion: My main question is: where are these functions needed? Are users expected to call them?

devmotion: Additionally, we already have functions such as `params`, `metadata`, and `getlogevidence` for querying transitions.

src/inference/Inference.jl
devmotion: I don't understand, can you explain why or how MCMCTempering would call these methods?

src/inference/tempering.jl
devmotion: Sorry, I don't understand. If you call `mcmcsample` you circumvent any implementation of `sample` for `MCMCTempering.TemperedSampler` and just call the fallback definition in AbstractMCMC directly which you would hit anyway if `sample` is not implemented for these types. So it's still not clear to me what you gain from calling `mcmcsample` here apart from possibly missing a special implementation of `sample` for these types.

src/inference/Inference.jl
HarrisonWilde: Yep you're right, haven't worked much with abstract functions before in Julia and had just misunderstood, will delete these from Inference and Turing

src/Turing.jl
```diff
@@ -138,5 +138,10 @@ export  @model,                 # modelling
         elementwise_loglikelihoods,
         generated_quantities,
         logprior,
-        logjoint
+        logjoint,
+
+        get_tempered_loglikelihoods_and_params,     # Tempering 
+        make_tempered_model,
+        make_tempered_loglikelihood,
+        get_params
```
devmotion, line 146: Sorry for being annoying but it is still unclear to me why these functions have to be exported - `get_params` is an internal function and should really not become part of the official API.

Maybe it's easier to understand and discuss if you show an example of how this is used in practice? BTW one should add such examples to the tests in any case :slightly_smiling_face:

src/inference/tempering.jl
HarrisonWilde: This is the last outstanding question for this PR, @devmotion @mohamed82008 any idea on how best to move forward with this? Also, thanks for all the help + comments David :)

src/Turing.jl
HarrisonWilde: Ah yes, that was another thing, I meant to ask, where is the `params` functions defined? If it is then yeah I can remove this altogether, and I perhaps resolved the comment and hid my own comment but I realised that yes these don't need to be exported, about to push an amended commit where they are no longer exported

src/inference/tempering.jl
devmotion: Ah I missed your replies here, let me check the implementation again.

BTW there's still https://github.com/TuringLang/Turing.jl/pull/1628#discussion_r642933638 that has to be resolved :slightly_smiling_face:

src/Turing.jl
devmotion: It is defined in Inference.jl.

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+end
+
+
+
+###########################
+#   MODEL
+###########################
+
+struct TemperedEval{T<:Real}
+    model :: DynamicPPL.Model
```
devmotion, line 67: This is part of my suggestions:
```suggestion
struct TemperedEval{T<:Real,F}
    f::F
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+    varinfo::DynamicPPL.AbstractVarInfo,
+    sampler::DynamicPPL.AbstractSampler,
+    context::DynamicPPL.AbstractContext,
+    args...
+)
+    minibatchcontext = DynamicPPL.MiniBatchContext(
+        context,
+        te.β
+    )
+    te.model.f(rng, model, varinfo, sampler, minibatchcontext, args...)
```
devmotion, line 83:
```suggestion
    te.f(rng, model, varinfo, sampler, minibatchcontext, args...)
```

src/Turing.jl
HarrisonWilde: I see yeah, I suppose maybe my version is justified if I am not exporting it, as in general we are taking in either a state (could be `HMCState` etc.) or a Transition, and then getting theta out, this is why I defined the `get_vi` helper as some states *are* the VarInfo whilst others contain it, then I just figured the sampler was the way to access the theta, but you are suggesting I instead should just call `getparams`?

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+end
+
+"""
+    MCMCTempering.make_tempered_model
+
+Uses a `TemperedEval` to temper a `model`'s `f` by using a `DynamicPPL.MiniBatchContext` to inject `β` as a multiplier on evaluation
+"""
+function MCMCTempering.make_tempered_model(model::DynamicPPL.Model, β::Real)
+    return DynamicPPL.Model(model.name, TemperedEval(model, β), model.args, model.defaults)
```
devmotion, line 92:
```suggestion
    return DynamicPPL.Model{DynamicPPL.getmissings(model)}(
        model.name,
        TemperedEval(model, β),
        model.args,
        model.defaults,
    )
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+    θkp1 = MCMCTempering.get_params(states[k + 1][2], sampler)
+    
+    return logπk, logπkp1, θk, θkp1
+end
+
+
+"""
+    MCMCTempering.make_tempered_loglikelihood
+
+Constructs the likelihood density function for a `model` weighted by inverse temperature `β`
```
devmotion, line 123:
```suggestion
Construct the likelihood density function for a `model` weighted by inverse temperature `β`.
```

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+    
+    θk = MCMCTempering.get_params(states[k][2], sampler)
+    θkp1 = MCMCTempering.get_params(states[k + 1][2], sampler)
+    
+    return logπk, logπkp1, θk, θkp1
+end
+
+
+"""
+    MCMCTempering.make_tempered_loglikelihood
```
devmotion, line 121: According to the convention and Julia style guide this should be the function signature (and usually lists of arguments etc. should be avoided).

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+
+    return logπ
+end
+
+
+"""
+    get_vi
+
+Returns the `VarInfo` given whatever a sampler's second return component is on an `AbstractMCMC.step`, in some cases this is the `VarInfo` itself, in others it must be accessed as a property of the state
+"""
```
devmotion, line 155: Also here the style could be fixed.

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+get_vi(state::Union{HMCState,GibbsState,EmceeState,SMCState}) = state.vi
+get_vi(vi::DynamicPPL.VarInfo) = vi
+
+
+"""
+MCMCTempering.get_params
+
+Uses the `sampler` to index the `VarInfo` and return the associated `θ` proposal
+"""
```
devmotion, line 164: Same here.

src/inference/tempering.jl
```diff
@@ -0,0 +1,166 @@
...
+MCMCTempering.get_params(state, sampler::DynamicPPL.Sampler) = get_vi(state)[sampler]
+# get_vi(states, k)[DynamicPPL.SampleFromPrior()]
```
devmotion, line 166:
```suggestion
```

devmotion (changes requested):
I made some suggestions regarding `TemperedEval`. I think it is the easiest approach for what you want to achieve but it touches some of the internals so one has to be careful (and changes in https://github.com/TuringLang/DynamicPPL.jl/pull/253 might break it).

Can you add some tests as well?

src/Turing.jl
devmotion: As long as it does not duplicate some existing functionality and it is internal, I am fine with it :slightly_smiling_face:

HarrisonWilde: Actually though, `getparams` returns a named tuple type thing, whereas I just want the ordered theta, so I think indexing by `sampler` is the only way to achieve this

src/inference/tempering.jl
HarrisonWilde: How should I fix the style for this one, is it acceptable to do something like:

```julia
"""
    get_vi(state) / get_vi(vi)

Returns the `VarInfo` given whatever a sampler's second return component is on an `AbstractMCMC.step`,
in some cases this is the `VarInfo` itself, in others it must be accessed as a property of the state.
"""
get_vi(state::Union{HMCState,GibbsState,EmceeState,SMCState}) = state.vi
get_vi(vi::DynamicPPL.VarInfo) = vi
```

HarrisonWilde:
I've tried to improve the style a bit, please comment again if there is more to be done as I am not sure what the optimal is here as I see a lot of different approaches to function docs in Julia packages.

Will write some tests later today and add them also.

src/inference/tempering.jl
devmotion: We follow the Julia standards for documentation and docstrings (or at least we should): https://docs.julialang.org/en/v1/manual/documentation/

In this example maybe:
```suggestion
"""
    get_vi(state)

Return the `VarInfo` object of the `state` of a sampler.

The `state` is the second component returned by `AbstractMCMC.step`. The `VarInfo` object can
be the `state` itself or a property of the `state`.
"""
```

src/inference/Inference.jl
```diff
@@ -62,7 +63,7 @@ export  InferenceAlgorithm,
         resume,
         predict,
         isgibbscomponent
-
+
```
devmotion, line 66:
```suggestion

```

devmotion:
> I've tried to improve the style a bit, please comment again if there is more to be done as I am not sure what the optimal is here as I see a lot of different approaches to function docs in Julia packages.

I tried to explain it in https://github.com/TuringLang/Turing.jl/pull/1628#discussion_r642977269.

test/inference/tempering.jl
```diff
@@ -0,0 +1,26 @@
...
+    @turing_testset "sampling" begin
+        Random.seed!(0)
+        N = 1000
+        samplers = (
+            HMC(0.1, 7),
+            MH(),
+            NUTS(0.65)
+        )
+        for s in samplers
+            ts = Tempered(s, 4)
```
devmotion, line 11: A Turing-unrelated comment: IMO the uppercase `T` suggests incorrectly that this is a constructor and `ts isa Tempered`. Usually, lowercase `tempered` is used in such situations. E.g., compare `Truncated` and `truncated` in Distributions: the first one always creates an object of type `Truncated` whereas the latter returns a truncated distribution that possibly is optimized for the unconstrained distribution and only falls back to `Truncated` if no better implementation exists. Users are explicitly advised to not use `Truncated`.

HarrisonWilde: That is a fair point, will perhaps change the functional form to `tempered` then in next MCMCTempering release. More tests to come by the way, been having technical issues for the past day so have some more stuff locally

test/inference/tempering.jl
```diff
@@ -0,0 +1,26 @@
...
+            @test typeof(ts) <: TemperedSampler
+            @test typeof(ts.internal_sampler) <: typeof(s)
+            
+            Random.seed!(0)
+            c1 = sample(gdemo_default, Tempered(s, 2), N)
+            # check_gdemo(c1)
```
devmotion, line 17: Does it fail currently?

torfjelde:
I'm going to close this for now, as we'll need a completely new PR after https://github.com/TuringLang/MCMCTempering.jl/pull/133 anyways.

PR #1636, opened 2021-06-10
Update to new DPPL version

torfjelde:
This PR makes Turing compatible with DynamicPPL@0.12.

EDIT: This should go after #1633 .


yebai:
I'm a bit surprised that we need to overlead `tilde_*` functions for ESS and MAP. It seems these algorithms only require a subset of three operations on the model, i.e. `logprior`, `loglikelihood` and `logdensity/logjoint`. Since these operations are already defined in `DynamicPPL` using contexts, is there a way to remove such overloads, or only overload sampler-facing functions like `assume/observe`?

torfjelde:
Likely, yeah, but IMO this should be done in a separate PR. This PR should be "minimal change necessary to make Turing.jl compatible with new DPPL".

torfjelde:
Also, it seems like SMC samplers are now broken? o.O Do any of you @yebai @devmotion have any idea as to why this is? I'm not at all familiar with APS so it's a bit difficult to debug.

yebai:
> Likely yeah, but IMO this should be done in a separate PR? This PR should be "minimal change necessary to make Turing.jl compatible with new DPPL".

Sounds good. 

> EDIT 2: Figured it out. observe(::Samler, ...) is never hit because DynamicPPL.tilde_observe(::DefaultContext, sampler, ...) = observe(...), i.e. requires a change in DPPL.

Great - I'll take a look at the DPPL PR.

torfjelde:
Ready with the :+1: once the last test passes?

devmotion (approved):
Looks good, just bump the Turing version number? Do we want it to be a breaking release as well?

torfjelde:
Good quesh. _Technically_ nothing that we export is breaking, buuuut we're exporting `DynamicPPL` so I don't know how that works.

torfjelde:
Maybe make it breaking just on the off-chance someone is actually overloading `Turing.tilde`, etc.?

devmotion:
As long as we do not export the methods that were broken in DynamicPPL, technically it is non-breaking. E.g., if we would reexport DynamicPPL or any of these methods it would be clear. It is a bit unclear how to deal with an exported module such as DynamicPPL... But generally if methods are only available internally as `Turing.tilde` etc. it does not require a breaking release.

torfjelde:
The boundary is blurred further by the fact that most people don't about DPPL but are using stuff from there directly through `Turing.whatever`.

BUT I was mistaken; we don't have `Turing.tilde`, etc. available, so we should be good :+1:

Issue #1648, opened 2021-06-17
Docstring for NUTS has a wrong name for a named arg


quildtide:
https://github.com/TuringLang/Turing.jl/blob/dd5ce7eded5bd804377bd67cfb84307aeec0aea8/src/inference/hmc.jl#L337

Under the existing API, the last named arg `ϵ`, should actually be `init_ϵ` when given as a named arg (and for that specific constructor).

On the other hand, the actual field in the struct is called `ϵ`, but changing this around would break the existing API.

EDIT: an alternative solution that possibly (?) improves the API while not breaking it involves allowing that particular constructor to also understand `ϵ` as a named argument, in addition to `init_ϵ`

yebai:
Fixed by https://github.com/TuringLang/Turing.jl/pull/1649

PR #1649, opened 2021-06-17
Fix docstring for epsilon named arg in NUTS

quildtide:
Simplest way to address #1648 (but maybe not the best)

Rather straightforward pull request if this is the route chosen to resolve the issue.

### DIFF
--- a/src/inference/hmc.jl
+++ b/src/inference/hmc.jl
@@ -334,7 +334,7 @@ end
 
 
 """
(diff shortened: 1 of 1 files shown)


cpfiffer:
I'm fine with this, anyone opposed?

yebai (approved):

yebai:
I think it's fine - although it's better to improve internal variable names in the future .

PR #1650, opened 2021-06-17
Fix PG

devmotion:
Follow-up to https://github.com/TuringLang/AdvancedPS.jl/pull/29.


cpfiffer (approved):

yebai:
It's surprising that the bugfix breaks some numerical tests.

devmotion:
It seems this PR got lost in my inbox.

The numerical errors are a bit annoying but since the corrected implementation leads to different samples maybe we were just unlucky. I increased the number of particles slightly from 10 to 15 and the number of iterations from 1000 to 5000, and now the test passes locally without adjusting the tolerance. In addition, I was able to lower tolerances in almost all other tests of Gibbs sampling, so it seems there's no general trend of worse or broken convergence.

yebai:
Sounds good, thanks @devmotion!

Issue #1654, opened 2021-06-29
Thinning and burnin isn't properly displayed in resulting chain


torfjelde:
```julia
julia> @model function demo()
           m ~ Normal()
       end
demo (generic function with 1 method)

julia> sample(demo(), NUTS(), 1_000, thinning=10, discard_initial=100)
┌ Info: Found initial step size
└   ϵ = 3.2
Sampling 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:04
Chains MCMC chain (1000×13×1 Array{Float64, 3}):

Iterations        = 1:1000
Thinning interval = 1
Number of chains  = 1
Samples per chain = 1000
...
```

Expected output:
```julia
Iterations        = 101:1000
Thinning interval = 10
Number of chains  = 1
Samples per chain = 900
```

Can probably be fixed by setting these values accordingly in the `bundle_samples`?

devmotion:
Yep, I think one just has to fix `bundle_samples`. IIRC they were not updated when these features were introduced in AbstractMCMC.

luiarthur:
Should the expected output actually be the following?
```julia
Iterations        = 101:1100
Thinning interval = 10
Number of chains  = 1
Samples per chain = 1000
```

cpfiffer:
Yeah, that's what I'd think it should be.

devmotion:
Hmm it's a bit inconsistent though since in both examples the `iterations` do not actually correspond to the original iterations? I think it should be
```julia
Iterations        = 101:10:10091
Thinning interval = 10
Number of chains  = 1
Samples per chain = 1000
```

devmotion:
The example also shows that the "thinning interval" in MCMCChains is a bit redundant, isn't it?

cpfiffer:
> The example also shows that the "thinning interval" in MCMCChains is a bit redundant, isn't it?

It really does, maybe it's worth just removing it? It's never been particularly relevant information IMO.

luiarthur:
Right, counting is difficult for me, apparently...

Yeah, it's probably redundant. However, I feel like having `Thinning interval` in the output is clearer. One could also say that `Samples per Chain` is redundant since it also can be inferred from `Iterations`.

Maybe another option is to remove `Iterations`? And instead show:

```julia
Discard initial = 100
Thinning interval = 10
Number of chains  = 1
Samples per chain = 1000
```

devmotion:
[...]

PR #1655, opened 2021-06-30
Fix iteration numbers of Chains and bugs in chains for Emcee

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1654 and two bugs related to Emcee (sorting of chains and parameter names).


cpfiffer (approved):
Happy to merge when tests pass.

PR #1656, opened 2021-07-02
Remove comment about dependencies

rikhuijzer:
This comment was added in 2017 (b696b57c05d12877d4038dcbd33abc58f5a5aef6). The codebase is not adhering to the comment at all, so it's probably better to remove the comment in order to not confuse readers.

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -1,13 +1,5 @@
 module Turing
 
-##############
-# Dependency #
-########################################################################
-# NOTE: when using anything from external packages,                    #
-#       let's keep the practice of explictly writing Package.something #
-#       to indicate that's not implemented inside Turing.jl            #
-########################################################################
-
 using Requires, Reexport, ForwardDiff
 using DistributionsAD, Bijectors, StatsFuns, SpecialFunctions
 using Statistics, LinearAlgebra


devmotion:
Actually, I change this e.g. in https://github.com/TuringLang/Turing.jl/blob/4449b56800e80060ecfaa6d9ec9c1f07ab759fdb/src/inference/Inference.jl#L23-L27 since it became too confusing and was not clear anymore which packages defined the different functions. So maybe it would be better to update the code :slightly_smiling_face:

yebai:
yes, I agree that it's helpful to gradually update the code to follow this guide. But this guide should probably live somewhere else, so also happy to merge this PR.

rikhuijzer:
> Actually, I changed this e.g. in
> 
> https://github.com/TuringLang/Turing.jl/blob/4449b56800e80060ecfaa6d9ec9c1f07ab759fdb/src/inference/Inference.jl#L23-L27
> 
> since it became too confusing and was not clear anymore which packages defined the different functions. So maybe it would be better to update the code 🙂

Ah okay. Yeah I didn’t check. My claim was just based on an earlier PR that I did here.

yebai (approved):

PR #1660, opened 2021-07-14
Updated HMC implementation for new AHMC version

torfjelde:
This PR makes Turing.jl compatible with the AdvancedHMC.jl's master branch, which will soon become a new release.


yebai:
@torfjelde It seems that the CI time increased due to some recent PRs, which leads to more often timeouts. Maybe splitting the `Turing-CI` workflow into two workflows to resolve the time limits of a single workflow?

torfjelde:
> @torfjelde It seems that the CI time increased due to some recent PRs, which leads to more often timeouts. Maybe splitting the `Turing-CI` workflow into two workflows to resolve the time limits of a single workflow?

Am fully aware; been trying to fix it for a couple of days now. I'm also not certain it's due to any of the recent PRs; it might be some other dependencies that slow things down.

torfjelde:
E.g. integration tests in DPPL passed within an hour in our last PR-merge just 3 days ago: https://github.com/TuringLang/DynamicPPL.jl/actions/runs/1031412449. Then it suddenly started timing out without any changes made to master: https://github.com/TuringLang/DynamicPPL.jl/actions/runs/1034731795.

torfjelde:
Uhhh I have no idea why but it seems like https://github.com/JuliaStats/Distributions.jl/pull/1362 is causing the freeze. I noticed it would halt when testing

https://github.com/TuringLang/Turing.jl/blob/d02919825f816712cd61bad31f3b720464d4612f/test/test_utils/models.jl#L73-L77

for `ESS` sampler. Removing this and everything runs as per usual, it seems.

yebai:
> Uhhh I have no idea why but it seems like JuliaStats/Distributions.jl#1362 is causing the freeze. I noticed it would halt when testing

cc @devmotion

torfjelde:
David is one vacation so not sure he'll respond :confused: 

Also, instead look at #1663 where I've added some more info.

Should we maybe upper-bound Turing.jl to Distributions@0.12.10 too, rather than just the tests? Worried peoples code might be running into similar scenarios (though I still don't understand why it fails).

yebai:
> Should we maybe upper-bound Turing.jl to Distributions@0.12.10 too, rather than just the tests? Worried peoples code might be running into similar scenarios (though I still don't understand why it fails).

Happy to upper-bound Distributions before we figure out why CI times out.

Issue #1674, opened 2021-08-04
ADVI fails for single parameter models


bgroenks96:
It appears that, if a model has only one parameter, `ADVI` currently fails. MWE:

```julia
using Turing
@model function simple_model(x)
         μ ~ Normal(0,1)
         x ~ Normal(μ,1)
end
model = simple_model(randn(100))
advi = ADVI(10,10_000)
q = vi(model, advi)
```
Error:

```
ERROR: BoundsError: attempt to access 0-element Vector{Vector{ForwardDiff.Dual{ForwardDiff.Tag{AdvancedVI.var"#f#19"{AdvancedVI.ELBO, ADVI{AdvancedVI.ForwardDiffAD{40}}, MultivariateTransformed{DistributionsAD.TuringDiagMvNormal{Vector{Float64}, Vector{Float64}}, Stacked{Tuple{Identity{0}}, Vector{UnitRange{Int64}}}}, Turing.Variational.var"#logπ#2"{DynamicPPL.Model{var"#1#2", (:x,), (), (), Tuple{Vector{Float64}}, Tuple{}}, DynamicPPL.TypedVarInfo{NamedTuple{(:μ,), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:μ, Tuple{}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:μ, Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}}}, Float64}}, Tuple{Int64}}, Float64}, Float64, 2}}} at index [1]
Stacktrace:
  [1] getindex
    @ ./array.jl:801 [inlined]
  [2] mapvcat
    @ ~/.julia/packages/Bijectors/Wa0sY/src/Bijectors.jl:95 [inlined]
[...]

PR #1677, opened 2021-08-05
Fix for #1674

torfjelde:


src/variational/advi.jl
```diff
@@ -6,9 +6,9 @@ denoting the dimensionality of the latent variables.
 """
 function Bijectors.bijector(
     model::DynamicPPL.Model,
-    ::Val{sym2ranges} = Val(false),
+    ::Val{sym2ranges} = Val(false);
+    varinfo::DynamicPPL.AbstractVarInfo = DynamicPPL.VarInfo(model)
```
torfjelde, line 10:
```suggestion
    varinfo = DynamicPPL.VarInfo(model)
```

torfjelde:
Could we also get this merged in the same go @cpfiffer @yebai ?

cpfiffer (approved):
Absolutely, LGTM

torfjelde:
Am I blind? I really can't tell why this is failing. It tells me that there's invalid syntax near `varname` in

https://github.com/TuringLang/Turing.jl/blob/ed82b4dabe20e2f590ecc104d0d175dbd4377358/src/variational/advi.jl#L7-L11

but I cannot see why this is the case o.O

cpfiffer:
Uh wow I have no idea?

PR #1679, opened 2021-08-12
Fix broken markup in variational_inference.md

feynmanliang:


torfjelde (approved):
Wonderful; thanks @feynmanliang !

torfjelde (changes requested):
Oh actually, I was a bit trigger happy here. So we don't want to make changes to the `markdown` files, but instead to the `.jmd` files in `tutorials/`.

torfjelde:
And I just had a look at the rendered version of the current docs; this section is completely messed up and requires quite a few changes to ensure that the math renders correctly. I might just take a stab at this myself (unless you want to:) ). Thanks for bringing this to my attention!

feynmanliang:
Yeah you are welcome to take over :)

Most of the markup errors are due to doubly escaped backslashes.

torfjelde:
>  when talking about .jmd files are you referring to the ones in

Correct; the ones you changed is directly in the turing.ml website. Tutorials in TuringTutorials are also rendered on the website, but these are independent and here we use `.jmd` files as the "source" and `.md` is just a target we export to (for example to allow easy rendering on the website).

> the "Edit me" link is broken - would it be possible to change that so it links to the right place in the TuringTutorials repo instead?

Ah, wasn't aware of this! @cpfiffer you know if this can be fixed?

cpfiffer:
The fix for the "Edit Me" link needs to be here:

https://github.com/TuringLang/turing.ml/blob/1837e6921c65e3f05ad17f98217e03a726ecad33/_includes/sidebar.html#L126

I suspect with a couple if statements it could be made to redirect to the correct location.

yebai:
@ParadaCarleton Could you take a look at what's missing before it got merged?

PR #1686, opened 2021-08-24
Upper bound Reexport.jl to fix compilation issue

torfjelde:
https://github.com/simonster/Reexport.jl/pull/33#issuecomment-904564731


Project.toml
```diff
@@ -50,7 +50,7 @@ ForwardDiff = "0.10.3"
 Libtask = "0.4, 0.5.3"
 MCMCChains = "4"
 NamedArrays = "0.9"
-Reexport = "0.2, 1"
+Reexport = "< 1.2.1"
```
devmotion, line 53: This reintroduces compatibility with Reexport 0.1. We should better state explicitly
```suggestion
Reexport = "0.2, ~1.0, ~1.1, =1.2.0"
```
(or something similar)

torfjelde: I realize but don't you think that doesn't matter at all? As in nobody uses Reexport@0.1.

devmotion: But people that still use it might be offered this version which (it seems) is actually not compatible.

devmotion:
Another fix would be to rename `Turing.Core` to something else I assume :stuck_out_tongue:

torfjelde:
> Another fix would be to rename Turing.Core to something else I assume

But isn't this a bug on the end of Reexport.jl though? In the sense that this is allowed by the language.

devmotion:
Yes, I guess so. I assume it would be sufficient to interpolate `Core` in the expression.

Project.toml
torfjelde: True, but how long do we keep that compat entry? Like, in 10 years, do we still have it around because someone might try to do that? :sweat_smile: 

I'll commit your change though, np :+1:

torfjelde:
> Yes, I guess so. I assume it would be sufficient to interpolate Core in the expression.

In which expression?

devmotion:
In the following way: https://github.com/simonster/Reexport.jl/pull/35

devmotion:
Maybe we can skip this PR if the PR to Reexport is merged and released without too much delay. The problem with such bounds in Turing in general is that it is not clear that it will help users without manual intervention. If they just `] up`, the Pkg resolver might decide that it's best to not downgrade Reexport and they would not end up with the new version of Turing. To fix the issue they should therefore perform a more explicit `] add Turing@0.17.2` - but then they could equally well just run an `] add Reexport@1.2.0` :shrug:

torfjelde:
> In the following way

Ah gotcha!:)

> Maybe we can skip this PR if the PR to Reexport is merged and released without too much delay. 

But even then we still need to ensure that we're not compat with 1.2.1, right? So yes but we still need a PR to fix bounds.

>  the Pkg resolver might decide that it's best to not downgrade Reexport and they would not end up with the new version of Turing.

Sure, but the big thing here is "might". Most will have the same bound for Reexport as we currently have, i.e. `1`. This will then work nicely with this version bound.

devmotion:
The main point is: this PR does not help if users/packages/etc. don't end up with the new Turing version. And if they already use Reexport@1.2.1 since e.g. they updated their environment, then it is not guaranteed that they end up with the new version since Reexport would have to be downgraded.

The Reexport bugfix release would solve these issues since the problem would be solved by just running `] up` without downgrading any packages.

The only way to ensure that Reexport@1.2.1 never shows up in any installations is to change the compatibilities in the general registry. Even this PR won't guarantee that some users don't end up with Reexport@1.2.1 since older versions of Turing are compatible with it (at least officially).

torfjelde:
> The main point is: this PR does not help if users/packages/etc. don't end up with the new Turing version. And if they already use Reexport@1.2.1 since e.g. they updated their environment, then it is not guaranteed that they end up with the new version since Reexport would have to be downgraded.

Hmm, how would the following end up with you did `] up`?

```toml
Turing = "0.17"
Reexport = "1"
```

Surely `Pkg` gives precedence to the package which explicitly restricts the other, no? I.e. it results in the most up-to-date Turing version.

devmotion (approved):
I don't know, Pkg and its decisions are still a mystery to me :smile: I just spent almost an hour trying to debug a Pkg weirdness and I still don't know what's going on. My gut feeling is that upper bounding package versions in new releases while they are unbounded in previous ones is no golden bullet. For instance, people opened issues all the time about some compilation error that was fixed in recent versions but somehow they all ended up with outdated versions of Turing and actually had to force the update to a more recent one. In the end I just fixed all compatibility bounds in the registry and it seems this worked :smile: 

That being said, I am not generally opposed to this PR but I think the better solution would be a bugfix release of Reexport :slightly_smiling_face:

torfjelde:
Aaah, okay :sweat_smile: Yeah I kind of just assumed it would do the above. And the isuse you're talking about sounds awful..

But I'll merge for now since it's holding up PRs in other packages :+1:

ararslan:
FYI, Reexport v1.2.2 has been released which fixes the issue with Turing, and v1.2.1 has been yanked from General. Since Turing requires Julia 1.3, it should be safe to revert to a compat entry of `Reexport = "1"`

devmotion:
Thanks, I already opened a PR 🙂

PR #1689, opened 2021-08-24
Reduce and speed up CI

devmotion:
Similar to https://github.com/TuringLang/DynamicPPL.jl/pull/317. Julia 1.3 is only tested with Ubuntu, and macOS and Windows are only tested with `JULIA_NUM_THREADS = 2`. Moreover, avoids unnecessary coverage analysis in cases where the result is not uploaded.


torfjelde (approved):
:heart:

PR #1692, opened 2021-08-28
Update CompatHelper.yml

devmotion:
I was surprised that CompatHelper did not open a PR for MCMCChains 5 and noticed that the action failed. This PR updates the action to the recommended action for CompatHelper 3 (https://github.com/JuliaRegistries/CompatHelper.jl#github). I also enable automatic version bumps by CompatHelper.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -2,17 +2,32 @@ name: CompatHelper
 
 on:
   schedule:
-    - cron: '00 00 * * *'
+    - cron: 0 0 * * *
   workflow_dispatch:
 
 jobs:
   CompatHelper:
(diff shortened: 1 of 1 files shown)


devmotion:
I ran the action manually with this PR to check if it works: https://github.com/TuringLang/Turing.jl/runs/3451980931?check_suite_focus=true

devmotion:
I opened an issue in the CompatHelper repo since even with the official configuration an error is thrown.

devmotion:
The problem is fixed in CompatHelper 3.0.8, the action runs successfully now: https://github.com/TuringLang/Turing.jl/runs/3452227042?check_suite_focus=true

cpfiffer (approved):

Issue #1725, opened 2021-11-09
Invalid test case for `Gibbs`?


torfjelde:
I'm currently updating Turing.jl to be compatible with the new version of DynamicPPL.jl and ran into something that seems like a bug.

Currently, we implement the logdensity-evaluation required for MH here:

https://github.com/TuringLang/Turing.jl/blob/d1d4071b603d60d6ba969f3b14bfb347795644b7/src/inference/mh.jl#L251-L263

But now that we have a clear separation between sampling (`SamplingContext`) and evaluation (`DefaultContext`), this seems like it should now be:

```julia
function (f::MHLogDensityFunction)(x)
    sampler = f.sampler
    vi = f.vi

    x_old, lj_old = vi[sampler], getlogp(vi)
    set_namedtuple!(vi, x)
    f.model(vi, DefaultContext()) # CHANGED
    lj = getlogp(vi)
    vi[sampler] = x_old
    setlogp!(vi, lj_old)

    return lj
end
```

To ensure that we're only evaluating, not sampling.

But making this change results in the following test-case to fail:

https://github.com/TuringLang/Turing.jl/blob/d1d4071b603d60d6ba969f3b14bfb347795644b7/test/inference/gibbs.jl#L91-L111

It complains that a component of `m` is missing. The issue is that once `z` has been sampled, the number of components has changed and so the size of `m` has changed. Currently this test will pass because `MH` will just silently sample this component from the prior _when evaluating the logpdf of the proposed state_.

It's unclear whether:
1. This was intended in the first place.
2. If it's even valid to do.

cpfiffer:
Uh yeah that seems quite wrong. The line you added

```julia
f.model(vi, DefaultContext()) # CHANGED
```

seems correct to me and it should be merged in. 

As to the failing test, I am not sure -- I don't recall working with this model but I would expect this behavior to be supported.

torfjelde:
@yebai Mentioned that it _might_ "accidentally" end up doing the correct thing, but even so we probably should do this in a separate/explicit way rather than implicit like currently done.

yebai:
Closed in favour of https://github.com/TuringLang/Turing.jl/issues/1904

PR #1726, opened 2021-11-09
Update DynamicPPL.jl version

torfjelde:
This is a sibling-PR of https://github.com/TuringLang/DynamicPPL.jl/pull/309.

The aim here is to make the bare-minimum changes to ensure that everything works as before. I've not made any significant attempts at making samplers compatible with immutable implementations of `AbstractVarInfo`, e.g. `SimpleVarInfo`, but rather just marked these lines with a `# TODO: make compatible with immutable vi` which we can then address in a separate PR.

Related: https://github.com/TuringLang/Turing.jl/issues/1725.

EDIT: Tests are passing locally, but they won't pass here until the PR in DPPL has been merged and released.


src/inference/AdvancedSMC.jl
```diff
@@ -342,17 +342,17 @@ function DynamicPPL.assume(
...
             r = rand(rng, dist)
-            push!(vi, vn, r, dist, DynamicPPL.Selector(:invalid))
+            push!!(vi, vn, r, dist, DynamicPPL.Selector(:invalid))
         end
         lp = logpdf_with_trans(dist, r, istrans(vi, vn))
-        acclogp!(vi, lp)
+        acclogp!!(vi, lp)
     end
-    return r, 0
+    return r, 0, _vi
```
torfjelde, line 350: Is this the right thing to do @yebai @devmotion ?

Essentially the issue is that when we're using `PG` with threads, `_vi` is a `ThreadSafeVarInfo` while `vi = AdvancedPS.current_trace().f.varinfo` is _not_. And so if we return `vi` instead of `_vi`, subsequent code in `evaluate_threadsafe!!` will fail.

I'm not too familiar with the AdvancedPS stuff so it's a bit unclear to me what I should be returning here, e.g. is the above okay?

torfjelde: Or should I overload `evaluate!!` and all the corresponding methods?

If we're doing this, we might want to consider making the "outer" calls, e.g. `(::Model)(...)` a bit more general, e.g. dispatch on `AbstractProbabilisticProgram` instead and then _just_ overloading `evaluate_threadsafe!!` and `evaluate_threadunsafe!!` for `TracedModel`.

yebai: Let's perhaps disallow the use of `ThreadSafeVarInfo` with `PG`/`SMC` for now. These samplers are not yet ready to run in threads parallelism due to the interaction of particles (i.e. the `resampling` step). I suggest that we simply throw an error here explaining that `ThreadSafeVarInfo` support for `PG` is to be supported in the near future.

torfjelde: Well, there currently isn't a way to allow/disallow specific combinations like this. And erroring or even just telling people not to use `PG`/`SMC` just because they are running with `--threads=4` seems wrong since they might not even be using threading inside of the model.

We _could_ add a `isthreadsafe(varinfo, context)` check in https://github.com/TuringLang/DynamicPPL.jl/blob/57c50f1f46aa5e51ec2125f3ae4883c61e59e9c3/src/model.jl#L391 and then implement this specifically for `SamplingContext{<:PG}`, etc.

devmotion: > We _could_ add a `isthreadsafe(varinfo, context)` check in https://github.com/TuringLang/DynamicPPL.jl/blob/57c50f1f46aa5e51ec2125f3ae4883c61e59e9c3/src/model.jl#L391 and then implement this specifically for `SamplingContext{<:PG}`, etc.

Yes, I think this would be a better approach. Then we can always choose the non-threadsafe VarInfo for these samplers instead of erroring if they run Julia with multiple threads (e.g. I always do, so I think it would be very inconvenient to throw an error here).

The only possible problem is that then it would be possible to use `@threads` with a non-threadsafe VarInfo which can lead to incorrect result silently. I don't think there's a clean way to detect if a model tries to use multiple threads (since it is not as simple as detecting a `@threads` macro call). I wonder if it is possible to add a locking mechanism to the default non-threadsafe VarInfo that makes it threadsafe while not impacting performance if it is run with a single thread. Surely it would be slower if multiple threads are used but with multiple threads we would use the more performant ThreadSafeVarInfo anyway - if we don't use PG etc. The main motivation would be to avoid silent problems with multithreaded models when PG + a non-threadsafe VarInfo is used - it seems fine if it is less performant since it is not recommended.

torfjelde: > Yes, I think this would be a better approach. Then we can always choose the non-threadsafe VarInfo for these samplers instead of erroring if they run Julia with multiple threads (e.g. I always do, so I think it would be very inconvenient to throw an error here).

I'll do that then :+1:

> The only possible problem is that then it would be possible to use @threads with a non-threadsafe VarInfo which can lead to incorrect result silently.

Yup :confused: We could potentially address this by using atomics (though it would probably require very recent Julia versions to be able to do this properly) which I expect would behave just like a `Ref` in the single-threaded case, or even just making the default `VarInfo` threadsafe by using an array instead of `Ref` for the logp field.

But it's worth mentioning that currently `PG`, etc. already silently does the wrong thing, doesn't it?

torfjelde:
Tests are now passing :+1: This should now be ready for a review.

yebai (approved):
Looks good! To clarify, this PR only updates the DynamicPPL API, but does not make use of `SimpleVarInfo` yet, right?

torfjelde:
Correct!

src/inference/hmc.jl
```diff
@@ -492,11 +492,10 @@ function DynamicPPL.dot_assume(
...
     vi,
 )
-    @assert length(dist) == size(var, 1)
-    DynamicPPL.updategid!.(Ref(vi), vns, Ref(spl))
-    r = vi[vns]
-    var .= r
-    return var, sum(logpdf_with_trans(dist, r, istrans(vi, vns[1])))
+    foreach(vns) do vn
+        DynamicPPL.updategid!(vi, vn, spl)
+    end
```
torfjelde, line 497:
```suggestion
    DynamicPPL.updategid!.(Ref(vi), vn, Ref(spl))
```

src/inference/emcee.jl
```diff
@@ -38,23 +38,26 @@ function AbstractMCMC.step(
...
 
     # Compute initial transition and states.
     transition = map(vis) do vi
         Transition(vi)
     end
+    # TODO: Make compatible with immutable `AbstractVarInfo`.
+    foreach(vis) do vi
+        # Transform to unconstrained space.
+        DynamicPPL.link!(vi, spl)
+    end
```
torfjelde, line 57:
```suggestion
    # Transform to unconstrained space.
    DynamicPPL.link!.(vis, spl)
```

src/inference/hmc.jl
```diff
@@ -506,10 +505,10 @@ function DynamicPPL.dot_assume(
     var::AbstractArray,
     vi,
 )
-    DynamicPPL.updategid!.(Ref(vi), vns, Ref(spl))
-    r = reshape(vi[vec(vns)], size(var))
-    var .= r
-    return var, sum(logpdf_with_trans.(dists, r, istrans(vi, vns[1])))
+    foreach(vns) do vn
+        DynamicPPL.updategid!(vi, vn, spl)
+    end
```
torfjelde, line 510:
```suggestion
    DynamicPPL.updategid!(Ref(vi), vns, Ref(spl))
```

torfjelde:
@devmotion you happy with this?

Btw, should we also re-export `condition`, `decondition`, and `conditioned` from Turing now?

devmotion:
I'll have a look this afternoon 👍

I guess we don't have any prominent documentation yet and don't use it in any tutorials or examples? I think maybe we could add them and export these function in a single PR?

torfjelde:
> 

Bueno :+1: 

And yeah, that's a fair point. Next on my TODO list is writing a tutorial/some docs on the recent features, e.g. `@submodel`, `condition`, etc. and so I guess we can export it then. Though the only thing is that those PRs will be going to different projects than Turing itself. But anyhow, it should probably be a separate PR so let's leave it out of this one :+1:

src/core/ad.jl
```diff
@@ -108,9 +108,11 @@ function gradient_logp(
     logp_old = getlogp(vi)
     function f(θ)
         new_vi = VarInfo(vi, sampler, θ)
-        model(new_vi, sampler, ctx)
+        new_vi = last(DynamicPPL.evaluate!!(model, new_vi, sampler, ctx))
         logp = getlogp(new_vi)
-        setlogp!(vi, ForwardDiff.value(logp))
+        # Don't need to capture the resulting `vi` since this is only
+        # needed if `vi` is mutable.
+        setlogp!!(vi, ForwardDiff.value(logp))
```
devmotion, line 115: I guess ideally we would want a no-op here when `setlogp!!` returns a new object - if we just call `setlogp!!` with a SimpleVarInfo it would create a new VarInfo object even though it's never used. Maybe one could check if `vi === new_vi` and only call `setlogp!!` if it is `true` (assuming that in this case `setlogp!!` would not return a new object)?

But I guess this should be left for the SimpleVarInfo PR?

src/core/container.jl
```diff
@@ -43,7 +43,7 @@ function AdvancedPS.reset_model(f::TracedModel)
 end
 
 function AdvancedPS.reset_logprob!(f::TracedModel)
-    DynamicPPL.resetlogp!(f.varinfo)
+    DynamicPPL.resetlogp!!(f.varinfo)
```
devmotion, line 46: This also relies on the fact that it mutates `f.varinfo`.

src/inference/AdvancedSMC.jl
```diff
@@ -110,8 +110,8 @@ function DynamicPPL.initialstep(
     # Reset the VarInfo.
     reset_num_produce!(vi)
     set_retained_vns_del_by_spl!(vi, spl)
-    resetlogp!(vi)
-    empty!(vi)
+    resetlogp!!(vi)
+    empty!!(vi)
```
devmotion, line 114: Does `empty!!` not call `resetlogp!!`? Maybe this would be reasonable? In any case, also here it is mandatory that the calls are actually mutating.

devmotion: Similar to the calls below.

src/inference/emcee.jl
```diff
@@ -38,23 +38,24 @@ function AbstractMCMC.step(
...
         end
     end
 
     # Compute initial transition and states.
     transition = map(vis) do vi
         Transition(vi)
     end
+    # TODO: Make compatible with immutable `AbstractVarInfo`.
+    # Transform to unconstrained space.
+    DynamicPPL.link!.(vis, Ref(spl))
```
devmotion, line 55: Why did you extract it? It seems there are no performance gains but rather now there are both a broadcasting and a map operation whereas before we just used a single `map`?

devmotion (commented):
Just had some minor questions, but looks good overall :+1: I guess most questions are related to the SimpleVarInfo integration which will be done in a separate PR.

src/core/ad.jl
torfjelde: Will this not be optimized away in the case where `vi` is immutable?
I sort of assumed so. But I guess we might as well give the compiler a helping hand just to be sure.

src/core/container.jl
torfjelde: Yup, but `TracedModel` isn't compatible with immutable `AbstractVarInfo` :confused:

src/inference/emcee.jl
torfjelde: I extracted it because at first I intended to make it compatible with immutable `AbstractVarInfo`, but then I realized this shouldn't be done in this PR :upside_down_face: I'll revert this change I guess :+1:

src/inference/AdvancedSMC.jl
torfjelde: Same response as above: these samplers aren't compatible with immutable `VarInfo` :confused:

torfjelde:
Thanks! I'll merge once tests pass.

devmotion (approved):

PR #1728, opened 2021-11-21
Fix incorrect docs use init_params to initialize sampling

dlakelan:
fix the outdated reliance on init_theta to specify initial parameter values in favor of up to date init_params


yebai (approved):

yebai:
many thanks @dlakelan!

PR #1739, opened 2021-12-10
Bump DynamicPPL to 0.16

torfjelde:


Project.toml
```diff
@@ -46,7 +45,7 @@ DataStructures = "0.18"
 Distributions = "0.23.3, 0.24, 0.25"
 DistributionsAD = "0.6"
 DocStringExtensions = "0.8"
-DynamicPPL = "0.15"
+DynamicPPL = "0.15, 0.16"
```
rikhuijzer, line 48: Possibly a stupid question, but shouldn't this be pinned to only `0.16` to avoid having to support two versions?

yebai:
Is this PR a duplicate of https://github.com/TuringLang/Turing.jl/pull/1726?

torfjelde:
> Is this PR a duplicate of #1726?

No. That's for DPPL 0.17. I just realized we never actually updated Turing to allow DPPL 0.16, which is what this PR does.

yebai (approved):

src/inference/ess.jl
```diff
@@ -137,24 +137,24 @@ function (ℓ::ESSLogLikelihood)(f)
     return getlogp(varinfo)
 end
 
-function DynamicPPL.tilde_assume(rng::Random.AbstractRNG, ctx::DefaultContext, sampler::Sampler{<:ESS}, right, vn, inds, vi)
+function DynamicPPL.tilde_assume(rng::Random.AbstractRNG, ctx::DefaultContext, sampler::Sampler{<:ESS}, right, vn, vi)
```
devmotion, line 140: I would assume these changes break support of DynamicPPL 0.15 so it should be removed from the compat entry?

torfjelde: Ah yes, good catch :+1:

Project.toml
torfjelde: It's not an issue as long as it's not a lie, i.e. we're actually compatible with both.

And the idea isn't to have to support two versions: if it turns out that 0.15 doesn't work anymore, then we just remove the compat entry.

PR #1740, opened 2021-12-11
Restrict Julia version in Project.toml

torfjelde:
Do we actually support Julia 1.4 and 1.5? Or is it only Julia 1.3 and 1.6?


Project.toml
```diff
@@ -59,4 +59,4 @@ StatsBase = "0.32, 0.33"
 StatsFuns = "0.8, 0.9"
 Tracker = "0.2.3"
 ZygoteRules = "0.2"
-julia = "1.3"
+julia = "1.3, 1.4, 1.5, 1.6"
```
rikhuijzer, line 62:
```suggestion
julia = "1.6"
```

rikhuijzer:
Usually, Julia developers run tests for some lower bound and upper bound version and if those pass assume that everything in between also works.

devmotion:
AFAIK we supported Julia >= 1.3, including 1.4 and 1.5, we just only tested with the oldest compatible version and the stable Julia version only to reduce CI.

Even with this PR users with Julia 1.7 will be able to install older releases of Turing. If we want to avoid this we would have to change the compat entry for older versions in the registry directly.

Project.toml
devmotion: We don't have to be more restrictive than necessary 🤷‍♂️ We can easily drop support for < 1.6 when it requires special effort to support older versions. As long as it just works, I don't think we have to remove them.

devmotion (approved):

rikhuijzer:
The package [RetroCap.jl](https://github.com/JuliaRegistries/RetroCap.jl) might be useful.

PR #1749, opened 2021-12-15
Set `DropEntry` for CompatHelper

rikhuijzer:
When looking through recent PRs, old compat entries are always dropped when updating compat entries. This PR sets  `entry_type=DropEntry()`, so that CompatHelper will drop existing compat entries.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -25,8 +25,11 @@ jobs:
         shell: julia --color=yes {0}
       - name: "Run CompatHelper"
         run: |
-          import CompatHelper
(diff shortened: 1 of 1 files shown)


devmotion:
Hm I don't think that we usually want to do this - the recent PRs just happened to be updates of packages with many breaking changes that also affected Turing. If possible, I think we should keep compatibilities with previous versions and hence not dropping them would be the better default. For instance, there is no good reason for dropping support for SpecialFunctions 1 when upgrading to SpecialFunctions 2. In fact, it would be very disruptive for the package ecosystem if packages would remove compatibility with SpecialFunctions 1.

rikhuijzer:
Ok :+1:

Issue #1756, opened 2021-12-23
UndefVarError: CodeInfo not defined


andreasko:
Hey all,

after switching Julia from 1.6.2 to 1.6.5, I experience problems when using ReverseDiff in combination with Memoization, The following MWE fails:

```julia
using Turing
using Memoization
using ReverseDiff
setadbackend(:reversediff)
Turing.setrdcache(true)


@model function demo(xs)
   s ~ InverseGamma(2, 3)
   m ~ Normal(10, √s)
   xs .~ Normal(m, √s)
end
model = demo(randn(10));
chain = sample(model, NUTS(), 10);
```

The full error message is

```
UndefVarError: CodeInfo not defined
in eval at base/boot.jl:360 
in top-level scope at test_regional.jl
in top-level scope at base/timing.jl:210 
in macro expansion at test_regional.jl:112 
in  at Turing/uMQmD/src/inference/Inference.jl:132
in #sample#1 at Turing/uMQmD/src/inference/Inference.jl:132 
in sample##kw at Turing/uMQmD/src/inference/Inference.jl:142 
in #sample#2 at Turing/uMQmD/src/inference/Inference.jl:142 
in sample##kw at Turing/uMQmD/src/inference/hmc.jl:116 
in var"#sample#40" at Turing/uMQmD/src/inference/hmc.jl:133
in  at AbstractMCMC/BPJCW/src/sample.jl:107
in var"#mcmcsample#20" at AbstractMCMC/BPJCW/src/sample.jl:114
in macro expansion at AbstractMCMC/BPJCW/src/logging.jl:8 
in macro expansion at ProgressLogging/6KXlp/src/ProgressLogging.jl:328 
in macro expansion at AbstractMCMC/BPJCW/src/sample.jl:123 
in step##kw at DynamicPPL/RcfQU/src/sampler.jl:62 
in var"#step#17" at DynamicPPL/RcfQU/src/sampler.jl:87
in  at Turing/uMQmD/src/inference/hmc.jl:153
in var"#initialstep#41" at Turing/uMQmD/src/inference/hmc.jl:167
in phasepoint at AdvancedHMC/HQHnm/src/hamiltonian.jl:153 
in phasepoint at AdvancedHMC/HQHnm/src/hamiltonian.jl:76 
in ∂H∂θ at AdvancedHMC/HQHnm/src/hamiltonian.jl:31
in ∂logπ∂θ at Turing/uMQmD/src/inference/hmc.jl:433 
[...]

PR #1758, opened 2022-01-07
Fix for HMCs `dot_assume`

torfjelde:
Currently `dot_assume` for HMC is broken, which this PR fixes.

This case is covered by tests, but (AFAIK) only in DynamicPPL's integration tests, _not_ in Turing's tests.

Should I copy-paste these or is there a way we can just run the integration tests from DPPL with this Turing version?


devmotion:
> Should I copy-paste these or is there a way we can just run the integration tests from DPPL with this Turing version?

I think the test setup is simpler and more reliable if you just copy the tests (I guess we could use the test models but probably that's what's done already in DynamicPPL?).

torfjelde:
Eeeh weird. I can't reproduce the errors locally :flushed:

torfjelde:
Aaaaight so I finally figured out what's going wrong:

``` julia
julia> using Tracker, Distributions

julia> m = first(param(zeros(1)))
0.0 (tracked)

julia> s = first(param(ones(1)))
1.0 (tracked)

julia> typeof(Normal.(m, s))
Tracker.Tracked{Normal{Float64}}
```

And DistributionsAD.jl doesn't address this issue:

``` julia
julia> using DistributionsAD

julia> typeof(Normal.(m, s))
Tracker.Tracked{Normal{Float64}}
```

It comes down to Tracker not handling `broadcast` properly. I'm just going to mark this test as broken for now (don't think anyone uses Tracker with Turing anyways) and raise an issue over at Tracker.jl.

devmotion:
Maybe related to or even the same as https://github.com/FluxML/Tracker.jl/issues/65?

torfjelde:
Possibly! I've made a new issue as it's not completely obvious to me whether it's the same or not atm: https://github.com/FluxML/Tracker.jl/issues/119

torfjelde:
Now running into https://github.com/TuringLang/Turing.jl/issues/1756

torfjelde:
Renamed `Core` to `TuringCore`. Should we maybe bump the minor version? Seems like this could break people's code (e.g. I know lots of my snippets will break).

torfjelde:
This should be good to go once #1762 has gone through :+1:

cpfiffer (approved):

PR #1762, opened 2022-01-14
Rename `Turing.Core` to `Turing.Essential`

devmotion:
Fix https://github.com/TuringLang/Turing.jl/issues/1756 by renaming `Turing.Core` to `Turing.Essential` (maybe there's a better name? Shouldn't choose `Base` or `Main` though :stuck_out_tongue:).


torfjelde:
Ah `Essential` is nicer than what I had in mind (e.g. `TuringCore`) :+1: 

_But_ should we maybe bump the minor version? I know `Core` is technically not exported, but I know a lot of people's code will break, e.g. I have snippets benchmarking AD backends that will now break.

torfjelde (approved):
LGTM! As long as tests pass :+1: 

Only thing I'd like to change is bumping minor version instead of patch version (as explained in the comment above).

devmotion:
Hmm yeah, probably safer. I wonder though if we could maybe even deprecate it with `@deprecate_binding` only in the top level `Turing` module, I hope that this would still fix the issue AND we could distribute this fix to all people with `Turing.Core` snippets (otherwise the have to pin their Julia versions manually to an older release). But I'm not sure if it's worth it (and of course I have to check first if the issue is still fixed).

torfjelde:
Uuuh I did not know about `@deprecate_binding`; neat! I just tried it and it works:) We could just do that :+1:

src/Turing.jl
```diff
@@ -30,8 +30,8 @@ include("stdlib/distributions.jl")
 include("stdlib/RandomMeasures.jl")
 include("utilities/Utilities.jl")
 using .Utilities
-include("core/Core.jl")
-using .Core
+include("essential/Essential.jl")
+using .Essential
```
torfjelde, line 34:
```suggestion
include("essential/Essential.jl")
Base.@deprecate_binding Core Essential
using .Essential
```

devmotion:
I checked and deprecating `Turing.Core` with `@deprecate_binding` seems to work, the issue is still fixed.

devmotion:
> I just tried it and it works:)

Hahahaha beat me again :smile:

src/Turing.jl
torfjelde: Actually, it does _export_ `Code` :confused: 

```julia
julia> @macroexpand Base.@deprecate_binding Core Essential
:($(Expr(:toplevel, :(export Core), :(const _dep_message_Core = ", use Essential instead."), :(const Core = Essential), :(Base.deprecate(Main, :Core)))))
```

devmotion: We shouldn't export it though:
```suggestion
include("essential/Essential.jl")
Base.@deprecate_binding Core Essential false
using .Essential
```

devmotion:
```julia
julia> @macroexpand Base.@deprecate_binding Core Essential false
:($(Expr(:toplevel, nothing, :(const _dep_message_Core = ", use Essential instead."), :(const Core = Essential), :(Base.deprecate(Main, :Core)))))
```

torfjelde: Haha, yeah I just realized :sweat_smile:

PR #1757, opened 2021-12-30
New Turing-libtask integration 

yebai:
This PR introduces a simpler way of evaluating models. It is to avoid the current complicated calling stack encountered in #1744. The benefit is that we no longer need to trace many boiler-plate functions such as `evaluate!!`. For example,  `model(...) ==> evaluate!!  ==> _evaluate!!`  becomes 

```julia

julia> using Turing, DynamicPPL, AdvancedPS

julia> @model gdemo(x, y) = begin
         # Assumptions
         σ ~ InverseGamma(2,3)
         μ ~ Normal(0,sqrt(σ))
         # Observations
         x ~ Normal(μ, sqrt(σ))
         y ~ Normal(μ, sqrt(σ))
       end
gdemo (generic function with 2 methods)

# Case 1: Sample from the prior. 

julia> m = Turing.Core.TracedModel(gdemo(1.5, 2.), SampleFromPrior(), VarInfo()) 

julia> f = m.evaluator[1];

julia> args = m.evaluator[2:end];

julia> f(args...) 
(2.0, VarInfo (2 variables (μ, σ), dimension 2; logp: -6.162))


using Libtask

t = Libtask.CTask(f, args...) 

consume(t) # work fine!


# Case 2: SMC sampler 

julia> m = Turing.Core.TracedModel(gdemo(1.5, 2.), Sampler(SMC(50)), VarInfo());

julia> t = Libtask.CTask(m.evaluator[1], m.evaluator[2:end]...);

julia> consume(t)
ERROR: KeyError: key :__trace not found
Stacktrace:
 [1] consume(ttask::TapedTask)
   @ Libtask ~/.julia/dev/Libtask/src/tapedtask.jl:164
 [2] top-level scope
   @ REPL[58]:1


```

So, with this PR, the integration between Turing-Libtask should be exactly the same as that for AdvancedPS-Libtask.

<details>
  <summary>More tests (not needed until above example is fixed) </summary>

```julia
using Turing, Test, AbstractMCMC, DynamicPPL, Random

import AbstractMCMC.AbstractSampler

function check_numerical(chain,
    symbols::Vector,
    exact_vals::Vector;
    atol=0.2,
    rtol=0.0
)
    for (sym, val) in zip(symbols, exact_vals)
    E = val isa Real ?
    mean(chain[sym]) :
    vec(mean(chain[sym], dims=1))
    @info (symbol=sym, exact=val, evaluated=E)
    @test E ≈ val atol=atol rtol=rtol
    end
end

[...]

PR #1766, opened 2022-01-25
Libtask Integration

yebai:
Replacement of https://github.com/TuringLang/Turing.jl/pull/1757. Awaiting for a new Libtask [release](https://github.com/TuringLang/Libtask.jl/pull/107) before merging.


test/Project.toml
```diff
@@ -12,6 +12,7 @@ DynamicPPL = "366bfd00-2699-11ea-058f-f148b4cae6d8"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
+Libtask = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
```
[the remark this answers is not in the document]
yebai: It's added by CompatHelper https://github.com/TuringLang/Turing.jl/pull/1765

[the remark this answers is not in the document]
devmotion: But CompatHelper only added the compact entry. I am surprised there is a Libtask dependency at all - it shows up as a new addition in this PR but I missed any change in the tests themselves that would explain this dependency.

yebai:
With https://github.com/TuringLang/Libtask.jl/pull/107, only one test is failing now!

Project.toml
```diff
@@ -59,4 +59,4 @@ StatsBase = "0.32, 0.33"
 StatsFuns = "0.8, 0.9"
 Tracker = "0.2.3"
 ZygoteRules = "0.2"
-julia = "1.3, 1.4, 1.5, 1.6"
+julia = "1.6, 1.7"
```
devmotion, line 62: This is the same as
```suggestion
julia = "1.6"
```
i.e., Julia 1.6 and all newer versions are supported. (The current compat entry seems weird, I guess it was supposed to bound it to < 1.7 but is incorrect.)

test/Project.toml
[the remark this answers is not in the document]
devmotion: It seems the entry here was added in https://github.com/TuringLang/Turing.jl/pull/1757 (without compat bound, hence CompatHelper opened a PR). But also there it seems it was not needed, so probably both the entry here and the compat entry below could be removed?
```suggestion
```

test/Project.toml
```diff
@@ -42,6 +43,7 @@ DynamicPPL = "0.17.2"
 FiniteDifferences = "0.10.8, 0.11, 0.12"
 ForwardDiff = "0.10.12"
 GalacticOptim = "0.4, 1, 2"
+Libtask = "0.6"
```
devmotion, line 46:
```suggestion
```

yebai:
A few tests are still failing for Julia 1.6 due to tight error bounds. Basically, the random seed that works for Julia 1.7 turns out not to work well for Julia 1.6.  The results are quite close, but still erroring due to current tight error bounds. Maybe we should relax the error a bit threshold for the sensitive tests, or increase the number of samples (which leads to a longer CI time).

devmotion:
Maybe we could use StableRNGs? There was a major change of the RNG in 1.7, and generally random numbers are not guaranteed to be reproducible with different Julia versions, OS, or architectures.

.github/workflows/Documentation.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Julia
         uses: julia-actions/setup-julia@v1
         with:
-          version: '1.6'
+          version: '1.7'
```
devmotion, line 18: Maybe
```suggestion
          version: '1'
```
so always the stable version is tested and we don't have to adjust CI for Julia 1.8 etc. manually?

.github/workflows/DynamicHMC.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1.6'
+          - '1.7'
```
devmotion, line 16:
```suggestion
          - '1'
```

.github/workflows/Numerical.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1.6'
+          - '1.7'
```
devmotion, line 16:
```suggestion
          - '1'
```

.github/workflows/StanCI.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1.6'
+          - '1.7'
```
devmotion, line 16:
```suggestion
          - '1'
```

.github/workflows/TuringCI.yml
```diff
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1.6'
+          - '1.7'
```
devmotion, line 17:
```suggestion
          - '1'
```

PR #1772, opened 2022-02-08
Create CITATION.cff

rikhuijzer:
This file makes it easier to people to cite Turing.jl. For example, people with the Zotero browser plugin can click on "Save to Zotero" and get a nicely formatted entry.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,29 @@
+# Based on https://github.com/citation-file-format/citation-file-format/blob/main/examples/1.2.0/pass/key-complete/CITATION.cff.
+
+cff-version: "1.2.0"
+message: "If you use this software, please cite it as below."
+
+authors:
+  - family-names: Ge
+    given-names: Hong
+  - family-names: Xu
+    given-names: Kai
+  - family-names: Ghahramani
+    given-names: Zoubin
+
+keywords:
+  - "probabilistic programming"
+  - "Julia language"
+  - "Bayesian inference"
+  - "Probabilistic models"
+  - "Markov chain Monte Carlo"
+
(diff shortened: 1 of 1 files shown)


devmotion:
The convention for Julia packages (also done e.g. by SciML and Distributions) is to add a CITATION.bib file. This is also supported by Github. Citations can be retrievd programmatically with PkgCite.jl.

rikhuijzer:
Okay. The convention doesn't work with Zotero and similar tools though. This is what I get on this repository currently:

image

rikhuijzer:
For comparison, this is for https://github.com/JuliaLang/julia/ which has a `CITATION.cff` file:

image

devmotion:
Seems like Zotero does not support all citation files that are officially supported by Github?

devmotion:
IMO the proper solution is that Zotero adds support for this standard instead of duplicating existing information. And for a correct and complete list of Julia references one has to use something like PkgCite anyway.

rikhuijzer:
> Seems like Zotero does not support all citation files that are officially supported by Github?

Even GitHub doesn't parse these alternative standards (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#other-citation-files).

Also, assuming that everyone will use BibTeX is unfortunately problematic. Social scientist could definitely benefit from using Turing but do not use Word. But whatever. I don't have time for this

devmotion:
I don't think you can avoid a tool like PkgCite if you want to obtain a list of your software references? It's the Julia pendant to `citation` in R, so it seems reasonable to follow this standard (IIRC it also existed before Github came up with CITATION.cff). I assume support for non-Bibtex references would have to be added to such tools such as PkgCite.

PR #1779, opened 2022-02-21
Switch to Libtask 0.7

rikhuijzer:
Pinning explicitly to Libtask 0.7 to avoid the an abstract field in `TapedTask`. See https://github.com/TuringLang/Libtask.jl/pull/123 for details.


devmotion:
IMO this should not be merged and we should wait until AdvancedPS is updated (`Trace` struct etc.). I think it's a mistake that AdvancedPS is declared to be compatible with Libtask 0.7.

yebai:
Closed in favour of #1783

PR #1781, opened 2022-02-21
Fix #1780

### LINKED ISSUE #1780: type Trace has no field ctask
Hey all,

after moving from Turing 19.2 to 20.3 on Julia 1.6.5 the following error appears in the simple example from the docs:

```julia
using Turing
@model function gdemo(x, y)
    s² ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s²))
    x ~ Normal(m, sqrt(s²))
    y ~ Normal(m, sqrt(s²))
end

sample(gdemo(1.5, 2), PG(10), 1000)
```

> type Trace has no field ctask

Same happens for the `SMC` sampler

devmotion:
Fixes https://github.com/TuringLang/Turing.jl/issues/1780.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.20.3"
+version = "0.20.4"
 
 [deps]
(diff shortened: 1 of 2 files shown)


yebai (approved):

yebai:
Thanks, @devmotion, for the quick fix!

PR #1787, opened 2022-03-03
Show timing and allocations for tests

rikhuijzer:
Given that we already run a lot of tests on a lot of systems, it makes sense to collect some allocations and time information. This PR collects the information during running the tests into an ordered dictionary and once the test is done, it will convert that into a DataFrame. Next, this can be pretty printed at the end of the tests and in a separate step in the GitHub Action. That separate Actions step makes it easy to see the output, that is, without having to scroll through all the logs.

This PR will also make detecting performance regressions easier when working on upstream packages with integration tests such as Libtask.jl.

In response to a suggestion by David in another place about using [BenchmarkCI.jl](https://github.com/tkf/BenchmarkCI.jl). I've tried to apply it here too, but didn't see how that package can be based on tests. They only mention [PkgBenchmark.jl](https://github.com/JuliaCI/PkgBenchmark.jl) which is based on putting things inside `benchmark/` (https://juliaci.github.io/PkgBenchmark.jl/stable/define_benchmarks/). Also, the code that I've added here isn't that much and can now easily be tweaked.


test/runtests.jl
```diff
@@ -44,16 +47,19 @@ setprogress!(false)
 
 include(pkgdir(Turing)*"/test/test_utils/AllUtils.jl")
 
+# Collect timing and allocations information to show in a clear way.
+include(pkgdir(Turing)*"/test/test_utils/timing.jl")
+
 @testset "Turing" begin
     @testset "essential" begin
-        include("essential/ad.jl")
+        time_include("essential/ad.jl")
```
devmotion, line 55: I'm a bit unhappy about these special commands and additional logic. It took me a while to clean the Turing test suite and remove all custom commands that one had to use for defining and running tests :smile:

devmotion (commented):
Could we maybe use something less intrusive, without all the special commands? I think this leads to a quite unidiomatic and surprising test structure. Maybe we could use https://github.com/KristofferC/TimerOutputs.jl or something similar?

It is also problematic in general to write into the package directory, it might be read-only (IIRC there were also discussions about possibly making them immutable at some point in the future).

test/runtests.jl
rikhuijzer: I get your point. That's a very suitable package for this kind of problem. The reason that I created `time_include` was to avoid duplication in the description and path. Also with TimerOutputs.jl, it will become
```julia
@timeit to "essential/ad.jl" include("essential/ad.jl")
```
which is verbose due to the duplicate "essential/ad.jl". It's slightly less surprising though indeed. Would we have still a less verbose alternative?

> It is also problematic in general to write into the package directory, it might be read-only (IIRC there were also discussions about possibly making them immutable at some point in the future).

It was already in the `.gitignore` so that's why I chose that path. Why is it problematic? What folder would not be problematic? I could do `_performance` since the trailing underscore means "generated" for Franklin and Literate

devmotion: > Why is it problematic?

Because it is not clear if one has write permissions. Running tests is a bit different from generating output for Franklin, Literate, or Documenter: There the intention is clear and the user can be expected to ensure sufficient file system permissions.

> What folder would not be problematic?

E.g. a temporary directory.

rikhuijzer:
This is now the output for `DynamicHMC-CI / test (1, ubuntu-latest, x64)`:

```
──────────────────────────────────────────────────────────────────
                                        Time          Allocations  
                                  ───────────────   ───────────────
          Total measured:              2066s             283GiB    

 Section                  ncalls     time    %tot     alloc    %tot
 ──────────────────────────────────────────────────────────────────
 essential/ad                  1     194s    9.4%   29.5GiB   10.4%
 inference/AdvancedSMC         1   12.2ms    0.0%    239KiB    0.0%
 inference/emcee               1    8.90s    0.4%   2.12GiB    0.7%
 inference/ess                 1   46.4ms    0.0%   2.03MiB    0.0%
 inference/is                  1   2.71ms    0.0%   52.2KiB    0.0%
 inference: forwarddiff        1     642s   31.1%    101GiB   35.6%
   mh                          1   10.6ms    0.0%    222KiB    0.0%
   sghmc                       1   6.52ms    0.0%    123KiB    0.0%
   dynamichmc                  1     134s    6.5%   10.7GiB    3.8%
   Inference                   1     507s   24.6%   90.0GiB   31.8%
   hmc                         1   17.2ms    0.0%   1.21MiB    0.0%
   gibbs_conditional           1   5.73ms    0.0%    104KiB    0.0%
   gibbs                       1   8.20ms    0.0%    148KiB    0.0%
 variational/advi              3   11.2ms    0.0%    240KiB    0.0%
 ModeEstimation                3    21.1s    1.0%   2.38GiB    0.8%
 OptimInterface                3    34.9s    1.7%   6.25GiB    2.2%
 inference: tracker            1     526s   25.5%   65.5GiB   23.1%
   mh                          1   8.95ms    0.0%    222KiB    0.0%
   sghmc                       1   6.01ms    0.0%    123KiB    0.0%
   dynamichmc                  1     118s    5.7%   12.3GiB    4.4%
   Inference                   1     408s   19.8%   53.1GiB   18.8%
   hmc                         1   11.0ms    0.0%    235KiB    0.0%
   gibbs_conditional           1   5.41ms    0.0%    104KiB    0.0%
   gibbs                       1   7.39ms    0.0%    148KiB    0.0%
 inference: reversediff        1     540s   26.2%   60.0GiB   21.2%
   mh                          1   9.60ms    0.0%    222KiB    0.0%
   sghmc                       1   6.43ms    0.0%    123KiB    0.0%
   dynamichmc                  1     118s    5.7%   10.4GiB    3.7%
   Inference                   1     422s   20.5%   49.6GiB   17.5%
   hmc                         1   11.2ms    0.0%    235KiB    0.0%
   gibbs_conditional           1   5.59ms    0.0%    104KiB    0.0%
   gibbs                       1   7.52ms    0.0%    148KiB    0.0%
 optimisers                    1    9.24s    0.4%   5.30GiB    1.9%
 distributions                 1   6.71ms    0.0%    111KiB    0.0%
 RandomMeasures                1    11.6s    0.6%   1.60GiB    0.6%
 utilities                     1    76.4s    3.7%   9.56GiB    3.4%
 ──────────────────────────────────────────────────────────────────
```

> It is also problematic in general to write into the package directory, it might be read-only (IIRC there were also discussions about possibly making them immutable at some point in the future).

I'm not sure how I can write to a temporary directory without either hardcoding the temporary filename or passing it via GitHub Actions env from one step to the next. Both seem not so ideal, so I now removed the functionality altogether. The time and allocations information is now only shown at the end of the `julia-actions/julia-runtest@latest` step.

rikhuijzer:
From `Stan-CI / test (1.6, ubuntu-latest, x64)`:

```
 ──────────────────────────────────────────────────────────────────
                                        Time          Allocations  
                                  ───────────────   ───────────────
          Total measured:              2405s             277GiB    

 Section                  ncalls     time    %tot     alloc    %tot
 ──────────────────────────────────────────────────────────────────
 essential/ad                  1     288s   12.0%   34.5GiB   12.4%
 inference/AdvancedSMC         1   11.6ms    0.0%    335KiB    0.0%
 inference/emcee               1    13.6s    0.6%   2.69GiB    1.0%
 inference/ess                 1   51.0ms    0.0%   2.25MiB    0.0%
 inference/is                  1   2.78ms    0.0%   65.0KiB    0.0%
 inference: forwarddiff        1     812s   33.8%    106GiB   38.3%
   OptimInterface              1    39.0s    1.6%   5.72GiB    2.1%
   ModeEstimation              1    25.9s    1.1%   2.48GiB    0.9%
   variational/advi            1   3.96ms    0.0%    117KiB    0.0%
   mh                          1   10.3ms    0.0%    299KiB    0.0%
   sghmc                       1   6.37ms    0.0%    188KiB    0.0%
   dynamichmc                  1   1.08ms    0.0%   13.2KiB    0.0%
   Inference                   1     747s   31.1%   97.9GiB   35.3%
   hmc                         1   18.9ms    0.0%   0.99MiB    0.0%
   gibbs_conditional           1   5.99ms    0.0%    131KiB    0.0%
   gibbs                       1   7.53ms    0.0%    204KiB    0.0%
 inference: tracker            1     575s   23.9%   59.3GiB   21.4%
   OptimInterface              1    8.05s    0.3%   0.96GiB    0.3%
   ModeEstimation              1    755ms    0.0%   51.7MiB    0.0%
   variational/advi            1   3.88ms    0.0%    117KiB    0.0%
   mh                          1   10.0ms    0.0%    299KiB    0.0%
   sghmc                       1   6.20ms    0.0%    188KiB    0.0%
   dynamichmc                  1   1.03ms    0.0%   13.2KiB    0.0%
   Inference                   1     566s   23.6%   58.3GiB   21.0%
   hmc                         1   13.3ms    0.0%    313KiB    0.0%
   gibbs_conditional           1   6.09ms    0.0%    131KiB    0.0%
   gibbs                       1   7.82ms    0.0%    204KiB    0.0%
 inference: reversediff        1     577s   24.0%   56.8GiB   20.5%
   OptimInterface              1    8.26s    0.3%   0.95GiB    0.3%
   ModeEstimation              1    1.30s    0.1%   71.8MiB    0.0%
   variational/advi            1   3.85ms    0.0%    117KiB    0.0%
   mh                          1   10.0ms    0.0%    299KiB    0.0%
   sghmc                       1   6.16ms    0.0%    188KiB    0.0%
   dynamichmc                  1   1.02ms    0.0%   13.2KiB    0.0%
   Inference                   1     567s   23.6%   55.8GiB   20.2%
   hmc                         1   12.8ms    0.0%    313KiB    0.0%
   gibbs_conditional           1   5.94ms    0.0%    131KiB    0.0%
   gibbs                       1   7.60ms    0.0%    204KiB    0.0%
 optimisers                    1    14.8s    0.6%   5.39GiB    1.9%
 distributions                 1   6.85ms    0.0%    156KiB    0.0%
 RandomMeasures                1    15.4s    0.6%   1.68GiB    0.6%
 utilities                     1     107s    4.5%   10.5GiB    3.8%
 ──────────────────────────────────────────────────────────────────
```

test/runtests.jl
```diff
@@ -35,6 +35,7 @@ using DynamicPPL: getval, getlogp
 using ForwardDiff: Dual
 using MCMCChains: Chains
 using StatsFuns: binomlogpdf, logistic, logsumexp
+using TimerOutputs: TimerOutput, @timeit, print_timer
```
devmotion, line 38: It doesn't matter but sometimes I use as rule of thumb to only load stuff that I call >= 10 times, or not even in these cases. Since `TimerOutput` and `print_timer` are only used once, we could eg only import
```suggestion
using TimerOutputs: TimerOutputs, @timeit
```
In any case, it seems `print_timer` and `show` are the same (apart from a newline?), so it's not needed:
```suggestion
using TimerOutputs: TimerOutput, @timeit
```

test/runtests.jl
```diff
@@ -44,61 +45,68 @@ setprogress!(false)
 
 include(pkgdir(Turing)*"/test/test_utils/AllUtils.jl")
 
+# Collect timing and allocations information to show in a clear way.
+const to = TimerOutput()
```
devmotion, line 49: In that case we have to qualify
```suggestion
const to = TimerOutputs.TimerOutput()
```

test/runtests.jl
```diff
@@ -44,61 +45,68 @@ setprogress!(false)
...
 
     @testset "utilities" begin
         # include("utilities/stan-interface.jl")
-        include("inference/utilities.jl")
+        @timeit to "utilities" include("inference/utilities.jl")
     end
 end
+
+# Hiding `avg` column via `compact=true` because we do only one run per entry.
+print_timer(to; compact=true, sortby=:firstexec)
```
devmotion, line 112: And

```suggestion
show(to; compact=true, sortby=:firstexec)
```

rikhuijzer:
The wrong ordering of the sub-elements is an upstream bug. I'll try to open a PR there. MWE:

```julia
julia> using TimerOutputs

julia> to = TimerOutput();

julia> @timeit to "group" begin
           @timeit to "sub1" sleep(0.1)
           @timeit to "sub2" sleep(0.1)
       end

julia> print_timer(to; compact=true, sortby=:firstexec)
 ───────────────────────────────────────────────────
                         Time          Allocations
                   ───────────────   ───────────────
  Total measured:       46.0s            2.29MiB

 Section   ncalls     time    %tot     alloc    %tot
 ───────────────────────────────────────────────────
 group          1    202ms  100.0%   2.55KiB  100.0%
   sub2         1    101ms   50.0%      320B   12.3%
   sub1         1    101ms   50.0%      320B   12.3%
 ───────────────────────────────────────────────────
```

devmotion (approved):
I like this much more 🙂 I think the summary can be useful, and have only minor suggestions.

yebai (approved):

KristofferC:
You could also take a look at https://github.com/Ferrite-FEM/Tensors.jl/blob/f2d296bc4f75f803f5dca931b6dfd642637733d1/test/runtests.jl#L8-L16. Then you can just use `@testsection` as a replacement of `@testset` and it will get added to the timer.

devmotion:
Since we only use it in `test/runtests.jl` I guess it's sufficiently clear and unsurprising for a reader if we define a similar convenience macro for `@timeit to ... include(...)` at the top of that file. Maybe even more obvious if we would call it eg `@timeit_include`.

test/runtests.jl
```diff
@@ -44,61 +45,70 @@ setprogress!(false)
 
 include(pkgdir(Turing)*"/test/test_utils/AllUtils.jl")
 
+# Collect timing and allocations information to show in a clear way.
+const TIMEROUTPUT = TimerOutputs.TimerOutput()
+macro timeit_include(path::AbstractString)
+    return :(@timeit TIMEROUTPUT $path include($path))
+end
```
devmotion, line 52: I guess it even fits in one line?

```suggestion
macro timeit_include(path::AbstractString) = :(@timeit TIMEROUTPUT $path include($path))
```

test/runtests.jl
```diff
@@ -44,61 +45,70 @@ setprogress!(false)
...
-                include("inference/gibbs_conditional.jl")
-                include("inference/hmc.jl")
-                include("inference/Inference.jl")
-                include("contrib/inference/dynamichmc.jl")
-                include("contrib/inference/sghmc.jl")
-                include("inference/mh.jl")
+        @timeit TIMEROUTPUT "inference: $adbackend" begin
+            Turing.setadbackend(adbackend)
+            @info "Testing $(adbackend)"
+            start = time()
```
devmotion, line 71: Maybe not needed anymore?

```suggestion
```

test/runtests.jl
```diff
@@ -44,61 +45,70 @@ setprogress!(false)
...
+            @testset "inference: $adbackend" begin
```
devmotion, line 72: We could also use a `@testset` for loop I guess? Not really relevant for this PR though.

test/runtests.jl
```diff
@@ -44,61 +45,70 @@ setprogress!(false)
...
+                @timeit_include("modes/ModeEstimation.jl")
+                @timeit_include("modes/OptimInterface.jl")
+            end
 
-        # Useful for
-        # a) discovering performance regressions,
-        # b) figuring out why CI is timing out.
-        @info "Tests for $(adbackend) took $(time() - start) seconds"
+            # Useful for figuring out why CI is timing out.
+            @info "Tests for $(adbackend) took $(time() - start) seconds"
```
devmotion, line 94: Still needed?

```suggestion
```

devmotion (approved):
Looks good, I left some final comments.

test/runtests.jl
rikhuijzer: There is still the 
```julia
# Useful for figuring out why CI is timing out.
@info "Tests for $(adbackend) took $(time() - start) seconds"
```
comment, so I guess that it is still useful

test/runtests.jl
rikhuijzer: Ah yes no indeed not. I thought so, but only the `@info "Testing $(adbackend)" ` should be enough.

test/runtests.jl
rikhuijzer: That's not valid syntax. It would be possible to do
```suggestion
macro timeit_include(path::AbstractString) :(@timeit TIMEROUTPUT $path include($path)) end
```
but the 3-line version seemed more readable.

rikhuijzer: Ah well it's not too bad

test/runtests.jl
rikhuijzer: Let's do it in another

PR #1795, opened 2022-03-10
Unify notation for variance in the model and chain

PeetoomHeida:
The term for variance in the @model call was changed to s², but extracting this value from the resulting changes was left as get(chain, [:s, :m, :lp]. Update this and val.s to val.s², as well as vi.metadata.s.vns[1] to vi.metadata.s².vns[1].


yebai:
thanks, @PeetoomHeida!

Issue #1804, opened 2022-03-14
Reconsider the default chunksize for ForwardDiff


ChrisRackauckas:
https://github.com/TuringLang/Turing.jl/blob/master/src/essential/ad.jl#L29

That's a bit crazy. ForwardDiff itself limits the duals to have a maximum chunksize of 12:

https://github.com/JuliaDiff/ForwardDiff.jl/blob/241829d80fa50810e859e884184da00eadeb14d8/src/prelude.jl#L13

This was because larger chunking was not seen to be beneficial to performance, while substantially increasing compile times. On top of that, very large chunk sizes will cause codes to expand fast which can lead to segfaults during inference time. Here's an example of that:

```julia
using DifferentialEquations
using Turing
using Distributions
using Random
using Turing: Variational

function vanderpol(du,u,p,t)
    x1,x2 = u
    θ,ϕ = p

    du[1] = x2
    du[2] = θ*(1-x1^2)*x2 - x1
end

function add_noise(du,u,p,t)
    x1,x2 = u
    θ,ϕ = p

    du[1] = ϕ #   0	
    du[2] = ϕ
end

u0 = [0.1, 0.1]
tspan = [0.0, 20]
p = [1.0,0.1]

dt=0.1
prob1 = SDEProblem(vanderpol, add_noise, u0, tspan, p)
sol = solve(prob1,EM(),dt=dt)
time = sol.t

# add noise to the solution
x1 = sol[1,:]
x1n = x1 .+ rand(Normal(0,.3),length(x1))
x2 = sol[2,:]
x2n = x2 .+ rand(Normal(0,.3),length(x2))

@model function fitvpEMn(datax, datay)
    σ = 0.3
    θ ~ Gamma(0.25,4.0)
    ϕ ~ Uniform(0,0.2)
    # create arrays for hidden variables
    xh = zero(datax)
    yh = zero(datay)
    xh[1] = 0.1
    yh[1] = 0.1
    for i in 2:length(datax)
        dxh = yh[i-1]
        dyh = θ * (1.0 - xh[i-1]^2) * yh[i-1] - xh[i-1]
        xh[i] ~ Normal(xh[i-1] + dxh * dt,ϕ*sqrt(dt))
        yh[i] ~ Normal(yh[i-1] + dyh * dt,ϕ*sqrt(dt))
    end
    datax ~ MvNormal(xh,σ)
    datay ~ MvNormal(yh,σ)
end

[...]

PR #1806, opened 2022-03-15
Set default chunk size to 0

devmotion:
Playing around with the idea in https://github.com/TuringLang/Turing.jl/issues/1804 and checking if/how it impacts the tests.


src/essential/ad.jl
```diff
@@ -26,13 +25,11 @@ function setadsafe(switch::Bool)
     ADSAFE[] = switch
 end
 
-const CHUNKSIZE = Ref(40) # default chunksize used by AD
+const CHUNKSIZE = Ref(0) # default chunksize used by AD
```
xukai92, line 28:
```suggestion
const CHUNKSIZE = Ref(0) # 0 means letting ForwardDiff set it automatically
```

xukai92 (approved):

devmotion:
I compared the test times of the initial and final commit in this PR with the last two tests on the master branch. I couldn't observe any clear regressions of the (of course fluctuating) test times.

yebai (approved):

yebai:
many thanks @devmotion!

PR #1816, opened 2022-04-14
Fixing Math in "For Developers" section.

tomroesch:
The math was not rendering correctly in the "For Developers" section. I went through it at fixed the formatting. Nothing in the actual math was changed.

### DIFF
--- a/docs/src/for-developers/how_turing_implements_abstractmcmc.md
+++ b/docs/src/for-developers/how_turing_implements_abstractmcmc.md
@@ -167,7 +167,7 @@ struct Transition{T, F<:AbstractFloat}
 end
 ```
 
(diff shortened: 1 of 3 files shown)


yebai:
thanks, @tomroesch for the fixes - several people have attempted to fix these math typesetting issues, but with only partial success. So it would be really useful if we can preview the changes before merging. However, I'm not sure how to set up a preview workflow on Github. An alternative route is to make sure this math can be built successfully locally.

tomroesch:
I tried tested the "Variational Inference" site on a different Jekyll based website, and I was not aware of the issues that come up there, so the pages will not render correctly with the fixes that I added. I will give it another run and open a new PR with new fixes that I have tested at least elsewhere.

PR #1834, opened 2022-06-01
Add support for GalacticOptim 3 and fix test errors

devmotion:
This PR adds support for GalacticOptim 3 in the tests and fixes the CI errors.

The correct `solve` dispatch was missing since GalacticOptimJL was not loaded. This is required since the optimization algorithms and dispatches were moved to subpackages to avoid Requires and its issues (https://github.com/SciML/Optimization.jl/pull/220).

The timing of the PR is a bit unfortunate since GalacticOptim is about to be renamed to Optimization, the pull request to the registry is already opened. That is, we should rename the test dependencies once it is released.

Fixes #1822.


yebai (approved):
thanks, @devmotion - looks good to me!

PR #1838, opened 2022-06-18
Fix minor inconsistency in HMC doc

yebai:
The actual default max tree depth is 10, see https://github.com/TuringLang/Turing.jl/blob/master/src/inference/hmc.jl#L402


storopoli:
Sorry I just saw this know.
Thanks for the PR.

PR #1841, opened 2022-06-20
Improve stacktraces by using custom tag for ForwardDiff

devmotion:
This should improve stacktraces and still be safe. Was suggested by @ChrisRackauckas [on discourse](https://discourse.julialang.org/t/first-function-call-produced-nans-when-using-nuts-but-not-mh/83024/4?u=devmotion).

Ref: https://www.stochasticlifestyle.com/improved-forwarddiff-jl-stacktraces-with-package-tags/

I'll add an example with failing AD in the PR to show if and how it improves stacktraces.

PR #1846, opened 2022-06-25
Unify log density function types

devmotion:
Follow up to #1841 that unifies different log density function types.

Requires https://github.com/TuringLang/AdvancedMH.jl/pull/68.


devmotion:
Seems I broke the MLE/MAP optimization accidentally. Maybe some incorrect dispatches?

src/modes/ModeEstimation.jl
```diff
@@ -78,39 +78,30 @@ end
 """
     OptimLogDensity{M<:Model,C<:Context,V<:VarInfo}
 
-A struct that stores the log density function of a `DynamicPPL` model.
+A struct that stores the negative log density function of a `DynamicPPL` model.
```
torfjelde, line 81: Is this change correct? Doesn't it now "store" the actual log density function rather than "negative"?

Though the current docstring is confusing anyways.

torfjelde (approved):
Awesome stuff David:) I only have one comment regarding one of the docstrings, but other than that this looks like a slam-dunk.

src/modes/OptimInterface.jl
```diff
@@ -206,13 +206,13 @@ function _optimize(
...
 )
-    return _optimize(model, f, f.vi[DynamicPPL.SampleFromPrior()], Optim.LBFGS(), args...; kwargs...)
+    return _optimize(model, f, f.varinfo[f.sampler], Optim.LBFGS(), args...; kwargs...)
 end
 
 function _optimize(
     model::Model, 
     f::OptimLogDensity, 
-    init_vals::AbstractArray = f.vi[DynamicPPL.SampleFromPrior()], 
+    init_vals::AbstractArray = f.varinfo[f.sampler],
```
torfjelde, line 215: L226 needs to be changed in a similar way to this line

torfjelde (changes requested):
Found a bug

src/modes/ModeEstimation.jl
devmotion: Well, currently `OptimLogDensity` stores the negative log density function (see, e.g., https://github.com/TuringLang/Turing.jl/blob/9951638916a4372bef31f20c723224d49e1f4124/src/modes/ModeEstimation.jl#L113). This makes sense since basically all optimization libraries expect a minimization problem. So I thought it would be best to keep the same behaviour for the `OptimLogDensity` alias and correct the docstring. I'm not completely happy though since it seems somewhat confusing and possibly cleaner to always return the log density when `LogDensityFunction` is evaluated.

Thus I thought about instead changing the implementations of the tilde pipeline for `OptimizationContext` to return the negative log density values instead (and hence adjust for the negative values there) but I think this would break if someone calls `@addlogprob!` or DynamicPPL.acclogp!! in a model.

devmotion: But I think even though this change is correct it is responsible for the test errors: In the implementation of the Optim API below currently we set everything to negative values manually but when unifying the types we already get the correct gradients etc. So the negative signs are wrong in the Optim implementation I think.

devmotion:
Tests pass now it seems :tada: 

Currently the Turing CI tests time out due to the recently introduced deprecation of `PDMats.dim`. This will be fixed by https://github.com/JuliaStats/Distributions.jl/pull/1572.

src/modes/OptimInterface.jl
devmotion: Thanks, I fixed it!

yebai (approved):
many thanks @devmotion!

torfjelde (approved):
Only one comment but otherwise happy with it if tests pass:) Thanks @devmotion !

src/modes/ModeEstimation.jl
```diff
@@ -309,8 +292,7 @@ function optim_function(
     else
         OptimizationFunction(
             l;
-            grad = (G,x,p) -> obj(nothing, G, nothing, x),
-            hess = (H,x,p) -> obj(nothing, nothing, H, x),
```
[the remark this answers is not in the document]
devmotion: Yes. It does not work - currently we explicitly throw an error if a non-`nothing` Hessian `H` is provided. So it's much more straightforward (and simpler for us) to let Optim etc. handle this case (throw an error, do some fancy AD ...) and just define `f(F, G, x)` instead of `f(F, G, H, x)` (in the Optim API) and only provide `grad` but not `hessian` (in the SciML API).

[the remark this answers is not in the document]
devmotion: @torfjelde Do you think we should keep it?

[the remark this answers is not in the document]
torfjelde: Ah gotcha, if it doesn't actually currently work then I'm happy to drop it :+1:

PR #1849, opened 2022-06-30
Remove implementations of `logpdf_with_trans` for `NoDist`

devmotion:
This should be hidden in stdlib/distributions.jl and at most be defined in DynamicPPL (see https://github.com/TuringLang/DynamicPPL.jl/pull/414#issuecomment-1170516505). I wonder, are they used anywhere and what breaks if we remove them completely?


torfjelde (approved):
Happy to merge if tests pass:)

devmotion:
Seems to be the case :slightly_smiling_face:  I'll bump the version and merge once all tests pass successfully.

Issue #1827, opened 2022-05-08
Math doesn't display on some docs pages


gdalle:
This is the case for instance with https://turing.ml/dev/docs/for-developers/variational_inference
I'll try to submit a PR

devmotion:
Probably relevant: https://github.com/TuringLang/turing.ml/issues/8

tomroesch:
Before #1817 the website looked a lot worse, but there is still work to do. I'll see if I can fix the remaining issues using the mentioned post, it looks very relevant.

filchristou:
is there a temporary way to read through this page somehow, e.g. compiling it locally ?

tomroesch:
There are instructions to build the website locally [here](https://github.com/TuringLang/Turing.jl/tree/master/docs).  However, there were some issues for me, which I explained with possible fixes in #1855.

I guess to test this PR one could clone the [repository](https://github.com/tomroesch/Turing.jl) I requested the PR from and build that locally (it also includes the fixes to build the website locally).

filchristou:
thanks it works. I didn't manage to get the CSS, but it's readable !

devmotion:
I rebuilt the webpage, and it seems https://github.com/TuringLang/Turing.jl/pull/1869 finally fixed the VI docs :tada: Thanks a lot, @tomroesch!

PR #1857, opened 2022-07-07
locally tested latex

tomroesch:
Following up on #1827. I locally build the website from `Turing.jl/docs` and tested that the math displays correctly now.

### DIFF
--- a/docs/src/for-developers/how_turing_implements_abstractmcmc.md
+++ b/docs/src/for-developers/how_turing_implements_abstractmcmc.md
@@ -31,28 +31,28 @@ Turing sampling methods (most of which are written [here](https://github.com/Tur
 
 First, we explain how Importance Sampling works in the abstract. Consider the model defined in the first code block. Mathematically, it can be written:
 
-$$
+```math
 \begin{align}
 s &\sim \text{InverseGamma}(2, 3), \\
 m &\sim \text{Normal}(0, \sqrt{s}), \\
 x &\sim \text{Normal}(m, \sqrt{s}), \\
 y &\sim \text{Normal}(m, \sqrt{s}),
 \end{align}
-$$
+```
 
-The **latent** variables are $s$ and $m$, the **observed** variables are $x$ and $y$. The model **joint** distribution $p(s,m,x,y)$ decomposes into the **prior** $p(s,m)$ and the **likelihood** $p(x,y \mid s,m).$ Since $x = 1.5$ and $y = 2$ are observed, the goal is to infer the **posterior** distribution $p(s,m \mid x,y).$
+The **latent** variables are ``$s$`` and ``$m$``, the **observed** variables are ``$x$`` and ``$y$``. The model **joint** distribution ``$p(s,m,x,y)$`` decomposes into the **prior** ``$p(s,m)$`` and the **likelihood** ``$p(x,y \mid s,m).$`` Since ``$x = 1.5$`` and ``$y = 2$`` are observed, the goal is to infer the **posterior** distribution ``$p(s,m \mid x,y).$``
 
-Importance Sampling produces independent samples $(s\_i, m\_i)$ from the prior distribution. It also outputs unnormalized weights
+Importance Sampling produces independent samples ``$(s_i, m_i)$`` from the prior distribution. It also outputs unnormalized weights
 
-$$
+```math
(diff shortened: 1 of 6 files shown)


tomroesch:
I accidentally committed the changes I made to run the website locally as well.

yebai (approved):
Many thanks, @tomroesch for fixing these annoying incorrect math-typesetting errors. 

I am cc'ing @xukai92  as we are in the process of reorganising documentation into its own [repo](https://github.com/TuringLang/Documentation). The goal is to create something like `TuringTutorials` which is easier to modify and improve. Hopefully, it will help gradually improve the documentation quality for all docs.

tomroesch:
@yebai Does the live website already contain these changes? It still does not look right, while the math typesetting looks fine  in my local build.

devmotion:
It seems the website was rebuilt 17 hours ago: https://github.com/TuringLang/turing.ml/actions So it should contain the updates in this PR, I think.

docs/src/for-developers/how_turing_implements_abstractmcmc.md
```diff
@@ -140,20 +140,20 @@ end
...
+```math
 \frac 1 N \sum_{j=1}^N w_i = \frac 1 N \sum_{j=1}^N p(x,y \mid s_i, m_i)
-$$ 
+``` 
 
 of the sample weights is a particularly important quantity:
 
 * it is used to **normalize** the **empirical approximation** of the posterior distribution
-* its logarithm is the importance sampling **estimate** of the **log evidence** $\log p(x, y)$
+* its logarithm is the importance sampling **estimate** of the **log evidence** ``$\log p(x, y)$``
```
devmotion, line 150: Is this actually correct? That's two math delimiters it seems, both ``` `` ``` and `$`?

tomroesch: I had to add the additional `$` to make it work when building the website locally (using `Turing.jl/docs/make.jl`). Here are the differences I observe:

This is how it looks locally when writing ` ``$\log p(x, y)$`` `:
image

However, here is what it looks like live:
image

So I am assuming that the live build works differently than the local build, maybe because it is not build from the same `make.jl` command? But that's just a guess. 

If someone can point me to the "right" way to build the website locally, I'll give this another try.

devmotion: > If someone can point me to the "right" way to build the website locally, I'll give this another try.

I haven't tried myself yet unfortunately, but reproducing the build locally would involve running the build script in the (locally checked out) turing.ml repo, I assume. I guess you could use a local version of Turing.jl (with your changes) by replacing some hardcoded links to the Turing.jl repo or dev'ing Turing.jl (`] dev Turing`) in the build environment.

yebai: The actual building workflow for the docs is described [here](https://github.com/TuringLang/turing.ml#building-the-documentation-site-locally) - it is likely that the `turing.ml` [repo](https://github.com/TuringLang/turing.ml/blob/master/make.jl) is using a slightly different setup from `Turing.jl`

tomroesch: Yes I think the issue is in the differences in the markdown processing, which runs through `kramdown` in `Turing.jl`, while in `turing.ml` `CommonMarkGhPages` is used for markdown. That makes it obsolete to test `Turing.jl` with a local build from the repo itself. I'll try to run it through `turing.ml` and in the meantime update the instructions for local website build on `turing.ml` (slightly inaccurate with directory names) and `Turing.jl` (how to build it through `turing.ml`.

devmotion: According to https://github.com/TuringLang/turing.ml/issues/8#issue-824225056, here in Turing.jl we should just use Documenter-style syntax, ie. ``` `` ``` for inline math (no `$`) and ```` ```math ```` for display math (no `$$`). Of course, it would be nice to test it first though.

tomroesch: Ok I figured out how to build the website through `turing.ml` locally and was able to finally test the latex display. I think it is finally fixed now. I will add a PR to `turing.ml` for the local build and a PR here to resolve the math issues.

PR #1858, opened 2022-07-09
Match AdvancedPS 0.4 API

FredericWantiez:
Tweaking Turing to work with the new interface of `AdvancedPS`

Closes https://github.com/TuringLang/Turing.jl/pull/1903 Closes https://github.com/TuringLang/Turing.jl/pull/1901 Closes https://github.com/TuringLang/Turing.jl/pull/1902


FredericWantiez:
Still running into one annoying issue:
```
dynamic model: Error During Test at /Users/frederic/Documents/projects/turing/Turing.jl/test/test_utils/staging.jl:42
  Got exception outside of a @test
  TypeError: in Type{...} expression, expected UnionAll, got a value of type typeof(TArray)
  Stacktrace:
    [1] get_matching_type(spl::Sampler{PG{(:z,), ResampleWithESSThreshold{typeof(resample_systematic), Float64}}}, vi::DynamicPPL.TypedVarInfo{NamedTuple{(:z, :m), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:z, Setfi
eld.IndexLens{Tuple{Int64}}}, Int64}, Vector{ChineseRestaurantProcess{DirichletProcess{Float64}, Vector{Int64}}}, Vector{AbstractPPL.VarName{:z, Setfield.IndexLens{Tuple{Int64}}}}, Vector{Int64}, Vector{Set{DynamicPPL.Select
or}}}, DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:m, Setfield.IndexLens{Tuple{Int64}}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:m, Setfield.IndexLens{Tuple{Int64}}}}, Vector{Float64}, Vector{Set{Dynam
icPPL.Selector}}}}}, Float64}, #unused#::Type{Vector{Float64}})
```
Which is caused by this `get_matching_type` here 
https://github.com/TuringLang/Turing.jl/blob/5990fae9f176e84d83bd119fa4a6b0e68f028493/src/inference/Inference.jl#L446-L452

and the new logic in `Libtask` in `0.8`.

yebai:
I think we don't need `get_matching_type` anymore because `Libtask@0.8` removed `TArray`. I will take a more careful look tomorrow.

FredericWantiez:
Apart from some numerical errors, the `infiniteGMM` model errors on a weird bound error in Libtask

FredericWantiez:
Does Libtask support random array size ? 

```julia
@model function infiniteGMM(x)
    # Hyper-parameters, i.e. concentration parameter and parameters of H.
    α = 1.0
    μ0 = 0.0
    σ0 = 1.0

    # Define random measure, e.g. Dirichlet process.
    rpm = DirichletProcess(α)

    # Define the base distribution, i.e. expected value of the Dirichlet process.
    H = Normal(μ0, σ0)

    # Latent assignment.
    z = zeros(Int, length(x))

    # Locations of the infinitely many clusters.
    μ = zeros(Float64, 0)

    for i in 1:length(x)

        # Number of clusters.
        K = maximum(z)
        nk = Vector{Int}(map(k -> sum(z .== k), 1:K)) # This will have a random size K
        println(nk)

        # Draw the latent assignment.
        z[i] ~ ChineseRestaurantProcess(rpm, nk)
        
        # Create a new cluster?
        if z[i] > K
            push!(μ, 0.0)

            # Draw location of new cluster.
            μ[z[i]] ~ H
        end
                
        # Draw observation.
        x[i] ~ Normal(μ[z[i]], 1.0)
    end
end
```

yebai:
Yes, `Libtask` should work correctly for these Vector/Array types. Do we deep-copy the `ChineseResturantProcess` struct here? It contains an array internally. I don't think `Libtask` inspect Arrays nested in structs by default -- it only looks for Arrays referenced directly by a variable in a `TapedFunction`. 

https://github.com/TuringLang/Turing.jl/blob/master/src/stdlib/RandomMeasures.jl#L54

FredericWantiez:
I'm not quite sure what's happening but looks like some references to $\mu$ and $z$ are not handled properly when forking the trace.

FredericWantiez:
The issue here is that `z` gets boxed and ends being of type `Core.Box(Array{Int})` which is then shallow copied across forks:

```
1 ──        (x@_56 = x@_5)::Vector{Float64}
│   (__varinfo__@_55 = __varinfo__@_3)::DynamicPPL.TypedVarInfo{....}, Float64}
│   Core.NewvarNode(:(retval#1789))::Any
│   (z@_9 = Core.Box())::Core.Box # <------ Here z = Core.Box()
│   (α = 1.0)::Core.Const(1.0)
│   (μ0 = 0.0)::Core.Const(0.0)
│   (σ0 = 1.0)::Core.Const(1.0)
│   (rpm = Main.DirichletProcess(α::Core.Const(1.0)))::Core.Const(DirichletProcess{Float64}(1.0))
│   (H = Main.Normal(μ0::Core.Const(0.0), σ0::Core.Const(1.0)))::Core.Const(Normal{Float64}(μ=0.0, σ=1.0))
```

KDr2:
> The issue here is that `z` gets boxed and ends being of type `Core.Box(Array{Int})` which is then shallow copied across forks:

But we have the content of Boxes deep or shallow copied here accordingly: https://github.com/TuringLang/Libtask.jl/blob/master/src/tapedfunction.jl#L479

Is it possible to wrap an MWE to reproduce this issue?

FredericWantiez:
Here's an example that fails on this branch

```julia

using Random
using Turing
using Turing.RandomMeasures
using Libtask
using AdvancedPS
using DynamicPPL

@model function infiniteGMM(x)
    # Hyper-parameters, i.e. concentration parameter and parameters of H.
    α = 1.0
    μ0 = 0.0
    σ0 = 1.0

    # Define random measure, e.g. Dirichlet process.
    rpm = DirichletProcess(α)

    # Define the base distribution, i.e. expected value of the Dirichlet process.
    H = Normal(μ0, σ0)

    # Latent assignment.
    z = zeros(Int, length(x))

    # Locations of the infinitely many clusters.
    μ = zeros(Float64, 0)

    for i in 1:length(x)
        # Number of clusters.
        K = maximum(z)
        nk = Vector{Int}(map(k -> sum(z .== k), 1:K))

        #println("i: $i, K: $K, Mu $(μ) $([objectid(μ)]) z: $z, $([objectid(z)]) $(typeof(z)) $(typeof(μ))")

        # Draw the latent assignment.
        z[i] ~ ChineseRestaurantProcess(rpm, nk)
        
        # Create a new cluster?
        if z[i] > K
            push!(μ, 0.0)

            # Draw location of new cluster.
            μ[z[i]] ~ H
        end
                
        # Draw observation.
        x[i] ~ Normal(μ[z[i]], 1.0)
    end
end

# Generate some test data.
Random.seed!(2);
m = 3
data = vcat(randn(5), randn(10) .- 5, randn(10) .+ 10);
data .-= mean(data);
data /= std(data);

model_fun = infiniteGMM(data);

# MCMC sampling
Random.seed!(54);
iterations = 300;
chain = sample(model_fun, SMC(), iterations; progress=false)

```

FredericWantiez:
I'll add some tests but this is mostly working now, still not quite sure why it fails on convergence for Ubuntu/Julia1.7 but not on the other images.

torfjelde:
Seems like it's just a difference in RNGs, no? As in, it's getting very close to the correct result but _just_ outside of the allowed tolerance. So try increasing iterations/particles for this particular test.

FredericWantiez:
Agree, but if everything is seeded explicitly it shouldn't vary across images ?

yebai (approved):

yebai:
@devmotion Are you happy with the changes in this PR? It (hopefully temporarily) dropped support for Julia 1.6. It is largely due to Libtask, which only supports Julia 1.7 and above -- substantial efforts are required to support Julia 1.6 due to major Julia compiler changes. 

In addition, @FredericWantiez and I discussed moving the general-purpose particle Gibbs sampler implementation (requires Libtask) into an experimental namespace and loading `Libtask` on demand. As an alternative, we are implementing a new package, [`SSMProblems`](https://github.com/TuringLang/SSMProblems.jl), to specify common-state-space models, which will be recommended if the user wants to specify a model as SSM and run particle filtering-related sampling algorithms.

devmotion:
Just saw the comment now (I'm very busy until Christmas and not too responsive on Github and Slack these days): An unfortunate consequence of the Julia compat change is also that we can't keep using Julia LTS in TuringTutorials (by doing so we avoid having to update the whole repo every time a new Julia version is available).

> In addition, @FredericWantiez and I discussed moving the general-purpose particle Gibbs sampler implementation (requires Libtask) into an experimental namespace and loading Libtask on demand

I assume you're talking about AdvancedPS here? If one can split the package cleanly in a part with and without Libtask that might be a good idea. Maybe one could move it to a subpackage (e.g., AdvancedPSLibtask) in the AdvancedPS repo to make the separation even clearer but avoid the friction of having to update two separate repos all the time.

> As an alternative, we are implementing a new package, [SSMProblems](https://github.com/TuringLang/SSMProblems.jl), to specify common-state-space models, which will be recommended if the user wants to specify a model as SSM and run particle filtering-related sampling algorithms.

If the dependencies of AdvancedPS are manageable, I wonder if it would be better to define the problem type there together with the algorithms. Otherwise it seems it is easy to introduce circular dependencies in the tests. Maybe an alternative could be to make it a subpackage, again to avoid having to re-iterate between both repos too much.

torfjelde:
Personally I'm _really_ not a big fan of dropping support for Julia 1.7. Julia 1.6 is still heavily used by Turing users, so this bump is very excluding.

torfjelde:
For example, we are in the process of bumping Setfield compat entry to 1 in Turing deps. Now this change won't benefit people using Julia 1.6.

devmotion:
Yes, I agree. I think support for Julia LTS is a good common ground throughout the ecosystem in general. It seems quite annoying for people that can't update Julia as easily (or don't want to, e.g., to avoid hunting bugs or performance regressions with newer versions). Since so many packages depend on each other (which is a good thing in general IMO) it, bumping the Julia compat can have a much larger effect than originally anticipated (see, e.g., also https://github.com/tpapp/DynamicHMC.jl/pull/172).

torfjelde:
Would it be possible to revert this change @yebai ? I have some PRs I'd like to make, but basing it on the current master worries me.

torfjelde:
I agree the decoupling is a good idea, but decoupling can be done without
merging this PR; you can make a PR to a non-master branch. This is probably
the better route as it does not hold back other updates to Turing.

torfjelde:
Do we have any updates on this? There are people asking in the slack channel regarding why we dropped support for 1.6

PR #1865, opened 2022-07-23
Minor update to tests after new DPPL update

torfjelde:


test/Project.toml
```diff
@@ -38,10 +38,10 @@ AdvancedPS = "0.3"
 AdvancedVI = "0.1"
 Clustering = "0.14"
 CmdStan = "6.0.8"
-Distributions = "0.25"
+Distributions = "<0.25.65"
```
devmotion, line 41: Due to the deprecation issues in 0.25.65, I assume? They are fixed in 0.25.66 so I guess this can be reverted:

```suggestion
Distributions = "0.25"
```

yebai (approved):
thanks @torfjelde!

PR #1869, opened 2022-07-26
Latex in `for-developers` section tested through `turing.ml`.

tomroesch:
Following up on my previous PR  #1857 which did not fix the issue. This time I build the website locally through `turing.ml` and was able to resolve the issues. I will add a PR to `turing.ml` that allows for the website to be built locally.


docs/src/for-developers/variational_inference.md
```diff
@@ -11,7 +11,7 @@ Here we'll have a look at the theory behind VI, but if you're interested in how
 
 # Motivation
 
-In Bayesian inference one usually specifies a model as follows: given data ``$\{x_i\}_{i = 1}^n$``, 
+In Bayesian inference one usually specifies a model as follows: given data ``\\{x_i\\}_{i = 1}^n``,
```
devmotion, line 14: I'm surprised `\` is escaped here but not in other places?

docs/src/for-developers/variational_inference.md
```diff
@@ -20,257 +20,257 @@ In Bayesian inference one usually specifies a model as follows: given data ``$\{
 \end{align*}
 ```
 
-where ``$\overset{\text{i.i.d.}}{\sim}$`` denotes that the samples are identically independently distributed. Our goal in Bayesian inference is then to find the _posterior_
+where ``\overset{\text{i.i.d.}}{\sim}`` denotes that the samples are identically independently distributed. Our goal in Bayesian inference is then to find the _posterior_
 
 ```math
-p(z \mid \{ x_i \}_{i = 1}^n) = \prod_{i=1}^{n} p(z \mid x_i).
+p(z \mid \\{ x\_i \\}\_{i = 1}^n) = \prod\_{i=1}^{n} p(z \mid x\_i).
```
devmotion, line 26: Similarly, why is `\_` (and `\\`) used here but other formulas use `_` and `\`?

docs/README.md
```diff
@@ -30,28 +30,34 @@ locally and review your changes, you will need to do the following:
 2. Install Julia by following [the relevant
    guide](https://julialang.org/downloads/).
 
-3. Navigate in a terminal to `Turing.jl/docs/`. Run `julia make.jl`,
+3. Clone the [`turing.ml` repository](https://github.com/TuringLang/turing.ml), 
+   where the website is built from.
+
+4. Add your local version of `Turing.jl` to the `turing.ml` environment by running
+   `using Pkg; Pkg.dev("path/to/Turing.jl")`.
```
devmotion, line 37: I don't understand this instruction, what do you have in mind? Where's the `turing.ml` environment? It seems you just add Turing.jl to your global environment since no environment was activated and you only navigate to turing.ml in the next step?

I guess also commonly one would run `] dev Turing` as initial command to check out and work on the Turing repo. Then it will be cloned to .julia/dev/Turing automatically.

docs/src/for-developers/variational_inference.md
tomroesch: Honestly I do not fully understand this either. However, to make curly braces appear in the equation, it had to be escaped. This was also true with line breaks when using the `align` mode.

docs/src/for-developers/variational_inference.md
tomroesch: The underscore has to be escaped if there are multiple ones in the same line, since when rendering the pages those will be interpreted as italic font first, which then breaks the rendering of the equation.

docs/README.md
tomroesch: When testing, I activated the `turing.ml` environment, by going into the repo and running `] activate .` This is what I usually do when working on packages. Also, the first two lines in the `make.jl` file say `import Pkg; Pkg.activate(".")`, so I thought it would be a more appropriate test. Do you think this is unnecessary? 

I forked the repo and then added it by running `] dev .` when being in the repo. However, then the `.julia/dev/` folder was empty. Anyways, I assume the instructions could be simpler so I am open to any suggestions.

devmotion: To avoid confusion and make it clearer that one has to use turing.ml, maybe we could remove all instructions and files other than the md sources, and only add a note that links to the instructions in the turing.ml repo?

Of course, then the question is still how to set it up and explain it in the turing.ml repo but we would reduce redundancies, inconsistencies, and possibly avoid some confusion.

tomroesch: Yes I think this is the way to go. I'll add a commit to this PR later today.

docs/src/for-developers/variational_inference.md
```diff
@@ -20,257 +20,257 @@ In Bayesian inference one usually specifies a model as follows: given data ``$\{
...
 ```
 
-In general one cannot obtain a closed form expression for <span>``$p(z \mid \{ x_i \}_{i = 1}^n)$``</span>, but one might still be able to _sample_ from ``$p(z \mid \{ x_i \}_{i = 1}^n)$`` with guarantees of converging to the target posterior ``$p(z \mid \{ x_i \}_{i = 1}^n)$`` as the number of samples go to ``$\infty$``, e.g. MCMC.
+In general one cannot obtain a closed form expression for <span>``p(z \mid \\{ x\_i \\}\_{i = 1}^n)``</span>, but one might still be able to _sample_ from ``p(z \mid \\{ x\_i \\}\_{i = 1}^n)`` with guarantees of converging to the target posterior ``p(z \mid \\{ x\_i \\}\_{i = 1}^n)`` as the number of samples go to ``\infty``, e.g. MCMC.
```
devmotion, line 29: I think `<span>` is another workaround for the underscore issue and could be removed now:
```suggestion
In general one cannot obtain a closed form expression for ``p(z \mid \\{ x\_i \\}\_{i = 1}^n)``, but one might still be able to _sample_ from ``p(z \mid \\{ x\_i \\}\_{i = 1}^n)`` with guarantees of converging to the target posterior ``p(z \mid \\{ x\_i \\}\_{i = 1}^n)`` as the number of samples go to ``\infty``, e.g. MCMC.
```

yebai:
Many thanks, @tomroesch for the help! I have followed up with a new PR that cleans up some obsolete files https://github.com/TuringLang/Turing.jl/pull/1872

PR #1872, opened 2022-08-10
Remove obsolete files in the docs folder.

yebai:
These files were moved to https://github.com/TuringLang/turing.ml quite a while ago. I suggest that we remove them to minimise redundancy and confusion.


devmotion (approved):
:100: I guess we should also remove https://github.com/TuringLang/Turing.jl/blob/hg/clean-docs/.github/workflows/Documentation.yml.

PR #1880, opened 2022-08-23
Remove outdated and unused Stan interface and tests

devmotion:
The code is not used anymore, and in particular the additional CI tests seem unnecessary. If needed, we could recover the code from the git history.


yebai (approved):
Thanks @devmotion!

PR #1881, opened 2022-08-25
feat: muliple iterations within Gibbs

xukai92:
Add the original Gibbs interface back

### DIFF
--- a/src/inference/gibbs.jl
+++ b/src/inference/gibbs.jl
@@ -41,20 +41,27 @@ Tips:
 methods like Particle Gibbs. You can increase the effectiveness of particle sampling by including
 more particles in the particle sampler.
 """
-struct Gibbs{space, A<:Tuple} <: InferenceAlgorithm
+struct Gibbs{space, A<:Tuple, B<:Tuple} <: InferenceAlgorithm
     algs::A   # component sampling algorithms
-
-    function Gibbs{space, A}(algs::A) where {space, A<:Tuple}
+    iterations::B
+    function Gibbs{space, A}(algs::A, iterations) where {space, A<:Tuple}
         all(isgibbscomponent, algs) || error("all algorithms have to support Gibbs sampling")
-        return new{space, A}(algs)
+        return new{space, A}(algs, iterations)
     end
 end
 
-function Gibbs(algs...)
+function Gibbs(args...)
+    if eltype(args) <: InferenceAlgorithm
+        algs = args
+        iterations = tuple(fill(length(args), 1))
+    else
+        algs = tuple(map(first, args))
+        iterations = tuple(map(last, args))
+    end
     # obtain space of sampling algorithms
     space = Tuple(union(getspace.(algs)...))
 
-    Gibbs{space, typeof(algs)}(algs)
+    Gibbs{space, typeof(algs)}(algs, iterations)
 end
 
 """
@@ -229,7 +236,7 @@ function AbstractMCMC.step(
     # Iterate through each of the samplers.
     vi = state.vi
     samplers = state.samplers
-    states = map(samplers, state.states) do _sampler, _state
+    states = map(samplers, spl.iterations, state.states) do _sampler, iteration, _state
         # Recompute `vi.logp` if needed.
         if _sampler.selector.rerun
             vi = last(DynamicPPL.evaluate!!(model, rng, vi, _sampler))
@@ -239,7 +246,10 @@ function AbstractMCMC.step(
         current_state = gibbs_state(model, _sampler, _state, vi)
 
         # Step through the local sampler.
-        _, newstate = AbstractMCMC.step(rng, model, _sampler, current_state; kwargs...)
+        newstate = current_state
+        for _ in 1:iteration
+            _, newstate = AbstractMCMC.step(rng, model, _sampler, current_state; kwargs...)
+        end
 
         # Update `VarInfo` object.
         vi = gibbs_varinfo(model, _sampler, newstate)
--- a/test/inference/gibbs.jl
+++ b/test/inference/gibbs.jl
@@ -54,6 +54,10 @@
         gibbs = Gibbs(PG(15, :z1, :z2, :z3, :z4), HMC(0.15, 3, :mu1, :mu2))
         chain = sample(MoGtest_default, gibbs, 5_000)
         check_MoGtest_default(chain, atol=0.15)
+
+        alg = Gibbs(MH(:s) => 10, HMC(0.2, 4, :m) => 10)
+        chain = sample(gdemo(1.5, 2.0), alg, 5_000)
+        check_numerical(chain, [:s, :m], [49/24, 7/6], atol=0.1)
     end
 
     @turing_testset "transitions" begin


yebai:
@torfjelde  Can you take a look at this PR? The aim is to make Gibbs a bit more general.

src/inference/gibbs.jl
```diff
@@ -41,20 +41,27 @@ Tips:
...
 
-function Gibbs(algs...)
+function Gibbs(args...)
+    if eltype(args) <: InferenceAlgorithm
+        algs = args
+        iterations = tuple(fill(1, length(args))...)
+    else
+        algs = tuple(map(first, args)...)
+        iterations = tuple(map(last, args)...)
+    end
```
devmotion, line 60: Maybe it would be a bit cleaner to use different constructors for both use cases with differently typed Varargs. I also wonder if one should actually support `::Tuple{InferenceAlgorithm, Int}` or if `Gibbs(::NTuple{N,InferenceAlgorithm}, ::NTuple{N, Int})` would be sufficient.

I also think a default of `nothing` would be better than `tuple(fill(1, ...))` (which should rather be `ntuple(_ -> 1, length(args))` to be inferrable I guess). Statements such as `if x.iterations !== nothing` could be compiled away completely if one just uses standard Gibbs, one could dispatch on it, and it's a simpler type.

src/inference/gibbs.jl
```diff
@@ -239,7 +246,10 @@ function AbstractMCMC.step(
         current_state = gibbs_state(model, _sampler, _state, vi)
 
         # Step through the local sampler.
-        _, newstate = AbstractMCMC.step(rng, model, _sampler, current_state; kwargs...)
+        newstate = current_state
+        for _ in 1:iteration
+            _, newstate = AbstractMCMC.step(rng, model, _sampler, current_state; kwargs...)
```
devmotion, line 251:
```suggestion
            _, newstate = AbstractMCMC.step(rng, model, _sampler, newstate; kwargs...)
```

src/inference/gibbs.jl
xukai92: For interfaces, we have a few options:
1. `Gibbs(HMC(0.1, 2, :x), PG(30, :y), MH(:z), (10, 20, 30))
2. `Gibbs((HMC(0.1, 2, :x), 10), (PG(30, :y), 20), (MH(:z), 30)
3. `Gibbs(HMC(0.1, 2, :x) => 10, PG(30, :y) => 20, MH(:z), => 30)
The current implementation supports both 2 and 3.
I prefer it for 2 reasons:
1. It's similar to the old interface syntax: `Gibbs(HMC(0.1, 2, 10, :x), PG(30, 20, :y, MH(30, :z)`
2. The relation between each Gibbs component and the corresponding number of iterations is easy to read as they are next to each other. If we do Option 1, one need to do `zip` in their mind to see how many iterations each component needs to run.

I'd perfer to set iterations to be `1` if it's not specified, as it is what the Gibbs loop really does: iterates each component for 1 iteration, which is just a special case of what's added in this PR.

devmotion: > The relation between each Gibbs component and the corresponding number of iterations is easy to read in Option 2 and 3 as they are next to each other

I'm fine with that but I still think that it would be better to add specific constructors for the supported options with `Vararg`s of the desired types. I.e., one with `Gibbs(alg1::InferenceAlgorithm, algs::InferenceAlgorithm...)` and one with `Gibbs(alg1::Tuple{InferenceAlgorithm, Int}, algs::Tuple{InferenceAlgorithm, Int}...)` (the first algorithm is written separately since otherwise one (re-)defines `Gibbs()`). Or with `Pair{...}` - but also I don't think one should support multiple options but rather decide for one to keep the API clean.

> I'd perfer to set iterations to be 1 if it's not specified, as it is what the Gibbs loop really does

Sure, that's how the "standard" case can be represented in the more general formulation. But it's not clear that this is the best way to implement the default case. The main motivation for using `nothing` instead is that 1) it allows the compiler to optimize the generated code for the default case (without looping and additional type stability analysis and potential problems) and 2) allows developers to dispatch on the default vs general case (not sure when this would be helpful but it's at least an option). So I think that `nothing` would be a better choice.

xukai92: > I'm fine with that but I still think that it would be better to add specific constructors for the supported options with `Varargs` of the desired types.

I agree with you the type signatures shall be improved.
Let me do what you suggest.

> Sure, that's how the "standard" case can be represented in the more general formulation. But it's not clear that this is the best way to implement the default case.

I'm not sure how much performance gain this optimization would actually give. The main computation happens in each Gibbs component and I don't feel getting rid of the for loop actually does much.

torfjelde:
I've had a look and am happy with it once the constructor-issue @devmotion raised has been addressed:)

IMO we'd ideally address this by just implementing composition for samplers instead, so that you could just construct kernels by doing, say, `sampler_1^10` to get a sampler that will apply `sampler_1` 10 times. Then there would be no need to specify the number of times to apply each sampler in the Gibbs.

But for now this is bueno:)

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
 alg = Gibbs(HMC(0.2, 3, :v1), PG(20, :v2))
 ```
 
+One can also pass the number of iterations for each Gibbs component using the following syntax:
+- `alg = Gibbs((HMC(0.2, 3, :v1), n_hmc), (PG(20, :v2), n_pg))`
+- `alg = Gibbs(HMC(0.2, 3, :v1) => n_hmc, PG(20, :v2) => n_pg)`
```
devmotion, line 41: Could we support only one syntax? Maybe only pairs as it more clearly illustrates the key-value-like relationship (but I'm fine with tuples as well)? I don't think there's a big advantage in supporting both syntaxes, and only supporting one would keep the API simpler and it would be clear what's the recommended/official way.

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+where `n_hmc` and `n_pg` are the number of iterations for HMC and PG in each Gibbs iteration.
+
 Tips:
 - `HMC` and `NUTS` are fast samplers, and can throw off particle-based
 methods like Particle Gibbs. You can increase the effectiveness of particle sampling by including
 more particles in the particle sampler.
 """
-struct Gibbs{space, A<:Tuple} <: InferenceAlgorithm
+struct Gibbs{space, A<:Tuple, B<:Tuple} <: InferenceAlgorithm
```
devmotion, line 49: Could even restrict it to

```suggestion
struct Gibbs{space,N, A<:NTuple{N,InferenceAlgorithm}, B<:NTuple{N,Int}} <: InferenceAlgorithm
```

BTW any change of the type parameters is in principle breaking as it can break downstream dispatches. So regardless of how the type parameters are changed, we should tag a breaking release.

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
     algs::A   # component sampling algorithms
-
-    function Gibbs{space, A}(algs::A) where {space, A<:Tuple}
+    iterations::B
+    function Gibbs{space, A, B}(algs::A, iterations::B) where {space, A<:Tuple, B<:Tuple}
```
devmotion, line 52: Similar this could be made more restrictive and safer:

```suggestion
    function Gibbs{space, N, A, B}(algs::A, iterations::B) where {space, N, A<:NTuple{N,InferenceAlgorithm}, B<:NTuple{N,Int}}
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
         all(isgibbscomponent, algs) || error("all algorithms have to support Gibbs sampling")
-        return new{space, A}(algs)
+        return new{space, A, B}(algs, iterations)
```
devmotion, line 54:
```suggestion
        return new{space, N, A, B}(algs, iterations)
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
     end
 end
 
-function Gibbs(algs...)
+function Gibbs(algs::InferenceAlgorithm...)
```
devmotion, line 58: This also defines `Gibbs`. To ensure that at least one algorithm is specified, we should use

```suggestion
function Gibbs(alg::InferenceAlgorithm, algs::Vararg{InferenceAlgorithm,N}) where {N}
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+    iterations = tuple(fill(1, length(algs))...)
```
devmotion, line 59: It would be better to use

```suggestion
    iterations = ntuple(_ -> 1, Val(N + 1))
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
     # obtain space of sampling algorithms
     space = Tuple(union(getspace.(algs)...))
```
devmotion, line 61:
```suggestion
    allalgs = (alg, algs...)
    space = Tuple(union(map(getspace, allalgs)...))
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+    return Gibbs{space, typeof(algs), typeof(iterations)}(algs, iterations)
```
devmotion, line 62:
```suggestion
    return Gibbs{space, N + 1, typeof(allalgs), typeof(iterations)}(allalgs, iterations)
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+end
 
-    Gibbs{space, typeof(algs)}(algs)
+function Gibbs(args::T...) where {T <: Union{Tuple{InferenceAlgorithm,Int}, Pair{InferenceAlgorithm,Int}}}
```
devmotion, line 65: Same here, this defines (or rather redefines `Gibbs()`. We should rather use something like:

```suggestion
function Gibbs(arg::Pair{InferenceAlgorithm,Int}, args::Vararg{Pair{InferenceAlgorithm,Int},N}) where {N}
```

One problem with the current code is also that it seems it is possible to mix tuples and pairs.

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+    algs = tuple(map(first, args)...)
+    iterations = tuple(map(last, args)...)
```
devmotion, line 67:
```suggestion
    allargs = (arg, args...)
    algs = map(first, allargs)
    iterations = map(last, allargs)
```

src/inference/gibbs.jl
```diff
@@ -36,25 +36,38 @@ end
...
+    # obtain space of sampling algorithms
+    space = Tuple(union(getspace.(algs)...))
+    return Gibbs{space, typeof(algs), typeof(iterations)}(algs, iterations)
```
devmotion, line 70:
```suggestion
    space = Tuple(union(map(getspace, algs)...))
    return Gibbs{space, N + 1, typeof(algs), typeof(iterations)}(algs, iterations)
```

test/inference/gibbs.jl
```diff
@@ -54,6 +54,14 @@
         gibbs = Gibbs(PG(15, :z1, :z2, :z3, :z4), HMC(0.15, 3, :mu1, :mu2))
         chain = sample(MoGtest_default, gibbs, 5_000)
         check_MoGtest_default(chain, atol=0.15)
+
+        for alg in [
+            Gibbs((MH(:s), 10), (HMC(0.2, 4, :m), 10)),
+            Gibbs(MH(:s) => 10, HMC(0.2, 4, :m) => 10),
+        ]
+            chain = sample(gdemo(1.5, 2.0), alg, 5_000)
+            check_numerical(chain, [:s, :m], [49/24, 7/6], atol=0.1)
```
devmotion, line 63: Is it necessary to specify the analytical values? I thought there's a function for it already, but maybe I misremember.

src/inference/gibbs.jl
xukai92: I'm fine with keeping only the `Pair` version but Hong prefers the `Tuple` version.

src/inference/gibbs.jl
xukai92: This looks more like a refactoring of `Gibbs` for me. Not sure if we want to do it in this PR.

src/inference/gibbs.jl
xukai92: The exisiting `Gibbs` constructor has to be parametric (i.e. `Gibbs{space, A, B}(algs, iterations)`) so I don't think it defines `Gibbs`.

xukai92:
It looks like the signature `Gibbs(args::T...) where {T <: Union{Tuple{InferenceAlgorithm,Int}, Pair{InferenceAlgorithm,Int}}}` is borken when the passed in algorithms have different types. Any idea what's the correct way to implement it?

src/inference/gibbs.jl
devmotion: The PR is breaking in any case.

Even without `N` it would be good to ensure at some point that the lengths are equal and we could restrict type parameters to
```suggestion
struct Gibbs{space, A<:Tuple{Vararg{InferenceAlgorithm}}, B<:Tuple{Vararg{Int}}} <: InferenceAlgorithm
```

src/inference/gibbs.jl
devmotion: A simple example that shows it does:
```julia
julia> struct Gibbs{space, A<:Tuple, B<:Tuple}
           algs::A
           iterations::B
           function Gibbs{space,A,B}(algs::A, iterations::B) where {space,A<:Tuple,B<:Tuple}
               return new{space,A,B}(algs, iterations)
           end
       end

julia> function Gibbs(algs::String...)
           iterations = ntuple(_ -> 1, Val(length(algs)))
           return Gibbs{:x,typeof(algs),typeof(iterations)}(algs, iterations)
       end
Gibbs

julia> Gibbs()
Gibbs{:x, Tuple{}, Tuple{}}((), ())
```

It's a bit annoying but every time you use `Vararg`s in Julia, you also define a method without any `Vararg`s and potentially redefine existing functions.

devmotion: And the definition below redefines it:
```julia
julia> function Gibbs(args::Pair{Real,Int}...)
           algs = map(first, args)
           iterations = map(last, args)
           return Gibbs{:x,typeof(algs),typeof(iterations)}(algs, iterations)
       end
Gibbs

julia> Gibbs()
ERROR: MethodError: Gibbs() is ambiguous. Candidates:
  Gibbs(algs::String...) in Main at REPL[3]:1
  Gibbs(args::Pair{Real, Int64}...) in Main at REPL[5]:1
Possible fix, define
  Gibbs()
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```

devmotion:
`Pair{<:InferenceAlgorithm, Int}` should work.

src/inference/gibbs.jl
devmotion: Then let's only use `Tuple`?

xukai92:
> Pair{<:InferenceAlgorithm, Int} should work.

Seems no? See minimal example below.

```julia
julia> struct A{T}
           v::T
           function A{T}(v::T) where {T}
               new{T}(v)
           end
       end

julia> function A(args::T...) where {T<:Tuple{Real,Int}}
           return A(args)
       end

julia> A((1.0, 2), (3, 4))
ERROR: MethodError: no method matching A(::Tuple{Float64, Int64}, ::Tuple{Int64, Int64})
Closest candidates are:
  A(::T...) where T<:Tuple{Real, Int64} at REPL[3]:1
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```

devmotion:
> > Pair{<:InferenceAlgorithm, Int} should work.
> 
> Seems no? See minimal example below.

It did for me. The main point of my suggestion was to use `Pair{<:InferenceAlgorithm, Int}` instead of `Pair{InferenceAlgorithm, Int}`. Thus similarly in your example you should use `Tuple{<:Real,Int}` instead of `Tuple{Real,Int}`.

xukai92:
It was a typo in my reply. I actually tried what you suggested:

```julia
julia> struct A{T}
           v::T
           function A{T}(v::T) where {T}
               new{T}(v)
           end
       end

julia> function A(args::T...) where {T<:Tuple{<:Real,Int}}
           return A(args)
       end
A

julia> A((1.0, 2), (3, 4))
ERROR: MethodError: no method matching A(::Tuple{Float64, Int64}, ::Tuple{Int64, Int64})
Closest candidates are:
  A(::T...) where T<:(Tuple{var"#s36", Int64} where var"#s36"<:Real) at REPL[2]:1
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

devmotion:
> EDIT: Ignore me. Basically it has to be A(args::Tuple{\u003C:Real,Int}...) instead of the signature above.

Yes, that's what I tried. But I did not realize that it makes a difference.

test/inference/gibbs.jl
devmotion: Found it: https://github.com/TuringLang/Turing.jl/blob/5990fae9f176e84d83bd119fa4a6b0e68f028493/test/test_utils/numerical_tests.jl#L56

```suggestion
            check_gdemo(chain; atol=0.1)
```

xukai92:
I added `N` to the Gibbs struct and its constuctors now.

torfjelde:
What is this PR? It seems quite big.

src/inference/gibbs_conditional.jl
```diff
@@ -49,7 +49,7 @@ m = inverse_gdemo(x)
 sample(m, Gibbs(GibbsConditional(:λ, cond_λ), GibbsConditional(:m, cond_m)), 10)
 ```
 """
-struct GibbsConditional{S, C}
+struct GibbsConditional{S, C} <: InferenceAlgorithm
```
devmotion, line 52: Can this change be reverted
```suggestion
struct GibbsConditional{S, C}
```
and instead the type restrictions of the Gibbs sampler and its constructors above be relaxed (`isgibbscomponent` in the inner constructor should be sufficient, I assume)? The problem is that any InferenceAlgorithm can be used in the `sample` calls but this really is not a standalone sampler.

torfjelde:
> @torfjelde @devmotion This PR is ready for another look. I am probably too involved to approve it myself.

Sorry, but was this approved in the end? It seems @devmotion only reviewed, but didnt' hit the approve button?

torfjelde:
That day probably won't come soon :sweat_smile: But I believe it's still preferable to wait until reviewers hit :+1:

PR #1892, opened 2022-10-17
Conditionally remove BernoulliLogit

devmotion:
This PR prepares Turing for https://github.com/JuliaStats/Distributions.jl/pull/1623.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.21.12"
+version = "0.21.13"
 
 [deps]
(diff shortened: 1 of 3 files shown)


devmotion:
Bump :slightly_smiling_face: 

All test failures should be unrelated.

yebai (approved):
Thanks, @devmotion - shouldn't we wait for https://github.com/JuliaStats/Distributions.jl/pull/1623 to be merged first?

devmotion:
This PR ensures that merging the Distributions PR does not break Turing (at least not the latest version). As long as one uses current and older versions of Distributions, BernoulliLogit is defined in Turing. But if it already exists since it is defined in Distributions, then we do not try to redefine it.

PR #1893, opened 2022-10-17
Better describe the restrictions when using reversediff

dlakelan:
When compiling the tape with the ReverseDiff system you can't use certain kinds of models. Previously the documentation had just said no to runtime if statements. However this is a HUGE restriction and not actually precise. The key is that at each step of sampling, the same instructions need to be executed each time, thus the same branches need to get executed. This will occur when the if statement can be determined at compile time, but will also occur whenever the if statements depend only on the data passed to the model, as the data is constant throughout the sampling. This is quite significant, because in the absence of the ability to use if statements based on the data, you can't even express models where you treat say males differently from females using an if to test for sex. (or any number of similar situations, such as treating different measurement instruments with different measurement error, or the like). 

By making the documentation more precise users will have less trouble determining what kinds of models they can use and will be able to safely use ReverseDiff on a much expanded set of models.

See discussion in this thread: https://discourse.julialang.org/t/clarification-on-setrdcache-in-turing/88767/6?u=dlakelan


devmotion:
Maybe one should be even clearer and say that one should not use compiled tapes if one is not absolutely certain that the computations (and hence in particular branches of if statements) are the same every time the model is run - including all implicitly called functions, log pdf computations etc.

docs/src/using-turing/autodiff.md
```diff
@@ -10,8 +10,7 @@ title: Automatic Differentiation
...
-
+To switch between the different AD backends, one can call function `Turing.setadbackend(backend_sym)`, where `backend_sym` can be `:forwarddiff` (`ForwardDiff`), `:tracker` (`Tracker`), `:zygote` (`Zygote`) or `:reversediff` (`ReverseDiff.jl`). When using `ReverseDiff`, to compile the tape only once and cache it for later use, the user has to call `Turing.setrdcache(true)`. However, note that the use of caching in certain types of models can lead to incorrect results and/or errors. Models for which the compiled tape can be safely cached are models where the computation doesn't change between different samples of the sampler. Thus, all loops should be fixed size, if-statements should only depend on conditions that can be determined at compile time (constants) or conditions that depend only on the data (which is constant throughout the sampling). If-statements that depend on the parameter values will vary in which branch they take from one sample to another and thus the record of the calculations compiled into the ReverseDiff cached tape will be incorrect. Thus you can not use ReverseDiff when your model makes decisions based on the value of parameters.
```
devmotion, line 13: Some suggestions:
```suggestion
To switch between the different AD backends, one can call function `Turing.setadbackend(backend_sym)`, where `backend_sym` can be `:forwarddiff` (`ForwardDiff`), `:tracker` (`Tracker`), `:zygote` (`Zygote`) or `:reversediff` (`ReverseDiff.jl`). When using `ReverseDiff`, to compile the tape only once and cache it for later use, the user has to call `Turing.setrdcache(true)`. However, note that the use of caching in certain types of models can lead to incorrect results and/or errors.
Compiled tapes should only be used if you are absolutely certain that the computation doesn't change between different executions of your model.
Thus, e.g., in the model definition and all im- and explicitly called functions in the model all loops should be of fixed size and and `if`-statements should always execute the same branches.
For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on the data will always execute the same branches during sampling (if the data is constant throughout sampling and, e.g., no mini-batching is used).
However, `if`-statements that depend on the model parameters can take different branches during sampling and hence the compiled tape might be incorrect.
Thus you must not use compiled tapes when your model makes decisions based on the model parameters.
```

dlakelan:
Hmm. Now I'm lost, I committed your suggested changes, but I have no idea where they actually are now. How do I see and make further edits?

dlakelan:
Ok! I think I figured it out. I added some additional text to expand the exposition of what is and isn't allowed. Not sure how to show that here?

PR #1897, opened 2022-10-24
added generated_quantities to docs

JasonPekos:
Currently there's no mention of generated_quantities in the docs (at least, in the Using Turing section).

This PR attempts to add a quick explanation of the function along with an example use case — non-centred parametrization of Neal's funnel.

### DIFF
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -595,6 +595,60 @@ plot(chn) # Plots statistics of the samples.
 
 There are numerous functions in addition to `describe` and `plot` in the `MCMCChains` package, such as those used in convergence diagnostics. For more information on the package, please see the [GitHub repository](https://github.com/TuringLang/MCMCChains.jl).
 
+### Reparametrization and generated_quantities
+
+Often, the most natural parameterization for a model is not the most computationally feasible. Consider the following
+(efficiently reparametrized) implementation of Neal's funnel [(Neal, 2003)](https://arxiv.org/abs/physics/0009028):
+
+```julia
+@model function Neal()
+    # Raw draws
+    y_raw ~ Normal(0,1)
+    x_raw ~ arraydist([Normal(0, 1) for i in 1:9])
+
+    # Transform:
+    y = 3*y_raw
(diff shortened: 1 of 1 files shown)


devmotion:
It's documented a bit in the docs of DynamicPPL (I assume since it is defined there): https://beta.turing.ml/DynamicPPL.jl/dev/api/#DynamicPPL.generated_quantities

JasonPekos:
Yeah, it's also used (but not explained) in the time series case study. Felt it was worth writing this up here because of [this](https://discourse.julialang.org/t/how-to-do-non-centered-parameterization-in-turing/44702) post, which is the top result right now (for me) when I search "reparametrization in turing".

The thread begins with discussion before `generated_quantities` was added:

> We currently don’t have a good way to record the transformed variable, but you can apply the transformation after the inference. We will likely add a mechanism to track arbitrary variables soon.

It seems like there's a lot of support for adding the specific example of Neal's funnel:

> Really nice. We should also add this to the Turing docs.

> Yeah, I’d be delighted to see this added!

and then the post ends with a recent-ish post from *after* `generated_quantities` was added:

> Is this possible now? I was just looking for the same thing. :slight_smile:
I don’t think I found it in the turing documentation.

So felt it would be worth to add, as I had a [slightly more complicated version of this](https://jasonpekos.github.io/pages/neal/) sitting around in my notes already. Also willing to implement the more complicated version (w notes on checking divergence diagnostics + images of the funnel) as a tutorial if that seems like a better fit.

yebai:
Many thanks, @JasonPekos - I agree these are useful for the general documentation. I would suggest that we add a cross-reference to the `DynamicPPL` docs so the user can find more details there. 

In addition, we are in the process of migrating all docs to `jmd` format and also moving them into the `TuringTutorials` repo (we might rename `TuringTutorials` to `TuringDocumentation` later to reflect this). Please see [here](https://github.com/TuringLang/TuringTutorials/pull/327) for an example. It might be better to close this PR and open a new one against TuringTutorials to be future-proof. @cgbotta should be able to help with the new `jmd` format and/or workflow if needed.

PR #1900, opened 2022-11-07
Compatibility with new DPPL version

torfjelde:
The goal of this PR is just to make Turing work with the new DynamicPPL version, _not_ to obtain full feature-parity of `SimpleVarInfo` and `VarInfo` (though it is a significant step towards exactly this).

Main changes are:
1. `unflatten` is now used to convert a vector into a `AbstractVarInfo`.
2. Use `(inv)link!!` instead of deprecated `(inv)link!`.
3. `setindex!!` instead of `setindex!`.
4. Use `evaluate` and capture resulting `AbstractVarInfo` to also support immutable impls of `AbstractVarInfo`.

Closes https://github.com/TuringLang/Turing.jl/pull/1899 and https://github.com/TuringLang/Turing.jl/pull/1898


src/inference/mh.jl
```diff
@@ -403,12 +400,7 @@ function propose!(
     densitymodel = AMH.DensityModel(Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext()))
     trans, _ = AbstractMCMC.step(rng, densitymodel, mh_sampler, prev_trans)
 
-    # TODO: Make this compatible with immutable `VarInfo`.
-    # Update the values in the VarInfo.
-    setindex!!(vi, trans.params, spl)
-    setlogp!!(vi, trans.lp)
-
-    return vi
+    return setlogp!!(DynamicPPL.unflatten(vi, spl, trans.param), trans.lp)
```
torfjelde, line 403:
```suggestion
    return setlogp!!(DynamicPPL.unflatten(vi, spl, trans.params), trans.lp)
```

torfjelde:
So it seems like on MacOS, the estimates are a bit more noisy, e.g. some tests are failing because now the error is slightly above the `atol=0.2` in the https://github.com/TuringLang/Turing.jl/blob/master/test/inference/Inference.jl#L76-L80.

And there's a bug with `MH` (it's producing completely incorrect results). Trying to figure that out now.

torfjelde:
> So it seems like on MacOS, the estimates are a bit more noisy, e.g. some tests are failing because now the error is slightly above the atol=0.2 in the https://github.com/TuringLang/Turing.jl/blob/master/test/inference/Inference.jl#L76-L80.

@yebai @devmotion Regarding this, should I just increase the threshold? It def seems like the samplers are doing the right thing (and locally on linux, the tests are passing).

yebai:
Sure

torfjelde:
I believe tests should be passing now, so a look-over for the PR would be ideal:)

Project.toml
```diff
@@ -55,6 +56,7 @@ NamedArrays = "0.9"
 Reexport = "0.2, 1"
 Requires = "0.5, 1.0"
 SciMLBase = "1.37.1"
+Setfield = "0.7.1, 0.8"
```
devmotion, line 59: Is Setfield 0.7 actually tested and needed? Maybe
```suggestion
Setfield = "0.8"
```
is sufficient?

src/contrib/inference/dynamichmc.jl
```diff
@@ -60,8 +60,8 @@ function DynamicPPL.initialstep(
 )
     # Ensure that initial sample is in unconstrained space.
     if !DynamicPPL.islinked(vi, spl)
-        DynamicPPL.link!(vi, spl)
-        model(rng, vi, spl)
+        vi = DynamicPPL.link!!(vi, spl, model)
+        vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
```
devmotion, line 64: Can you remind me why we run the model here?

devmotion: And why we use the `SamplingContext`?

src/contrib/inference/sghmc.jl
```diff
@@ -56,8 +56,8 @@ function DynamicPPL.initialstep(
 )
     # Transform the samples to unconstrained space and compute the joint log probability.
     if !DynamicPPL.islinked(vi, spl)
-        DynamicPPL.link!(vi, spl)
-        model(rng, vi, spl)
+        vi = DynamicPPL.link!!(vi, spl, model)
+        vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
```
devmotion, line 60: Same here?

src/contrib/inference/dynamichmc.jl
```diff
@@ -109,8 +109,8 @@ function AbstractMCMC.step(
     Q, _ = DynamicHMC.mcmc_next_step(steps, state.cache)
 
     # Update the variables.
-    vi[spl] = Q.q
-    DynamicPPL.setlogp!!(vi, Q.ℓq)
+    vi = DynamicPPL.setindex!!(vi, Q.q, spl)
+    vi = DynamicPPL.setlogp!!(vi, Q.ℓq)
```
devmotion, line 113: Do we have to redefine the log-density function since `l.vi` is potentially a different object?

src/contrib/inference/sghmc.jl
```diff
@@ -90,8 +90,8 @@ function AbstractMCMC.step(
     newv = (1 - α) .* v .+ η .* grad .+ sqrt(2 * η * α) .* randn(rng, eltype(v), length(v))
 
     # Save new variables and recompute log density.
-    vi[spl] = θ
-    model(rng, vi, spl)
+    vi = DynamicPPL.setindex!!(vi, θ, spl)
+    vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
```
devmotion, line 94: Same here

src/contrib/inference/sghmc.jl
```diff
@@ -238,8 +238,8 @@ function AbstractMCMC.step(
     θ .+= (stepsize / 2) .* grad .+ sqrt(stepsize) .* randn(rng, eltype(θ), length(θ))
 
     # Save new variables and recompute log density.
-    vi[spl] = θ
-    model(rng, vi, spl)
+    vi = DynamicPPL.setindex!!(vi, θ, spl)
+    vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
```
devmotion, line 242: Do we need a new logdensity function?

devmotion: And is sampling needed?

src/contrib/inference/sghmc.jl
```diff
@@ -209,8 +209,8 @@ function DynamicPPL.initialstep(
 )
     # Transform the samples to unconstrained space and compute the joint log probability.
     if !DynamicPPL.islinked(vi, spl)
-        DynamicPPL.link!(vi, spl)
-        model(rng, vi, spl)
+        vi = DynamicPPL.link!!(vi, spl, model)
+        vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
```
devmotion, line 213: Same here

torfjelde:
Hmm, one thing to note here: it seems the tests takes ~2X to run from before (I just looked at some previously closed PRs) :confused: I'm uncertain if this is compile-time or runtime, but nonetheless maybe something to be aware of.

devmotion:
Hmm that's very unfortunate. It would be good to know where the regression comes from, if it is run-time or compilation time, if it is AD-backend specific etc. Maybe comparing the timings and allocations in the logs (e.g., https://github.com/TuringLang/Turing.jl/actions/runs/3441679090/jobs/5741483837#step:6:825) could give some hints?

torfjelde:
Good news! It looks like the `Emcee` sampler is the cause.
1. Current
    - Windows: https://github.com/TuringLang/Turing.jl/actions/runs/3441679090/jobs/5741483751#step:6:828. Emcee tests takes **1.8hrs**.
    - MacOS: https://github.com/TuringLang/Turing.jl/actions/runs/3441679090/jobs/5741483837#step:6:825. Emcee tests takes **1275s**.
3. Old 
    - On Windows: https://github.com/TuringLang/Turing.jl/actions/runs/2955691229/jobs/4725904554#step:6:817. Emcee tests takes **24.7s**.
    - On MacOS: https://github.com/TuringLang/Turing.jl/actions/runs/2955691229/jobs/4725904639#step:6:814. Emcee tests takes **28.0s**.

I need to have a look at what's causing this though. It's also "interesting" that the effect is significantly different between architectures.

torfjelde:
> whether we accidentally evaluate the model twice somewhere in the flatten/unflatten pipeline?

And just for the record, we don't do any execution of the model in this. In particular, for `VarInfo`, nothing has changed in the flatten/unflatten functionality, i.e. `unflatten(vi, spl, x) = VarInfo(vi, spl, x)`.

src/contrib/inference/dynamichmc.jl
torfjelde: I _think_ this was done to update the logp after `link!`. _But_ now `link!!` also includes the log-absdet-jacobian correction, so we could potentially drop it.

And regarding why we use the `SamplingContext`: didn't want to make changes to something I wasn't 100% certain would have no side-effects. But yes, I also think this is no longer necessary.

The question is if we should make these changes _now_ or in another PR.

src/contrib/inference/dynamichmc.jl
torfjelde: Oooo good point! Though, in that case we should probably just do away with the `state.vi` and use `state.logdensity.varinfo`, no?

src/contrib/inference/sghmc.jl
torfjelde: Do we not need the `SamplingContext` in case this is used as part of a gibbs kernel?

torfjelde: As in, in that case `spl` should only be dealing with a subset of the variables.

torfjelde:
Okay, so this is all very weird.

If I run the tests locally, then the same slowdown occurs when I hit `include("inference/emcee.jl")`.

_But_ if I copy-paste the code from `runtests.jl` up until and including the `include("inference/emcee.jl")` and execute it in the REPL (using the env created by `TestEnv.activate`), then it runs in 30s, as before.

So I'm very confused.

Have you seen anything like this before @devmotion?

devmotion:
Maybe, I'm not sure. But I've definitely seen cases where `Pkg.test` behaves differently from running the same code in the REPL. 

The more obvious difference is that `Pkg.test` starts a Julia process with some specific commandline options (by default), leading to differences such as described in https://discourse.julialang.org/t/erratic-test-failure-difference-between-test-and-include-test-runtests-jl/72342 and similar discourse posts. So the first thing to check might be to start the Julia REPL with exactly the same commandline options as the ones used by `Pkg.test` (you could e.g. inspect `Base.julia_cmd()`).

I've also seen differences due to the use of Revise, so it might be good to also check the behaviour if Revise is not loaded (e.g., by starting a clean Julia process with `--startup-file=no`).

I guess you already checked if the same package versions are used?

And then, of course, scoping is different in the REPL: https://docs.julialang.org/en/v1/manual/variables-and-scoping/

torfjelde:
After reading your reply @devmotion I was reminded of how `depwarn=yes` has often be the cause of perf regressions like this, aaaand lo and behold it indeed was the case. Once https://github.com/TuringLang/DynamicPPL.jl/pull/433 is through, we should be good here.

src/contrib/inference/dynamichmc.jl
devmotion: Maybe safer to move these questions in a separate issue and potentially address them in a separate PR :+1:

test/runtests.jl
```diff
@@ -58,7 +58,7 @@ macro timeit_include(path::AbstractString) :(@timeit TIMEROUTPUT $path include($
 
     @testset "samplers (without AD)" begin
         @timeit_include("inference/AdvancedSMC.jl")
-        @timeit_include("inference/emcee.jl")
+        # @timeit_include("inference/emcee.jl")
```
yebai, line 61:
```suggestion
        @timeit_include("inference/emcee.jl")
```

src/contrib/inference/dynamichmc.jl
devmotion: Yes, I guess. It seems `state.vi` is not even used in the DynamicHMC calls above but only state.logendensity.varinfo implicitly. I assume currently the values in the state.logdensity.varinfo are not completely crucial though since in the DynamicHMC calls its values are updated anyway.

src/contrib/inference/sghmc.jl
devmotion: Do we not have a sampler-specific evaluation context? It's mainly because we just evaluate (it seems) that I think passing around `rng` seems unintuitive.

torfjelde:
Oh you made reverted some of my changes @yebai

Increasing the number of samples helped but didnt' solve it. I guess I'll just increase further.

test/inference/mh.jl
```diff
@@ -21,7 +21,7 @@
     @numerical_testset "mh inference" begin
         Random.seed!(125)
         alg = MH()
-        chain = sample(gdemo_default, alg, 2000)
+        chain = sample(gdemo_default, alg, 7000)
```
torfjelde, line 24:
```suggestion
        chain = sample(gdemo_default, alg, 10_000)
```

yebai (approved):

PR #1909, opened 2022-11-14
Bump minor version

torfjelde:
After #1900 we should, IMO, bump the minor version given that we did so in DPPL. _Technically_ this should be non-breaking, but I'm personally worried that there will be some unforeseen consequences OR people are simply using some internal functionality from DPPL (which _did_ have a breaking release).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.21.14"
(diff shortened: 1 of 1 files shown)


yebai (approved):
thanks @torfjelde - looks sensible to make a breaking release!

FYI @ChrisRackauckas, in case SciML hook into any `DynamicPPL` internals. We are in the process of making DynamicPPL core data structure more transparent, performant and hackable, but it might cause some disruptions along the process.

Issue #1923, opened 2022-12-23
 StableRNG in Turing is not yielding reproducible output 


sourish-cmi:
Hi All:

In the following code, we are using StableRNGs. However, the results are not reproducing. Can anybody help us? So that we can reproduce the results.

```
using Turing, StableRNGs

@model function example(x)
    m ~ Normal(0, 1)
    x ~ Normal(m, sqrt(1))
end

res = sample(StableRNG(123), example(1), NUTS(), 1000)
```

**First Run**

```
julia> res = sample(StableRNG(123), example(1), NUTS(), 1000)
┌ Info: Found initial step size
└   ϵ = 1.6
Chains MCMC chain (1000×13×1 Array{Float64, 3}):

Iterations        = 501:1:1500
Number of chains  = 1
Samples per chain = 1000
Wall duration     = 0.54 seconds
Compute duration  = 0.54 seconds
parameters        = m
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters      mean       std   naive_se      mcse        ess      rhat   ess_per_sec 
      Symbol   Float64   Float64    Float64   Float64    Float64   Float64       Float64 

           m    0.4905    0.6835     0.0216    0.0314   490.8002    1.0054      903.8677

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5% 
      Symbol   Float64   Float64   Float64   Float64   Float64 

           m   -0.8832    0.0403    0.4865    0.9456    1.7660

```

**Second Run**

```
julia> res = sample(StableRNG(123), example(1), NUTS(), 1000)
┌ Info: Found initial step size
└   ϵ = 3.2
Chains MCMC chain (1000×13×1 Array{Float64, 3}):

Iterations        = 501:1:1500
Number of chains  = 1
Samples per chain = 1000
Wall duration     = 0.55 seconds
Compute duration  = 0.55 seconds
parameters        = m
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters      mean       std   naive_se      mcse        ess      rhat   ess_per_sec 
[...]

PR #1924, opened 2022-12-23
Fix non-reproducible step sizes

devmotion:
Might be sufficient to fix https://github.com/TuringLang/Turing.jl/issues/1923.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.23.1"
+version = "0.23.2"
 
 [deps]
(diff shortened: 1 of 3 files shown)


torfjelde:
Sorry, didn't see this before making my own :sweat_smile: 

Should we maybe add a short test? E.g.

```julia
    @turing_testset "reproducibility" begin
        alg = NUTS(1000, 0.8)
        res1 = sample(StableRNG(123), gdemo_default, alg, 1000)
        res2 = sample(StableRNG(123), gdemo_default, alg, 1000)
        @test Array(res1) == Array(res2)
    end
```

to `test/inference/hmc.jl`?

torfjelde (approved):
LGTM but a test would be nice:)

yebai:
thanks @devmotion!

PR #1927, opened 2022-12-28
Fix expectation of s² and m in gdemo

rdiaz02:
The expectation of m cannot be anything different from 0, given the definition. Likewise for an inverse gamma random variable. Former sentence, "the expectation of s is 49/24 (2.04166666...) and the expectation of m is 7/6 (1.16666666...)." cannot be correct.

### DIFF
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -42,7 +42,7 @@ end
 ```
 
 
-Note: As a sanity check, the expectation of `s²` is 3 (theta/(alpha - 1)) and the expectation of `m` is 0. This can be easily checked using Prior:
+Note: As a sanity check, the prior expectation of `s²` is 3 (= theta/(alpha - 1), where theta is 3 and alpha is 2) and the prior expectation of `m` is 0. This can be easily checked using `Prior`:
 
 ```julia
 p1 = sample(gdemo(missing, missing), Prior(), 100000)
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -42,7 +42,12 @@ end
 ```
 
 
-Note: As a sanity check, the expectation of `s` is 49/24 (2.04166666...) and the expectation of `m` is 7/6 (1.16666666...).
+Note: As a sanity check, the expectation of `s²` is 3 (theta/(alpha - 1)) and the expectation of `m` is 0. This can be easily checked using Prior:
+
+```julia
+p1 = sample(gdemo(missing, missing), Prior(), 100000)
+```
+
 
 
 We can perform inference by using the `sample` function, the first argument of which is our probabilistic program and the second of which is a sampler. More information on each sampler is located in the [API]({{site.baseurl}}/docs/library).
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -42,7 +42,7 @@ end
 ```
 
 
-Note: As a sanity check, the prior expectation of `s²` is 3 (= theta/(alpha - 1), where theta is 3 and alpha is 2) and the prior expectation of `m` is 0. This can be easily checked using `Prior`:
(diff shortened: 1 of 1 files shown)


yebai (approved):
thanks, @rdiaz02 for the catch -- it is likely copy-and-pasted from some other places where we might be sampling from an actual posterior rather than the prior.

docs/src/using-turing/guide.md
```diff
@@ -42,7 +42,12 @@ end
 ```
 
 
-Note: As a sanity check, the expectation of `s` is 49/24 (2.04166666...) and the expectation of `m` is 7/6 (1.16666666...).
+Note: As a sanity check, the expectation of `s²` is 3 (theta/(alpha - 1)) and the expectation of `m` is 0. This can be easily checked using Prior:
```
devmotion, line 45: theta and alpha are not defined, are they? Maybe an alternative could be
```suggestion
Note: As a sanity check, the prior expectation of `s²` is `3/(2 - 1) = 3` and the prior expectation of `m` is 0. This can be easily checked using `Prior`:
```

rdiaz02: theta and alpha I took to be implicitly defined by the model. I think your wording is better but maybe it is even clearer if we still use theta and alpha explicitly in the expression as those are the names of the parameters given in the doc.

```suggestion
Note: As a sanity check, the prior expectation of `s²` is 3 (theta/(alpha - 1), where theta is 3 and alpha is 2) and the prior expectation of `m` is 0. This can be easily checked using `Prior:
```

devmotion: Which doc? I couldn't find theta and alpha in the guide, maybe I missed it?

rdiaz02: Sorry for the confusion, I meant the help for InverseGamma (?InverseGamma).

My argument, in full, for using alpha and theta is the following.

The user sees `InverseGamma(2, 3)`.  The user might not know/remember what the expected value of a random variable with an inverse gamma is.  But upon doing `?InverseGamma` they can see that in this model `alpha = 2, theta = 3`.  And they can go to the Wikipedia page linked from that same help for InverseGamma ---https://en.wikipedia.org/wiki/Inverse-gamma_distribution) that the expression for the mean is theta/(alpha - 1).   (Note that the Wikipedia uses beta instead of theta).

So by giving the value of 3 as an explicit function of alpha and theta (and saying what values alpha and theta are taking), a user can quickly check where that 3 and that 3/(2-1) are coming from.

devmotion: Thank you for the explanatoon! While I can see the motivation for alpha and theta now, I think that their use rather creates confusion. In the guide alpha and theta are not explained and not used in any other place, so readers do not know that they are supposed to refer to the namea lls of the parameters in the docstring of `InverseGamma`.

docs/src/using-turing/guide.md
```diff
@@ -42,7 +42,12 @@ end
 ```
 
 
-Note: As a sanity check, the expectation of `s` is 49/24 (2.04166666...) and the expectation of `m` is 7/6 (1.16666666...).
+Note: As a sanity check, the prior expectation of `s²` is 3 (= theta/(alpha - 1), where theta is 3 and alpha is 2) and the prior expectation of `m` is 0. This can be easily checked using `Prior`:
```
devmotion, line 45: Maybe another alternative could be
```suggestion
Note: As a sanity check, the prior expectation of `s²` is `mean(InverseGamma(2, 3)) = 3/(2 - 1) = 3` and the prior expectation of `m` is 0. This can be easily checked using `Prior`:
```

yebai:
thanks, @devmotion  and @rdiaz02 for the nice improvements!

docs/src/using-turing/guide.md
rdiaz02: Yes, you are right, that seems good too.

PR #1928, opened 2022-12-28
Fix error in example and add simpler example

rdiaz02:
As discussed in https://discourse.julialang.org/t/turing-guide-documentation-prob-examples-throws-errors/50319 the example of the use of `prob"..."` threw errors from at least:
- model was not instantiated
- some queries used `s` but the model used `s²`

In addition, it was unclear the role of the values in the instantiated model, which is now clarified with a simpler example that is easily checked against a standard normal.

### DIFF
--- a/docs/src/using-turing/guide.md
+++ b/docs/src/using-turing/guide.md
@@ -375,8 +375,35 @@ Similarly, when using a particle sampler, the Julia variable used should either 
 
 ### Querying Probabilities from Model or Chain
 
+Consider first the following simplified `gdemo` model:
+```julia
+@model function gdemo0(x)
+    s ~ InverseGamma(2, 3)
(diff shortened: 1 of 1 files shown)


yebai:
thanks @rdiaz02 -- it looks like a good improvement. 

FYI, the `prob` string macro will soon be depreciated in favour of a more intuitive and robust interface in https://github.com/TuringLang/DynamicPPL.jl/pull/438.

Related: https://beta.turing.ml/DynamicPPL.jl/dev/tutorials/prob-interface/

cc @torfjelde @devmotion @YongchaoHuang

yebai (approved):

rdiaz02:
Thank you!!! And thanks for letting me know about the new interface.

PR #1932, opened 2023-01-01
Fix test failure

devmotion:
It seems currently tests fail on e.g. Windows in #1929. This PR should fix the issue.


yebai (approved):

yebai:
thanks @devmotion!

PR #1936, opened 2023-01-13
Reduce usage of sampler

torfjelde:
Long-term we want to move away from using `DynamicPPL.Sampler` to decide which variables are being sampled, etc.
This PR reduces the explicit accessing and some usage of `DynamicPPL.Sampler` as one step towards this.
In addition, due to how `DynamicPPL.AbstractSampler` is handled differently for `SimpleVarInfo` and `VarInfo`, the current design of `LogDensityFunction` (which requires a sampler) means that we cannot use `SimpleVarinfo` in this regard (`SimpleVarInfo` + `AbstractSampler` results in always sampling variables, hence making it useless for computing gradients, etc.).


src/Turing.jl
```diff
@@ -27,25 +28,72 @@ function setprogress!(progress::Bool)
...
+hassampler(ctx::DynamicPPL.SamplingContext) = true
+hassampler(ctx::DynamicPPL.AbstractContext) = hassampler(DynamicPPL.NodeTrait(ctx), ctx)
+hassampler(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = false
+hassampler(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = hassampler(DynamicPPL.childcontext(ctx))
+
+# TODO: This is annoying. Should all this be moved to DynamicPPL so we can share the impl of `unflatten`?
+function unflatten(varinfo, context, θ)
+    return hassampler(context) ? unflatten(getsampler(context), varinfo, context, θ) : DynamicPPL.unflatten(varinfo, θ)
+end
+unflatten(sampler::DynamicPPL.AbstractSampler, varinfo, context, θ) = DynamicPPL.unflatten(varinfo, sampler, θ)
```
torfjelde, line 72: Should we maybe move all this to DPPL?

src/Turing.jl
```diff
@@ -27,25 +28,72 @@ function setprogress!(progress::Bool)
...
+end
+
+# Convenient for end-user.
+function LogDensityfunction(
+    model::DynamicPPL.Model,
+    varinfo::DynamicPPL.AbstractVarInfo=DynamicPPL.VarInfo(model),
+    context::DynamicPPL.AbstractContext=DynamicPPL.DefaultContext(),
+)
+    return LogDensityFunction(model, varinfo, context)
+end
```
devmotion: Should this be added at this point? IIRC it's not exported and not suggested that users use it (I think it's not even documented).

src/Turing.jl
```diff
@@ -27,25 +28,72 @@ function setprogress!(progress::Bool)
...
+getsampler(ctx::DynamicPPL.AbstractContext) = getsampler(DynamicPPL.NodeTrait(ctx), ctx)
+getsampler(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = getsampler(DynamicPPL.childcontext(ctx))
+
+hassampler(f::LogDensityFunction) = hassampler(f.context)
+hassampler(ctx::DynamicPPL.SamplingContext) = true
+hassampler(ctx::DynamicPPL.AbstractContext) = hassampler(DynamicPPL.NodeTrait(ctx), ctx)
+hassampler(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = false
+hassampler(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = hassampler(DynamicPPL.childcontext(ctx))
+
+# TODO: This is annoying. Should all this be moved to DynamicPPL so we can share the impl of `unflatten`?
```
devmotion, line 68: My plan was to move the whole LogDensityFunction to DynamicPPL at some point, as it only depends on things defined therein and is not Turing-specific.

src/essential/ad.jl
```diff
@@ -77,13 +77,17 @@ Find the autodifferentiation backend of the algorithm `alg`.
 """
 getADbackend(spl::Sampler) = getADbackend(spl.alg)
 getADbackend(::SampleFromPrior) = ADBackend()()
```
devmotion, line 79: Are these still needed? Or could we remove them and specialize `getADbackend(::SamplingContext{<:Sampler})` and `getADbackend(::SamplingContext{SampleFromPrior})`? I think mentally it's easier if `getADbackend` only operates on contexts.

src/essential/ad.jl
```diff
@@ -77,13 +77,17 @@ Find the autodifferentiation backend of the algorithm `alg`.
...
+getADbackend(ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.NodeTrait(ctx), ctx)
+getADbackend(ctx::DynamicPPL.SamplingContext) = getADbackend(ctx.sampler)
+getADbackend(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.childcontext(ctx))
+getADbackend(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = ADBackend()()
```
devmotion, line 83: Maybe it's easier to read if the version with traits is renamed
```suggestion
getADbackend(ctx::DynamicPPL.AbstractContext) = _getADbackend(DynamicPPL.NodeTrait(ctx), ctx)
getADbackend(ctx::DynamicPPL.SamplingContext) = getADbackend(ctx.sampler)
_getADbackend(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.childcontext(ctx))
_getADbackend(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = ADBackend()()
```
or it's reordered at least:
```suggestion
getADbackend(ctx::DynamicPPL.SamplingContext) = getADbackend(ctx.sampler)
getADbackend(ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.NodeTrait(ctx), ctx)

getADbackend(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = ADBackend()()
getADbackend(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.childcontext(ctx))
```

src/Turing.jl
```diff
@@ -27,25 +28,72 @@ function setprogress!(progress::Bool)
...
 end
 
+_get_indexer(ctx::DynamicPPL.AbstractContext) = _get_indexer(DynamicPPL.NodeTrait(ctx), ctx)
+_get_indexer(ctx::DynamicPPL.SamplingContext) = ctx.sampler
+function _get_indexer(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext)
+    return _get_indexer(DynamicPPL.childcontext(ctx))
+end
+function _get_indexer(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext)
+    return Colon()
+end
```
devmotion, line 92: Reorder to make it easier to read?
```suggestion
_get_indexer(ctx::DynamicPPL.SamplingContext) = ctx.sampler
_get_indexer(ctx::DynamicPPL.AbstractContext) = _get_indexer(DynamicPPL.NodeTrait(ctx), ctx)

_get_indexer(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = Colon()
function _get_indexer(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext)
    return _get_indexer(DynamicPPL.childcontext(ctx))
end
```

src/Turing.jl
```diff
@@ -27,25 +28,72 @@ function setprogress!(progress::Bool)
...
+getparams(f::LogDensityFunction) = f.varinfo[_get_indexer(f.context)]
```
devmotion, line 93: Should we define `getindex(::AbstractVarInfo, ::AbstractContext)` in DynamicPPL instead?

src/inference/mh.jl
```diff
@@ -375,6 +375,9 @@ function propose!!(
 
     # Make a new transition.
     densitymodel = AMH.DensityModel(Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext()))
+    densitymodel = AMH.DensityModel(
+        Base.Fix1(LogDensityProblems.logdensity, Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext()))
```
devmotion, line 379:
```suggestion
        Base.Fix1(LogDensityProblems.logdensity, Turing.LogDensityFunction(vi, model, DynamicPPL.SamplingContext(spl)))
```

src/modes/ModeEstimation.jl
```diff
@@ -46,36 +46,30 @@ DynamicPPL.childcontext(context::OptimizationContext) = context.context
...
-    return DynamicPPL.dot_tilde_assume(ctx, sampler, right, left, vns, vi)
-end
-
-function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext{<:LikelihoodContext}, sampler::SampleFromPrior, right, left, vns, vi)
+function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext{<:LikelihoodContext}, right, left, vns, vi)
     # Values should be set and we're using `SampleFromPrior`, hence the `rng` argument shouldn't
     # affect anything.
-    r = DynamicPPL.get_and_set_val!(Random.GLOBAL_RNG, vi, vns, right, sampler)
+    # TODO: Stop using `get_and_set_val!`.
+    r = DynamicPPL.get_and_set_val!(Random.GLOBAL_RNG, vi, vns, right, SampleFromPrior())
```
devmotion, line 64: Unrelated, but these `Random.GLOBAL_RNG` seem surprising and a bit fishy. Maybe at least use
```suggestion
    r = DynamicPPL.get_and_set_val!(Random.default_rng(), vi, vns, right, SampleFromPrior())
```
?

src/modes/ModeEstimation.jl
```diff
@@ -46,36 +46,30 @@ DynamicPPL.childcontext(context::OptimizationContext) = context.context
...
     return r, 0, vi
 end
 
-function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext, sampler::SampleFromPrior, right, left, vns, vi)
+function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext, right, left, vns, vi)
     # Values should be set and we're using `SampleFromPrior`, hence the `rng` argument shouldn't
     # affect anything.
-    r = DynamicPPL.get_and_set_val!(Random.GLOBAL_RNG, vi, vns, right, sampler)
+    # TODO: Stop using `get_and_set_val!`.
+    r = DynamicPPL.get_and_set_val!(Random.GLOBAL_RNG, vi, vns, right, SampleFromPrior())
```
devmotion, line 72:
```suggestion
    r = DynamicPPL.get_and_set_val!(Random.default_rng(), vi, vns, right, SampleFromPrior())
```

src/modes/ModeEstimation.jl
devmotion: Ah, I missed the comment above.

devmotion: But that means ideally we would not use a `SamplingContext`.

src/Turing.jl
torfjelde: I figured:) And as I added all these additional methods I was thinking "I'm probably going to just make a sibling-PR to DPPL with all of this". Given your comment too, I'll just do that :+1:

src/Turing.jl
torfjelde: IMO yeah; in fact, I kind of want to export it. 
Now that we're adopting the LogDensityProblems.jl interface more aggressively, I also find myself using it _very_ often. I also very much like it because most users will find the "full" description necessary to run the model, i.e. `model` + `varinfo` + `context` too "scary". If we can say "if you want to play around with your model, you can convert it to a simple thing which is compatible with LogDensityProblems", I think that's quite nice.

src/essential/ad.jl
torfjelde: Yeah I can try doing that! I was worried this would break too many tests and discourage me from completing the PR, but now that the tests are passing I can see if it's worth it.

src/Turing.jl
torfjelde: Agreed :+1:  Am moving all this to DPPL.

src/modes/ModeEstimation.jl
torfjelde: Yeah, this `get_and_set_val!` shouldn't actually be doing any sampling, but because of how it's implemented it still requires an `rng`. It basically samples if a value is not present in `varinfo` or if we're using `SampleFromUniform`.

src/Turing.jl
devmotion: I like it as well and I think we want to expose it somehow. My main questions are just whether the right time to do is during the refactoring and whether we want to expose it via a different API. For instance, we could also use something like
```julia
logjoint(model, ::Type{T}=VarInfo) where {T<:AbstractVarInfo} = LogDensityFunction(model, T(model), DefaultContext())
logprior(model, ::Type{T}=VarInfo) where {T<:AbstractVarInfo} = LogDensityFunction(model, T(model), PriorContext())
loglikelihood(model, ::Type{T}=VarInfo) where {T<:AbstractVarInfo} = LogDensityFunction(model, T(model), LikelihoodContext())
```
and keep `LogDensityFunction` an internal detail. Maybe that would be even easier for users.

torfjelde: Though I do like this interface, I feel like we still need to inform the end-user about it and might as well export it given that they'll be seeing it "a lot"?

devmotion: I'm not sure if they have to know about `LogDensityFunctions` at all if `logjoint` etc would be the API. The only relevant point for thr uaers would be (and should be documented) that these functions return something that follows the LogDensityProblems interface. Fields, names etc. should not matter for users.

devmotion: A bit similar to how users don't know (and don't have to know) what `truncated` returns. Similarly, `logjoint` would make it easier to use the same API for other models for which we might need something different than `LogDensityFunctions`.

yebai: I have seen some cases where the user manually constructs these `LogDensityFunctions`.

 It is an intuitive interface for working with models, compared to treating them as a somewhat black-box object where users can only interact with a set of APIs. It allows users to explore new sampling algorithms, for instance, or plug a Turing model inside a large manually coded models.

yebai (approved):
Good idea to phase out filtering `VarInfo` variables by `sampler`, the new `context` mechanism is strictly better, I think.

src/essential/ad.jl
```diff
@@ -77,13 +77,18 @@ Find the autodifferentiation backend of the algorithm `alg`.
 """
 getADbackend(spl::Sampler) = getADbackend(spl.alg)
 getADbackend(::SampleFromPrior) = ADBackend()()
+getADbackend(ctx::DynamicPPL.SamplingContext) = getADbackend(ctx.sampler)
+getADbackend(ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.NodeTrait(ctx), ctx)
+
+getADbackend(::DynamicPPL.IsLeaf, ctx::DynamicPPL.AbstractContext) = ADBackend()()
+getADbackend(::DynamicPPL.IsParent, ctx::DynamicPPL.AbstractContext) = getADbackend(DynamicPPL.childcontext(ctx))
 
 function LogDensityProblemsAD.ADgradient(ℓ::Turing.LogDensityFunction)
```
yebai, line 86: Most codes in this `ad` file can be safely transferred to `DynamicPPL` too. One benefit is that users of `DynamicPPL` get the gradient feature of `LogDensityFunction` without loading the `Turing` package.

src/Turing.jl
torfjelde: I think @devmotion agrees with this, but he's just saying that generally the user shouldn't need to use the constructor itself, but instead use the `logjoint`, etc. methods for constructing it. 

I think I'll agree with not exporting it for now, but instead define these "nicer" constructors just because exporting a new thing is non-breaking but removing an export _is_ breaking, hence we can always just export it later if we feel like it's useful.

devmotion: I don't question its usefulness (actually I think it's very useful and not surprised that it's used even though it's not documented) but solely the two points mentioned: whether we should add to the API in a PR that otherwise only performs internal refactoring and whether a more generic API without exposing `LogDensityFunctions` directly would be better.

src/essential/ad.jl
torfjelde: Yup, though I think I'll leave it here for the time being just because I want to keep the PRs simple.

src/Turing.jl
torfjelde: Btw, is there a reason why we prefer something like:
```julia
logjoint(model::Model, ::Type{T} = VarInfo) = LogDensityFunction(VarInfo(model), model, ...)
```
vs.
```julia
logjoint(model::Model, varinfo::AbstractVarInfo = VarInfo(model)) = LogDensityFunction(varinfo, model, ...)
```
/

torfjelde: This constructor is actually quite a bit annoying as it doesn't let you re-use an existing `AbstractVarInfo` :confused:

devmotion: It was just the first thing I thought of, maybe there are better approaches. But maybe that's also an indication that a separate PR and discussion about the API could be uaeful.

torfjelde: Could always add `make_logjoint`, etc.?

torfjelde: I agree with a separate PR, but worried it's going to delay exposing the LogDensityProblems.jl-functionality to the users. But I'll make separate PR for now, and if it takes too long we can just export `LogDensityFunction` :shrug:

devmotion: Or something like `logjointof` etc., similar to `(log)densityof`.

On a second thought though, I think generally (and particularly when dealing with other model type) it would be nice if there would be some curried form of `logjoint` etc. that gives you a function that you can evaluate for different variables and that would be consistent with the exiating `logjoint` etc. And generally there one would only want to provide the variables when evaluating the curried form. Maybe one could add a keyword argument for the LogDensityFunction case (similar to e.g. how one can adjust behaviour of AD backends in LogDensityProblemsAD by providing an example of the values that log density function will be evaluated with). But to summarize, to me the desire to provide `VarInfo(model)` when constructing the LogDensityFunction seems very specific to this case (and maybe not something many users would want to do?) but not something you would want to do for every kind of models, and hence maybe it should not be first-class part of a generic API but a bit more specific to this case such as maybe a keyword argument.

Project.toml
```diff
@@ -47,7 +47,7 @@ DataStructures = "0.18"
 Distributions = "0.23.3, 0.24, 0.25"
 DistributionsAD = "0.6"
 DocStringExtensions = "0.8, 0.9"
-DynamicPPL = "0.21"
+DynamicPPL = "0.22"
```
yebai, line 50:
```suggestion
DynamicPPL = "0.21"
```

test/Project.toml
```diff
@@ -40,7 +40,7 @@ Clustering = "0.14"
 Distributions = "0.25"
 DistributionsAD = "0.6.3"
 DynamicHMC = "2.1.6, 3.0"
-DynamicPPL = "0.21"
+DynamicPPL = "0.22"
```
torfjelde, line 43:
```suggestion
DynamicPPL = "0.21"
```

src/inference/mh.jl
```diff
@@ -374,7 +374,9 @@ function propose!!(
     prev_trans = AMH.Transition(vt, getlogp(vi))
 
     # Make a new transition.
-    densitymodel = AMH.DensityModel(Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext()))
+    densitymodel = AMH.DensityModel(
```
devmotion, line 377: Isn't it easier to use a `LogDensityModel`? Then you don't need Fix1.

torfjelde: Oh true!

torfjelde: Wait, does that work though? AFAIK we don't support wrapping something that implements `LogDensityModel` in `AMH.DensityModel`?

src/Turing.jl
```diff
@@ -8,6 +8,7 @@ using Libtask
 using Tracker: Tracker
 
 import AdvancedVI
+using DynamicPPL: DynamicPPL, LogDensityFunction
 import DynamicPPL: getspace, NoDist, NamedDist
 import LogDensityProblems
```
devmotion, line 13: Maybe this can be removed now?
```suggestion
```

src/inference/mh.jl
devmotion: We don't want to wrap it in a `DensityModel`, do we? `step` seem to support `LogDensityModel`s: https://github.com/TuringLang/AdvancedMH.jl/blob/1638f068261f936141404e232dd7f099b2cdde95/src/mh-core.jl#L74-L113

src/Turing.jl
torfjelde: Still need it for all the optimation related stuff.

src/modes/ModeEstimation.jl
```diff
@@ -46,45 +46,42 @@ DynamicPPL.childcontext(context::OptimizationContext) = context.context
...
     # affect anything.
-    r = DynamicPPL.get_and_set_val!(Random.GLOBAL_RNG, vi, vns, right, sampler)
+    # TODO: Stop using `get_and_set_val!`.
+    r = DynamicPPL.get_and_set_val!(Random.default_rng(), vi, vns, right, SampleFromPrior())
     return r, 0, vi
 end
 
-function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext, sampler::SampleFromPrior, right, left, vns, vi)
+_loglikelihood(dist::Distribution, x) = loglikelihood(dist, x)
+_loglikelihood(dists::AbstractVector{<:Distribution}, x) = loglikelihood(arraydist(dists), x)
```
torfjelde, line 69: Is this going to break on stuff like `x::Matrix .~ right::MultivariateDistribution`? :confused:

PR #1942, opened 2023-01-30
Fix test errors with LogDensityProblemsAD >= 1.2

devmotion:
Fixes test failures introduced by https://github.com/tpapp/LogDensityProblemsAD.jl/pull/5.


devmotion:
The (IMO internal) structs were moved to the extensions submodules.

PR #1943, opened 2023-01-31
Preserve `context` in `LogDensityFunction`

torfjelde:
After https://github.com/TuringLang/DynamicPPL.jl/pull/452 we can properly preserve the context related to the model.

For example, at the moment something like

``` julia
sample(DynamicPPL.contextualize(model, DynamicPPL.PriorContext()), NUTS(), 1000)
```

won't work because `LogDensityFunction` is constructed with `DefaultContext`, which then replaces the leaf-context in `model` when we call `evaluate!!`.

After this PR, stuff like the above works as intended.


src/inference/hmc.jl
```diff
@@ -538,7 +549,12 @@ function HMCState(
 
     # Get the initial log pdf and gradient functions.
     ∂logπ∂θ = gen_∂logπ∂θ(vi, spl, model)
-    logπ = Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext())
+    logπ = Turing.LogDensityFunction(
```
devmotion, line 552: Seems this is not tested (codecov complains)?

src/inference/mh.jl
```diff
@@ -375,7 +375,14 @@ function propose!!(
 
     # Make a new transition.
     densitymodel = AMH.DensityModel(
-        Base.Fix1(LogDensityProblems.logdensity, Turing.LogDensityFunction(vi, model, DynamicPPL.SamplingContext(rng, spl)))
+        Base.Fix1(
```
devmotion, line 378: It seems MH sampler should be tested as well?

devmotion (approved):
Looks good to me, I would just suggest adding a few additional tests to cover all changes. Feel free to merge afterwards.

src/inference/mh.jl
torfjelde: Done! Though it seems to really struggle with here, even with the prior (despite sampling from the prior; am really confused about this..)

src/inference/hmc.jl
torfjelde: Tests have been added :+1:

PR #1976, opened 2023-04-11
Insane memory-usage when using `Emcee` with large number of iterations

torfjelde:
On this branch:

``` julia
julia> using Turing

julia> @model demo() = x ~ Normal()

julia> sample(demo(), Emcee(10), 10_000)
Sampling 100%|██████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:05
Chains MCMC chain (10×2×10000 Array{Float64, 3}):

Iterations        = 1:1:10
Number of chains  = 10000
Samples per chain = 10
parameters        = x
internals         = lp

Summary Statistics
  parameters      mean       std      mcse      ess_bulk      ess_tail      rhat   ess_per_sec 
      Symbol   Float64   Float64   Float64       Float64       Float64   Float64       Missing 

           x   -0.0091    1.0040    0.0026   147424.4897   137431.1963    1.0026       missing

Quantiles

           x   -1.9819   -0.6879   -0.0095    0.6660    1.9589
```

On master I can't run the above code without blowing up my 32GB mem machine.

All because of one pesky splat.

### DIFF
--- a/src/inference/emcee.jl
+++ b/src/inference/emcee.jl
@@ -124,7 +124,8 @@ function AbstractMCMC.bundle_samples(
     # Extract names & construct param array.
     nms = [nms; extra_params]
     parray = map(x -> hcat(x[1], x[2]), zip(vals_vec, extra_values_vec))
-    parray = cat(parray..., dims=3)
+    # NOTE: Use `reduce` instead of splatting to avoid stack overflow.
+    parray = reduce((x, y) -> cat(x, y; dims=3), parray)
 
     # Get the average or final log evidence, if it exists.
     le = getlogevidence(samples, state, spl)


torfjelde:
Also, while we're here. Why is it that the `bundle_samples` for `Emcee` results in `num_walkers` taking the place of iterations and the actual iterations being treated as `num_chains`?

This means that `chain[end, :, :]` results in the full trajectory for _one_ walker, rather than, at least what I'd expect, the last iteration for all the walkers.

Is it because it makes more sense with downstream calculations?

torfjelde:
Tests are breaking with some `ReadOnlyError`? :confused:

cpfiffer:
> > Also, while we're here. Why is it that the bundle_samples for Emcee results in num_walkers taking the place of iterations and the actual iterations being treated as num_chains?
> 
> I also find this slightly strange. Maybe our mighty @cpfiffer can clarify?

I think this was an oversight, or at the very least it wasn't particularly thoughtfully chosen. I think I agree that it should probably be the other way around -- I would want the chain iterations to be actual iterations and the different walkers to be different "chains". I suspect this is honestly a typo because I can't find a reason to rationalize why it is the way it currently is.

torfjelde:
Should I switch the order then? And if so, I should make this a breaking release I suppose.

cpfiffer:
> Should I switch the order then? And if so, I should make this a breaking release I suppose.

I think so, on both counts.

torfjelde:
I'm getting a strange error in the tests when run on CI :confused: It results in a `ReadOnlyMemoryError`

torfjelde:
It seems the `ReadMemoryOnlyError` is basically the job running out of memory.

PR #1979, opened 2023-04-18
Dump DynamicPPL.jl version

torfjelde:
What the title says


test/runtests.jl
```diff
@@ -65,7 +65,7 @@ macro timeit_include(path::AbstractString) :(@timeit TIMEROUTPUT $path include($
     end
 
     Turing.setrdcache(false)
-    for adbackend in (:forwarddiff, :tracker, :reversediff)
+    for adbackend in (:forwarddiff, :reversediff)
```
devmotion, line 68: @torfjelde I just noticed this change - this seems quite dangerous because things might start breaking silently without us noticing it while still stating officially in the package (compat entry, docs etc.) that Tracker is supported.

So I think in case this is the official policy (which IMO is a bit sad because Tracker is quite stable, much simpler than ReverseDiff or Zygote, and started to receive updates more frequently again) then IMO it should be removed from the package, docs, etc., ideally before removing tests for it.

PR #1999, opened 2023-06-01
Warn user if we're struggling to find good init params

torfjelde:
Currently, if HMC can't find good initial params, we don't tell the user anything. This can be very confusing as you might just think something is broken.

This PR starts warning the user if we fail to initialize within 10 tries.


src/inference/hmc.jl
```diff
@@ -178,13 +178,22 @@ function DynamicPPL.initialstep(
     # If no initial parameters are provided, resample until the log probability
     # and its gradient are finite.
     if init_params === nothing
+        init_attempt_count = 0
+        init_user_warned = false
```
devmotion, line 182: Why is this needed? It seems it would be equivalent to only check `init_attempt_count == 10` below?
```suggestion
```

src/inference/hmc.jl
```diff
@@ -178,13 +178,22 @@ function DynamicPPL.initialstep(
     # If no initial parameters are provided, resample until the log probability
     # and its gradient are finite.
     if init_params === nothing
+        init_attempt_count = 0
```
devmotion, line 181: Wasn't the algorithm above already the first attempt?
```suggestion
        init_attempt_count = 1
```

src/inference/hmc.jl
```diff
@@ -178,13 +178,22 @@ function DynamicPPL.initialstep(
...
+        init_user_warned = false
         while !isfinite(z)
+            if !init_user_warned && init_attempt_count == 10
```
devmotion, line 184:
```suggestion
            if init_attempt_count == 10
```

src/inference/hmc.jl
```diff
@@ -178,13 +178,22 @@ function DynamicPPL.initialstep(
...
+                @warn "failed to find valid initial parameters in $(init_attempt_count) tries; consider providing explicit initial parameters using the `init_params` keyword"
+                init_user_warned = true
```
devmotion, line 186:
```suggestion
```

devmotion (commented):
Would be good, I assume, to add a simple test as well.

torfjelde:
Ah yes, I had the `have_warned_user` variable there because at first I was thinking I'd compare using inequality, but then I realized I should just do `==` instead but forgot to remoe the warned-user-variable :facepalm: 

Cheers @devmotion !

torfjelde:
Tests added :+1:

test/inference/hmc.jl
```diff
@@ -223,4 +223,18 @@
...
+    @turing_testset "warning for difficult init params" begin
+        attempt = 0
+        @model function demo_warn_init_params()
+            x ~ Normal()
+            if (attempt += 1) < 20
+                Turing.@addlogprob! -Inf
+            end
+        end
+
+        @test_warn "failed to find valid initial parameters in 10 tries; consider providing explicit initial parameters using the `init_params` keyword" begin
```
devmotion, line 236: This can't be tested with `@test_warn` (checks `stderr`) but one has to use `@test_logs`:
```suggestion
        @test_logs (:warn, "failed to find valid initial parameters in 10 tries; consider providing explicit initial parameters using the `init_params` keyword") begin
```

torfjelde:
I've never used `@test_logs` before, so a bit uncertain what's going on here. You seen this error before @devmotion ?

torfjelde:
This should be good to go now :+1:

test/inference/hmc.jl
```diff
@@ -223,4 +223,21 @@
...
+            x ~ Normal()
+            if (attempt += 1) < 30
+                Turing.@addlogprob! -Inf
+            end
+        end
+
+        @test_logs (
+            :warn,
+            "failed to find valid initial parameters in 10 tries; consider providing explicit initial parameters using the `init_params` keyword",
+        ) min_level = Logging.Warn begin
```
devmotion, line 239: Apparently Logging is not defined. But based on the docs, I think you could avoid it by using
```suggestion
        ) (:info,) match_mode=:any begin
```

torfjelde:
I'm a bit confused why CodeCoverage is complaining; given that the warning occurs, it clearly is hitting those lines?

torfjelde:
Sooo do we just ignore it? :confused:

torfjelde:
Looks like we're good! @devmotion @yebai

PR #2008, opened 2023-06-13
[WIP] Allow usage of `AbstractSampler`

torfjelde:
This is a very rough idea of how we could go about supporting `AbstractMCMC.AbstractSampler` to reduce the necessary glue code.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -148,8 +148,8 @@ end
 
 function AbstractMCMC.sample(
     rng::AbstractRNG,
-    model::AbstractModel,
-    sampler::Sampler{<:InferenceAlgorithm},
+    model::DynamicPPL.Model,
+    sampler::AbstractSampler,
     N::Integer;
     chain_type=MCMCChains.Chains,
     resume_from=nothing,
@@ -166,7 +166,7 @@ end
 
 function AbstractMCMC.sample(
     rng::AbstractRNG,
-    model::AbstractModel,
+    model::DynamicPPL.Model,
     alg::Prior,
     N::Integer;
     chain_type=MCMCChains.Chains,
@@ -209,8 +209,8 @@ end
 
 function AbstractMCMC.sample(
     rng::AbstractRNG,
-    model::AbstractModel,
-    sampler::Sampler{<:InferenceAlgorithm},
+    model::DynamicPPL.Model,
+    sampler::AbstractSampler,
     ensemble::AbstractMCMC.AbstractMCMCEnsemble,
     N::Integer,
     n_chains::Integer;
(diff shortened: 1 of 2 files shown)


src/inference/Inference.jl
```diff
@@ -148,8 +148,8 @@ end
 
 function AbstractMCMC.sample(
     rng::AbstractRNG,
-    model::AbstractModel,
-    sampler::Sampler{<:InferenceAlgorithm},
+    model::DynamicPPL.Model,
+    sampler::AbstractSampler,
```
torfjelde, line 152: This will inevitably lead to method ambiguities though, so we need a better solution.

torfjelde:
As demonstrated in #2011 it seems supporting this properly is somewhat non-trivial, and so for the moment I've decided to just introduce a simple `SamplerWrapper <: InferenceAlgorithm` which wraps a `AbstractMCMC.AbstractSampler` but inherits then inherits everything else that comes with being an `InferenceAlgorithm`.

The annoying thing (other than it being ugly) is that the user has to wrap the `AbstractMCMC.AbstractSampler` themselves, since otherwise we run into ambuigity issues as before. The positives: any sampler using `LogDensityFunction` is now usable and integrates "nicely" with Turing.

torfjelde:
I believe the tests are sometimes failing simply because of numerical issues.

Some of the test models, e.g. `demo_dot_assume_observe_index_literal`, have one variance parameter for every observation. This means that the MLE estimate will have the variance parameter `s` be 0. _Most_ of the time this seems to work just fine, i.e. LBFGS finds something close to 0, but sometimes it fails.

src/contrib/inference/abstractmcmc.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+# NOTE: Only thing that depends on the underlying sampler.
+# Something similar should be part of AbstractMCMC at some point:
+# https://github.com/TuringLang/AbstractMCMC.jl/pull/86
+getparams(transition::AdvancedHMC.Transition) = transition.z.θ
+getstats(transition::AdvancedHMC.Transition) = transition.stat
+
+getparams(transition::AdvancedMH.Transition) = transition.params
+getstats(transition) = NamedTuple()
```
devmotion, line 37: @torfjelde Isn't this obsolete now that #2026 was merged?

torfjelde: Ah gotcha. @JaimeRZP you can always just do `]add Turing#torfjelde/allow-abstractsampler-draft` if you want to try out recent developments. And if you want to develop features based on this branch, just create a branch based on this PR and then continue from there, as you did with #2028 :)

PR #2014, opened 2023-06-17
Restrict tests to FillArrays 1.0.0

torfjelde:
Attempt to fix tests as per https://github.com/TuringLang/DistributionsAD.jl/pull/248


devmotion:
Shouldn't we also (even retro-actively) bound FillArrays in the Turing package itself to avoid that users run into these issues?

torfjelde:
Possibly! My worry about that is that:
1. We're losing out on 1.1. and 1.2 just because of these somewhat specific scenarios. This doesn't seem to be an issue that occurs very widely.
2. Could quickly run into an issue where one of our deps bump their compat bound because of some bug that was fixed after 1.0.0, in which case the user won't be able to stay on the most recent version of Turing.

IMO we don't for now, and then if we get people running into this issue in practice, we bound Turing too.

devmotion:
In my impression this FillArrays issue is a major problem as it affects (log) normal distributions (see, e.g., https://github.com/TuringLang/DistributionsAD.jl/actions/runs/5258560449/jobs/9502944457). I guess there are just not more test failures in Turing since we don't test as exhaustively as in DistributionsAD. Hence I think we should completely remove any use of FillArrays > 1.0.0 from the Turing ecosystem until these upstream issues are debugged and fixed.

torfjelde:
Aight, let's do that then :+1:

test/inference/hmc.jl
```diff
@@ -218,10 +218,16 @@
     end
 
     @turing_testset "prior" begin
+        @model function demo_hmc_prior()
+            # NOTE: Used to use `InverseGamma(2, 3)` but this has infinite variance
+            # which means that it's _very_ difficult to find a good tolerance in the test below:)
```
torfjelde, line 223: No wonder this test has been causing so much annoyance :sweat_smile:

torfjelde:
Happy with this now @devmotion ?:)

PR #2016, opened 2023-06-19
Bugfix for `condition` + optimization

torfjelde:
As reported in https://discourse.julialang.org/t/two-equivalent-conditioning-syntaxes-giving-different-likelihood-values/100320/2 the following fails on the current release:

``` julia
@model function model1(x)
    μ ~ Uniform(0, 2)
    x ~ LogNormal(μ, 1)
end;

@model function model2()
    μ ~ Uniform(0, 2)
    x ~ LogNormal(μ, 1)
end;

model2(x) = model2() | (; x)

v = 1.0
w = [1.0]

ctx = Turing.OptimizationContext(DynamicPPL.LikelihoodContext())
Turing.OptimLogDensity(model1(v), ctx)(w)  # = 1.4189385332046727
Turing.OptimLogDensity(model2(v), ctx)(w)  # = 2.112085713764618
```

The problem here is that `OptimizationContext` assumes that it is always the immediate parent-context of a leaf-context, e.g. `LikelihoodContext`, and then it overrides the behavior of the leaf-context. This causes issues if a `model` has been `contextualize`ed since in that scenario the context attached to the model will be inserted at the leaf node of the `context` passed to `evaluate!!`, which in this case is `OptimizationContext`. 

This PR therefore replaces `OptimizationContext` with two leaf-contexts, which, since `OptimizationContext` currently alters leaf-behaviors, is the more correct way to approach it. And in doing so, we fix the bug above.


src/modes/ModeEstimation.jl
```diff
@@ -18,7 +18,8 @@ export  constrained_space,
         MAP,
         MLE,
         OptimLogDensity,
-        OptimizationContext,
+        LikelihoodOptimizationContext,
```
devmotion, line 21: Could we avoid defining more contexts? 🤔 Or at least not expose it to users?

src/modes/ModeEstimation.jl
```diff
@@ -37,45 +53,38 @@ The `OptimizationContext` transforms variables to their constrained space, but
...
+function DynamicPPL.dot_tilde_assume(ctx::DefaultOptimizationContext, right, left, vns, vi)
     # Values should be set and we're using `SampleFromPrior`, hence the `rng` argument shouldn't
     # affect anything.
     # TODO: Stop using `get_and_set_val!`.
     r = DynamicPPL.get_and_set_val!(Random.default_rng(), vi, vns, right, SampleFromPrior())
     return r, _loglikelihood(right, r), vi
 end
 
+# Shared.
+const OptimizationContext = Union{LikelihoodOptimizationContext,DefaultOptimizationContext}
```
devmotion, line 78: This means it's a breaking change.

torfjelde: But `OptimizationContext` isnn't being exported? 

It's exported from the module, but not from Turing.

src/modes/ModeEstimation.jl
torfjelde: Happy to not export; I just did it because it was already being done.

But it's worth pointing out that it's not being exported to the user; this is from the `ModeEstimation` module. You still want me to remove the export?

devmotion: Ideally, I would like to stick with a single OptimizationContext, that's my main complaint.

I think it should be possible to just combine it with the existing contexts. It seems orthogonal to LikelihoodContext etc., so it would be nice to combine them easily. This would also scale better than having dedicated optimization contexts for every other context you want to combine it with.

src/modes/ModeEstimation.jl
devmotion: Ah, good point, I missed that 👍

src/modes/ModeEstimation.jl
torfjelde: > This would also scale better than having dedicated optimization contexts for every other context you want to combine it with.

In what sense do you mean here? Even if we allow it to wrap other contexts (as is done in the current release), the scaling is the same, no?

torfjelde: I'd also just mention that this PR is somewhat urgent since it is silently breaking all MLE estimation with models using `condition` :confused:  So if we're talking about making improvements to how we deal with hierarchies of contexts _in general_ (i.e. beyond addressing this bug), we should defer that.

devmotion: > As in, what do you have in mind for

Sorry, I didn't have time to think about it more carefully earlier today. I opened a PR with what I imagined: https://github.com/TuringLang/Turing.jl/pull/2022

> In what sense do you mean here?

I thought about allowing all kind/types of other leaf contexts. But I guess you're right that it's not possible anyway (currently) (and also not supported in #2022).

torfjelde:
Closed in favour of #2022

PR #2020, opened 2023-06-20
SaveCSV simple callback

JaimeRZP:
Callback designed to allow users to save samples on the go to a CSV file. 
We might want to move it to the `TuringCallback`s in the future.


JaimeRZP:
moved to https://github.com/TuringLang/TuringCallbacks.jl/pull/44

PR #2022, opened 2023-06-20
Fix MLE with ConditionContext

devmotion:
Alternative to #2016 that does not introduce a new context.

Something along these lines I imagined in the discussion in https://github.com/TuringLang/Turing.jl/pull/2016/files#r1235430996.

~~I copied your tests @torfjelde from #2016 but just noticed that I forgot to give you credit for these. I'll try to fix that.~~

Edit: Fixed and made @torfjelde a co-author of the commit.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.26.1"
+version = "0.26.2"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/src/modes/ModeEstimation.jl
+++ b/src/modes/ModeEstimation.jl
@@ -39,41 +39,51 @@ intended to allow an optimizer to sample in R^n freely.
 """
 struct OptimizationContext{C<:AbstractContext} <: AbstractContext
     context::C
+
+    function OptimizationContext{C}(context::C) where {C<:AbstractContext}
+        leaf = DynamicPPL.leafcontext(context)
+        if !(leaf isa Union{DefaultContext,LikelihoodContext})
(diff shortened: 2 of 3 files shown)


torfjelde:
> Edit: Fixed and made @torfjelde a co-author of the commit.

Oh that's very unnecessary, but I apprecate it:)

Regarding the PR, I'm not quite convinced regarding the short-cutting of the tilde-pipeline :confused:

Examples of failures:

devmotion:
Nice examples, I hadn't considered other scenarios 👍 My computer is already turned off but it seems this could be fixed by making OptimizationContext a leaf context, couldn't it? Then the difference between the PRs would basically boil down to OptimizationContext{<:LikelihoodContext} vs LikelihoodOptimizatiomContext.

torfjelde:
> Then the difference between the PRs would basically boil down to OptimizationContext{<:LikelihoodContext} vs LikelihoodOptimizatiomContext.

Exactly. I felt two separate contexts made more sense instead doing`OptimizationContext{<:LikelihoodContext}`, etc

torfjelde:
Btw, no matter which option we go, we should add the above to tests.

devmotion:
> Btw, no matter which option we go, we should add the above to tests.

Yes, I thought the same but it took me a while to copy and edit it on my phone 😅 Hopefully I did not introduce too many typos.

I think a neat thing about just fixing the existing types is that the diff is quite small (could even be more minimal without merging the tilde definitions - but I liked that it became IMO more obvious that the only difference is the log density calculation) and the type structure is unchanged. I also think it reflects a bit better that optimization contexts are wrappers of existing leafs such as DefaultContext and LikelihoodContext with small modifications.

torfjelde:
That makes sense :+1: I went with the other option because I thought it was a bit "disengineous" to have a leaf context wrap leaf contexts :sweat_smile: 

But let's go with your approach :+1: I'll have a look now.

torfjelde (approved):
LGTM!

PR #2026, opened 2023-06-27
Transition

JaimeRZP:
Some Samplers define their own `Transition` structures. 
This PR unifies all of these particular structures into a single one.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -105,15 +105,22 @@ end
 # Default Transition #
 ######################
 
-struct Transition{T, F<:AbstractFloat}
-    θ  :: T
-    lp :: F
+struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
+    θ     :: T
+    lp    :: F
+    stat  :: S
 end
 
 function Transition(vi::AbstractVarInfo, nt::NamedTuple=NamedTuple())
     theta = merge(tonamedtuple(vi), nt)
     lp = getlogp(vi)
-    return Transition{typeof(theta), typeof(lp)}(theta, lp)
+    return Transition{typeof(theta), typeof(lp)}(theta, lp, nothing)
+end
+
+function Transition(vi::AbstractVarInfo)
+    theta = tonamedtuple(vi)
+    lp = getlogp(vi)
+    return ransition{typeof(theta), typeof(lp)}(theta, lp, nothing)
 end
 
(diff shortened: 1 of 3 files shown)


src/inference/Inference.jl
```diff
@@ -105,18 +105,23 @@ end
...
+    θ = merge(tonamedtuple(vi), nt)
     lp = getlogp(vi)
-    return Transition{typeof(theta), typeof(lp)}(theta, lp)
+    return Transition(θ, lp, nothing)
 end
 
-metadata(t::Transition) = (lp = t.lp,)
+function metadata(t::Transition)
+    return merge((lp = t.lp,), t.stat)
+end
```
devmotion, line 124: This will throw an error if `t.stat === nothing`. I think you need something like
```suggestion
function metadata(t::Transition)
    stat = t.stat
    if stat === nothing
        return (lp = t.lp,)
    else
        return merge((lp = t.lp,), t.stat)
    end
end
```

src/inference/Inference.jl
```diff
@@ -105,18 +105,28 @@ end
...
+    return Transition(θ, lp, nothing)
 end
 
-metadata(t::Transition) = (lp = t.lp,)
+function metadata(t::Transition)
+    stat = t.stat
+    if stat === nothing
+        return (lp = t.lp,)
+    else
+        return merge((lp = t.lp,), t.stat)
```
devmotion, line 127: Sorry, a tiny "mistake" in my suggestion: I wanted to use
```suggestion
        return merge((lp = t.lp,), stat)
```
here.

src/inference/Inference.jl
```diff
@@ -105,18 +105,28 @@ end
 # Default Transition #
 ######################
 
-struct Transition{T, F<:AbstractFloat}
-    θ  :: T
-    lp :: F
+struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
+    θ     :: T
+    lp    :: F
+    stat  :: S
```
devmotion, line 111: I wonder if we should simplify this to just
```suggestion
    stat  :: S
```
which would include `lp`. Then we also don't have to deal with the `nothing` case below and `metadata(t::Transition) = t.stat` (maybe that would motivate renaming the field even to `metadata`?).

devmotion (approved):
Looks good to me. I guess in the future it could/should maybe even be moved to AbstractMCMC?

torfjelde (approved):
LGTM too:)

PR #2027, opened 2023-07-03
tap --> 0.8

JaimeRZP:


src/contrib/inference/abstractmcmc.jl
```diff
@@ -58,6 +59,11 @@ function AbstractMCMC.step(
     # Link the varinfo.
     f = setvarinfo(f, DynamicPPL.link!!(getvarinfo(f), model))
 
+    # Make init_params
+    if init_params === nothing
+        init_params = 4 .* rand(rng, LogDensityProblems.dimension(f)) .- 2
```
devmotion, line 64: Is this supposed to replicate the default initialization that already exists in DynamicPPL? The motivation is a bit unclear to me since 1) initial parameters will already be set using this approach by the code in DynamicPPL, 2) IIRC `init_params` is supposed to be given in the original untransformed space of the variables and hence this initialization might produce values outside of the support (in DynamicPPL this initialization is performed in the transformed space where these values are always valid!), and 3) `init_params` is not used below.

PR #2030, opened 2023-07-05
Unify transition also in external samplers

JaimeRZP:
We no longer need `TuringTransition`

### DIFF
--- a/src/contrib/inference/abstractmcmc.jl
+++ b/src/contrib/inference/abstractmcmc.jl
@@ -3,20 +3,8 @@ struct TuringState{S,F}
     logdensity::F
 end
 
-struct TuringTransition{T,NT<:NamedTuple,F<:AbstractFloat}
-    θ::T
-    lp::F
-    stat::NT
-end
-
-function TuringTransition(vi::AbstractVarInfo, t)
-    theta = tonamedtuple(vi)
-    lp = getlogp(vi)
-    return TuringTransition(theta, lp, getstats(t))
-end
-
-metadata(t::TuringTransition) = merge((lp = t.lp,), t.stat)
-DynamicPPL.getlogp(t::TuringTransition) = t.lp
+metadata(t::Transition) = merge((lp = t.lp,), t.stat)
+DynamicPPL.getlogp(t::Transition) = t.lp
 
 state_to_turing(f::DynamicPPL.LogDensityFunction, state) = TuringState(state, f)
 function transition_to_turing(f::DynamicPPL.LogDensityFunction, transition)
@@ -24,7 +12,7 @@ function transition_to_turing(f::DynamicPPL.LogDensityFunction, transition)
     varinfo = DynamicPPL.unflatten(f.varinfo, θ)
     # TODO: `deepcopy` is overkill; make more efficient.
     varinfo = DynamicPPL.invlink!!(deepcopy(varinfo), f.model)
-    return TuringTransition(varinfo, transition)
+    return Transition(varinfo, transition)
 end
 
 # NOTE: Only thing that depends on the underlying sampler.


src/contrib/inference/abstractmcmc.jl
```diff
@@ -3,28 +3,19 @@ struct TuringState{S,F}
...
 end
 
-struct TuringTransition{T,NT<:NamedTuple,F<:AbstractFloat}
-    θ::T
-    lp::F
-    stat::NT
-end
-
-function TuringTransition(vi::AbstractVarInfo, t)
+function Transition(vi::AbstractVarInfo, t)
```
devmotion, line 6: Two comments: This constructor should be defined next to https://github.com/TuringLang/Turing.jl/blob/e41f58cabf0022c53d7c6c4a0b1e73368c6b4e38/src/inference/Inference.jl#L135 (otherwise it will be confusing for anyone reading the code) - and it seems it is not tested, so could you add a test that covers it?

JaimeRZP: Agreed! However, I am wrong to think this `Transition` would be tested when testing the external samplers?

src/inference/Inference.jl
```diff
@@ -130,6 +130,12 @@ struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
     stat  :: S
 end
 
+function Transition(vi::AbstractVarInfo, t)
+    theta = tonamedtuple(vi)
+    lp = getlogp(vi)
+    return Transition(theta, lp, getstats(t))
```
devmotion, line 136: Could we define a default fallback
```
getstats(t) = nothing
```
here? Then one could read and understand this method, and `getstats` would be extended in `src/contrib/inference/abstractmcmc.jl`. Additionally, we could set `t = nothing` as default argument above and unify it with the `Transition` constructor below.

BTW I think it is wrong that #2008 used a default `getstats(t) = NamedTuple()` since this is inconsistent with the existing `Transition` constructor below.

src/contrib/inference/abstractmcmc.jl
devmotion: I assumed that as well, and I assumed we test some of them - but apparently these lines are not tested.

src/inference/Inference.jl
JaimeRZP: Nice suggestions! 
Yes, I am also a little puzzled now that you point it out: @torfjelde do we have a good reason to default `getstats(t) = NamedTuple()` in contrib/abstractmcmc.jl?

src/inference/Inference.jl
```diff
@@ -123,19 +123,26 @@ end
 ######################
 # Default Transition #
 ######################
+# Default
+# Extended in contrib/inference/abstractmcmc.jl
+getstats(t) = nothing
 
-struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
+Base.@kwdef struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
```
devmotion, line 130: IMO in most cases you don't want to use `Base.@kwdef`, in particular if it's a non-user facing struct. I think defining the desired constructors is simpler and clearer here (and we don't need an all-keyword argument constructor).
```suggestion
struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
```

src/inference/Inference.jl
```diff
@@ -123,19 +123,26 @@ end
...
     lp    :: F # TODO: merge `lp` with `stat`
-    stat  :: S
+    stat  :: S = nothing
 end
 
-Transition(θ, lp) = Transition(θ, lp, nothing)
+function Transition(vi::AbstractVarInfo, t)
+    θ = tonamedtuple(vi)
+    lp = getlogp(vi)
+    return Transition(θ=θ, lp=lp, stat=getstats(t))
```
devmotion, line 139: Just make this
```suggestion
function Transition(vi::AbstractVarInfo, t=nothing; nt::NamedTuple=NamedTuple())
    θ = merge(tonamedtuple(vi), nt)
    lp = getlogp(vi)
    return Transition(θ, lp, getstats(t))
```
and everything is covered.

src/inference/Inference.jl
```diff
@@ -123,19 +123,26 @@ end
...
+end
 
 function Transition(vi::AbstractVarInfo; nt::NamedTuple=NamedTuple())
     θ = merge(tonamedtuple(vi), nt)
     lp = getlogp(vi)
-    return Transition(θ, lp, nothing)
+    return Transition(θ=θ, lp=lp)
 end
```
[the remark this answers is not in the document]
JaimeRZP: done!

src/inference/Inference.jl
torfjelde: Exactly. My intention was to implement the external sampler stuff without touching any other code, deferring these sort of changes to another PR. But given that we didn't make a release, we probably could have just put them all in the same PR :grimacing:

torfjelde:
@JaimeRZP are you not also changing `HMCTransition` in this PR? Isn't that redundant now?

src/inference/Inference.jl
```diff
@@ -123,19 +123,20 @@ end
...
+# Extended in contrib/inference/abstractmcmc.jl
+getstats(t) = nothing
 
 struct Transition{T, F<:AbstractFloat, S<:Union{NamedTuple, Nothing}}
     θ     :: T
     lp    :: F # TODO: merge `lp` with `stat`
     stat  :: S
 end
 
-Transition(θ, lp) = Transition(θ, lp, nothing)
```
devmotion, line 133: I think this is still needed and should not have been removed (maybe I suggested this incorrectly?) - it covers the cases where the first argument is not an `AbstractVarInfo`.

Project.toml
```diff
@@ -39,7 +39,6 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 [compat]
 AbstractMCMC = "4"
 AdvancedHMC = "0.3.0, 0.4"
-AdvancedMH = "0.6.8, 0.7"
```
devmotion, line 42: Just noticed now, seems this was removed by accident?

PR #2031, opened 2023-07-06
Extract Priors

JaimeRZP:
Implements the tool to extract the priors from a model proposed by @torfjelde in #2009 and applies to create a constructor for AdvancedMH samplers based on the model priors like:
```Julia
function StaticMH(model::Model)
    priors = extract_priors(model)
    return AMH.MetropolisHastings(AMH.StaticProposal(priors))
end

function RWMH(model::Model)
    priors = extract_priors(model)
    return AMH.MetropolisHastings(AMH.RandomWalkProposal(priors))
end
```
Where `extract_priors` is a new method defined in `infernece/inference.jl`

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -68,7 +68,8 @@ export  InferenceAlgorithm,
         resume,
         predict,
         isgibbscomponent,
-        externalsampler
+        externalsampler,
+        extract_priors
 
 #######################
 # Sampler abstraction #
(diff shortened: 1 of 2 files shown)


src/inference/Inference.jl
```diff
@@ -485,6 +486,54 @@ end
 # Utilities  #
 ##############
 
+Base.@kwdef struct PriorExtractorContext{D,Ctx} <: AbstractContext
```
devmotion, line 489: I think there's nothing Turing.jl specific in this context and the helper function below? Then IMO DynamicPPL would be the right place to define them.

JaimeRZP: @torfjelde what do you think? I have never played with upstream libs such as AbstractMCMC or DynamicPPL.

torfjelde: Yeah, I agree with @devmotion here

Sorry, I should have been explicit about it; I just assumed it was a PR to DPPL :grimacing:

JaimeRZP:
We also have to make LogDensityFunction take a named tupple as an input. 
It would make this [code](https://github.com/TuringLang/Turing.jl/blob/6725e4a887275201ed0edc93492316f8da4a2186/src/inference/mh.jl#L208-L274C4) much nicer.

JaimeRZP:
Moved the DynamicPPL bit to [#496](https://github.com/TuringLang/DynamicPPL.jl/pull/496)

PR #2032, opened 2023-07-10
Re-add AdvancedMH compat

devmotion:
Seems this was removed in #2030 by accident.


JaimeRZP:
Yes! I don't know how that happend but it was not intentional. Thank you @devmotion!

PR #2037, opened 2023-07-14
MH Constructor

JaimeRZP:
Allows the user to construct a MH sampler using the priors of a Turing model as proposals.

### DIFF
--- a/src/inference/mh.jl
+++ b/src/inference/mh.jl
@@ -188,6 +188,16 @@ function MH(space...)
     return MH{tuple(syms...), typeof(proposals)}(proposals)
 end
 
+function StaticMH(model::Model)
+    priors = extract_priors(model)
+    return AMH.MetropolisHastings(AMH.StaticProposal(priors))
+end
+
+function RWMH(model::Model)
+    priors = extract_priors(model)
+    return AMH.MetropolisHastings(AMH.RandomWalkProposal(priors))
+end
+
 #####################
 # Utility functions #
 #####################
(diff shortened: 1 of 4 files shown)


src/inference/mh.jl
```diff
@@ -188,6 +188,22 @@ function MH(space...)
     return MH{tuple(syms...), typeof(proposals)}(proposals)
 end
 
+function StaticMH(model::Model)
```
devmotion, line 191: It seems this shadows `AMH.StaticMH` (and the same below for `RWMH`)? This can't really be fixed by qualifying these constructors since then it would be type piracy - neither the function nor any of its arguments are owned by Turing.

More generally, my understanding from the PR in AMH was that these functions will be deprecated and renamed as they misleadingly suggest that something of type `StaticMH` will be constructed here.

JaimeRZP: Ok agreed, let's not make the confusion bigger in this PR. 
The user can pass the type of proposal for the priors with a kwarg.

torfjelde:
I'm not certain these failing tests are a good indication of whether there's something wrong with SimpleVarInfo (for example, the models for which it fails are already tested in DPPL) as there are too many moving parts here (for example, it doesn't seem like a good idea to just use `SimplevarInfo` in the logdensity evaluation, completely disregarding the varinfo present in `f`).

I'm working on this PR now though, so will push some updates soon.

torfjelde:
One identified issue is that external samplers are by default working with linked varinfo, while AMH expects unlinked or linked depending on which sampler is being used.

torfjelde:
Issues with RandomWalkProposal should now be fixed.

torfjelde:
Once #2018 is good to go, CI should hopefully pass here too.

PR #2042, opened 2023-07-18
Revert dependencies on FillArrays

yebai:
Revert https://github.com/TuringLang/Turing.jl/pull/2014 which was only introduced to restrict FillArrays version.


devmotion (approved):
LGTM, given that tests pass.

PR #2043, opened 2023-07-18
Fix redundant definition of `getstats(t::Any)`

yebai:
The `getstats(t::Any)` is defined twice in the following lines:

https://github.com/TuringLang/Turing.jl/blob/e1d3fef4633096e1e73872967b1281c31eee0f0b/src/inference/Inference.jl#L128

https://github.com/TuringLang/Turing.jl/blob/e1d3fef4633096e1e73872967b1281c31eee0f0b/src/contrib/inference/abstractmcmc.jl#L22

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -125,7 +125,7 @@ end
 ######################
 # Default
 # Extended in contrib/inference/abstractmcmc.jl
-getstats(t) = nothing
+function getstats end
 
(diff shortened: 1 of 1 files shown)


devmotion (commented):
IIRC we discussed in #2030 that the definition in abstractmcmc.jl should be removed since it was not consistent with the existing behaviour of setting metadata to `nothing`?

PR #2044, opened 2023-07-18
Fix redundant definition of `getstats`

devmotion:
Alternative to #2043.

### DIFF
--- a/src/inference/Inference.jl
+++ b/src/inference/Inference.jl
@@ -124,7 +124,6 @@ end
 # Default Transition #
 ######################
 # Default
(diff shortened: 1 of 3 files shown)


src/inference/Inference.jl
```diff
@@ -124,7 +124,6 @@ end
 # Default Transition #
 ######################
 # Default
-# Extended in contrib/inference/abstractmcmc.jl
```
devmotion, line 127: @yebai It's still extended there in line 19 for AdvancedHMC, so I suggest keeping this comment.

PR #2045, opened 2023-07-18
DynamicHMC

JaimeRZP:
Reworking DynamicHMC interface using external sampler support.

### DIFF
--- a/src/contrib/inference/dynamichmc.jl
+++ b/src/contrib/inference/dynamichmc.jl
@@ -3,23 +3,6 @@
 ###
 
 """
-    DynamicNUTS
-
-Dynamic No U-Turn Sampling algorithm provided by the DynamicHMC package.
-
-To use it, make sure you have DynamicHMC package (version >= 2) loaded:
-```julia
-using DynamicHMC
-```
-""" 
-struct DynamicNUTS{AD,space} <: Hamiltonian{AD} end
-
-DynamicNUTS(args...) = DynamicNUTS{ADBackend()}(args...)
-DynamicNUTS{AD}(space::Symbol...) where AD = DynamicNUTS{AD, space}()
-
-DynamicPPL.getspace(::DynamicNUTS{<:Any, space}) where {space} = space
-
-"""
     DynamicNUTSState
 
 State of the [`DynamicNUTS`](@ref) sampler.
@@ -27,94 +10,57 @@ State of the [`DynamicNUTS`](@ref) sampler.
 # Fields
 $(TYPEDFIELDS)
 """
-struct DynamicNUTSState{L,V<:AbstractVarInfo,C,M,S}
+struct DynamicNUTSState{L,C,M,S}
     logdensity::L
-    vi::V
     "Cache of sample, log density, and gradient of log density evaluation."
     cache::C
     metric::M
     stepsize::S
 end
 
-# Implement interface of `Gibbs` sampler
-function gibbs_state(
-    model::Model,
-    spl::Sampler{<:DynamicNUTS},
-    state::DynamicNUTSState,
-    varinfo::AbstractVarInfo,
-)
-    # Update the log density function and its cached evaluation.
-    ℓ = LogDensityProblemsAD.ADgradient(Turing.LogDensityFunction(varinfo, model, spl, DynamicPPL.DefaultContext()))
-    Q = DynamicHMC.evaluate_ℓ(ℓ, varinfo[spl])
-    return DynamicNUTSState(ℓ, varinfo, Q, state.metric, state.stepsize)
-end
-
-DynamicPPL.initialsampler(::Sampler{<:DynamicNUTS}) = SampleFromUniform()
-
-function DynamicPPL.initialstep(
+function AbstractMCMC.step(
     rng::AbstractRNG,
-    model::Model,
-    spl::Sampler{<:DynamicNUTS},
-    vi::AbstractVarInfo;
+    model::LogDensityModel,
+    spl::DynamicNUTS;
+    init_params = nothing,
     kwargs...
 )
-    # Ensure that initial sample is in unconstrained space.
-    if !DynamicPPL.islinked(vi, spl)
-        vi = DynamicPPL.link!!(vi, spl, model)
-        vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
-    end
-
-    # Define log-density function.
-    ℓ = LogDensityProblemsAD.ADgradient(Turing.LogDensityFunction(vi, model, spl, DynamicPPL.DefaultContext()))
+    # Unpack model
+    ℓ = model.logdensity
 
     # Perform initial step.
-    results = DynamicHMC.mcmc_keep_warmup(
+    results = mcmc_keep_warmup(
         rng,
         ℓ,
         0;
-        initialization = (q = vi[spl],),
-        reporter = DynamicHMC.NoProgressReport(),
+        initialization = (q = init_params,),
+        reporter = NoProgressReport(),
     )
-    steps = DynamicHMC.mcmc_steps(results.sampling_logdensity, results.final_warmup_state)
-    Q, _ = DynamicHMC.mcmc_next_step(steps, results.final_warmup_state.Q)
+    steps = mcmc_steps(results.sampling_logdensity, results.final_warmup_state)
+    transition, _ = mcmc_next_step(steps, results.final_warmup_state.Q)
+    state = DynamicNUTSState(ℓ, transition, steps.H.κ, steps.ϵ)
 
-    # Update the variables.
-    vi = DynamicPPL.setindex!!(vi, Q.q, spl)
-    vi = DynamicPPL.setlogp!!(vi, Q.ℓq)
-
-    # Create first sample and state.
-    sample = Transition(vi)
-    state = DynamicNUTSState(ℓ, vi, Q, steps.H.κ, steps.ϵ)
-
-    return sample, state
+    return transition, state
 end
 
 function AbstractMCMC.step(
     rng::AbstractRNG,
-    model::Model,
-    spl::Sampler{<:DynamicNUTS},
+    model::LogDensityModel,
+    spl::DynamicHMC.NUTS,
     state::DynamicNUTSState;
     kwargs...
 )
     # Compute next sample.
-    vi = state.vi
     ℓ = state.logdensity
     steps = DynamicHMC.mcmc_steps(
         rng,
-        DynamicHMC.NUTS(),
+        spl,
         state.metric,
         ℓ,
         state.stepsize,
     )
-    Q, _ = DynamicHMC.mcmc_next_step(steps, state.cache)
-
-    # Update the variables.
-    vi = DynamicPPL.setindex!!(vi, Q.q, spl)
-    vi = DynamicPPL.setlogp!!(vi, Q.ℓq)
-
-    # Create next sample and state.
-    sample = Transition(vi)
+    transition, _ = DynamicHMC.mcmc_next_step(steps, state.cache)
     newstate = DynamicNUTSState(ℓ, vi, Q, state.metric, state.stepsize)
 
-    return sample, newstate
+    return transition, newstate
 end


JaimeRZP:
I think the PR is ready for review!
There's a bit of code which I am not very happy with myself. 
Basically the problem stems from the fact that the external sampler support rightfully assumes the sampler to be `AbstractMCMC.AbstractSampler`. However, `DynamicHMC.NUTS()` is not an `AbstractMCMC.AbstractSampler`.
The solution that I found is to create a convenience constructor within the Turing interface that wraps it. 
But I find it pretty ugly, see below:

```julia

# Wraps DynamicHMC as an AbstractSampler
struct DynamicNUTS{S<:DynamicHMC.NUTS} <: AbstractSampler
    sampler::S
end

DynamicNUTS() = DynamicNUTS(DynamicHMC.NUTS())
externalsampler(spl::DynamicHMC.NUTS) = ExternalSampler(DynamicNUTS(spl))
externalsampler(spl::DynamicNUTS) = ExternalSampler(spl)
```

src/inference/Inference.jl
```diff
@@ -463,7 +463,6 @@ include("gibbs_conditional.jl")
 include("gibbs.jl")
 include("../contrib/inference/sghmc.jl")
 include("emcee.jl")
-include("../contrib/inference/dynamichmc.jl")
```
JaimeRZP: I think we needed this. 
 Otherwise, It cannot find the functions defined in `"../contrib/inference/dynamichmc.jl"`

ext/dynamichmc.jl
```diff
@@ -0,0 +1,86 @@
+###
+### DynamicHMC backend - https://github.com/tpapp/DynamicHMC.jl
+###
+
+import DynamicHMC
+
+# Wraps DynamicHMC as an AbstractSampler
```
devmotion, line 7: Is this really needed now that `externalsampler` exists?

ext/dynamichmc.jl
```diff
@@ -0,0 +1,86 @@
...
+struct DynamicNUTS{S<:DynamicHMC.NUTS} <: AbstractSampler
+    sampler::S
+end
+
+DynamicNUTS() = DynamicNUTS(DynamicHMC.NUTS())
+externalsampler(spl::DynamicHMC.NUTS) = ExternalSampler(DynamicNUTS(spl))
```
devmotion, line 13: Could we just use
```suggestion
externalsampler(spl::DynamicHMC.NUTS) = DynamicNUTS(spl)
```
?

devmotion (commented):
It seems the `Hamiltonian` supertype and all type info about AD and parameter spaces is removed?

ext/dynamichmc.jl
JaimeRZP: The problem comes from the fact that `ExternalSampler `expects an `AbstractMCMC.AbstractSampler` but DynamicHMC doesn't follow that interface. So first I have to create a wrapper for DynamicHMC.NUTS that turns it into an `AbstractMCMC.AbstractSampler`  and then apply the `ExternalSampler` wrapper to use the external sampler interface.

Ideally, we make a PR on DynamicHMC and make the samplers follow the interface but we don't own that repo.

ext/dynamichmc.jl
JaimeRZP: I think we no longer want to support Gibbs sampling right @yebai?

ext/dynamichmc.jl
devmotion: But do we actually want a `ExternalSampler` type here? Can't we always dispatch purely on `DynamicNUTS`? It's owned by us so there would be no type piracies involved.

src/Turing.jl
```diff
@@ -68,6 +56,15 @@ using .ModeEstimation
     export optimize
 end
 
+if !isdefined(Base, :get_extension)
+    using Requires
+end
+function __init__()
+    @static if !isdefined(Base, :get_extension)
```
devmotion, line 63: This should be moved outside of the `__init__` function to avoid defining it when the default is sufficient.

ext/dynamichmc.jl
devmotion: I guess your comment @JaimeRZP was related to me mentioning parameter spaces in https://github.com/TuringLang/Turing.jl/pull/2045#pullrequestreview-1541635867? What about the AD backend choice?

Project.toml
```diff
@@ -66,3 +66,12 @@ StatsBase = "0.32, 0.33, 0.34"
...
 julia = "1.7"
+
+[extensions]
+DynamicHMCExt = "DynamicHMC"
+
+[extras]
+DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb"
+
+[weakdeps]
+DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb"
```
devmotion, line 77: We need a compat entry now that it's a (weak) dependency (which IMO is good since it allows us to rule out incompatible DynamicHMC versions in a much better way) 🙂

JaimeRZP: done!

src/Turing.jl
JaimeRZP: Like this?(See last commit)

ext/dynamichmc.jl
JaimeRZP: You are correct. We own `DynamicNUTS `and we could dispatch on it. 
However, I think the current approach is closer to what we would like to do if we had access to DynamicHMC and preserves the logic of the ExternalSampler.

JaimeRZP:
> @JaimeRZP There are still some minor issues with this PR. It is mostly related to how to setup extensions robustly. You might find the following PR helpful for learning how to use Julia extensions
> 
> [TuringLang/DistributionsAD.jl#235](https://github.com/TuringLang/DistributionsAD.jl/pull/235)

I am still facing the same issue.
If I try to run:
``` 
using Random, DynamicHMC, LinearAlgebra, AbstractMCMC
using Revise, Turing

d = 10
@model function funnel()
    θ ~ Truncated(Normal(0, 3), -3, 3)
    z ~ MvNormal(zeros(d-1), exp(θ)*I)
    x ~ MvNormal(z, I)
end

Random.seed!(1)
(;x) = rand(funnel() | (θ=0,))
funnel_model = funnel() | (;x)

spl = DynamicHMC.NUTS()

external_spl = externalsampler(spl)
```
I get the error:
```
MethodError: no method matching externalsampler(::DynamicHMC.NUTS{Val{:generalized}})

Closest candidates are:
  externalsampler(::AbstractMCMC.AbstractSampler)
   @ Turing [~/Cambridge/Turing.jl/src/inference/](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jaimerz/Cambridge/Turing.jl/~/Cambridge/Turing.jl/src/inference/)Inference.jl:100


Stacktrace:
 [1] top-level scope
   @ In[5]:1
``` 

Which means that the extension is not loading even if I am `using DynamicHMC`.
Any ideas what might be the problem?

devmotion:
AFAICT probably the problem is not the loading of the extension but rather that you define a new function `externalsampler` instead of extending the existing one.

ext/dynamichmc.jl
devmotion: But as long as DynamicHMC does not implement the AbstractMCMC interface (and it's unclear if DynamicHMC ever wants to support AbstractMCMC) there's arguably no external sampler involved here. IMO also `sample(..., DynamicNUTS(), ...)` is a bit more convenient than `sample(...., externalsampler(DynamicNUTS()), ...)`.

src/Turing.jl
```diff
@@ -68,6 +56,13 @@ using .ModeEstimation
     export optimize
 end
 
+if !isdefined(Base, :get_extension)
+    using Requires
+end
+@static if !isdefined(Base, :get_extension)
+    @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" include("../ext/DynamicHMCExt.jl")
```
devmotion, line 63: Sorry if my previous suggestion was misleading: This line should be inside of `___init___` - but the `@static` check should be wrap `__init___`. So if extensions are supported, we don't define a custom __init__ function at all.

JaimeRZP:
> AFAICT probably the problem is not the loading of the extension but rather that you define a new function `externalsampler` instead of extending the existing one.

That was indeed the problem. Thank you so much @devmotion!
The code should be working now. 
We only have to iron out some design decisions.

src/Turing.jl
JaimeRZP: ok but if I try `@static function __init__()` it won't compile.
Is there something I am missing? Sorry for being dense.

ext/dynamichmc.jl
JaimeRZP: @yebai do you want to break the stale mate? I see the points of @devmotion, so I am happy to go down this route if we all agree.

ext/DynamicHMCExt.jl
```diff
@@ -0,0 +1,99 @@
...
+###
+
+
+if isdefined(Base, :get_extension)
+    using Turing
+    import DynamicHMC
+    import AbstractMCMC
+    import Random
+    import LogDensityProblems
+    import LogDensityProblemsAD
```
devmotion, line 13: BTW this is not OK - extensions should only load the main package (Turing) and the weak dependency (DynamicHMC). Everything else (including stdlibs) has to be loaded via those or made an additional weak dependency.

src/Turing.jl
devmotion: It's a pattern used all over the Julia ecosystem, so in prinicple it has to work. The problem here, however, might be the additional `@init`s - they also change the `__init__` function and I don't know how this interacts with this approach.

ext/DynamicHMCExt.jl
JaimeRZP: Ok let me know if the current approach is better.

ext/DynamicHMCExt.jl
```diff
@@ -0,0 +1,95 @@
...
+    spl::DynamicNUTS;
+    init_params = nothing,
+    kwargs...
+)
+    # Unpack model
+    # We wrap it again in ADgradient 
+    ℓ = LogDensityProblemsAD.ADgradient(model.logdensity)
+
+    # Make init params if nothing
+    if init_params == nothing
```
devmotion, line 52:
```suggestion
    if init_params === nothing
```
is generally preferrable.

ext/DynamicHMCExt.jl
```diff
@@ -0,0 +1,95 @@
...
+    using ..Turing.Inference: LogDensityProblemsAD
+end
+
+# Wraps DynamicHMC as an AbstractSampler
+struct DynamicNUTS{S<:DynamicHMC.NUTS} <: AbstractMCMC.AbstractSampler
+    sampler::S
+end
+
+DynamicNUTS() = DynamicNUTS(DynamicHMC.NUTS())
+Turing.externalsampler(spl::DynamicHMC.NUTS) = Turing.externalsampler(DynamicNUTS(spl))
```
torfjelde, line 25: Do we need to re-wrap in external here?

ext/DynamicHMCExt.jl
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+    DynamicNUTSState
+
+State of the [`DynamicNUTS`](@ref) sampler.
```
torfjelde, line 30:
```suggestion
State of the [`DynamicNUTS`](@ref) sampler.

# Fields
$(DocStringExtensions.FIELDS)
```

ext/DynamicHMCExt.jl
```diff
@@ -0,0 +1,95 @@
+module DynamicHMCExt
+###
+### DynamicHMC backend - https://github.com/tpapp/DynamicHMC.jl
+###
+
+
+if isdefined(Base, :get_extension)
+    import DynamicHMC
+    using Turing
+    using Turing: AbstractMCMC, Random, LogDensityProblems
```
torfjelde, line 10:
```suggestion
    using Turing: AbstractMCMC, Random, LogDensityProblems, DocStringExtensions
```

torfjelde (changes requested):
Made a few minor comments.

If we're indeed dropping support for usage of this in Gibbs sampling, then we need to make this a breaking release

ext/DynamicHMCExt.jl
JaimeRZP: Both you and @devmotion pointed this out. 
I will change it!

ext/DynamicHMCExt.jl
devmotion: Seems good :slightly_smiling_face:

JaimeRZP:
Ok I think we have one last disagreement and we are ready.
Basically the question is whether we need to do the following wrapping:
`ExternalSampler(DynamicNUTS(DynamicHMC.NUTS))`
where DynamicNUTS just turns DynamicHMC.NUTS into an AbstractSampler. 
The reason why this is currently done is two fold:

- DynamicHMC doesn't return an AbstractMCMC.AbstractSampler
- The external sampler interface explicity requires an  AbstractMCMC.AbstractSampler to trigger. 

Since both @torfjelde and @devmotion brought this up, I now have had a look at how you would do it otherwise.

- **Option a**: We try to get rid of DynamicNUTS. In order to do we would have to allow ExternalSampler to take in non-AbstractSamplers which in my opinion is not a good idea. Note that we cannot simply create our ExternalSampler structure which takes in DynamicHMC.NUTS since that would clash with the general structure which explicitly expects an AbstractSampler. 
- **Option b**: We reproduce the current abstractmcmc interface for DynamicHMC.NUTS. This in my opinion duplicates a lot of code and adds a lot of room for error. 

I hope I have convinced you that doing the current wrapping, while a little bit akward, is the most elegant solution that allows us to reuse the largest amount of code and preserve the types of the current interface. 

Let me know if I have missed something!

ext/DynamicHMCExt.jl
JaimeRZP: I think so! See post below

devmotion:
I think the simplest option is to just keep DynamicHMC as an internal sampling algorithm of type `InferenceAlgorithm` and continue dispatching on `Sampler{<:DynamicNUTS}`. Why should we complicate the convenient existing interface?

JaimeRZP:
> I think the simplest option is to just keep DynamicHMC as an internal sampling algorithm of type `InferenceAlgorithm` and continue dispatching on `Sampler{<:DynamicNUTS}`. Why should we complicate the convenient existing interface?

Well, I think the premise of the PR was to adapt the current DynamicHMC interface to the recently developed externalsampler interface to reduce code and homogenize the design. If DynamicHMC returned an AbstractSampler this could be done very elegantly. Since it doesn't we have to do, as you point out, some akward wrappings.

I thik it is @yebai call whether we want to change the goal of the PR and retain the current interface.

JaimeRZP:
> I think the simplest option is to just keep DynamicHMC as an internal sampling algorithm of type `InferenceAlgorithm` and continue dispatching on `Sampler{<:DynamicNUTS}`. Why should we complicate the convenient existing interface?

But then I think we shouldn't even make DynamicHMC an extension since  we would need to export `DynamicNUTS`.
I think if we really go down this route we really shouldn't do this PR at all which is probaly what @devmotion is politely trying to get to. 

I think what we have to get to is whether we are happy to exchange some elegance in the interface in order to unify the treatment of DynamicHMC with other external samplers.

devmotion:
> But then I think we shouldn't even make DynamicHMC an extension since we would need to export DynamicNUTS.

If we want to unify it with `externalsampler`, we could make `DynamicNUTS` an internal implementation details and define `externalsampler(alg::DynamicHMC.NUTS) = Sampler(DynamicNUTS(alg))`, couldn't we? I think this could be done without introducing new samplers and without changing the overall implementation apart from moving it to an extension.

torfjelde:
> I think the simplest option is to just keep DynamicHMC as an internal sampling algorithm of type InferenceAlgorithm and continue dispatching on Sampler{<:DynamicNUTS}. Why should we complicate the convenient existing interface?

Seconded.

> If we want to unify it with externalsampler, we could make DynamicNUTS an internal implementation details and define externalsampler(alg::DynamicHMC.NUTS) = Sampler(DynamicNUTS(alg)), couldn't we? I think this could be done without introducing new samplers and without changing the overall implementation apart from moving it to an extension.

This is basically what I had in mind when I asked whether we wanted to also wrap `DynamicNUTS` in `ExternalSampler`; seems easiest to just keep the existing way we use DynamicHMC and just have `externalsampler` return something that isn't actually `AbstractSampler`.

Project.toml
```diff
@@ -68,11 +69,14 @@ StatsFuns = "0.8, 0.9, 1"
 Tracker = "0.2.3"
 julia = "1.7"
 
-[weakdeps]
-Optim = "429524aa-4258-5aef-a3af-852621145aeb"
-
 [extensions]
+DynamicHMCExt = "DynamicHMC"
```
devmotion, line 73: Following the common convention, can you rename the extension to
```suggestion
TuringDynamicHMCExt = "DynamicHMC"
```
?

JaimeRZP:
Ok following the feedback I have brought back the previous DynamicHMC interface. 
Now `externalsampler(DynamicHMC.NUTS(args...))` returns the old DynamicNUTS object which doesn't trigger the external sampler interface. Instead it goes through `Inference/hmc`.
I have also kept the Gibbs interface for the sake of continuity. 
Nonetheless, I have modified the old DynamicNUTS structure to take an extra field, `sampler`, which is the particular initialization of `DynamicHMC.NUTS` the user provided to `externalsampler`. 
Previously, the user was forced to use the default `DynamicHMC.NUTS()` initialization which seems very restrictive. 

Hopefully this more aligned with your vision @torfjelde @devmotion!

src/Turing.jl
```diff
@@ -135,21 +135,13 @@ export  @model,                 # modelling
         optim_function,
         optim_problem
 
+if !isdefined(Base, :get_extension)
+    using Requires
+end
+
 function __init__()
     @static if !isdefined(Base, :get_extension)
```
devmotion, line 143: Can you re-try to move the `@static` check outside of `__init__()`? Maybe it works now since we do not use the `@init` macro anymore.

test/contrib/inference/dynamichmc.jl
```diff
@@ -1,15 +1,10 @@
-@stage_testset "dynamichmc" "dynamichmc.jl" begin
-    import DynamicHMC
+@testset "DynamicHMCExt" begin
+    using DynamicHMC
```
devmotion, line 2:
```suggestion
    import DynamicHMC
```
should still be sufficient.

ext/DynamicHMCExt.jl
```diff
@@ -12,10 +26,15 @@ To use it, make sure you have DynamicHMC package (version >= 2) loaded:
 using DynamicHMC
 ```
 """ 
-struct DynamicNUTS{AD,space} <: Hamiltonian{AD} end
+struct DynamicNUTS{AD,space} <: Turing.Inference.Hamiltonian{AD} 
+    sampler::DynamicHMC.NUTS
+end
```
devmotion, line 31: This will cause type inference issues, can we use
```suggestion
struct DynamicNUTS{AD,space,T<:DynamicHMC.NUTS} <: Turing.Inference.Hamiltonian{AD} 
    sampler::T
end
```
instead?

src/Turing.jl
JaimeRZP: I am afraid it still results in compilation errors see below:

ERROR: LoadError: ArgumentError: invalid @static macro
```
Stacktrace:
 [1] var"@static"(__source__::LineNumberNode, __module__::Module, ex::Any)
   @ Base ./osutils.jl:37
 [2] include
   @ ./Base.jl:457 [inlined]
 [3] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2010
 [4] top-level scope
   @ stdin:2
in expression starting at /home/jaimerz/Cambridge/Turing.jl/src/Turing.jl:142
in expression starting at /home/jaimerz/Cambridge/Turing.jl/src/Turing.jl:1
in expression starting at stdin:2
```

ext/DynamicHMCExt.jl
JaimeRZP: I implemented something like what you proposed but I feel like it is a little bit hacked together. 

```
struct DynamicNUTS{AD,space,T<:DynamicHMC.NUTS} <: Turing.Inference.Hamiltonian{AD} 
    sampler::T
end

DynamicNUTS(args...) = DynamicNUTS{Turing.ADBackend()}(args...)
DynamicNUTS{AD}(spl::DynamicHMC.NUTS, space::Tuple) where AD = DynamicNUTS{AD, space, DynamicHMC.NUTS}(spl)
DynamicNUTS{AD}(spl::DynamicHMC.NUTS) where AD = DynamicNUTS{AD}(spl, ())
DynamicNUTS{AD}() where AD = DynamicNUTS{AD}(DynamicHMC.NUTS())
Turing.externalsampler(spl::DynamicHMC.NUTS) = DynamicNUTS(spl)

DynamicPPL.getspace(::DynamicNUTS{<:Any, space}) where {space} = space
```

Any suggestions for improvements?

JaimeRZP:
After some thought I have actually removed the support for Gibbs

torfjelde:
> After some thought I have actually removed the support for Gibbs

Why?

src/Turing.jl
```diff
@@ -135,21 +135,15 @@ export  @model,                 # modelling
...
-                error("Please update DynamicHMC, v1.x is no longer supported")
-            end
-        end
-    end
+        @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" include("../ext/TuringDynamicHMCExt.jl")
+  end
 end
 
-end
+end # module
```
torfjelde, line 149:
```suggestion
end
```

ext/TuringDynamicHMCExt.jl
```diff
@@ -12,10 +26,15 @@ To use it, make sure you have DynamicHMC package (version >= 2) loaded:
...
 """ 
-struct DynamicNUTS{AD,space} <: Hamiltonian{AD} end
+struct DynamicNUTS{AD,space,T<:DynamicHMC.NUTS} <: Turing.Inference.Hamiltonian{AD} 
+    sampler::T
+end
 
-DynamicNUTS(args...) = DynamicNUTS{ADBackend()}(args...)
-DynamicNUTS{AD}(space::Symbol...) where AD = DynamicNUTS{AD, space}()
+DynamicNUTS(args...) = DynamicNUTS{Turing.ADBackend()}(args...)
+DynamicNUTS{AD}(spl::DynamicHMC.NUTS, space::Tuple) where AD = DynamicNUTS{AD, space, DynamicHMC.NUTS}(spl)
```
torfjelde, line 34:
```suggestion
DynamicNUTS{AD}(spl::DynamicHMC.NUTS, space::Tuple) where AD = DynamicNUTS{AD, space, typeof(spl)}(spl)
```

What you are currently doing still leads to type-instability since `DynamicHMC.NUTS` is not a concrete type.

test/contrib/inference/dynamichmc.jl
```diff
@@ -1,15 +1,10 @@
-@stage_testset "dynamichmc" "dynamichmc.jl" begin
+@testset "DynamicHMCExt" begin
```
torfjelde, line 1:
```suggestion
@testset "TuringDynamicHMCExt" begin
```

torfjelde (changes requested):
Added a few minor suggestions.

But it seems like you haven't removed Gibbs support, as per your previous comment, or am I missing something?

JaimeRZP:
> > After some thought I have actually removed the support for Gibbs
> 
> Why?

The API would have to be something like `externalsampler(DynamicHMC.NUTS(), :x)` which I think is very ugly and muddies what the input of external sampler should be.
Since, we decided to remove Gibbs support anyways I think not including it is the cleanest option.

torfjelde:
Ah okay, but this is only for the `externalsampler` interface, right? It still works if I construct the `DynamicNUTS` sampler manually?

src/Turing.jl
devmotion: What exactly did you try?

devmotion:
> We can no longer export DynamicNUTS from an extension.

You can export it from the main package. The common approach with extensions in such cases is to define the struct in the main package and possibly add an error hint if the extension/DynamicHMC is not loaded.

src/Turing.jl
JaimeRZP:
```
@static  function __init__()
    if !isdefined(Base, :get_extension)
        @require Optim="429524aa-4258-5aef-a3af-852621145aeb" include("../ext/TuringOptimExt.jl")
        @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" include("../ext/TuringDynamicHMCExt.jl")
  end
end
```

ext/DynamicHMCExt.jl
JaimeRZP: sorted!

src/Turing.jl
devmotion: It should be
```julia
@static if !isdefined(Base, :get_extension)
    function __init__()
        @require Optim="429524aa-4258-5aef-a3af-852621145aeb" include("../ext/TuringOptimExt.jl")
        @require DynamicHMC="bbc10e6e-7c05-544b-b16e-64fede858acb" include("../ext/TuringDynamicHMCExt.jl")
  end
end
```
Can you try that?

torfjelde:
Btw this needs to be made another breaking release.

mrazomej:
Dear Turing.jl team,

Having this change without updating the documentation simultaneously threw me into several wasted hours trying to understand why my code was not working anymore. Would it be possible to update the documentation for these breaking changes? I would do it myself if I knew what changed. But this goes beyond my passive-user knowledge of the structure of the package.

mrazomej:
My apologies. I don't know why the version of the website I was looking at had the previous method. Thank you very much.

PR #2047, opened 2023-07-19
Replace internal AD backend types with ADTypes

devmotion:
This PR is a draft proposal for replacing our internal AD backend types with ADTypes.

Needs support of ADTypes in LogDensityProblemsAD: https://github.com/tpapp/LogDensityProblemsAD.jl/pull/17


torfjelde:
Love it!

torfjelde:
We'll also need to make equivalent changes in AdvancedVI, I believe.

Red-Portal:
Hi, yes that's already up to date!

src/essential/ad.jl
```diff
@@ -36,21 +36,16 @@ function setchunksize(chunk_size::Int)
     AdvancedVI.setchunksize(chunk_size)
 end
 
-abstract type ADBackend end
-struct ForwardDiffAD{chunk,standardtag} <: ADBackend end
+# TODO: Upstream support for tags to ADTypes?
+struct AutoForwardDiff{chunk,standardtag} <: ADTypes.AbstractADType end
```
devmotion, line 40: I added this in https://github.com/SciML/ADTypes.jl/pull/12.

Project.toml
```diff
@@ -61,7 +63,7 @@ EllipticalSliceSampling = "0.5, 1, 2"
 ForwardDiff = "0.10.3"
 Libtask = "0.7, 0.8"
 LogDensityProblems = "2"
-LogDensityProblemsAD = "1.4"
+LogDensityProblemsAD = "1.6"
```
devmotion, line 66: Version 1.6.0 was broken:
```suggestion
LogDensityProblemsAD = "1.6.1"
```

src/essential/Essential.jl
```diff
@@ -11,6 +11,7 @@ using Bijectors: PDMatDistribution
 using AdvancedVI
 using StatsFuns: logsumexp, softmax
 @reexport using DynamicPPL
+using ADTypes: ADTypes, AutoTracker, AutoReverseDiff, AutoZygote
```
devmotion, line 14:
```suggestion
using ADTypes: ADTypes, AutoForwardDiff, AutoTracker, AutoReverseDiff, AutoZygote
```

Project.toml
```diff
@@ -44,6 +45,7 @@ TuringDynamicHMCExt = "DynamicHMC"
 TuringOptimExt = "Optim"
 
 [compat]
+ADTypes = "0.1.5"
```
devmotion, line 48: ADTypes 0.1.5 is quite old, we should support at least
```suggestion
ADTypes = "0.2"
```

devmotion:
I opened https://github.com/tpapp/LogDensityProblemsAD.jl/pull/21.

src/essential/Essential.jl
```diff
@@ -40,10 +46,10 @@ export  @model,
...
         setadbackend,
         setadsafe,
-        ForwardDiffAD,
-        TrackerAD,
-        ZygoteAD,
-        ReverseDiffAD,
+        AutoForwardDiff,
+        AutoTracker,
+        AutoZygote,
+        AutoReverseDiff,
```
devmotion, line 52: Do we actually want to export these types? Or should we tell users to use `ADTypes.AutoForwardDiff` etc. (in particular when these types would be used in other packages such as e.g. AdvancedVI as well).

src/essential/ad.jl
```diff
@@ -36,21 +36,10 @@ function setchunksize(chunk_size::Int)
...
-getchunksize(::ForwardDiffAD{chunk}) where chunk = chunk
-
-standardtag(::ForwardDiffAD{<:Any,true}) = true
-standardtag(::ForwardDiffAD) = false
-
-struct TrackerAD <: ADBackend end
-struct ZygoteAD <: ADBackend end
-
-struct ReverseDiffAD{cache} <: ADBackend end
+standardtag(::AutoForwardDiff{<:Any,nothing}) = true
```
devmotion, line 41:
```suggestion
standardtag(::AutoForwardDiff{<:Any,Nothing}) = true
```

test/essential/ad.jl
```diff
@@ -204,8 +204,8 @@
         f = DynamicPPL.LogDensityFunction(gdemo_default)
         θ = DynamicPPL.getparams(f)
 
-        f_rd = LogDensityProblemsAD.ADgradient(Turing.Essential.ReverseDiffAD{false}(), f)
-        f_rd_compiled = LogDensityProblemsAD.ADgradient(Turing.Essential.ReverseDiffAD{true}(), f)
+        f_rd = LogDensityProblemsAD.ADgradient(Turing.Essential.ReverseDiffAD(false), f)
+        f_rd_compiled = LogDensityProblemsAD.ADgradient(Val(:ReverseDiff), f; compile=Val(true), x=θ) # need to compile with non-zero inputs
```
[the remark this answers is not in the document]
devmotion: I think we should not add a keyword argument but just continue overloading `ADgradient` for `TuringLogDensityFunction`. I wouldn't want users to have to deal with such a keyword argument.

Project.toml
```diff
@@ -61,7 +63,7 @@ EllipticalSliceSampling = "0.5, 1, 2"
 ForwardDiff = "0.10.3"
 Libtask = "0.7, 0.8"
 LogDensityProblems = "2"
-LogDensityProblemsAD = "1.4"
+LogDensityProblemsAD = "1.6.1"
```
devmotion, line 66:
```suggestion
LogDensityProblemsAD = "1.7.0"
```

src/mcmc/hmc.jl
```diff
@@ -59,30 +61,23 @@ sample(gdemo([1.5, 2]), HMC(0.1, 10), 1000)
...
+function HMC(
     ϵ::Float64,
     n_leapfrog::Int,
     space::Symbol...;
-    metricT=AHMC.UnitEuclideanMetric
+    metricT=AHMC.UnitEuclideanMetric,
+    adtype::ADTypes.AbstractADType = ADBackend(),
 ) where AD
-    return HMC{AD}(ϵ, n_leapfrog, metricT, space)
+    return HMC(ϵ, n_leapfrog, metricT, space; adtype = adtype)
```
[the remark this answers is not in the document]
devmotion: And set the default AD type for each algorithm/sampler to a specific default such as AutoForwardDiff, you mean? Or would you like users to always specify the AD type explicitly?

src/mcmc/hmc.jl
```diff
@@ -59,30 +61,23 @@ sample(gdemo([1.5, 2]), HMC(0.1, 10), 1000)
...
-    return HMC{AD, space, metricT}(ϵ, n_leapfrog)
-end
-function HMC{AD}(
-    ϵ::Float64,
-    n_leapfrog::Int,
-    ::Tuple{};
-    kwargs...
-) where AD
-    return HMC{AD}(ϵ, n_leapfrog; kwargs...)
+function HMC(ϵ::Float64, n_leapfrog::Int, ::Type{metricT}, space::Tuple; adtype::ADTypes.AbstractADType = ADBackend()) where {metricT <: AHMC.AbstractMetric}
```
devmotion, line 70: Should we remove this constructor and only support `metricT` as keyword argument? Or make all arguments keyword arguments?

Generally, these HMC constructors are quite messy...

src/mcmc/hmc.jl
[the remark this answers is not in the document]
devmotion: I would like to get rid of the global flag 👍 If we make it in a separate PR we should maybe wait with tagging a breaking release until this follow-up PR is merged as well, to avoid two breaking releases in a row.

storopoli:
So the tape gets compiled by default now when using `ReverseDiff`?
No need more for `Turing.setrdcache(true)`?

storopoli:
Yes, but still is not clear. The example has only `adtype=AutoForwardDiff(; chunksize))` in the sampler constructor.
I had no idea that `AutoReverseDiff` took positional arguments, and why `AutoForwardDiff` doesn't take any positional arguments?

Also, https://github.com/SciML/ADTypes.jl doesn't have docs, so it is even harder to figure it out.

storopoli:
I can do that if you want to. I am hours away from my Holiday break.
At least the docs part.

src/mcmc/hmc.jl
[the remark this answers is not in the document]
ya0: This is implemented in 0.30 ?? I updated turing and my get and error now. could you update the docs? in particular [autodiff](https://turinglang.org/v0.30/docs/using-turing/autodiff)

PR #2050, opened 2023-07-24
bump compat of AdvancedHMC

JaimeRZP:
Bump compatibility of AdvancedHMC to include the latest version that includes the interface with external samplers.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -38,7 +38,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 
 [compat]
 AbstractMCMC = "4"
-AdvancedHMC = "0.3.0, 0.4"
+AdvancedHMC = "0.3.0, 0.4, 5"
 AdvancedMH = "0.6.8, 0.7"
 AdvancedPS = "0.4"
 AdvancedVI = "0.2"


Project.toml
```diff
@@ -38,7 +38,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 
 [compat]
 AbstractMCMC = "4"
-AdvancedHMC = "0.3.0, 0.4"
+AdvancedHMC = "0.3.0, 0.4, 5"
```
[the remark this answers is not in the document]
JaimeRZP: Yes, we want to merge https://github.com/TuringLang/AdvancedHMC.jl/pull/329, make a release and then bump the version here.

devmotion:
Tests use AdvancedHMC 0.4 it seems.

JaimeRZP:
> Tests use AdvancedHMC 0.4 it seems.

image
 
 I think it is fixed!

Project.toml
```diff
@@ -38,7 +38,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 
 [compat]
 AbstractMCMC = "4"
-AdvancedHMC = "0.3.0, 0.4"
+AdvancedHMC = "0.3.0, 0.4, 0.5"
```
devmotion, line 41: Do to the changes below and since the PR requires https://github.com/TuringLang/AdvancedHMC.jl/pull/342, this should be changed to
```suggestion
AdvancedHMC = "0.5.2"
```

src/inference/Inference.jl
```diff
@@ -189,10 +189,10 @@ function AbstractMCMC.sample(
     kwargs...
 )
     if resume_from === nothing
-        return AbstractMCMC.mcmcsample(rng, model, sampler, N;
-                                       chain_type=chain_type, progress=progress, kwargs...)
+        return AbstractMCMC.mcmcsample(rng, model, sampler, N + min(div(N, 10), 1_000);
```
devmotion, line 192: This means users will get more than the requested `N` samples - I think this should be reverted?

Project.toml
devmotion: Doesn't the change in inference/hmc.jl add a function call that does not work on 0.3 and 0.4? Did HMCKernel exist in these old versions?

src/inference/Inference.jl
devmotion: Sure, we have to take into account the adaption steps - but changing this line here means that a call such as `sample(model, NUTS(), 10)` will return not 10 but 1010 samples? These discarded steps only have to be specified as keyword argument but not added to the positional argument (number of samples). AbstractMCMC will added `discard_initial` to these internally to the requested number of samples automatically.

PR #2051, opened 2023-07-25
Move Optim support to extension

devmotion:


src/Turing.jl
```diff
@@ -11,6 +11,12 @@ import AdvancedVI
 using DynamicPPL: DynamicPPL, LogDensityFunction
 import DynamicPPL: getspace, NoDist, NamedDist
 import LogDensityProblems
+import NamedArrays
```
JaimeRZP, line 14: I understand that we are importing these here so we can then grab them in the extension. 
Are they used somewhere else? If so that's fine. 
Otherwise, does the extension really need them?

JaimeRZP:
Just one question but looks very good and straight forward otherwise!!

JaimeRZP (approved):
Just one question but looks very good and straight forward otherwise!!

PR #2058, opened 2023-07-31
Re-organise tests.

yebai:
This PR makes it easy to run each test file individually locally:

```julia
shell> pwd
/Users/hg344/projects/Turing.jl

(test) pkg> activate test
  Activating project at `~/projects/Turing.jl/test`

julia> using Turing; include(pkgdir(Turing)*"/test/test_utils/test_utils.jl")

julia> include("test/modes/OptimInterface.jl")
```

This PR doesn't contain any functionality changes.


test/test_utils/test_utils.jl
```diff
@@ -0,0 +1,311 @@
...
+using StatsFuns: binomlogpdf, logistic, logsumexp
+using TimerOutputs: TimerOutputs, @timeit
+using Turing: BinomialLogit, ForwardDiffAD, Sampler, SampleFromPrior, NUTS, TrackerAD,
+                Variational, ZygoteAD, getspace
+using Turing.Essential: TuringDenseMvNormal, TuringDiagMvNormal
+import LogDensityProblems
+import LogDensityProblemsAD
+
+
+#
```
torfjelde, line 50: Do we feel this is more readable than what we did before?

As in, why don't we just keep the separate files but include all of them in a single file?

torfjelde (commented):
I'm uncertain if this is an issue of Turing.jl or rather an issue of "standard Julia workflow".

What I usually do is that I take this "header" for the `runtests.jl`, paste it into a `setup.jl` file (similar to what you've done, but using `include` instead), and then use this is a setup-file for my testruns. I then use a custom script which re-runs the files I've specified whenever changes are made to either the package or the tests themselves (script: https://gist.github.com/torfjelde/62c1281d5fc486d3a404e5de6cf285d4). You might find this useful:)

torfjelde:
> However, it is helpful to introduce a simple way to achieve a similar objective. Hence this PR.

Fair! But do we need to also remove all the files in the process? As in, why not just use `include` but add a single file that contains all these?

PR #2072, opened 2023-09-01
`FlattenIterator` is replaced by `varname_and_value_leaves`

yebai:


src/utilities/Utilities.jl
```diff
@@ -6,15 +6,4 @@ using Distributions, Bijectors
 using StatsFuns, SpecialFunctions
 import Distributions: sample
 
-export  vectorize,
```
torfjelde, line 9: Shouldn't this entire module be removed then?

PR #2078, opened 2023-09-06
Attach `varname_to_symbol` mapping to `Chains`

torfjelde:
After https://github.com/TuringLang/DynamicPPL.jl/pull/534 in DynamicPPL.jl, attaching a dict-like mapping from `VarName` to `Symbol` (as represented in the `Chains`) allows a much improved experience when working with `Model` and `Chains` together, e.g. `generated_quantities` that can handle things like `Cholesky` variates.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.29"
+version = "0.29.1"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"


torfjelde:
So I was just looking at this one myself, and I don't think upping the DPPL version should fix this. This PR should have been functional on it's own.

src/mcmc/emcee.jl
```diff
@@ -133,10 +134,9 @@ function AbstractMCMC.bundle_samples(
     le = getlogevidence(samples, state, spl)
 
     # Set up the info tuple.
+    info = (varname_to_symbol = OrderedDict(zip(varnames, varnames_symbol)),)
```
torfjelde, line 137: Should we make this inclusion optional? Forcing it could be annoying for serialization of the chains, etc. @yebai

torfjelde: If we skip it in serialization, then those methods won't work after deserialization anyways :confused: 

And yes I agree it should be enabled by default, but I was also thinking it would be useful to provide a way to disable it.

torfjelde: You can, but it means that you no longer can do `using MCMCChains; deserialize(...)` but you need to have done `using Turing` first. IMO this is quite annoying :confused:

torfjelde: Potentially in the longer run, but I don't think we should do that when MCMCChains doesn't use any functionality from AbstractPPL.

torfjelde:
Okay, so I can't actually reproduce the error from the tests :confused: 

But I've now made it optional to include the `varname_to_symbol` mapping, so it will be possible for users to opt-out of this inclusion if they prefer. I think that should be good enough :+1:

PR #2086, opened 2023-09-22
Bringback old getparams

JaimeRZP:
`getparams(t) = t.θ` is extremely useful to connect external samplers to Turing while 
```
# NOTE: Only thing that depends on the underlying sampler.
# Something similar should be part of AbstractMCMC at some point:
# https://github.com/TuringLang/AbstractMCMC.jl/pull/86
getparams(transition::AdvancedHMC.Transition) = transition.z.θ
getstats(transition::AdvancedHMC.Transition) = transition.stat

getparams(transition::AdvancedMH.Transition) = transition.params

getvarinfo(f::DynamicPPL.LogDensityFunction) = f.varinfo
getvarinfo(f::LogDensityProblemsAD.ADGradientWrapper) = getvarinfo(parent(f))

setvarinfo(f::DynamicPPL.LogDensityFunction, varinfo) = Setfield.@set f.varinfo = varinfo
setvarinfo(f::LogDensityProblemsAD.ADGradientWrapper, varinfo) = setvarinfo(parent(f), varinfo)
```
get properly pushed into abstractmcmc.

### DIFF
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -293,6 +293,7 @@ end
 
 Return a named tuple of parameters.
 """
+getparams(t) = t.θ
 getparams(model, t) = t.θ
 function getparams(model::DynamicPPL.Model, vi::DynamicPPL.VarInfo)
     # Want the end-user to receive parameters in constrained space, so we `link`.


JaimeRZP:
@torfjelde Could you quickly have a look? It is just one line of code that I need to connect some external samplers to Turing.

devmotion:
Maybe it would be better to finish https://github.com/TuringLang/AbstractMCMC.jl/pull/86?

JaimeRZP:
> Maybe it would be better to finish [TuringLang/AbstractMCMC.jl#86](https://github.com/TuringLang/AbstractMCMC.jl/pull/86)?

I completely agree that this is ultimately the correct solution. 
This is just an extremely quick patch to allow external libraries to connect to Turing. 
The latest change broke a lot of connections, this brings back that functionality while we work on the proper solution.

torfjelde:
Am away from computer atm; will have a look when I get back.

torfjelde:
Sorry, why can't we just provide the model and use the existing `getparams`?

torfjelde:
And I'm also very much in favor of what @devmotion is suggesting

JaimeRZP:
> Sorry, why can't we just provide the model and use the existing `getparams`?

Because that's not the signature of getparams used in TuringTransition. See code [here](https://github.com/TuringLang/Turing.jl/blob/ed410b14f0f91698d90cc804e37fddec5205194c/src/mcmc/abstractmcmc.jl#L7-L18)

torfjelde:
Sure, but there's nothing stopping us from just changing `TuringTransition` to use this signature, right? It would just be mapping `getparams(transition)` to `getparams(f.model, transition)` (and this should of course have been done in the PR that original introduced this new `getparams`)

JaimeRZP:
Sure.
 We can now close this PR in favour yours!

PR #2089, opened 2023-10-01
Fixed `externalsampler`

torfjelde:
We recently made some improvements to the construction of the chains (#2071), and in the process seem to have broken the `externalsampler` (though I'm somewhat confused as to how tests were passing).

### DIFF
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -104,7 +104,7 @@ externalsampler(sampler::AbstractSampler) = ExternalSampler(sampler)
 A log density function for the External sampler.
 
 """
-const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.DefaultContext}
+const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.SamplingContext{<:S}}
 function LogDensityProblems.logdensity(f::ESLogDensityFunction, x::NamedTuple)
     return DynamicPPL.logjoint(f.model, DynamicPPL.unflatten(f.varinfo, x))
 end
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -104,7 +104,7 @@ externalsampler(sampler::AbstractSampler) = ExternalSampler(sampler)
 A log density function for the External sampler.
 
 """
-const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.SamplingContext{<:S}}
+const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.DefaultContext}
 function LogDensityProblems.logdensity(f::ESLogDensityFunction, x::NamedTuple)
     return DynamicPPL.logjoint(f.model, DynamicPPL.unflatten(f.varinfo, x))
 end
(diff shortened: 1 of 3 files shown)


devmotion:
According to codecov, both the old and new implementation is not covered.

Related to external samplers, https://github.com/TuringLang/Turing.jl/blob/ed410b14f0f91698d90cc804e37fddec5205194c/src/mcmc/Inference.jl#L107 seems wrong - the sampler should be part of the right-hand side, shouldn't it? I think it should be changed to
```julia
const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.SamplingContext{<:S}}
```

torfjelde:
> According to codecov, both the old and new implementation is not covered.

This is quite strange though, as `test/mcmc/abstractmcmc.jl` is clearly included in `test/runtests.jl`: https://github.com/TuringLang/Turing.jl/blob/3780ca830fc3b49d68c60dcf3a5c711249e4a317/test/runtests.jl#L79C23-L79C24

torfjelde:
> seems wrong - the sampler should be part of the right-hand side, shouldn't it? I think it should be changed to

I applied your suggestion a bit too quickly here:) No, the original impl was correct.

`externalsampler` only works for samplers which requires nothing but the `LogDensityProblems` interface. The `SamplingContext` is not necessary here because these samplers will not be overloading any of the tilde statements; they all just need `logdensity` and/or `logdensity_and_gradient`, hence why we just use `DefaultContext`.

devmotion:
I see. The line still seems wrong to me - since the (external) sampler is not used on the RHS, it means that anything defined for `ESLogDensityFunction` will in fact be defined for any `TuringLogDensityFunction` with a `DefaultContext`, everything defined for `ESLogDensityFunction{M<:Model}` will be defined for any `TuringLogDensityFunction{M}` with a `DefaultContext`, and even `ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler}}` will be defined for any `TuringLogDensityFunction{M}` with a `DefaultContext`.

torfjelde:
Ah great point and 100% agree :+1: 
It seems the only point where it is used is here: 

https://github.com/TuringLang/Turing.jl/blob/f8876d6d65c4cc4266eeeb87dbd87b5e31d398ac/src/mcmc/Inference.jl#L108-L116

which is an unfortunate hack we need to be compatible with AdvancedMH.jl IIRC

src/mcmc/Inference.jl
```diff
@@ -98,19 +98,15 @@ Wrap a sampler so it can be used as an inference algorithm.
...
-
-A log density function for the External sampler.
-
-"""
-const ESLogDensityFunction{M<:Model,S<:Sampler{<:ExternalSampler},V<:AbstractVarInfo} = Turing.LogDensityFunction{V,M,<:DynamicPPL.DefaultContext}
-function LogDensityProblems.logdensity(f::ESLogDensityFunction, x::NamedTuple)
+function LogDensityProblems.logdensity(
+    f::Turing.LogDensityFunction{<:AbstractVarInfo,<:Model,<:DynamicPPL.DefaultContext},
+    x::NamedTuple
+)
```
devmotion, line 104: This is type piracy (I know it was already there 😉 - and BTW `DynamicPPL.unflatten` below is another instance of type piracy). In DynamicPPL, we explicitly decided to not add this overload: https://github.com/TuringLang/DynamicPPL.jl/pull/501 Can we just remove it completely?

torfjelde: Yeaaah I'm aware of both and I don't like them :confused: 

It's done to be compatible with AdvancedMH.jl which allows you to specify the proposal distribution using `NamedTuple` :confused: So we can't just remove it without messing with this; we'd have to do some more work for that, unfortunately.

devmotion: Ohh haha I missed that, I just remembered that the issue had been closed 😅

JaimeRZP:
How can I help to get this merged?
We need this to be able to finally merge https://github.com/TuringLang/TuringTutorials/pull/403

JaimeRZP (approved):
LGTM!

src/mcmc/Inference.jl
devmotion: Yes, let's address this separately.

Issue #2095, opened 2023-10-03
Asssigning to a Vector of Matrices produces BoundsError


tiemvanderdeure:
Sampling from `LKJ ` or `LKJCholesky` and assigning to a vector of matrices errors, regardless of AD backend.

The code compiles but errors with a `BoundsError`. Sampling from the Prior works does not produce an error.

MWE
```
@model function vector_of_cor_matrices(n, Type::TV=Matrix{Float64}) where {TV}
    mat = Vector{TV}(undef, n)
    for i = 1:n      
        mat[i] ~ LKJ(3, 3.0)
    end
end

sample(vector_of_cor_matrices(2), NUTS(), 200)
```

Full stacktrace

```
ERROR: BoundsError: attempt to access 6-element Vector{ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 6}} at index [10:12]
Stacktrace:
  [1] throw_boundserror(A::Vector{ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 6}}, I::Tuple{UnitRange{Int64}})
    @ Base .\abstractarray.jl:744
  [2] checkbounds
    @ .\abstractarray.jl:709 [inlined]
  [3] view
    @ .\subarray.jl:177 [inlined]
  [4] getval
    @ C:\Users\tsh371\.julia\packages\DynamicPPL\m0PXI\src\varinfo.jl:318 [inlined]
  [5] getval
    @ C:\Users\tsh371\.julia\packages\DynamicPPL\m0PXI\src\varinfo.jl:317 [inlined]
  [6] getval
    @ C:\Users\tsh371\.julia\packages\DynamicPPL\m0PXI\src\threadsafe.jl:194 [inlined]
  [7] invlink_with_logpdf
    @ C:\Users\tsh371\.julia\packages\DynamicPPL\m0PXI\src\abstract_varinfo.jl:679 [inlined]
  [8] assume
    @ C:\Users\tsh371\.julia\packages\DynamicPPL\m0PXI\src\context_implementations.jl:197 [inlined]
[...]

PR #2096, opened 2023-10-03
Partial fix for #2095

torfjelde:
This is a partial fix for #2095. Currently, HMC won't work with vectors, etc. of variables which have differently sized support in the transformed space, e.g.

``` julia
@model function vector_of_dirichlet(::Type{TV}=Vector{Float64}) where {TV}
    xs = Vector{TV}(undef, 3)
    for i = 1:3
        xs[i] ~ Dirichlet(ones(5))
    end
end

model = vector_of_dirichlet()
sample(model, NUTS(), 1000)
```

breaks with a similar error to that mentioned in #2095.

This occurs because `unflatten`, which is used in `DynamicPPL.LogDensityFunction`, attempts to reconstruct using a `VarInfo` which has a much larger vector for underlying storage than the one provided as input, e.g. `varinfo` is working with 3 x 5 dimensional `xs` but is only using a subset of the indexes, while the provided vector will be (3 - 1) x 5 dimensional since it's in linked space.

We can fix this by simply using the immutable `DynamicPPL.link`, which results in the "underlying" `varinfo` also using (3 - 1) x 5 dimensional storage for `xs`.

### DIFF
--- a/src/mcmc/hmc.jl
+++ b/src/mcmc/hmc.jl
@@ -132,13 +132,13 @@ function DynamicPPL.initialstep(
     rng::AbstractRNG,
     model::AbstractModel,
     spl::Sampler{<:Hamiltonian},
-    vi::AbstractVarInfo;
+    vi_original::AbstractVarInfo;
     init_params=nothing,
     nadapts=0,
     kwargs...
 )
     # Transform the samples to unconstrained space and compute the joint log probability.
-    vi = link!!(vi, spl, model)
+    vi = DynamicPPL.link(vi_original, spl, model)
 
     # Extract parameters.
     theta = vi[spl]


torfjelde:
Yeah I'll have a look at this later :+1:

torfjelde:
Okay, so I figured out what the issue is, and the fix is somewhat annoying.

`DynamicPPL.(inv)link` does not currently support specification using the `Sampler`, and so it links the entire `VarInfo`, even though only a sub-set of  the `VarInfo` needs linking.

We can add this quite "easily" for `TypedVarInfo`, but for `UntypedVarInfo` things become much more annoying :confused:

torfjelde:
Okay, so this should now be fixed by https://github.com/TuringLang/DynamicPPL.jl/pull/542

torfjelde:
Is there a reason why we're doing a lot of progress reporting during tests @yebai @devmotion ? It makes it quite annoying to determine which tests are actually failing by inspecting the logs :confused:

devmotion:
> Is there a reason why we're doing a lot of progress reporting during tests @yebai @devmotion ? It makes it quite annoying to determine which tests are actually failing by inspecting the logs 😕

I don"t recall any previous discussions. I think it would be good to check that progress logging "works" by sampling with progress logging in the tests - but I think this could be limited to a few selected examples and in the remaining `sample` calls we could disable it. Since generally progress logging introduces overhead and slows down sampling, this could even improve CI times.

PR #2098, opened 2023-10-06
Fix for CI

torfjelde:
We've added some new models to `DynamicPPL.TestUtils.DEMO_MODELS` at some point, and these are also examples where optimization might terminate early because of insane values (as noted before, this is because the parameters include variances for each data point, hence, when performing MLE, these should be 0).


torfjelde:
@yebai @sunxd3 would you mind hitting approve on this one?

CI is also breaking on v1.7 when using DynamicHMC.jl. This doesn't seem to be a bug introduced on our end, but haven't looked into it too much yet.

torfjelde:
The issue with DynamicHMC.jl might have gotten fixed by https://github.com/mcabbott/LazyStack.jl/pull/18 (my love for @devmotion is not describable in words)

PR #2099, opened 2023-10-06
New Gibbs sampler using `condition`

torfjelde:
This is an attempt at a new Gibbs sampler which makes use of the `condition` functionality from DynamicPPL.jl

This should, for the models which are compatible with `condition`, make for a much more flexible approach to Gibbs sampling.

The idea is basically as follows: instead of having a single `varinfo` which is shared between all the samplers, we instead have a different `varinfo` for every sampler involved in the composition. We then subsequently `condition` the `model` on the other `varinfo` for each "inner" `step`.

The current implementation is somewhat rough, but it does seem to work!

# Open issues
## TODO Issue with MH

```julia
using Turing
@model demo() = x ~ Normal()
sampler = Turing.Experimental.Gibbs(
  Turing.OrderedDict(
    @varname(x) => MH(AdvancedMH.RandomWalkProposal(Normal(0,0.003)))
  )
)
```

breaks, but if we make it a filldist or something then it works

```julia
sampler = Turing.Experimental.Gibbs(
  Turing.OrderedDict(
    @varname(x) => MH(AdvancedMH.RandomWalkProposal(filldist(Normal(0,0.003),1)))
  )
)
```

## TODO Linking isn't quite there (see last comment below)

TODO


torfjelde:
@yebai

torfjelde:
This PR is not ready though (hence why I put it in draft-mode). Some cases are functional, but it needs more work (which I'm currently doing:))

test/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,176 @@
...
+    end
+
+    @testset "gdemo with CSMC & ESS" begin
+        # `GibbsV2` does not work with SMC samplers, e.g. `CSMC`.
+        # FIXME: Oooor it is (see tests below). Uncertain.
+        Random.seed!(100)
+        alg = GibbsV2(@varname(s) => CSMC(15), @varname(m) => ESS(:m))
+        chain = sample(gdemo(1.5, 2.0), alg, 10_000)
+        @test_broken mean(chain[:s]) ≈ 49 / 24
+        @test_broken mean(chain[:m]) ≈ 7 / 6
```
torfjelde, line 118: @yebai Any idea why this fails? It does work with `CSMC` in the example below..

torfjelde: Main difference I see here is that this example uses CSMC on continuous rather than discrete.

torfjelde: (I've also tried without the `ESS(:m)` locally, and it doesn't make a difference)

torfjelde: As me and @yebai discussed, this issues is likely caused by the fact that SMC samplers will now also treat the conditioned variables as observations, which is not quite what we want.

But atm I don't quite see how we can work around this with the existing functionality we have in DynamicPPL :confused: 

Basically, we have `fix` and `condition`:
- `condition` makes the variables be treated as observations, which, as seen above, is bad for likelihood-tempered samplers.
- `fix` avoids variables being treated as observations _but_ it also doesn't include the log-prob of that variable. This then causes issues in models with hierarchical dependencies, e.g. `gdemo` where the prior on `m` depends on the value of `s`, and so changing `s` changes the joint also through the prior on `m`.

Effectively, we'd need something similar to `fix` which _also_ computes the log-prob.

torfjelde: One thing that confuses me slightly is that, _technically_, even if the SMC sampler sees one of the parameters as an observation, AFAIK, it shouldn't _break_ things, no? It might increase the variance, etc. but it should still converge.

torfjelde: And I just tried altering the implementation of `FixedContext` so it also computes the prior and adds it to the logjoint, and then swapped out usage of `condition` in the new Gibbs with `fix`.

Still giving me similarly incorrect answers :confused:

torfjelde: I genuinely don't understand what's going on here :confused: 

So, I've addressed the issue of `m` being treated as a observation when running `PG`, but this does not resolve the issue.

I've also tried to inspect the conditional distribution of the two different samplers to see if there's something obvious that pops out ; see below (`k` refers to the number of steps from initial state, and `k = 0` is only the initial state).

There are clearly differences, but I have not been able to identify the cause. Debugging the PG sampler is quite the pain :confused: 

Honestly, it might be worth it to just disallow PG samplers for now and open a corresponding issue which we link in the error message. 

Thoughts @yebai @devmotion ?

gdemo_m_k=0
gdemo_s_k=0
gdemo_m_k=1
gdemo_s_k=1
gdemo_m_k=5
gdemo_s_k=5

torfjelde: A few things:
1. As mentioned in one of the previous comments, the issue still persists even if I address this issue by using `fix` instead of `condition`, and then altering `fix` to also compute the log-prob (the plots above are generated using this fix). 
2. But AFAIK this should still work, _even without my fix_. It might have higher variance, but it should still converge to the correct thing. " we cannot treat m as data, since that would double-count information in D" I don't understand this. How are we double-counting the data? CSMC should work just fine here; it's just that now we've tacked on one additional distribution in our importance weight, i.e. $p(m \mid s)$, no? The fact that we have one additional resampling step in there shouldn't matter much.

torfjelde: Okay, so I think I might know what's up here.

1. SMC samplers uses a separate `VarInfo` for every particle.
2. SMC samplers relies on accumulating log-probs of _everything_ in the tilde-statements.
3. Because of (2), any accumulation of log-probs in anything that is _not_ in the tilde-statements, means that the task-local `varinfo` will _not_ contain this log-prob.

But then this also means that `PG` won't work with models using existing functionality too, e.g. `@addlogprob!` :confused: 

But this does not seem like something we should try to fix in this PR, but requires a more substantial change in DynamicPPL or something.

One potential avenue to address these issues might be to introduce a `addlogprob!!(context, varinfo, logp)`, which can then be overloaded for the scenario of particle samplers, which can then accumulate this in the trask-local `varinfo` instead of the "global" one.

torfjelde: Okay, okay, so I've fixed the above issue locally (will make a PR to DPPL), but _still_ this example fails.

_Now_ the issue is with ESS. ESS assumes the provided state (which is just a `VarInfo`) comes from one where only the log-likelihood has been accumulated, while the current impl of `GibbsV2` recomputes the _log-joint_. Hence we the rejection sampling in ESS ends up comparing the likelihood of the proposal to the _joint_ of the previous state.

Will have a look at this later.

torfjelde: Aight, so I think we'll need to adopt the same `rerun` functionality as for the current impl of `Gibbs`, since we need to re-evaluate the model in the same fashion as before but without sampling.

devmotion: I haven't followed the discussion in this thread but regarding the last point (ESS): This difference (log-likelihood vs log-joint) was one (the?) main reason for the current design of the re-evaluation in `Gibbs`, i.e., for `gibbs_rerun`: https://github.com/TuringLang/Turing.jl/blob/6649f10c48917a27531214f02777408d2ab82928/src/mcmc/gibbs.jl#L134-L142 We try to avoid re-evaluating `varinfo.logp` when we know that it is not needed but to be safe (and e.g. for `ESS`) the default is to re-evaluate.

torfjelde: Aaalrighty! Adopting the rerun mechanisms from the current `Gibbs` seems to have done the trick!

I've also tried a `acclogp!!(context, varinfo, logp)` implementation, and I think this is something we should adopt, as it will immediately address the issue of SMC samplers not currently working with `@acclogprob!!` + there are other scenarios where we might want to hook into this, e.g. `DebugContext` where we can to "record" these hard-coded acclogprobs for debugging and model checking purposes.

torfjelde:
Aaalrighty friends! I think we're good to go, as in tests should now be passing and so it might be worth giving this a review:)

One thing we should definitively discuss is the name: currently I've gone with `GibbsV2` but we shouldn't do that :upside_down_face:

test/mcmc/gibbs_new.jl
torfjelde: Btw, @devmotion didn't see your comment because was working on a train without internet :upside_down_face:

torfjelde: > One thing that confuses me slightly is that, technically, even if the SMC sampler sees one of the parameters as an observation, AFAIK, it shouldn't break things, no? It might increase the variance, etc. but it should still converge.

To make this absolutely clear, let's look at PG from Andrieu (2010):

Screenshot_20231121_111042

Looking at this, it might seem as if we shouldn't include the "prior" probability of $m$ in `gdemo`, since $p_{\theta}(x_{1:T} \mid y_{1:T})$ does _not_ include $p(\theta)$.

_But_ this is all done under the assumption that the joint nicely factorizes as
$$p(x_{1:T}, \theta \mid y_{1:T}) \propto p_{\theta}(x_{1:T}, y_{1:T}) p(\theta)$$
In this scenario, $p\big(\theta(i)\big)$ will be the same value for all particles in step (b), and so there's no need to include this in the weights.

_But_ in our scenario, the joint does _not_ nicely factorize as above. We have $\theta = m$ and $x_{1:T} = s$, so
$$p(s, m \mid y_{1:T}) = p(s) p(m \mid s) p(y_{1:T} \mid m, s)$$
To get a similar factorization as above, we can of course just define
$$p_m(s, y_{1:T}) = \frac{p(s, m, y_{1:T})}{p(m)} \propto p(s, m, y_{1:T})$$
where we can just drop the contribution of $p(m)$ since it's constant wrt. $s$ that we're sampling.

With this notation, mapping this to the algorithm above is immediate, in which case it's clear that we're targeting the full joint, and so, in the general scenario, we need to include the log-prob contribution of latent variables that are not being sampled by conditional SMC.

And more generally speaking, PG is just a Gibbs sampler with:
1. Using some method (in  Andrieu (2010) they assume it's available in closed-form)
   $$\theta(i) \sim p \big( \theta \mid y_{1:T}, X_{1:T} \propto p(\theta, x_{1:T}, y_{1:T})$$
2. Using CSMC
   $$X_{1:T}(i) \sim p \big( X_{1:T} \mid y_{1:T}, \theta \big) \propto p(\theta, x_{1:T}, y_{1:T})$$

And the reason why we're _not_ including the contribution from the variable we're targeting with CSMC is because we would then have to adjust the weights later by removing the prior contribution since this is our importance weight; that is, our proposal for $s$ is $p(s)$ and we're targeting $p(s, m, y_{1:T})$ so our IS weight is
$$w = \frac{p(s, m, y_{1:T})}{p(s)} = \frac{p(s) p(m \mid s) p(y_{1:T} \mid s, m)}{p(s)} = p(m \mid s) p(y_{1:T} \mid s, m)$$

Does that make sense @yebai ?

And finally, whether we're resampling when hitting $\theta$ or not, is not a correctness issue since we're still targeting the same distribution (though it probably does something to the variance of the estimator; not sure if it's good or bad :shrug: )

torfjelde: This also explains why we need to make `acclogp!!` accumulate log-probs in the task-local `varinfo` and not the "global" one.

src/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,396 @@
...
+    return condition_gibbs(
+        model,
+        filter(Base.Fix1(!==, target_varinfo), varinfos)...
+    )
+end
+
+wrap_algorithm_maybe(x) = x
+wrap_algorithm_maybe(x::InferenceAlgorithm) = Sampler(x)
+
+struct GibbsV2{V,A} <: InferenceAlgorithm
```
torfjelde, line 180: We need to rename this

src/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,396 @@
...
+    varnames::V
+    samplers::A
```
torfjelde, line 182: Make `NamedTuple` or `OrderedDict`

src/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,396 @@
...
+    model::Model,
+    spl::Sampler{<:GibbsV2},
+    state::GibbsV2State;
+    kwargs...,
+)
+    alg = spl.alg
+    samplers = alg.samplers
+    states = state.states
+    varinfos = map(varinfo, state.states)
+    @assert length(samplers) == length(state.states)
```
torfjelde, line 269:
```suggestion
```

src/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,396 @@
...
+            model_local,
+            sampler_local,
+            current_state_local;
+            kwargs...,
+        ),
+    )
+
+    # 3. Extract the new varinfo.
+    # Return the resulting state and invlinked `varinfo`.
+    varinfo_local_state = varinfo(new_state_local)
```
torfjelde, line 387: Open issue regarding checking whether next sampler needs linking or not

src/mcmc/gibbs_new.jl
```diff
@@ -0,0 +1,396 @@
...
+    # TODO: Check if this is known at compile-time if `varinfos isa Tuple`.
+    return condition_gibbs(
+        model,
+        filter(Base.Fix1(!==, target_varinfo), varinfos)...
+    )
+end
+
+wrap_algorithm_maybe(x) = x
+wrap_algorithm_maybe(x::InferenceAlgorithm) = Sampler(x)
+
```
torfjelde, line 179: Benchmarks

torfjelde, line 179: Experimental module

torfjelde:
So I've been trying to make `externalsampler` work with the new gibbs sampler, but I've realized that this won't be possible without some more extensive additions.

Effectively, we need the same as I've been playing around with in MCMCTempering.jl and as discussed in https://github.com/TuringLang/AbstractMCMC.jl/pull/86, i.e. we need something like
- `setparams!!(sampler_dst, sampler_src, state_dst, state_src)`: update `state_dst` with parameters from `state_src`
- `getparams(sampler, state)`: return the underlying realization as a `Vector` or something useable by other samplers.


 Me and @devmotion recently discussed this and arrived at the decision of first showcasing the utility of such methods before adopting them upstream, i.e. in AbstractMCMC.jl.

Originally I was planning on having MCMCTempering.jl be this demo, but maybe it's worth pursuing it here for the new Gibbs sampler instead.

_But_ this is just about adding compatibility with a broader class of samplers; the current implementation works nicely with the existing Turing.jl-specific samplers. Hence, I'm thinking we should maybe just get this merged and then I can open a separate PR where I pursue compat more broadly with `AbstractSampler`? @yebai @devmotion

torfjelde:
This should now be ready for review and merge.

There are some left-overs that we have discussed before, but which we decided to defer until we have better state-to-state interactions across samplers, etc.

Once we get to try this out in practice a bit, I'm positive we can fully replace the current `Gibbs` sampler with this approach, in which case we will then remove all the `selector` stuff from DynamicPPL, combined with a follow-up PR to Turing.jl where we can simplify the new `Gibbs` sampler (remove refs to `selector`, etc.) + remove the current `Gibbs` sampler.

But as I said, this is left for future work. I'll open issues accordingly.

test/experimental/gibbs.jl
```diff
@@ -0,0 +1,175 @@
...
+        @test any(Base.Fix2(DynamicPPL.subsumes, vn), parent_varnames)
+    end
+end
+
+# Okay, so what do we actually need to test here.
+# 1. (✓) Needs to be compatible with most models.
+# 2. (???) Restricted to usage of pairs for now to make things simple.
+
+# TODO: Don't require usage of tuples due to potential of blowing up compilation times.
+
```
torfjelde, line 21:
```suggestion
```

torfjelde:
It seems there are still some issues with linking, etc. Will have a look at this soon.

torfjelde:
Tests are at least passing now, but I have some local changes I'm currently working on, so no merge-y yet please.

src/Turing.jl
```diff
@@ -70,6 +71,7 @@ export  @model,                 # modelling
         ESS,
         Gibbs,
         GibbsConditional,
+        GibbsV2,
```
torfjelde, line 74:
```suggestion
```

src/experimental/gibbs.jl
```diff
@@ -0,0 +1,468 @@
...
+    # Needs to do a a few things.
+    sampler_local = samplers[index]
+    state_local = states[index]
+    varinfo_local = varinfos[index]
+
+    # Make sure that all `varinfos` are linked.
+    varinfos_invlinked = map(varinfos) do vi
+        # NOTE: This is immutable linking!
+        # TODO: Do we need the `istrans` check here or should we just always use `invlink`?
+        DynamicPPL.istrans(vi) ? DynamicPPL.invlink(vi, model) : vi
```
torfjelde, line 415:
```suggestion
        # TODO: Do we need the `istrans` check here or should we just always use `invlink`?
        # FIXME: Suffers from https://github.com/TuringLang/Turing.jl/issues/2195
        DynamicPPL.istrans(vi) ? DynamicPPL.invlink(vi, model) : vi
```

test/experimental/gibbs.jl
```diff
@@ -0,0 +1,195 @@
...
+        @testset "sample" begin
+            chain = sample(model, sampler, 1000; progress=false)
+            @test size(chain, 1) == 1000
+            display(mean(chain))
+        end
+    end
+
+    @testset "gdemo with CSMC & ESS" begin
+        # `Turing.Experimental.Gibbs` does not work with SMC samplers, e.g. `CSMC`.
+        # FIXME: Oooor it is (see tests below). Uncertain.
```
torfjelde, line 111:
```suggestion
```

torfjelde:
This is starting to take shape @devmotion @yebai @sunxd3 

After https://github.com/TuringLang/DynamicPPL.jl/pull/587 there's very little reason to use the "current" Gibbs sampler over this experimental one. It provides much more flexibility, in addition to being much easier to debug, etc. I've also encountered issues with linking, etc. in the current impl of Gibbs which then just become a pain to debug.

torfjelde:
Tests are passing here now :+1: 

Think it's worth merging and making a release with this before we make a breaking release after #2197

PR #2130, opened 2023-11-11
forward `progress`

FredericWantiez:


src/mcmc/Inference.jl
```diff
@@ -174,19 +174,21 @@ function AbstractMCMC.sample(
     model::AbstractModel,
     alg::InferenceAlgorithm,
     N::Integer;
+    progress=PROGRESS[],
```
devmotion, line 177: We should just remove Turing.PROGRESS and Turing.setprogress! and use the corresponding ones in AbstractMCMC

FredericWantiez: Looks like it's broken now, this would still trigger the logger:
```julia
using Turing
using Distributions
using Random

Turing.setprogress!(false)

@model function gdemo_d()
    s ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s))
    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))
    return s, m
end

model = gdemo_d()
alg = Gibbs(PG(3, :s), HMC( 0.4, 8, :m))
sampler = Turing.Sampler(alg, model)
sample(model, sampler, 500)
```

Issue #2132, opened 2023-11-16
Remove global AD flag `ADBACKEND` 


yebai:
See https://github.com/TuringLang/Turing.jl/pull/2047#discussion_r1394499393

PR #2134, opened 2023-11-24
Remove global AD flag "ADBACKEND" and function suite

sunxd3:
Addressing #2132, ref https://github.com/TuringLang/Turing.jl/pull/2047#discussion_r1394499393

This is likely to be quite hairy, as the ADBACKEND goes back.

Steps towards completion:
- [x] Replace `Turing.ADBackend()` with `AutoForwardDiff` as default for `adtype` argument in samplers
- [x] Remove `setadbackend` function, the main issue here is that tests use it quite extensively, so some refactoring is needed
- [x] Remove `ADSAFE`, `CHUNKSIZE`, and `RDCache` and related functions
- [x] Update documentation and register a breaking release

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.31.0"
+version = "0.30.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


torfjelde:
@sunxd3 can you undo the formatting changes? It makes it somewhat annoying to review + makes version history a less clear

PR #2165, opened 2024-02-03
Reenable PriorContext for Optimization

alyst:
This PR reenables the support for *PriorContext* in optimization, which was disabled by #2022.
While it is not directly supported by the current *optim_problem()* interface (only MLE/MAP are supported, although adding MaxPrior is possible), I think it is useful to support *PriorContext* for *Turing.OptimLogDensity()*.
In my code I use it for multi-objective Pareto optimization (prior + llh).


devmotion:
> This PR reenables the support for _PriorContext_ in optimization, which was disabled by #2022.

AFAICT support for `PriorContext` has never existed and has never been tested. The check in the PR was only introduced to communicate this more clearly and provide a better error message.

alyst:
> AFAICT support for PriorContext has never existed and has never been tested.

I see. It was working before though. Do you think it makes sense to support it for some of the user use cases?

torfjelde:
IMO if we 're allowing both `DefaultContext` and `LikelihoodContext`, I feel like allowing `PriorContext` also seems reasonable (i.e. _all_ the leaf-contexts).

src/optimisation/Optimisation.jl
```diff
@@ -83,6 +83,12 @@ function DynamicPPL.dot_tilde_assume(ctx::OptimizationContext, right, left, vns,
     return r, lp, vi
 end
 
+DynamicPPL.tilde_observe(::OptimizationContext{<:PriorContext}, right, left, vi) =
+    (0, vi)
+
+DynamicPPL.dot_tilde_observe(::OptimizationContext{<:PriorContext}, right, left, vi) =
+    (0, vi)
```
mhauru, line 90:
```suggestion
DynamicPPL.tilde_observe(ctx::OptimizationContext{<:PriorContext}, args...) =
    DynamicPPL.tilde_observe(ctx.context, args...)

DynamicPPL.dot_tilde_observe(ctx::OptimizationContext{<:PriorContext}, args...) =
    DynamicPPL.dot_tilde_observe(ctx.context, args...)
```
I think falling back on the `PriorContext` implementation is a bit neater. Result is the same.

mhauru (commented):
I had one minor style change to suggest.

I also got some extra peace of mind by, in the test, checking that
```
            true_prior_logpdf = -Distributions.logpdf(Uniform(0, 2), a[1])
            @test Turing.OptimLogDensity(m1, prictx)(a) ≈ true_prior_logpdf
```
Could incorporate a more explicit check like that.

Both of these are optional, I'd be okay with the code as is. I'll leave approving for @torfjelde since there might be context (pun unintended) to this that I lack.

alyst:
@mhauru I have incorporated your suggestions, thank you!

torfjelde:
Would be happy to include this:) But we need to fix the merge conflicts :confused:

mhauru:
Thanks @alyst!

torfjelde:
Lovely work @alyst :)

PR #2167, opened 2024-02-06
Fix error TuringOptimExt.jl with information matrix/vcov

smith-garrett:
I've been using the MLE/MAP estimation functionality ([defined here](https://github.com/TuringLang/Turing.jl/blob/master/ext/TuringOptimExt.jl)) with some Turing models, and I've come across an error in the `informationmatrix` function, extended from StatsBase.

In the function definition, the Hessian of the log likelihood or joint is first calculated on the unconstrained scale. Then that matrix is inverted and returned as the observed Fisher information matrix for the MLE/MAP. `StatsBase.vcov` is then extended to basically just be an alias for `informationmatrix`.

This seems to be incorrect. [According to Wikipedia](https://en.wikipedia.org/wiki/Observed_information), the observed information matrix is just the negative of the Hessian (not the inverse), and the variance-covariance matrix is the inverse of the information matrix. This pull request fixes these two spots in the code.

### DIFF
--- a/ext/TuringOptimExt.jl
+++ b/ext/TuringOptimExt.jl
@@ -86,7 +86,7 @@ function StatsBase.informationmatrix(m::ModeResult; hessian_function=ForwardDiff
     # Calculate the Hessian.
     varnames = StatsBase.coefnames(m)
     H = hessian_function(m.f, m.values.array[:, 1])
-    info = inv(H)
+    info = -H
 
     # Link it back if we invlinked it.
     if linked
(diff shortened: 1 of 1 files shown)


cpfiffer (approved):
Yep, this seems correct! Thanks for the fix 😁

smith-garrett:
My pleasure! I've been using Turing a lot in the last couple years, so I'm happy to be able to contribute a little bit. :)

It looks like there are a couple of tests still failing. I can have a look at those next week and update the pull request when they pass.

smith-garrett:
I've now updated the tests, but they still seem to be failing. Maybe I missed a step somewhere in getting the CI pipeline to run the new test versions?

smith-garrett:
@yebai Thanks for the update. I think my tests were off, but I think I have fixed them now. The vcov and information matrices now have separate tests based comparing the results to analytical calculations.

smith-garrett:
@yebai Sorry, there were some silly errors in my last commit. I think I've fixed everything now, as the tests are all passing on my computer. 🤞

smith-garrett:
There are still two failing tests, but they don't seem to have to do with the information matrix/vcov stuff as far as I can tell. I hope I haven't introduced a new error somewhere!

PR #2172, opened 2024-02-16
Add `accepted` arg to `AdvancedMH.Transition` calls

sunxd3:
Reflect https://github.com/TuringLang/AdvancedMH.jl/pull/88


devmotion:
As I wrote in TuringLang/AdvancedMH.jl#88 this approach can easily lead to misleading/incorrect acceptance rates?

PR #2176, opened 2024-02-26
Disable CI sampling progress logging

sunxd3:


src/Turing.jl
```diff
@@ -28,6 +29,9 @@ Enable progress logging in Turing if `progress` is `true`, and disable it otherw
 function setprogress!(progress::Bool)
     @info "[Turing]: progress logging is $(progress ? "enabled" : "disabled") globally"
     PROGRESS[] = progress
+    Logging.with_logger(Logging.NullLogger()) do
+        AbstractMCMC.setprogress!(progress)
```
devmotion, line 33: Should we just add a `quiet` or `silent` keyword argument to `AbstractMCMC.setprogress!` (and possibly the other `setprogress!` functions as well).

src/Turing.jl
```diff
@@ -28,6 +29,8 @@ Enable progress logging in Turing if `progress` is `true`, and disable it otherw
 function setprogress!(progress::Bool)
     @info "[Turing]: progress logging is $(progress ? "enabled" : "disabled") globally"
     PROGRESS[] = progress
+    AbstractMCMC.setprogress!(progress; silent=silent)
```
devmotion, line 32: I think we want
```suggestion
    AbstractMCMC.setprogress!(progress; silent=true)
```
regardless of whether the keyword argument `silent` of `Turing.setprogress!` is `true` or `false`: If it is `true`, then AbstractMCMC should be silent as well; and if it is `false`, then the info from Turing is already sufficient.

Issue #2180, opened 2024-03-11
Proposal for `MH` only works correctly with subtypes of `MvNormal`


torfjelde:
`MH` is implemented in a way where even if it receives a `AdvancedMH.RandomWalkProposal` it will still hit the "general" `propose!!` that assumes the proposal is a `NamedTuple`, i.e.

https://github.com/TuringLang/Turing.jl/blob/3a315cee636f8345a59b5f71a02c634d4a04631a/src/mcmc/mh.jl#L377-L383

instead of the one that is hit if we're a subtype of `MvNormal`

https://github.com/TuringLang/Turing.jl/blob/3a315cee636f8345a59b5f71a02c634d4a04631a/src/mcmc/mh.jl#L410-L417

For example, if I want to use a `filldist(..., n)` as a proposal instead of `product_distribution(Fill(...), n)`, stuff breaks:

``` julia
julia> using Turing, AdvancedMH, LinearAlgebra

julia> @model demo() = x ~ MvNormal(zeros(3), I)
demo (generic function with 2 methods)

julia> sample(
           demo(),
           MH(AdvancedMH.RandomWalkProposal(filldist(Normal(), 3))),
           10
       )
Sampling 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:00
ERROR: MethodError: no method matching _dist_tuple(::RandomWalkProposal{…}, ::DynamicPPL.TypedVarInfo{…}, ::@NamedTuple{…})

Closest candidates are:
  _dist_tuple(::NamedTuple{propnames}, ::DynamicPPL.VarInfo, ::NamedTuple{names}) where {names, propnames}
   @ Turing ~/.julia/packages/Turing/I5L1s/src/mcmc/mh.jl:335

Stacktrace:
  [1] dist_val_tuple(spl::DynamicPPL.Sampler{MH{(), RandomWalkProposal{…}}}, vi::DynamicPPL.TypedVarInfo{@NamedTuple{x::DynamicPPL.Metadata{…}}, Float64})
    @ Turing.Inference ~/.julia/packages/Turing/I5L1s/src/mcmc/mh.jl:317
  [2] propose!!(rng::Random.TaskLocalRNG, vi::DynamicPPL.TypedVarInfo{…}, model::DynamicPPL.Model{…}, spl::DynamicPPL.Sampler{…}, proposal::RandomWalkProposal{…})
[...]

PR #2181, opened 2024-03-11
Fix for #2180

torfjelde:
Fix for #2180


torfjelde:
Tests should pass after this

PR #2196, opened 2024-04-19
Bugfix for Optim.jl on models with different linked dimensionality

torfjelde:
See the test for an example that is currently breaking.

TL;DR: immutable versions of `invlink` and `link` avoid the issues of changing dimensionality between linked and invlinked reps that was causing this issue.

Probably related: https://github.com/TuringLang/DynamicPPL.jl/pull/525

NOTE: The Optim.jl interface and everything is due a rewrite IMO, but that is beyond the scope of this PR (because time is finite :pray:)


torfjelde:
@yebai this should be a quick merge

PR #2197, opened 2024-04-19
Bump DynamicPPL to v0.25

torfjelde:
Replaces Setfield.jl with Accessors.jl.

EDIT: Won't work until DPPL 0.25 is released.


torfjelde:
@sunxd3 it's good practice to let the author of a branch know if you contribute to it:) I was very confused as to what was going wrong here until I realized you had merged changes :sweat_smile:

torfjelde:
@sunxd3 any clue what's going on here?

torfjelde:
> No idea -- interesting it's only failing windows but not ubuntu

I think it's Julia 1.7 that's the issue, not windows vs linux

torfjelde:
Aye, it's the difference between 1.7 and 1.10.

In particular, on 1.7 we hit the following impl of `copyto!`

``` julia
function copy!(dst::AbstractVector, src::AbstractVector)
    if length(dst) != length(src)
        resize!(dst, length(src))
    end
    for i in eachindex(dst, src)
        @inbounds dst[i] = src[i]
    end
    dst
end
```

while on 1.10 we first hit

``` julia
function copy!(dst::AbstractVector, src::AbstractVector)
    firstindex(dst) == firstindex(src) || throw(ArgumentError(
        "vectors must have the same offset for copy! (consider using `copyto!`)"))
    if length(dst) != length(src)
        resize!(dst, length(src))
    end
    copyto!(dst, src)
end
```

which then eventually hits

``` julia
function _copyto_impl!(dest::Array, doffs::Integer, src::Array, soffs::Integer, n::Integer)
    n == 0 && return dest
    n > 0 || _throw_argerror("Number of elements to copy must be nonnegative.")
    @boundscheck checkbounds(dest, doffs:doffs+n-1)
    @boundscheck checkbounds(src, soffs:soffs+n-1)
    unsafe_copyto!(dest, doffs, src, soffs, n)
    return dest
end
```

i.e. we hit `unsafe_copyto!`, which I assume has no such check for `undef` and hence works without any issues. In contrast, the 1.7 version requires `getindex` impl.

Tbh, I'm a bit uncertain what's "correct" here (and it's definitively a BangBang.jl issue rather than something we're doing wrong).

> it's hitting NoBang 

Preferably not, but not surprised. BangBang.jl have fairly rudimentary checks to decide whether to mutate or not.

torfjelde:
Fix: https://github.com/JuliaFolds2/BangBang.jl/pull/22

torfjelde:
Note that this is a breaking change due to how the varnames are now shown, e.g. `x[1,2]` before is now `x[1, 2]`.

torfjelde:
I removed the failing test for Julia <1.8 for now since it's not "our fault". Once https://github.com/JuliaFolds2/BangBang.jl/pull/22 goes through, we'll bring it back

torfjelde:
Tests should be passing now

Issue #2195, opened 2024-04-18
Issues with constrained parameters depending on each other


torfjelde:
# Problem

```julia
julia> @model function buggy_model()
           lb ~ Uniform(0, 0.1)
           ub ~ Uniform(0.11, 0.2)
           # x ~ truncated(Normal(0, 1), lower=lb, upper=ub)
           x ~ transformed(Normal(0, 1), inverse(Bijectors.Logit(lb, ub)))
       end

buggy_model (generic function with 2 methods)

julia> model = buggy_model();

julia> chain = sample(model, NUTS(), 1000);
┌ Info: Found initial step size
└   ϵ = 3.2
Sampling 100%|█████████████████████████████████████████████████████████████████████████████| Time: 0:00:01

julia> results = generated_quantities(model, chain); # (×) Breaks!
ERROR: DomainError with -0.05206647177072762:
log was called with a negative real argument but will only return a complex result if called with a complex argument. Try log(Complex(x)).
DomainError detected in the user `f` function. This occurs when the domain of a function is violated.
For example, `log(-1.0)` is undefined because `log` of a real number is defined to only output real
numbers, but `log` of a negative number is complex valued and therefore Julia throws a DomainError
by default. Cases to be aware of include:

* `log(x)`, `sqrt(x)`, `cbrt(x)`, etc. where `x<0`
* `x^y` for `x<0` floating point `y` (example: `(-1.0)^(1/2) == im`)
...
````

In contrast, if we use `Prior` to sample, we're good:

```julia
julia> chain_prior = sample(model, Prior(), 1000);

Sampling 100%|█████████████████████████████████████████████████████████████████████████████| Time: 0:00:00

julia> results_prior = generated_quantities(model, chain_prior); # (✓) Works because no linking needed
```

The issue is caused by the fact that we use `DynamicPPL.invlink!!(varinfo, model)` when constructing a `transition`, which is what ends up in the chain rather than an issue with the inference itself.

For example, if we use AdvancedHMC.jl directly:
```julia
julia> using AdvancedHMC: AdvancedHMC

julia> f = DynamicPPL.LogDensityFunction(model);

julia> DynamicPPL.link!!(f.varinfo, f.model);

julia> chain_ahmc = sample(f, AdvancedHMC.NUTS(0.8), 1000);
[ Info: Found initial step size 3.2
Sampling 100%|███████████████████████████████| Time: 0:00:00
  iterations:                                   1000
[...]

PR #2202, opened 2024-04-20
Use `values_as_in_model` to extract the parameters from a `Transition`

torfjelde:
See https://github.com/TuringLang/Turing.jl/issues/2195 and https://github.com/TuringLang/DynamicPPL.jl/pull/588


torfjelde:
This PR depends on features introduced after the breaking release of DPPL and so tests will not pass until https://github.com/TuringLang/Turing.jl/pull/2202 has gone through

torfjelde:
For reference: used the new `DebugContext` to identify the bug in the tests and it was _very_ convenient:)

torfjelde:
@yebai @devmotion tests are passing and everything seems good (will bump patch version once it has gone through to avoid re-running tests)

You happy with it? I'd also like to point to #2215

PR #2204, opened 2024-04-20
Improvements to `externalsampler`

torfjelde:
Currently there's no way of choosing an ad backend for `externalsampler`, nor to specify whether it requires constrained or unconstrained parameters. This PR adds those options.

In addition, the PR adds a `default_adtype` method to be more consistent across the codebase.

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -18,6 +18,10 @@ import StatsBase
 import Printf
 import Random
 
+using ADTypes: ADTypes
+
+default_adtype() = ADTypes.AutoForwardDiff(; chunksize=0)
+
 const PROGRESS = Ref(true)
 
 # TODO: remove `PROGRESS` and this function in favour of `AbstractMCMC.PROGRESS`
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -79,22 +79,54 @@ abstract type StaticHamiltonian <: Hamiltonian end
 abstract type AdaptiveHamiltonian <: Hamiltonian end
 
 """
-    ExternalSampler{S<:AbstractSampler}
+    ExternalSampler{S<:AbstractSampler,Unconstrained}
+
+Represents a sampler that is not an implementation of `InferenceAlgorithm`.
+
+The `Unconstrained` type-parameter is to indicate whether the sampler requires unconstrained space.
 
 # Fields
 $(TYPEDFIELDS)
 """
-struct ExternalSampler{S<:AbstractSampler} <: InferenceAlgorithm
+struct ExternalSampler{S<:AbstractSampler,AD<:AbstractADType,Unconstrained} <: InferenceAlgorithm
     "the sampler to wrap"
     sampler::S
+    "the automatic differentiation (AD) backend to use"
+    adtype::AD
+
+    @doc """
+        ExternalSampler(sampler::AbstractSampler, unconstrained::Bool=true)
+
+    Wrap a sampler so it can be used as an inference algorithm.
+    """
+    function ExternalSampler(sampler::S, adtype::AD, unconstrained::Bool=true) where {S<:AbstractSampler,AD<:AbstractADType}
+        return new{S,unconstrained}(sampler)
+    end
 end
 
 """
-    externalsampler(sampler::AbstractSampler)
+    requires_unconstrained_space(sampler::ExternalSampler)
+
+Return `true` if the sampler requires unconstrained space, and `false` otherwise.
+"""
+requires_unconstrained_space(::ExternalSampler{<:Any,<:Any,Unconstrained}) where {Unconstrained} = Unconstrained
+
+"""
+    externalsampler(sampler::AbstractSampler; adtype=AutoForwardDiff(), unconstrained=true)
 
 Wrap a sampler so it can be used as an inference algorithm.
+
+# Arguments
+- `sampler::AbstractSampler`: The sampler to wrap.
+
+# Keyword Arguments
+- `adtype::ADTypes.AbstractADType=ADTypes.AutoForwardDiff()`: The automatic differentiation (AD) backend to use.
+- `unconstrained::Bool=true`: Whether the sampler requires unconstrained space.
 """
-externalsampler(sampler::AbstractSampler) = ExternalSampler(sampler)
+function externalsampler(sampler::AbstractSampler; adtype=Turing.default_adtype(), unconstrained::Bool=true)
+    return ExternalSampler(sampler, adtype, unconstrained)
+end
+
 
 getADType(spl::Sampler) = getADType(spl.alg)
 getADType(::SampleFromPrior) = AutoForwardDiff(; chunksize=0)
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -129,11 +129,11 @@ end
 
 
 getADType(spl::Sampler) = getADType(spl.alg)
-getADType(::SampleFromPrior) = AutoForwardDiff(; chunksize=0)
+getADType(::SampleFromPrior) = Turing.default_adtype()
 
 getADType(ctx::DynamicPPL.SamplingContext) = getADType(ctx.sampler)
 getADType(ctx::DynamicPPL.AbstractContext) = getADType(DynamicPPL.NodeTrait(ctx), ctx)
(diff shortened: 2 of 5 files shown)


src/Turing.jl
```diff
@@ -18,6 +18,10 @@ import StatsBase
 import Printf
 import Random
 
+using ADTypes: ADTypes
+
+default_adtype() = ADTypes.AutoForwardDiff(; chunksize=0)
```
torfjelde, line 23: Does anyone know why we specify `chunksize=0` instead of just using `chunksize=nothing`?

src/mcmc/Inference.jl
```diff
@@ -79,29 +79,61 @@ abstract type StaticHamiltonian <: Hamiltonian end
...
 """
-struct ExternalSampler{S<:AbstractSampler} <: InferenceAlgorithm
+struct ExternalSampler{S<:AbstractSampler,AD<:AbstractADType,Unconstrained} <: InferenceAlgorithm
     "the sampler to wrap"
     sampler::S
+    "the automatic differentiation (AD) backend to use"
+    adtype::AD
+
+    @doc """
+        ExternalSampler(sampler::AbstractSampler, unconstrained::Bool=true)
```
torfjelde, line 98:
```suggestion
        ExternalSampler(sampler::AbstractSampler, adtype::ADTypes.AbstractADType, unconstrained::Bool=true)
```

src/Turing.jl
```diff
@@ -18,6 +18,10 @@ import StatsBase
 import Printf
 import Random
 
+using ADTypes: ADTypes
+
+default_adtype() = ADTypes.AutoForwardDiff()
```
devmotion, line 23: Wouldn't a constant be simpler?
```suggestion
const DEFAULT_ADTYPE = ADTypes.AutoForwardDiff()
```

src/mcmc/Inference.jl
```diff
@@ -79,29 +79,61 @@ abstract type StaticHamiltonian <: Hamiltonian end
...
+    adtype::AD
+
+    @doc """
+        ExternalSampler(sampler::AbstractSampler, adtype::ADTypes.AbstractADType, unconstrained::Bool=true)
+
+    Wrap a sampler so it can be used as an inference algorithm.
+    """
+    function ExternalSampler(sampler::S, adtype::AD, unconstrained::Bool=true) where {S<:AbstractSampler,AD<:ADTypes.AbstractADType}
+        return new{S,AD,unconstrained}(sampler, adtype)
+    end
```
devmotion, line 104: Generally, the return type can't be inferred here. Maybe make unconstrained the first type parameter and only support
```julia
ExternalSampler{unconstrained}(sampler, adtype) where unconstrained = ExternalSampler{unconstrained,typeof(sampler),typeof(adtype)}(sampler, adtype)
ExternalSampler(sampler, adtype) = ExternalSampler{true}(sampler, adtype)
```

test/mcmc/abstractmcmc.jl
```diff
@@ -34,44 +34,91 @@ function initialize_nuts(model::Turing.Model)
...
 
 
-function initialize_mh(model)
+function initialize_mh_rw(model)
     f = DynamicPPL.LogDensityFunction(model)
     d = LogDensityProblems.dimension(f)
     return AdvancedMH.RWMH(MvNormal(Zeros(d), 0.1 * I))
 end
 
+# TODO: Should this go somewhere else?
```
devmotion, line 43: Probably DynamicPPL I'd say?

src/Turing.jl
torfjelde: Actually did that first :sweat_smile: 

Buuuut using a method _does_ allow an advanced user to actuallly overload the default AD backend that is used :shrug:

src/mcmc/Inference.jl
torfjelde: Am aware, and am happy to make this change :+1: Buuuut the user will be using `externalsampler` to construct it anyways, which will never be inferrable either :confused: 

Buuut I agree with you;; we should at least have the _option_ of using a type-inferrable constructor so will make the change :+1:

test/mcmc/abstractmcmc.jl
torfjelde: Was thinking the same. Only thing holding me back is that it's unclear to me if this is indeed how we want to implement something like this :confused: It was a quick--and-dirty way of doing it, hence why I figured I'd just leave it here for now.

src/Turing.jl
devmotion: I'm not completely convinced that this should be supported since it has such widereaching consequences. Some limited flexibility could be allowed by making the default a Preferences-constant that could be modified by users.

src/mcmc/Inference.jl
devmotion: Maybe use `Val(true)`/`Val(false)` in `externalsampler` to make it fully inferrable?

torfjelde: > Maybe use Val(true)/Val(false) in externalsampler to make it fully inferrable?

Don't like this as it requires users to be familiar with what `Val` is, which doesn't seem worth it given that `externalsampler` will never be used in "hot" code (it's really only going to be used before `sample(...)`, which is not type-stable anyways) :confused:

torfjelde: Made the changes you wanted to the constructor (though leeft `externalsampler` as is)

torfjelde: Making use of `Val` for constructor `ExternalSampler`, but leaving `externalsampler` as is I think

torfjelde:
You happy with the current version @devmotion ?

torfjelde:
I'll add one final thing to this PR: fix issues with initial params when working with unconstrained models.

torfjelde:
Added initial params handling + testing. Should pass as usual, at which point we can merge:)

torfjelde:
Actually, nvm, this is not breaking :facepalm: Will check that tests run after merge with master, and then will merge (unless there are objections)

src/Turing.jl
```diff
@@ -20,6 +20,10 @@ using Accessors: Accessors
 import Printf
 import Random
 
+using ADTypes: ADTypes
+
+DEFAULT_ADTYPE = ADTypes.AutoForwardDiff()
```
devmotion, line 25: Did you make it non-`const` on purpose?
```suggestion
const DEFAULT_ADTYPE = ADTypes.AutoForwardDiff()
```

src/mcmc/Inference.jl
```diff
@@ -78,30 +78,75 @@ abstract type Hamiltonian <: InferenceAlgorithm end
...
+    - `sampler::AbstractSampler`: The sampler to wrap.
+    - `adtype::ADTypes.AbstractADType`: The automatic differentiation (AD) backend to use.
+    - `unconstrained::Val=Val{true}()`: Value type containing a boolean indicating whether the sampler requires unconstrained space.
+    """
+    function ExternalSampler(
+        sampler::AbstractSampler,
+        adtype::ADTypes.AbstractADType,
+        unconstrained::Union{Val{false},Val{true}}=Val{true}()
+    )
+        return new{typeof(sampler),typeof(adtype),_bool_from_val(unconstrained)}(sampler, adtype)
```
devmotion, line 116: Wouldn't it be easier to write
```suggestion
       ::Val{unconstrained}=Val(true)
    ) where {unconstrained}
        if !(unconstrained isa Bool)
            throw(ArgumentError("Expected Val{true} or Val{false}, got Val{$unconstrained}"))
        end
        return new{typeof(sampler),typeof(adtype),unconstrained}(sampler, adtype)
```

src/mcmc/Inference.jl
```diff
@@ -78,30 +78,75 @@ abstract type Hamiltonian <: InferenceAlgorithm end
 abstract type StaticHamiltonian <: Hamiltonian end
 abstract type AdaptiveHamiltonian <: Hamiltonian end
 
+_bool_from_val(::Val{true}) = true
+_bool_from_val(::Val{false}) = false
+_bool_from_val(@nospecialize(x)) = throw(ArgumentError("Expected Val{true} or Val{false}, got $x"))
+
```
torfjelde, line 84:
```suggestion
```

torfjelde:
Accepted your changes @devmotion ; thanks!

devmotion:
I assume the errors are unrelated to this PR and caused by the latest release of AliasTables: https://github.com/LilithHafner/AliasTables.jl/issues/48

devmotion:
Should be fixed by AliasTables 1.1.1.

PR #2209, opened 2024-04-26
Bump ADTypes compat entry to include version 1

torfjelde:


torfjelde:
Will rerun checks once https://github.com/SciML/Optimization.jl/pull/738 has gone through 👍

PR #2217, opened 2024-05-08
Re-export certain functionality from DynamicPPL

torfjelde:
Given that Turing.jl has become a "public-facing" shell package, I think it's worth exporting some functionality commonly used with models from DynamicPPL.jl, so the end-user doesn't have to see DynamicPPL to be able to use these.

Amongst these I think the following should be included:
- `condition`
- `decondition`
- `fix`
- `unfix`
- `OrderedDict` (the only reliable way to call many methods, e.g. `rand`)
- `conditioned` / `observations` (though probably just one of these?)

Any objections? @yebai @devmotion

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -61,7 +61,7 @@ include("deprecated.jl") # to be removed in the next minor version release
 # Exports #
 ###########
 # `using` statements for stuff to re-export
-using DynamicPPL: pointwise_loglikelihoods, generated_quantities, logprior, logjoint
+using DynamicPPL: pointwise_loglikelihoods, generated_quantities, logprior, logjoint, condition, decondition, fix, unfix, OrderedDict
 using StatsBase: predict
 
 # Turing essentials - modelling macros and inference algorithms
@@ -114,7 +114,8 @@ export  @model,                 # modelling
         BernoulliLogit,         # Part of Distributions >= 0.25.77
         OrderedLogistic,
         LogPoisson,
-        NamedDist,
+
+        NamedDist,              # Exports from DynamicPPL
         filldist,
         arraydist,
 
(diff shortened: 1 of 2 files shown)


src/Turing.jl
```diff
@@ -61,8 +61,9 @@ include("deprecated.jl") # to be removed in the next minor version release
 # Exports #
 ###########
 # `using` statements for stuff to re-export
-using DynamicPPL: pointwise_loglikelihoods, generated_quantities, logprior, logjoint
+using DynamicPPL: pointwise_loglikelihoods, generated_quantities, logprior, logjoint, condition, decondition, fix, unfix, OrderedDict, conditioned
```
devmotion, line 64: It's a bit dangerous to re-export indirect dependencies. I think it would be better to use
```suggestion
using DynamicPPL: pointwise_loglikelihoods, generated_quantities, logprior, logjoint, condition, decondition, fix, unfix, conditioned
using OrderedCollections: OrderedDict
```

torfjelde: Good point; done!

torfjelde:
Okay, these tests failing are now becoming somewhat worrying (previously everything passed except on one run, but now it's also failing on other systems)

torfjelde:
Why was this merged?

torfjelde:
@yebai

torfjelde:
I literally said "please don't do this" just a few days ago.

torfjelde:
I see tests are at least passing after the merge, but still, please don't do this. Ping the creator of the PR to check first.

torfjelde:
But what arguments exist for merging vs. pinging the creator of the PR and asking them to merge @yebai ? I can't see any in favour of doing this. Do you have any?

In contrast, I can easily come up with arguments for _not_ doing this, e.g. we've had examples before where the PR creator realizes the test cases aren't sufficient, starts working on it, tries to push, only to realize the branch has already been merged and released _with bugs the PR creator is fixing_.

torfjelde:
But regardless of whether we want to make "don't merge someone else's PR without checking with them first" a team-wide rule or not, I ask that you at least don't merge PRs that _I_ have a created but instead ping me and ask if it's ready to merge. 
If I'm being very unresponsive, then sure, feel free to merge, but otherwise let me have the final say in whether a PR _I_ have made is ready or not.

PR #2218, opened 2024-05-08
Check model by default

torfjelde:
We now have the ability to perform some rudimentary model checks (https://github.com/TuringLang/DynamicPPL.jl/pull/540), and so we should probably perform a few of these at every call to `sample` with an inference alg.

This PR adds these checks by default. This means that now we will give informative errors if either of the following occurs:
- Varnames are repeated, e.g. https://github.com/TuringLang/DynamicPPL.jl/issues/427.
- Missings are found in observe statements.

We can also perform certain checks depending on the sampler, e.g. error early and informatively if the user attempts to use gradient based samplers for discrete parameters.

### DIFF
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -240,6 +240,15 @@ DynamicPPL.getlogp(t::Transition) = t.lp
 # Metadata of VarInfo object
 metadata(vi::AbstractVarInfo) = (lp = getlogp(vi),)
 
+# TODO: Implement additional checks for certain samplers, e.g.
+# HMC not supporting discrete parameters.
+function _check_model(model::DynamicPPL.Model)
+    return DynamicPPL.check_model(model; error_on_failure=true)
+end
+function _check_model(model::DynamicPPL.Model, alg::InferenceAlgorithm)
+    return _check_model(model)
+end
+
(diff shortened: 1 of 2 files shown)


torfjelde:
@yebai @devmotion

torfjelde:
> Should we consider x[1] and x[1:2] being a clash (and even weirder subsume cases). Also, after transitioning to Accessors, we lost some generality in equality testing. For instance, x[:] concretized on x = [1, 2, 3] is !== from x[1:3] (but it was == before).

We indeed should! But yes, this is something we should deal with in DPPL, not in this PR.

But yeah, you're very correct that we should probably check subsumes rather than equality :+1:

torfjelde:
This requires #2225 as some of the fixes are present there.

torfjelde:
I think the IS tests are all failiing because of previous changes which lead to evaluation of the model _twice_, buuut I don't understand why these weren't failiing before o.O These should have been failiing for a while.

torfjelde:
Think the random seeds have just been tuned carefully for those MH inference tests so because of the slight alterations in the RNG due to the model check run, they're now failiing.

I've tried to just provide decent initial params to `sample` which seems to have done the trick.

PR #2221, opened 2024-05-10
Optimization improvements

mhauru:
Implementation of the work started in #2214. Work in progress. Needs
* More and better tests.
* A general clean-up.
* Docstrings all over.
* Taking care of the TODOs littered around.

@torfjelde


mhauru:
This is at a stage where getting a review would be good. Please be pedantic and show no mercy, I need to learn the ways of the project. Some questions I've been wondering about myself:

1. Functions names, especially exported ones. `estimate_mode` and `link` were two that I hesitated on.
1. What's the convention for when to use implicit and explicit imports? I see a lot of source files mixing the two.
1. Do we want to keep using ModeResult? That's the type that OptimInterface.jl returns its results in. I made Optimization.jl return the same type for compatibility, but would probably have designed the return type differently if given free rein.
1. I've copied over tests from OptimInterface.jl's tests with minimal adaptation, resulting in some overlap. If the whole Optim.jl interface is on its way out though, might be worth leaving the overlap there, so that one can delete the OptimInterface.jl tests and not lose anything significant.
1. Depending on what we do with ModeResult and OptimInterface.jl tests, I might split off some of the tests to a new file of tests just for OptimizationCore.jl.

torfjelde:
Having a look now:)

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    ld = prob.log_density
+    if prob.linked
+        ld = Accessors.@set ld.varinfo = DynamicPPL.unflatten(ld.varinfo, solution_values)
+        ld = Accessors.@set ld.varinfo = DynamicPPL.invlink(ld.varinfo, ld.model)
+        solution_values = ld.varinfo[:]
+    end
+    # Store the parameters and their names in a NamedArray.
+    varnames = variable_names(prob.log_density)
+    vmat = NamedArray(solution_values, varnames)
+    return ModeResult(vmat, solution, -solution.objective, prob.log_density)
```
torfjelde, line 212: The following should do the trick:

```julia
    varinfo_new = DynamicPPL.unflatten(varinfo, solution.u)
    # `getparams` performs the invlinking, if needed, etc.
    vns_vals_iter = getparams(prob.model, varinfo_new)
    syms = map(Symbol ∘ first, vns_vals_iter)
    vals = map(last, vns_vals_iter)
    return ModeResult(
        NamedArray(vals, syms),
        solution,
        -solution.objective,
        prob.log_density
    )
```

Though probably need to fix some of the references.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-    vi = DynamicPPL.VarInfo(model)
+"""
+    estimate_mode(
+        model::Model,
+        estimator::ModeEstimator,
+        [init_value::Union{AbstractVector,Nothing}],
+        [solver];
+        adtype::AbstractADType=AutoForwardDiff(),
+        <constraints>,
+        <extra kwargs>
```
torfjelde, line 223: Same comment about docstrings (see below)

ext/TuringOptimExt.jl
```diff
@@ -2,105 +2,34 @@ module TuringOptimExt
...
+    import Turing:
+        DynamicPPL,
+        NamedArrays,
+        Accessors
+    import Turing.OptimisationCore:
+        ModeResult,
+        MLE,
+        MAP,
+        OptimLogDensity,
+        variable_names,
```
torfjelde, line 14:
```suggestion
```

Note needed anymore. See below

ext/TuringOptimExt.jl
```diff
@@ -2,105 +2,34 @@ module TuringOptimExt
...
+    import ..Turing:
+        DynamicPPL,
+        NamedArrays,
+        Accessors
+    import ..OptimisationCore:
+        ModeResult,
+        MLE,
+        MAP,
+        OptimLogDensity,
+        variable_names,
```
torfjelde, line 28:
```suggestion
```

ext/TuringOptimExt.jl
```diff
@@ -246,17 +175,11 @@ function _optimize(
...
 
-    # Make one transition to get the parameter names.
-    ts = [Turing.Inference.Transition(
-        Turing.Inference.getparams(model, f.varinfo),
-        DynamicPPL.getlogp(f.varinfo)
-    )]
-    varnames = map(Symbol, first(Turing.Inference._params_to_array(model, ts)))
-
     # Store the parameters and their names in an array.
+    varnames = map(Symbol ∘ first, Turing.Inference.getparams(model, f.varinfo))
```
torfjelde, line 179: Need to make corresponding changes to this objective; see comments on the similar `ModeResult` construction below.

src/Turing.jl
```diff
@@ -122,13 +124,13 @@ export  @model,                 # modelling
...
         MLE,
-        get_parameter_bounds,
-        optim_objective,
-        optim_function,
-        optim_problem
+        OptimLogDensity,
+        ModeResult,
+        estimate_mode,
+        maximum_a_posteriori,
+        maximum_likelihood
```
torfjelde, line 133: `MAP`, `MLE`, and `estimate_mode` no longer needs to be exported, right? `maximum_a_posteriori` and `maximum_likelihood` has us covered.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    ModeEstimator, variable_names
+using NamedArrays: NamedArray
+using Optimization
+using OptimizationOptimJL: LBFGS, IPNewton
+using DocStringExtensions: TYPEDFIELDS
 using Bijectors
 using Random
-using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD
+using SciMLBase: OptimizationFunction, OptimizationProblem, solve
+using ADTypes: AbstractADType
```
torfjelde, line 13:
```suggestion
using SciMLBase: SciMLBase
using ADTypes: ADTypes
```

When using stuff from other modules, almost always prefer qualified references, i.e. we explicitly write out `SciMLBase.solve`.

When working in larger projects like Turing.jl, it's always difficult to know exactly what is from where, so we should try to make this as easy as possible to tell for the reader.

(Note that this is unfortunately not a pattern we've followed well enough up until now)

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
 module Optimisation
 
 using ..Turing
+using ..OptimisationCore: ModeResult, MLE, MAP, OptimLogDensity, OptimizationContext,
+    ModeEstimator, variable_names
+using NamedArrays: NamedArray
+using Optimization
+using OptimizationOptimJL: LBFGS, IPNewton
+using DocStringExtensions: TYPEDFIELDS
```
torfjelde, line 9:
```suggestion
```

See comment below.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
 module Optimisation
 
 using ..Turing
+using ..OptimisationCore: ModeResult, MLE, MAP, OptimLogDensity, OptimizationContext,
```
torfjelde, line 4: What's the reasoning for not keeping all this in the same module? After this PR we sort of want to support all optimizers but _through_ Optimiziation.jl. Hence it seems a bit redundant to have another "core" module when we actually just want to rely on Optimization.jl anyways (and so our "core" becomes just interfacing with Optimization.jl).

Does that make sense?

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD
+using SciMLBase: OptimizationFunction, OptimizationProblem, solve
+using ADTypes: AbstractADType
 
 using Accessors: Accessors
-using DynamicPPL
-using DynamicPPL: Model, AbstractContext, VarInfo, VarName,
-    _getindex, getsym, getfield,  setorder!,
-    get_and_set_val!, istrans
+using DynamicPPL: Model, DefaultContext, LikelihoodContext
```
torfjelde, line 16: For stuff from `DynamicPPL` it's "less" important to qualify the names, _but_ I'm still preferential to doing so (at least for symbols not explicitly exported)

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
 
-import LogDensityProblems
-import LogDensityProblemsAD
+export estimate_mode, maximum_a_posteriori, maximum_likelihood
```
torfjelde, line 18:
```suggestion
export maximum_a_posteriori, maximum_likelihood
```

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    Iv<:AbstractVector,
+    Cons<:ModeEstimationConstraints,
+    OLD<:OptimLogDensity,
+    ADType<:AbstractADType,
+}
+    "The DynamicPPL model for which to estimate the mode."
+    model::M
+    """The mode estimator. Either `MLE()` for maximum likelihood or `MAP()` for maximum a
+    posteriori."""
+    estimator::E
```
torfjelde, line 100:
```suggestion
    """The mode estimator. Either [`MLE()` ](@ref) for maximum likelihood or [`MAP()`](@ref) for maximum a
    posteriori."""
    estimator::E
```

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-at the array `z`.
-"""
-function (f::OptimLogDensity)(z::AbstractVector)
-    varinfo = DynamicPPL.unflatten(f.varinfo, z)
-    return -getlogp(last(DynamicPPL.evaluate!!(f.model, varinfo, f.context)))
-end
-
-# NOTE: This seems a bit weird IMO since this is the _negative_ log-likelihood.
-LogDensityProblems.logdensity(f::OptimLogDensity, z::AbstractVector) = f(z)
+    ModeEstimationProblem(model, estimator, init_value, lb, ub, cons, lcons, ucons, adtype)
```
torfjelde, line 114: Generally want to keep method signatures to one line. If it overflows, just do `ModeEstimationProblem(args...)` and explain the args below.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-        # Save the gradient to the pre-allocated array.
-        copyto!(G, ∇neglogp)
-
-        # If F is something, the negative log joint is requested as well.
-        # We have already computed it as a by-product above and hence return it directly.
-        if F !== nothing
-            return neglogp
-        end
-    end
+`lb`, `ub`, `cons`, `lcons`, and `ucons` are constraints for the optimization problem as
```
torfjelde, line 118: Do stuff like

```julia
"""
    f(args...; kwargs...)

Return something something.

# Arguments
- `args...`: something

# Keyword Arguments
- `kwargs...`: something
"""
function f end
```

See https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation for official docs on method docstrings.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-    # Only negative log joint requested but no gradient.
-    if F !== nothing
-        return LogDensityProblems.logdensity(f, z)
-    end
+If `init_value` is `nothing`, it is generated by either sampling from the prior distribution
+or uniformly from the box constraints.
 
-    return nothing
+The problem is always created with `linked=false`. See [`link`](@ref) for how to transform
+it.
```
torfjelde, line 125: Is this needed? Is the linking not implicitly defined through the `varinfo`?

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+            "not yet implemented."
+        )
     end
-
-    return f
+    # Note that redefining obj and init_value out of place is intentional: It avoids
+    # issues with models for which linking changes the parameter space dimension.
+    ld = p.log_density
+    ld = Accessors.@set ld.varinfo = DynamicPPL.unflatten(ld.varinfo, copy(p.init_value))
+    ld = Accessors.@set ld.varinfo = DynamicPPL.link(ld.varinfo, ld.model)
```
torfjelde, line 164:
```julia
    ld = p.log_density
    varinfo = DynamicPPL.link(
        DynamicPPL.unflatten(ld.varinfo, copy(p.init_value))
        ld.model
    )
    ld = Accessors.@set ld.varinfo = varinfo
```

The constant `@set` just makes it unnecessarily difficult to read (know this is is from old code)

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    OptimizationProblem(me_prob::ModeEstimationProblem)
 
-function transform!!(p::AbstractArray, vi::DynamicPPL.VarInfo, model::DynamicPPL.Model, ::constrained_space{false})
-    linked = DynamicPPL.istrans(vi)
-    if linked
-        vi = DynamicPPL.invlink!!(vi, model)
+Create a SciML `OptimizationProblem` from a `ModeEstimationProblem`.
+"""
+function OptimizationProblem(me_prob::ModeEstimationProblem)
+    c = me_prob.constraints
```
torfjelde, line 177:
```suggestion
function OptimizationProblem(prob::ModeEstimationProblem)
    c = prob.constraints
```
IMO a bit nicer but no biggie :+1:

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    # Note that OptimLogDensity is a callable that evaluates the model with given
+    # parameters. Hence we can use it as the objective function.
+    f = OptimizationFunction((x, _) -> me_prob.log_density(x), me_prob.adtype; cons=c.cons)
+    if !has_constraints(me_prob)
+        opt_prob = OptimizationProblem(f, me_prob.init_value)
+    else
+        opt_prob = OptimizationProblem(
+            f, me_prob.init_value;
+            lcons=c.lcons, ucons=c.ucons, lb=c.lb, ub=c.ub
+        )
```
torfjelde, line 187:
```suggestion
    opt_prob = if !has_constraints(me_prob)
        OptimizationProblem(f, me_prob.init_value)
    else
        OptimizationProblem(
            f, me_prob.init_value;
            lcons=c.lcons, ucons=c.ucons, lb=c.lb, ub=c.ub
        )
```

Makes it clear that both branches are expected to set this variable.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+using ..OptimisationCore: ModeResult, MLE, MAP, OptimLogDensity, OptimizationContext,
+    ModeEstimator, variable_names
+using NamedArrays: NamedArray
+using Optimization
+using OptimizationOptimJL: LBFGS, IPNewton
+using DocStringExtensions: TYPEDFIELDS
 using Bijectors
 using Random
-using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD
+using SciMLBase: OptimizationFunction, OptimizationProblem, solve
```
torfjelde, line 12: This is strange: how is the current code working at all o.O

When you do `using SciMLBase: OptimizationProblem`, you can't overload it. So the re-definition below should cause issues AFAIK o.O

But this will be addressed when using qualifying references, as outlined in the comment below.

torfjelde (changes requested):
Alrighty, here comes comments! :grimacing:

Do let me know if anything is unclear, as there are quite a few of them. 

I think going forward I'll try to keep the reviews a bit "themed" so that there isn't too much feedback at once, but we instead can do several iterations. But given that I'm not feeling 100%, I figured I'd do as much reviewing as possible when I'm feeling okay-ish.

Overall, I also have a few comments:
1. See https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation for official docs on docstrings :+1:
2. Though I can see that the `ModeEstimationProblem` is in a way a nice abstraction, now that we're going to depend fully on Optimization.jl, it seems a bit overkill in terms of what is needed. IIUC it's only really used in `estimate_mode`, and so can't we just put all this functionailty in `estimate_mode` directly? IMO this will make the code a more readable and easier to follow, as it just becomes "1) set up `SciMLBase.OptimizationProblem`, 2) solve it, and 3) convert into human readable format".
3. I'd recommend merging the files `OptimisationCore.jl` and `Optimisation.jl` into a single one. Moreover, _in particular_ for the sake of reviewing, try to avoid copy-pasting code form existing files to other files before the majority of the reviewing has been done, as it makes it very difficult to see exactly what is _new_ code and what is simply moved code.

Hope this is helpful!

src/optimisation/Optimisation.jl
mhauru: The only reasoning was to not have OptimInterface.jl import from the Optimization.jl interface code, which seemed unaesthetical to me. If the OptimInterface.jl is going to go soon then I agree that there's no reason for the distinction.

Do we want to keep the distinction until we are confident that OptimInterface.jl is about to go though? Might want to make sure that we can e.g. turn every call to its exported functions into an equivalent call to `estimate_mode` behind the scenes, to avoid breaking user code.

src/optimisation/Optimisation.jl
mhauru: Nice, very happy to conform to that, would actually be my preference as well. The Julia convention of just doing `using X; using Y` and leaving users mystified as to which package some later-referenced thing comes from baffles me.

src/Turing.jl
mhauru: They don't need to, yes. I could imagine someone preferring them as an interface, but happy to minimise the exported surface if you prefer.

test/runtests.jl
```diff
@@ -11,7 +11,9 @@ using MCMCChains
 using NamedArrays
 using Optim
 using Optimization
-using OptimizationOptimJL
+using OptimizationBBO: BBO_separable_nes
+using OptimizationOptimJL: LBFGS, NelderMead, Fminbox, IPNewton
+using OptimizationNLopt: NLopt.LD_TNEWTON_PRECOND_RESTART
```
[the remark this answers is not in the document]
mhauru: No, it's actually used in the tests for Optimisation.jl, [here](https://github.com/TuringLang/Turing.jl/blob/472f6babd3c9c3ac1b77523627797df7aabab20f/test/optimisation/Optimisation.jl#L344). I wanted to check that things work for algorithms from a couple of different packages.

Relatedly, I might have a think about if we could restructure the tests into submodules, so that imports like these could only be in scope for the tests that actually need them, and be closer to where they are used.

src/optimisation/Optimisation.jl
torfjelde: > If the OptimInterface.jl is going to go soon then I agree that there's no reason for the distinction.

> Do we want to keep the distinction until we are confident that OptimInterface.jl is about to go though? 

IMO no :man_shrugging: I get the motivation that it's unaesthetic, but IMO for the uninitiated it makes it a bit more difficult to get a sense of what's going on. We want the user to only see `maximum_a_posteriori` and `maximum_likelihood`, with `optimize` just being a thing we expose if they do `using Optim` for the sake of backwards compat (but how we implement this behind the scenes, is just using Optimization.jl)

src/Turing.jl
torfjelde: Agree but I think those are sufficient "power users" where they can just import them or do the qualifier anyways :+1:

torfjelde:
> We can put all of it in estimate_mode, yes, though I would at least keep some subfunctions to separate concerns. I do find the abstraction clarifies it for me

I can see that; as I said, I think they make sense. _But_ it does add quite a bit of an overhead for a uninitiated reader who just wants to see what's up.

> he main reason for the struct was that whether the objective function and the initial value (and correspondingly the returned solution) have been transformed to unconstrained space or not need to go hand in hand. 

Makes a lot of sense! But I think after one of the changes suggested, this shouldn't really be a worry anymore.

Specifically, you can just do the following _once_:

``` julia
    ld = p.log_density
    varinfo = DynamicPPL.link(
        DynamicPPL.unflatten(ld.varinfo, copy(p.init_value))
        ld.model
    )
    ld = Accessors.@set ld.varinfo = varinfo
```

if we need to work in unconstrained space, and then use 

``` julia
    varinfo_new = DynamicPPL.unflatten(varinfo, solution.u)
    # `getparams` performs the invlinking, if needed, etc.
    vns_vals_iter = getparams(prob.model, varinfo_new)
    syms = map(Symbol ∘ first, vns_vals_iter)
    vals = map(last, vns_vals_iter)
    return ModeResult(
        NamedArray(vals, syms),
        solution,
        -solution.objective,
        prob.log_density
    )
```

(copy-paste from inline comments above) once you need to convert to a `ModeResult`. With this, there shouldn't be any other handling of unconstrained, etc.

Does that make sense?

> Fair point about copy-paste, sorry about that. I would usually make moving and editing separate commits, but this time the code went through a liquefy-and-reassemble stage in the beginning, and the first few commits do a bit of everything.

No worries:) I also should have mentioned this before, and I every much know what you mean:)

src/optimisation/Optimisation.jl
```diff
@@ -1,150 +1,25 @@
 module Optimisation
 
 using ..Turing
+using ..Turing: NamedArrays, ModeResult, MLE, MAP, OptimLogDensity, OptimizationContext, ModeEstimator
+using Optimization
+using OptimizationOptimJL: LBFGS
 using Bijectors
 using Random
-using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD
+using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD, solve, AbstractSciMLAlgorithm
```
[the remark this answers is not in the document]
mhauru: How do you feel about qualifying also references to imports from the standard library? Or at least only importing from them only the things you need, e.g. `using Random: rand`, to avoid polluting the namespace with a bazillion things not explicitly named in our module anywhere.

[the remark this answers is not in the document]
torfjelde: I think it sort of depends, e.g. `rand` is just used to much that we probably shouldn't bother qualifying it.

Similarly for `logpdf`, etc. Overall, we're probably happy with anything in `Base` being `using`ed in addition to Distributions.jl.

src/optimisation/Optimisation.jl
```diff
@@ -188,137 +60,142 @@ function transform!!(p::AbstractArray, vi::DynamicPPL.VarInfo, model::DynamicPPL
...
+    return estimate_mode(model, estimator, nothing, LBFGS(), args...; kwargs...)
 end
 
-function optim_objective(model::DynamicPPL.Model, estimator::Union{MLE, MAP}; constrained::Bool=true)
-    return _optim_objective(model, estimator, constrained_space{constrained}())
-end
+function estimate_mode(
+    model::DynamicPPL.Model,
+    estimator::ModeEstimator,
+    init_value::Union{AbstractVector,Nothing},
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
```diff
@@ -188,137 +60,142 @@ function transform!!(p::AbstractArray, vi::DynamicPPL.VarInfo, model::DynamicPPL
...
-function optim_function(
-    model::Model,
-    estimator::Union{MLE, MAP};
-    constrained::Bool=true,
-    adtype::Union{Nothing, AbstractADType}=NoAD(),
-)
-    if adtype === nothing
-        Base.depwarn("the use of `adtype=nothing` is deprecated, please use `adtype=SciMLBase.NoAD()`", :optim_function)
+    if init_value === nothing
+        init_value = init(model; ub=ub, lb=lb, lcons=lcons, ucons=ucons)
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/OptimisationCore.jl
```diff
@@ -0,0 +1,250 @@
...
+
+An abstract type to mark whether mode estimation is to be done with maximum a posteriori
+(MAP) or maximum likelihood estimation (MLE).
+"""
+abstract type ModeEstimator end
+struct MLE <: ModeEstimator end
+struct MAP <: ModeEstimator end
+
+# TODO(mhauru) Is the docstring out of date? I don't think this context does the
+# transformation.
```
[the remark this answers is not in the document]
mhauru: Fixed

src/optimisation/OptimisationCore.jl
```diff
@@ -0,0 +1,250 @@
...
+    init = VarInfo(model)
+    return Turing.LogDensityFunction(init, model, context)
+end
+
+"""
+    LogDensityProblems.logdensity(f::OptimLogDensity, z)
+
+Evaluate the negative log joint (with `DefaultContext`) or log likelihood (with `LikelihoodContext`)
+at the array `z`.
+"""
```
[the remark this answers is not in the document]
mhauru: Fixed.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    return estimate_mode(model, estimator, nothing, nothing, args...; kwargs...)
 end
 
+"""
+    maximum_a_posteriori(
+        model::Model,
+        [init_value::Union{AbstractVector,Nothing}],
+        [solver];
+        adtype::AbstractADType=AutoForwardDiff(),
+        <constraints>
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+    prob = ModeEstimationProblem(
+        model, estimator, init_value, lb, ub, cons, lcons, ucons, adtype
+    )
+    (solver === nothing) && (solver = default_solver(prob))
+    # TODO(mhauru) We currently couple together the questions of whether the user specified
+    # bounds/constraints and whether we transform the objective function to an
+    # unconstrained space. These should be separate concerns, but for that we need to
+    # implement getting the bounds of the prior distributions.
+    optimise_in_unconstrained_space = !has_constraints(prob)
+    optimise_in_unconstrained_space && (prob = link(prob))
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-    obj = transform!!(obj)
-    init = Init(obj.varinfo, model, constrained_space{false}())
-    t = ParameterTransform(obj.varinfo, model, constrained_space{true}())
+`solver` is the optimization algorithm to use. It can be any solver recognised by
+Optimization.jl. If it is `nothing` or omitted a default solver is used: LBFGS, or IPNewton
+if generic constraints are present.
 
-    return (obj=obj, init = init, transform=t)
-end
+`<constraints>` refers to the keyword arguments `lb`, `ub`, `cons`, `lcons`, and `ucons`
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
 
-struct MLE end
-struct MAP end
+A struct that holds constraints for mode estimation problems.
 
+The fields are the same as possible constraints supported by the Optimization.jl:
+`ub` and `lb` specify lower and upper bounds of box constraints. `cons` is a function that
+takes the parameters of the model and returns a list of derived quantities, which are then
+constrained by the lower and upper bounds set by `lcons` and `ucons`. We refer to these
+as generic constraints. Please see the documentation of Optimization.jl for more details.
```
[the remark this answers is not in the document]
mhauru: Added

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-            throw(ArgumentError("`OptimizationContext` supports only leaf contexts of type `DynamicPPL.DefaultContext` and `DynamicPPL.LikelihoodContext` (given: `$(typeof(context)))`"))
-        end
-        return new{C}(context)
-    end
+struct ModeEstimationConstraints{
+    Lb<:Union{Nothing,AbstractVector},
+    Ub<:Union{Nothing,AbstractVector},
+    Conns,
+    LConns<:Union{Nothing,AbstractVector},
+    UConns<:Union{Nothing,AbstractVector},
```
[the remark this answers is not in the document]
mhauru: Nope, pure typo. Fixed.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+        ))
     end
-    return r, lp, vi
+    if has_box_constraints(constraints)
+        return [
+            rand(Uniform(lower, upper))
+            for (lower, upper) in zip(constraints.lb, constraints.ub)
+        ]
+    end
+    return rand(model) |> values |> Iterators.flatten |> collect
```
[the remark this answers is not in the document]
mhauru: Nice, done.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-    return Turing.LogDensityFunction(init, model, context)
+struct ModeEstimationProblem{
+    M<:Model,
+    E<:ModeEstimator,
+    Iv<:AbstractVector,
+    Cons<:ModeEstimationConstraints,
+    OLD<:OptimLogDensity,
+    ADType<:AbstractADType,
+}
+    "The DynamicPPL model for which to estimate the mode."
```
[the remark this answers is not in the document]
mhauru: Done.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
-    return p
-end
+"""
+    OptimizationProblem(me_prob::ModeEstimationProblem)
 
-function transform!!(p::AbstractArray, vi::DynamicPPL.VarInfo, model::DynamicPPL.Model, ::constrained_space{false})
-    linked = DynamicPPL.istrans(vi)
-    if linked
-        vi = DynamicPPL.invlink!!(vi, model)
+Create a SciML `OptimizationProblem` from a `ModeEstimationProblem`.
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done. The commit doing all the code moving is the first one of the latest batch, so if you want to see changes other than the big move, skip 4b3fbf5e5a230714afef6f5774eae6095af8115b in the diff.

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Agreed that's nicer, done.

src/optimisation/Optimisation.jl
```diff
@@ -1,324 +1,327 @@
...
+"""
+function OptimizationProblem(me_prob::ModeEstimationProblem)
+    c = me_prob.constraints
+    # Note that OptimLogDensity is a callable that evaluates the model with given
+    # parameters. Hence we can use it as the objective function.
+    f = OptimizationFunction((x, _) -> me_prob.log_density(x), me_prob.adtype; cons=c.cons)
```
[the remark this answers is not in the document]
mhauru: Doing this with `OptimizationProblem` would constitute type piracy, no? Whereas doing it with `ModeEstimationProblem`... :)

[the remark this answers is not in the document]
mhauru: Alternatively could use the other parameters of the objective function like `f = OptimizationFunction((x, me_prob) -> me_prob.log_density(x), me_prob.adtype; cons=c.cons)`

src/optimisation/OptimisationCore.jl
```diff
@@ -0,0 +1,250 @@
...
+end
+
+"""
+    variable_names(lg::OptimLogDensity)
+
+Return the names of the variables in the model of an `OptimLogDensity` as symbols.
+"""
+function variable_names(lg::OptimLogDensity)
+    return map(Symbol ∘ first, Turing.Inference.getparams(lg.model, lg.varinfo))
+end
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Nice, done.

ext/TuringOptimExt.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Now qualifying these references, and importing them as e.g. `using DocStringExtension: DocStringExtension`.

src/optimisation/Optimisation.jl
mhauru: Agreed, done.

src/Turing.jl
mhauru: I left `MLE` and `MAP` since the Optim.jl interface needs for them to be exported. Can remove them later once we remove that extension, although will have to think about breakage and deprecation. Otherwise now down to just `maximum_likelihood` and `maximum_a_posteriori`.

ext/TuringOptimExt.jl
mhauru: Done.

ext/TuringOptimExt.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Good point, don't know what was going on there. Now using `import`.

src/optimisation/Optimisation.jl
mhauru: Not needed anymore, see other comments.

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
mhauru: I went with the latter option.

mhauru:
I think I've addressed all the points you raised @torfjelde. The big code-move where OptimisationCore.jl got merged into Optimisation.jl is in the first commit of the new batch, if you want to skip it in the diff. With `VarInfo` handling converting between constrained/unconstrained the way you pointed out above I agree that the `ModeEstimationProblem` struct became overkill, so I did remove it.

Let me know if you have any other comments. Before we merge, I'll merge main into this branch and add the `:=` test you added for the Optim.jl interface.

mhauru:
Actually, I'm gonna sort out the merge conflicts now, because otherwise CI isn't running, and it was failing earlier. Some of the failures seemed unrelated to this PR, but I need to investigate. If you don't want the merge commit in your diff, there's a button in the Files changed tab that allows you to select any range of commits to view the diff for.

torfjelde:
Awesome @mhauru :) Having a look now!

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
-    else
-        OptimizationFunction(
-            l;
-            grad = (G,x,p) -> obj(nothing, G, x),
-        )
-    end
-    
-    return (func=f, init=init, transform = t)
+This is a convenience function that calls `estimate_mode` with `MAP()` as the estimator.
+Please see the documentation of [`estimate_mode`](@ref) for more details.
```
torfjelde, line 494: `estimate_mode` is no longer exported right?
```suggestion
Please see the documentation of [`Turing.Optimisation.estimate_mode`](@ref) for more details.
```

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+function maximum_a_posteriori(model::DynamicPPL.Model, args...; kwargs...)
+    return estimate_mode(model, MAP(), args...; kwargs...)
 end
 
+"""
+    maximum_likelihood(
+        model::DynamicPPL.Model,
+        [initial_params::Union{AbstractVector,Nothing}],
+        [solver];
+        <keyword arguments>
```
torfjelde, line 505:
```suggestion
        kwargs...
```

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
-)
-    f, init, transform = optim_function(model, estimator; constrained=constrained, adtype=adtype)
-
-    u0 = init_theta === nothing ? init() : init(init_theta)
-    prob = OptimizationProblem(f, u0; kwargs...)
+Find the maximum likelihood estimate of a model.
 
-    return (; prob, init, transform)
+This is a convenience function that calls `estimate_mode` with `MLE()` as the estimator.
+Please see the documentation of [`estimate_mode`](@ref) for more details.
```
torfjelde, line 511:
```suggestion
Please see the documentation of [`Turing.Optimisation.estimate_mode`](@ref) for more details.
```

test/ext/OptimInterface.jl
```diff
@@ -1,35 +1,11 @@
...
-    value, logp, vi = DynamicPPL.tilde_assume(context.context, right, vn, vi)
-    return value, context.logprior_weight, vi
-end
-function DynamicPPL.tilde_observe(context::OverrideContext, right, left, vi)
-    logp, vi = DynamicPPL.tilde_observe(context.context, right, left, vi)
-    return context.loglikelihood_weight, vi
-end
-
-@numerical_testset "OptimInterface.jl" begin
+@numerical_testset "ext/OptimInterface.jl" begin
```
torfjelde, line 1:
```suggestion
@numerical_testset "TuringOptimExt" begin
```
?

test/optimisation/Optimisation.jl
```diff
@@ -0,0 +1,557 @@
+@testset "optimisation/Optimisation.jl" begin
```
torfjelde, line 1:
```suggestion
@testset "Optimisation" begin
```

torfjelde: You can argue that the pattern you're suggesting is a better one:) But AFAIK we're not following this pattern elsewhere in the code, so I think consistency is a bit more important atm.

test/optimisation/Optimisation.jl
```diff
@@ -0,0 +1,557 @@
...
+
+    # Issue: https://discourse.julialang.org/t/two-equivalent-conditioning-syntaxes-giving-different-likelihood-values/100320
+    @testset "OptimizationContext" begin
+        # Used for testing how well it works with nested contexts.
+        struct OverrideContext{C,T1,T2} <: DynamicPPL.AbstractContext
```
torfjelde, line 6: I wonder if this could maybe be replaced with `DynamicPPL.check_model_and_trace` and look at the captured trace :thinking: 

Don't do this for this PR, but might be interesting to consider for future testing.

test/optimisation/Optimisation.jl
```diff
@@ -0,0 +1,557 @@
...
+            m1 = Turing.Optimisation.estimate_mode(
+                gdemo_default, MLE()
+            )
+            m2 = maximum_likelihood(
+                gdemo_default, true_value, OptimizationOptimJL.LBFGS()
+            )
+            m3 = maximum_likelihood(
+                gdemo_default, OptimizationOptimJL.Newton()
+            )
+            # TODO(mhauru) How can we check that the adtype is actually AutoReverseDiff?
```
torfjelde, line 100: Annoyingly, I think the "best" approach is just

``` julia
types_used = []

@model function demo()
    global types_used
    push!(types_used, eltype(__varinfo__[:]))

    # Do other stuff.
end
```

and then check this.

The better approach is probably to generalize the `DynamicPPL.DebugContext` a bit to be a `CallbackContext`, which allows to just hook into internals of the models to extract the types used. Currently I think `DynamicPPL.check_model_and_trace` captures a bit too many details to be useful for this particular use-case.

test/optimisation/Optimisation.jl
```diff
@@ -0,0 +1,557 @@
...
+        end
+    end
+
+    @numerical_testset "gdemo" begin
+        @testset "MLE" begin
+            Random.seed!(222)
+            true_value = [0.0625, 1.75]
+            true_logp = loglikelihood(gdemo_default, (s=true_value[1], m=true_value[2]))
+
+            function check_success(result)
```
torfjelde, line 84: Are just redefining this method within every testset? If so, maybe we should define it once at the top of the file, and just have it take the `true_vals ` as an additional input

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+# Keyword arguments
+- `adtype::AbstractADType=AutoForwardDiff()`: The automatic differentiation type to use.
+- Keyword arguments `lb`, `ub`, `cons`, `lcons`, and `ucons` define constraints for the
+    optimization problem. Please see [`ModeEstimationConstraints`](@ref) for more details.
+- Any extra keyword arguments are passed to `Optimization.solve`.
+"""
+function estimate_mode(
+    model::DynamicPPL.Model,
+    estimator::ModeEstimator,
+    initial_params::Union{AbstractVector,Nothing},
```
torfjelde, line 405: I'd suggest we make this a kwarg, similar to what it is for `sample`.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+    args...;
+    kwargs...
+)
+    return estimate_mode(model, estimator, nothing, nothing, args...; kwargs...)
 end
 
+"""
+    maximum_a_posteriori(
+        model::DynamicPPL.Model,
+        [initial_params::Union{AbstractVector,Nothing}],
```
torfjelde, line 486: Same here: make a kwarg

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+Please see the documentation of [`estimate_mode`](@ref) for more details.
+"""
+function maximum_a_posteriori(model::DynamicPPL.Model, args...; kwargs...)
+    return estimate_mode(model, MAP(), args...; kwargs...)
 end
 
+"""
+    maximum_likelihood(
+        model::DynamicPPL.Model,
+        [initial_params::Union{AbstractVector,Nothing}],
```
torfjelde, line 503: Same here: make a kwarg

torfjelde (changes requested):
Starting to look real nice @mhauru ! Added some additional comments. Think after this we'll basically be there:)

src/optimisation/Optimisation.jl
```diff
@@ -37,25 +38,29 @@ The `OptimizationContext` transforms variables to their constrained space, but
...
+                (given: `$(typeof(context)))`
+            """
+            throw(ArgumentError(msg))
         end
         return new{C}(context)
     end
 end
 
-OptimizationContext(context::AbstractContext) = OptimizationContext{typeof(context)}(context)
+OptimizationContext(ctx::DynamicPPL.AbstractContext) = OptimizationContext{typeof(ctx)}(ctx)
```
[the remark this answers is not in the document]
mhauru: The qualified reference was about to push this above 92 chars, and `ctx` seems to be used as short-hand for context in other places.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
 
-    return lb, ub
+Create an `OptimizationProblem` for the objective function defined by `log_density`.
+"""
+function Optimization.OptimizationProblem(log_density::OptimLogDensity, adtype, constraints)
+    # Note that OptimLogDensity is a callable that evaluates the model with given
+    # parameters. Hence we can use it in the objective function as below.
+    f = Optimization.OptimizationFunction(
+        (x, _log_density) -> _log_density(x), adtype; cons=constraints.cons
+    )
```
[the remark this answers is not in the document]
mhauru: The `OptimizationFunction` constructor expects a callable that takes two arguments. The first is the values of parameters being optimised, the other is any other parameters that stay fixed during the optimisation. In our case, `_log_density` is the callable, hence the admittedly weird anonymous function there. This is an alternative to the earlier `(x, _) -> log_density(x)`, where `log_density` would be captured in a closure. Now, instead we pass `log_density` as an argument to `OptimizationProblem`, which then gives it as the second argument to our anonymous function.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
-function _optim_objective(model::DynamicPPL.Model, ::MAP, ::constrained_space{false})
-    ctx = OptimizationContext(DynamicPPL.DefaultContext())
-    obj = OptimLogDensity(model, ctx)
+"""
+    estimate_mode(
+        model::DynamicPPL.Model,
+        estimator::ModeEstimator,
+        [initial_params::Union{AbstractVector,Nothing}],
+        [solver];
+        <keyword arguments>
```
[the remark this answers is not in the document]
mhauru: Julia docs disagree though: https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+            DynamicPPL.LikelihoodContext()
+        end
+        ctx = OptimizationContext(inner_context)
+        ld = OptimLogDensity(model, ctx)
+        Accessors.@set ld.varinfo = DynamicPPL.unflatten(ld.varinfo, initial_params)
+    end
 
-    return (obj=obj, init = init, transform=t)
-end
+    # TODO(mhauru) We currently couple together the questions of whether the user specified
```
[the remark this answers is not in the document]
mhauru: Do you want the keyword arg now, even before we implement that? Would have to start throwing errors if the user sets it to `false` without giving explicit bounds, or conversely leaves it `true` when providing explicit bounds. I think that could be quite confusing, even if it would mean that our interface could be a bit more stable in the future.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+)
+    return estimate_mode(model, estimator, nothing, nothing, args...; kwargs...)
 end
 
+"""
+    maximum_a_posteriori(
+        model::DynamicPPL.Model,
+        [initial_params::Union{AbstractVector,Nothing}],
+        [solver];
+        <keyword arguments>
```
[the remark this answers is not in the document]
mhauru: See the other comment on this.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+    optim_result::O
+    "The final log likelihood or log joint, depending on whether `MAP` or `MLE` was run."
+    lp::Float64
+    "The evaluation function used to calculate the output."
+    f::M
+end
 
+#############################
+# Various StatsBase methods #
+#############################
```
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
+            "You must provide an initial value when using generic constraints."
+        ))
+    end
+    if has_box_constraints(constraints)
+        return [
+            rand(Distributions.Uniform(lower, upper))
+            for (lower, upper) in zip(constraints.lb, constraints.ub)
+        ]
+    end
+    return rand(Vector, model)
```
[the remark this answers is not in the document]
mhauru: Sure, done.

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,364 @@ function (f::OptimLogDensity)(F, G, z)
...
-    ctx = OptimizationContext(DynamicPPL.DefaultContext())
-    obj = OptimLogDensity(model, ctx)
-    
-    init = Init(obj.varinfo, model, constrained_space{true}())
-    t = ParameterTransform(obj.varinfo, model, constrained_space{true}())
-    
-    return (obj=obj, init = init, transform=t)
+    prob = Optimization.OptimizationProblem(log_density, adtype, constraints)
+    solution = Optimization.solve(prob, solver; kwargs...)
+    # TODO(mhauru) We return a ModeResult for compatibility with the older Optim.jl
```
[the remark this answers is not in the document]
mhauru: Done. #2232

src/optimisation/Optimisation.jl
mhauru: Good catch, fixed.

src/optimisation/Optimisation.jl
mhauru: Fixed.

test/ext/OptimInterface.jl
mhauru: Done.

test/optimisation/Optimisation.jl
mhauru: Done.

test/optimisation/Optimisation.jl
mhauru: Do you mind if I just open an issue and leave this for now?

test/optimisation/Optimisation.jl
mhauru: Yeah, no idea why I was doing that. Fixed now.

src/optimisation/Optimisation.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: Done.

src/optimisation/Optimisation.jl
mhauru: See the other comment on this.

test/runtests.jl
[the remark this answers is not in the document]
mhauru: Yup, doing that now as part of the move to qualified references.

test/optimisation/Optimisation.jl
mhauru: Fun side note, found this while fixing this: https://github.com/JuliaLang/julia/issues/54536

Project.toml
```diff
@@ -68,6 +70,8 @@ LogDensityProblems = "2"
 LogDensityProblemsAD = "1.7.0"
 MCMCChains = "5, 6"
 NamedArrays = "0.9, 0.10"
+Optimization = "3"
+OptimizationOptimJL = "0.1, 0.2"
```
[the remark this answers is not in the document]
mhauru: Thanks, I didn't realise 0.3 was a thing

test/optimisation/Optimisation.jl
torfjelde: Yep!

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
torfjelde: Is there a reason why we don't just do:
```julia
(log_density::OptimLogDensity)(u, p) = log_density(u)
```
and then
```julia
OptimizationFunction(log_density, adtype; cons=constraints.cons)
```
?

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
torfjelde: Uhmm _interesting_ :thinking: 

Throughout TuringLang, we always use `; kwargs...` :grimacing: E.g. https://github.com/TuringLang/AbstractMCMC.jl/blob/2a77f53114c4b3c5a11a3759e39233dab9e16f9a/src/interface.jl#L63-L74

THough I was clearly very wrong here, I'd say do `kwargs...` instead of `<keyword arguments>` to be consistent with the rest of the code-bases in TuringLang. If we want to move to `<keyword arguments>`, then we should do that in a single push to ensure everything is consistent :+1: 

But thank you for making me aware of this; did not know this:)

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
torfjelde: No, we introduce the kwarg with the functionality, not before :+1:

torfjelde (changes requested):
miiiinor final comments, but looking really good now:)

src/optimisation/Optimisation.jl
```diff
@@ -1,34 +1,35 @@
...
 using ..Turing
-using Bijectors
-using Random
-using SciMLBase: OptimizationFunction, OptimizationProblem, AbstractADType, NoAD
-
+using NamedArrays: NamedArrays
+using DynamicPPL: DynamicPPL
+import LogDensityProblems: LogDensityProblems
+import LogDensityProblemsAD: LogDensityProblemsAD
+import Optimization: Optimization
```
torfjelde, line 8:
```suggestion
using LogDensityProblems: LogDensityProblems
using LogDensityProblemsAD: LogDensityProblemsAD
using Optimization: Optimization
```

`import` isn't necessary if you're using qualified referneces:)

src/optimisation/Optimisation.jl
```diff
@@ -139,186 +154,330 @@ function (f::OptimLogDensity)(F, G, z)
...
+    # Create an OptimLogDensity object that can be used to evaluate the objective function,
+    # i.e. the negative log density. Set its VarInfo to the initial parameters.
+    log_density = let
+        inner_context = if estimator isa MAP
+            DynamicPPL.DefaultContext()
+        else
+            DynamicPPL.LikelihoodContext()
+        end
+        ctx = OptimizationContext(inner_context)
+        ld = OptimLogDensity(model, ctx)
```
torfjelde, line 428:
```suggestion
        ld = OptimLogDensity(model, ctx, adtype)
```

test/optimisation/Optimisation.jl
mhauru: Issue opened, see #2235

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
mhauru: No strong reason. It seemed unnecessary to me to define a whole new method for a type that, even though it isn't exported is being seen by some users, just to satisfy the needs of Optimization's interface. Can do that instead though if you prefer.

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
mhauru: Makes sense, done.

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Done

src/optimisation/Optimisation.jl
mhauru: Nice, had forgotten how these work. Done.

Is there a reason why we don't just use `import ModuleName`? Seems more concise that `using ModuleName: ModuleName`, and according to some docs I just read is equivalent. If we also wanted to make a rule out of not having mystery names in our namespaces, we could just say that the word `using` should not appear anywhere, and that would take care of it.

src/optimisation/Optimisation.jl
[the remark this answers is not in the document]
mhauru: I went and implemented this your way to speed things along. If I convinced you that my way is better you can just revert the last commit with a force push. :)

torfjelde (approved):
Fantastic stuff @mhauru :clap: I'm very happy about this change and you've done really great here:)

And sorry about the delayed reply; had a busy week with presentations, as you know. 

Feel free to merge (and we can bump the Project.toml afterwards to avoid having to re-run tests)

PR #2223, opened 2024-05-16
Fixes to AD backend usage in `externalsampler`

torfjelde:
I was playing around with an implementation of AutoMALA (see https://arxiv.org/abs/2310.16782 and Pigeons.jl) using AbstractMCMC.jl for a presentation I'm doing on Turing.jl-related stuff, and I ran into some issues with `externalsampler`.

Specifically, somewhere along the way, a `LogDensityProblemsAD.ADgradient(...)` of a `DynamicPPL.LogDensityModel` becomes _just_ a `DynamicPPL.LogDensityModel`, i.e. we lose the ability to compute gradients.

We didn't catch this in our tests because we both AbstractHMC.jl and AbstractMH.jl default to using `ForwardDiff` if needed, and so they would use the correct `adtype` in the very first iteration and then subsequently switch to using their own defaults :upside_down_smiley:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.31.4"
+version = "0.31.5"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/mcmc/abstractmcmc.jl
+++ b/src/mcmc/abstractmcmc.jl
@@ -12,7 +12,7 @@ function transition_to_turing(f::DynamicPPL.LogDensityFunction, transition)
     return Transition(f.model, varinfo, transition)
 end
 
-state_to_turing(f::LogDensityProblemsAD.ADGradientWrapper, state) = state_to_turing(parent(f), state)
+state_to_turing(f::LogDensityProblemsAD.ADGradientWrapper, state) = TuringState(state, f)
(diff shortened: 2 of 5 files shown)


torfjelde:
@yebai @devmotion @sunxd3 this should be a quick merge

torfjelde:
Had to disable a few tests for `MH`, which doesn't actually follow the LogDensityProblems.jl interfacr properly (calls the method using `NamedTuple` instead of `Vector`). Two things here: a) all the `NamedTuple` cases are already easily supported by the _internal_ `MH` sampler, so we're not losing any features here, and b) this is something that shoud be addressed in AdvancedMH.jl, not here.

torfjelde:
Uhmm the test failures seem very strange o.O I cannot reproduce this on my own machine; it's not windows, but I can't see how we can get `undef` behaviors _depending on the OS_ o.O

torfjelde:
It seems to be the same issue as https://github.com/JuliaFolds2/BangBang.jl/issues/21 but we're using the correct version of BangBang.jl, indicating that the fix I did in https://github.com/JuliaFolds2/BangBang.jl/pull/22 did not cover ReverseDiff.jl for some reason (though I'm also confused as to why we haven't caught this before).

torfjelde:
Okay, so _now_ this PR should be passing. Disabled the test as mentioned in https://github.com/TuringLang/DynamicPPL.jl/issues/612.

Once those tests have passed, this should be good to go.

PR #2224, opened 2024-05-17
Fix for `optimize` and `:=`

torfjelde:
`using Optim; optimize(model, MAP())` will currently result in an error if the model includes `:=`.

This PR fixes this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.31.4"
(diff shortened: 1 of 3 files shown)


torfjelde:
@yebai @sunxd3 should be a quick merge

torfjelde:
> Apologize @torfjelde, I don't understand why this error and how does this fix it. Could you elaborate a bit more?

Error is caused because now that we support `:=`, there's a discrepancy between what `Turing.Inference.getparams` returns (includes `:=` variables) and what `varinfo[:]` returns (does _not_ include `:=` variables)

PR #2225, opened 2024-05-17
Bump DPPL version to 0.27

torfjelde:
This includes a few breaking changes + some fixes.

Of greatest importance is:
- `prob"..."` macro is gone.
- Definitions of the form `@model demo(::Type{T}=...)` are now converted into `demo(::TypeWrap{T}=...)` to improve type-stability and thus performance. But this means that one can no longer call `demo(Vector{Float64})` but instead need to do `demo(TypeWrap{Vector{Float64}}())`.

Close https://github.com/TuringLang/Turing.jl/pull/2220


torfjelde:
Yep! Edited.

torfjelde:
But we need to merge https://github.com/TuringLang/DynamicPPL.jl/pull/609 first

torfjelde:
> Is the old syntax no longer supported, or just inefficient due to type instability?

Old syntax is still supported for _definition_, but not for calling.

torfjelde:
Anyone wants to give this the go-ahead?

torfjelde:
Waiting with the merge until I've gotten the other two PRs through, as they are not breaking (but need to fix tests)

PR #2231, opened 2024-05-18
Fixes and improvements to experimental `Gibbs`

torfjelde:
Turns out that there were some subtle bugs present in the impl of `Turing.Experimental.Gibbs` which is likely part of the reason why we were seeing some strange results here and there.

This PR does the following:
- Fixes these issues with, i.e. we're no longer accidentally hitting `condition` instead of `gibbs_condition`.
- Properly handles `initial_params`.
- More rigorous correctness testing of inference results.

Remaining TODOs:
- [ ] Make it work for `externalsampler` (part of the reason why I waited with this PR was it needed some functionality from #2223 )


torfjelde:
A few things:
1. Sampling from `gdemo` using `CSMC` in the gibbs sampler is _very_ slow. Removing those tests (requires a large number of iterations to something reasonable).
2. We're hitting OOM issues when using the SMC samplers. Seems somewhat crazy that this would happen.

torfjelde:
As there have been a few scenarios where we've hit some interesting snags wrt. failures of tests when only looking at "simple" statistics, e.g. mean, I'm trying to use something a bit more prinicpled. Specifically, I've added tests using Anderson-Darling tests (similar to Kolmogorov-Smirnov, but integrating over the entire ECDF instead of just considering the supremum) where we have tests to make sure that the significance level is set in such a way that a) that minor (both additive and mutliplicative) perturbations to the "true" samples are caught, but also b) tests between "true" samples and the samples of interest pass.

torfjelde:
Pfft I think the way of testing the marginals here is a really good way to go, and it's _almost_ there, but I'm starting to think that maybe Anderson-Darling is just a bit too strong of a test; it puts particular emphasis on the tails of the distributions, which of course can be a bit of a problem for any kind of MCMC output.

I'm thinking something like a Cramer-von Mises test would be perfect, as it doesn't inflate the differences in the tails of distribution, but is still more suitable than a Kolmogorov-Smirnov test (which only considers the maximum difference between the two distributions). Buuut no implementation of Cramer-von Mises exists in HypothesisTests.jl, so that's a bit annoying (see related issue: https://github.com/JuliaStats/HypothesisTests.jl/issues/201).

Don't have time to implement it myself right now (might be worth just piggy-backing off of scipy's implementation or something to give it a try), but something to keep in mind for future reference.

Note that this "investigation" all started because we've had experiences in the past where just testing means and std isn't really good enough, and so we start comparing quantiles. But if we're comparing quantiles, then we might as well just do a proper ECDF hypothesis test, where we make sure the acceptance threshold is sufficient to capture certain differences.

One possible (though seems lightly hacky) alternative might be to just compare the underlying test-statistics of samples from the inference method to be tested vs. test-statistics from "perturbations" of the "true" samples. E.g. you scale and shift the "true" samples, compute test-statistic, make sure that the test-statistic of samples from inference alg is better than these perturbed ones. Seems a bit hacky but also probably some way of theoretically motivating this.

torfjelde:
`externalsampler` will work with https://github.com/TuringLang/DynamicPPL.jl/pull/621

torfjelde:
What are all these optimization tests which are failing? Any idea @yebai @mhauru ?

mhauru:
Probably what has happened is that some dependency of ours has dropped support for ADTypes v0.2. We then install ADTypes v1 and that forces SciMLBase v1.92.0, which looks compatible because it puts no bounds on ADTypes, but is actually not compatible, because it simply doesn't use ADTypes.

What we would really like is to put a bound on SciMLBase that says `>1.92.0`, but it's not a direct dependency of ours. The "real" fix would be to bump Optimization.jl to v3.21, where it dropped support for the very old SciMLBase versions, but that one doesn't support Julia 1.7 any more.

Regardless I think this is independent of this PR, I would guess just rerunning CI on main would hit this issue.

mhauru:
I was correct that CI now fails on master too. I was wrong that SciMLBase is only an indirect dependency for us. Trying to fix this in #2267 by raising the lower bound for SciMLBase.

torfjelde:
Note that this should not be merged until https://github.com/TuringLang/DynamicPPL.jl/pull/621 has been merged! The tests currently contains the corresponding overload (which is type-piracy and no good). I'll remove those now so that the tests fail and we dont' accidentally merge.

torfjelde:
So _technically_ `Turing.Experimental.Gibbs` now works with `externalsampler`, which is nice.

Effectively, `Turing.Inference.recompute_logprob!!` is the method which (maybe) recomputes the logprob for the given sampler. Right now we _always_ recompute the logprob, even if the model hasn't changed (i.e. we haven't chnaged the values we're conditioing on since the previous iteration for this sampler) because that's always going to be correct. _But_ it is wasteful: in many scenarios we really don't need to recompute the logprob. Buuuut we'll leave that for future work.

_Buuuuut_ there is one **final** issue: we need to update the model (since we might condition on new values), which, in the case of gradient based samplers, might require recompiling the gradient tape, e.g. this is the case for `AutoReverseDiff` with `compile=true`.

Problem: since we have now moved the AD extensions to DPPL, we'd have to add extensions only for this implementation of `setmodel(f::LogDensityProblemsAD.ADGradientWrapper, model::DynmaicPPL.Model)`, which very much seems like overkill.

I therefore think we need something equivalent oof `setmodel` and `getmodel` in this PR present in DynamicPPL.jl, which subsequently needs to be overloaded for the specific AD backends which require recompilation of the tape.

torfjelde:
Yeah, this part from the previous comment:

> Buuuuut there is one final issue: we need to update the model (since we might condition on new values), which, in the case of gradient based samplers, might require recompiling the gradient tape, e.g. this is the case for AutoReverseDiff with compile=true.

> Problem: since we have now moved the AD extensions to DPPL, we'd have to add extensions only for this implementation of setmodel(f::LogDensityProblemsAD.ADGradientWrapper, model::DynmaicPPL.Model), which very much seems like overkill.

> I therefore think we need something equivalent oof setmodel and getmodel in this PR present in DynamicPPL.jl, which subsequently needs to be overloaded for the specific AD backends which require recompilation of the tape.

torfjelde:
> A issue is that the type of f::LogDensityProblemsAD.ADGradientWrapper will be thrown away, and a new wrapper will be created according to sampler's AD config. As far as I can tell, this is what we are doing right now anyway, but may worth some consideration.

So I did indeed consider dispatching on the adtype, but it's not quite enough for what we want to achieve here. Yes, it works with `ExternalSampler`, but we wanted to use this as a playground to enable Gibbs even outside of Turing.jl, in which case you're no longer  working with an `ExternalSampler`.

But I think the reasoning is fair, though it would be nice if, say, the `adtype` used in a gradient wrapper would also be stored with the gradient wrapper itself or something, so you could retrive this _from_ the gradient wrapper itself :confused:

torfjelde:
> hould there be an extra line that merge varinfos[1] into vi_base and then replace it with vi_base?

Very nice catch! From a first glance, it indeed looks like we're missing a `merge` :+1:

torfjelde:
Done 👍 

The bug was just in the initial step, so didn't really make much of a difference, but added the `merge` now.

torfjelde:
CI is failliing because of `check_model` issues on Julia 1.7 that somehow weren't caught before. Have opened PR to address this in DPPL.

torfjelde:
> @torfjelde to clarify, do we still need https://github.com/tpapp/LogDensityProblemsAD.jl/pull/33 after [7c4368e](https://github.com/TuringLang/Turing.jl/pull/2231/commits/7c4368e3c71586036bec64a05f70d7d343248c8f)?

We no longer _need_ it, no. But it would make things cleaner

torfjelde:
Yeah this should be good to go now:)

Red-Portal:
Great I'll look into incorporating the slice samplers using the new interface!

Issue #2179, opened 2024-03-11
refactor tests into multiple Github CI steps


yebai:
Similiar to https://github.com/TuringLang/JuliaBUGS.jl/issues/149. 

Let's remove the complex macros including `numerical_testset`, and `turing_testset` etc. Instead, let's simply group tests by files and run these test files in different CI runners. 

Useful resource for local test: https://gist.github.com/torfjelde/62c1281d5fc486d3a404e5de6cf285d4

torfjelde:
Very much agree with this :+1:

torfjelde:
> We only need to test gradient evaluation for the AD functionality test, but we currently run HMC, which is an overkill.

A counter point to this: https://github.com/TuringLang/DynamicPPL.jl/issues/612

Sometimes AD only breaks when we're sampling 😬

Issue #2234, opened 2024-05-22
Undeterministic test failure


mhauru:
I just had a case in #2221 where one of the CI test runs [failed](https://github.com/TuringLang/Turing.jl/actions/runs/9188559958/job/25268670695) but then after a trivial [commit](https://github.com/TuringLang/Turing.jl/commit/b6838de868fb834e688822b711493423986e706f) that shouldn't affect the failing tests in anyway it [passed](https://github.com/TuringLang/Turing.jl/pull/2221).

The failure is some approx comparison not being within tolerance in `test/experimental`. First suspicion would be an unfixed seed somewhere, but I haven't investigated yet.

PR #2237, opened 2024-05-28
Test restructuring

mhauru:
This PR makes two orthogonal changes to how tests are run and structured.

1. All test files included in `runtests.jl` are made into modules, and imports are moved into those modules. `runtests.jl` now has minimal imports. This cleans up the namespaces in which the individual tests are run, and makes it clearer where various references in tests point to.
2. Test files can now be included selectively using keyword arguments. If one runs e.g. `Pkg.test("Turing", test_args=["optim", "hmc", "--skip", "ext", "nelder"])` then only test files that have the substring `"optim"` or `"hmc"` in their path will be run, unless they also have one of the substrings `"ext"` or `"nelder"` in their path, in which case they are skipped. By default all tests are run. Substring comparisons are case-insensitive.

Point 2. makes the `@turing_testset` and `@numerical_testset` macros unnecessary, and they will be removed.

The selective running of tests is useful for both local development and splitting our CI to run tests for various parts of the Turing.jl in parallel (#2179). I'll try implementing the latter next to see that it goes smoothly.

This is an early draft, I'm opening it to get comments (@yebai, @torfjelde). I've only converted a couple of files to the submodule format, and the selective running lacks documentation. If this is deemed a good design I'll go and finish the implementation.


mhauru:
For the record, optimally I would have liked the selective test running to be based on testset names rather than file paths. I tried a couple of designs for that, but ran into various limitations and complexities in how `AbstractTestSet` and nesting macros work. This could be revisited later.

mhauru:
I finished the restructuring described in the opening message. I also removed the code coverage part of CI, because it seems to have been broken for the last > 1 year. If we want to reintroduce it I would propose running it as a separate job outside the current test matrix, since it requires running all the tests in one go and only on a single platform.

More work could be done harmonising the way we do imports and qualified references, but I think that could be a separate PR. In this one I avoided touching anything in the actual test code, and only modifying the "header" of each file.

The longest test CI job now takes around 25 minutes, so more could perhaps be done to make tests faster. Either making them more lightweight or splitting some of the heavy test files into several files.

test/essential/ad.jl
```diff
@@ -1,5 +1,24 @@
...
+import LogDensityProblems
+import LogDensityProblemsAD
+using Zygote
+
+using Turing
+using DynamicPPL: getval, getlogp
+using Turing: SampleFromPrior
+
+include(pkgdir(Turing)*"/test/test_utils/models.jl")
+include(pkgdir(Turing)*"/test/test_utils/ad_utils.jl")
```
willtebbutt, line 18:
```suggestion
include(pkgdir(Turing) * "/test/test_utils/models.jl")
include(pkgdir(Turing) * "/test/test_utils/ad_utils.jl")
```
We usually use a space around binary operators. I think this thing happens in a few more places as well.

test/essential/ad.jl
```diff
@@ -1,5 +1,24 @@
...
+using ForwardDiff
+using ReverseDiff
+using Distributions: logpdf
+import LogDensityProblems
+import LogDensityProblemsAD
+using Zygote
+
+using Turing
+using DynamicPPL: getval, getlogp
+using Turing: SampleFromPrior
```
willtebbutt, line 15: Could you expain a bit behind how you've chosen to format imports / usings? Skimming through the files, I feel like there's probably a structure to the way you've done things, but I'm not 100% sure what it is just by looking at it!

willtebbutt (commented):
I have no particular comments on the functionality, just a couple of comments on style! I've commented on representative examples, rather than going through and commenting on each thing manually.

test/essential/ad.jl
mhauru: At some point I think I started, out of old habit, doing a split into Stdlib - external packages - Turing.jl itself. Clearly I've deviated from that here. Do we have a convention within the project?

Same goes for `using X: X` vs `import X`. I think I've seen both, is one preferred?

willtebbutt: Ah okay. I'm not aware of existing conventions on this front. Maybe @yebai or @torfjelde have a view on this.

mhauru: The [Blue](https://github.com/JuliaDiff/BlueStyle?tab=readme-ov-file#module-imports) style guide recommends alphabetic ordering, so could go with that.

mhauru:
Thanks for the comment @devmotion, happy to discuss this. Here are reasons why I went with `ARGS` rather than env vars, not in any particular order:
1. I wanted to have the option to include and exclude multiple test files. This requires some parsing of the input, and doing that for command line arguments is more natural and more standard practice than for env vars, which are usually single values.
2. `Pkg.test` has features explicitly for this, which I think is a strong hint that this can be considered the standard way of doing test selection in Julia. `Pkg.test` [docs](https://pkgdocs.julialang.org/v1/api/#Pkg.test) mention exactly this as an intended use case of `test_args`.
3. I find command line arguments handier when running things manually: I'm more used to passing things args than setting env vars when I run things from a terminal. For CI I think both work equally well.

>It's also the approach taken throughout the SciML ecosystem and in e.g. Bijectors, DynamicPPL, and DistributionsAD.

That's a fair point that I hadn't considered. At the same time, the `ARGS` approach is what [Julia itself does](https://github.com/JuliaLang/julia/blob/6a10d03c5ec7ceaace325e2379ac1f0cac7c3761/test/runtests.jl#L16) (I took the name of `--skip` from the tests of stdlib).

>it's much easier to use ENV variables for test GROUPs (similar to what @willtebbutt uses in Tapir it seems) since this does (almost) not require any additional code

I would argue that using `ARGS` does not require any more extra code in `runtests.jl`. The only reason why there's any substantial extra code here (still <30 lines + docstrings) is because of parsing of include/exclude tests, which I would like to have regardless of whether we use env vars or `ARGS`.

> you can just update the ENV variable in the julia REPL and (re)run ] test

You can have a similar user experience with `ARGS`, by doing `ARGS .= ["my", "new", "--skip", "args"]; ] test`.

test/essential/ad.jl
mhauru: Good point, fixed.

devmotion:
I can see these points and I'm sure tests could be structured with `ARGS` as well - but I think it would be good to follow common practices in the ecosystem and use environment variables.

IMO one of the main problems with the current Turing test setup is that it does not follow common practices in the ecosystem and defines its own macros for structuring the tests. My impression is that this makes it unnecessarily complicated for new contributors to read, modify, and add tests and adds additional maintenance burden on our side. My worry is that with an uncommon practice such as `ARGS` (I haven't encountered a single Julia package yet that uses it to structure tests) we will still make it more complicated than necessary for new contributors (and the custom parsing requires additional maintenance on our side).

devmotion:
I'm still not convinced (but also don't care too much if everyone else has a different opinion) - I think it's good to adopt improvements even if they are not common practice yet, but I don't see why/how `ARGS` is superior than `ENV` and therefore I would go with the more established and more well-known approach.

test/essential/ad.jl
mhauru: I alphabetised the imports, and started using `import X` over `using X: X`. It's now violating the Blue guide though, because that one says `import`s and `using`s should be in separate blocks.

Maybe further edits on this would warrant first a broader discussion about setting a style on this, and then we could start following that style not just here, but all over the codebase?

mhauru:
I don't really have anything to add to my earlier comments on command line arguments vs env vars, except that I don't think the args setup would get in the way when adding/modifying existing tests. By default it runs all tests, and there's no manual grouping of tests so you can add new ones even if you're oblivious that there even is a way to run tests selectively. I appreciate that consistency with related packages has value though. I'm happy to say it's Hong's call, but that's probably because he seems to prefer what I prefer. :)

Regardless of the final design, I'll add a note in the [contributing](https://turinglang.org/dev/docs/contributing/guide) docs page on how to run the tests once this has been merged.

In other news, I took all the files in `test/test_utils` and either made them into modules that are now imported with relative imports, or, in cases where only one test file was using the utils, merged them into the test file. I also deleted a bunch of code from the utils that wasn't being used anywhere.

There's also `test/skipped` which is about 900 lines of code not included in runtests.jl. Might these tests be brought back at some point, or are they run manually in some particular situations?

willtebbutt:
@yebai should this not be a patch version bump? It looks non-breaking

mhauru:
> One suggestion for CI setup is that failures of particular CI tasks should not cancel other CI tasks so we can independently identify multiple CI failures in parallel.

Good point, should now be fixed.

Assuming CI passes (the previous error was random, see #2234), I have no other changes in mind that I would like to make. I also don't have merge rights, so someone else needs to make the call on when this is final.

torfjelde:
> In other words, what would be the env var equivalent of doing Pkg.test("Turing", test_args=["optim", "hmc", "--skip", "ext", "nelder"])?

This is a really good argument :+1: You could of course implement the same parsing of an `ENV` variable, but I very much agree that it's much, much more natural to do so using `ARGS`.

> Respecting Pkg.test's intended way for selecting tests to run.

I that the the docs mention that this is a use-case for args, but does it specifically recommend that over ENV variables? 

I'm am genuinely quite keen on giving this pattern a go, but I'm also somewhat worried about doing this in Turing.jl :confused: 

Has there been a similar discussion elsewhere? I.e. what's the reasoning of the rest of the ecosystem going with `ENV` variables?

mhauru:
>This is a really good argument 👍 You could of course implement the same parsing of an ENV variable, but I very much agree that it's much, much more natural to do so using ARGS.

Yeah, I think parsing complicated env vars gets awkward. You could define two env vars though, `INCLUDE_TESTS` and `EXCLUDE_TESTS`, both of which could be space separated lists. I don't know if that would satisfy the desire of consistency with SciML practices any more though.

> I that the the docs mention that this is a use-case for args, but does it specifically recommend that over ENV variables?

No, I don't think it says anything about environment variables.

mhauru:
The test failures seem unrelated to this PR. They seem to rather be about the recent ADTypes compat change interacting in funky ways with Optimization.jl version bounds when using Julia 1.7. I haven't gotten to the bottom of it yet.

devmotion:
> In other words, what would be the env var equivalent of doing

Probably something similar to `JULIA_LOADPATH` etc. where a single environment variable is split into a vector of paths/arguments based on a delimiter such as `:` (using e.g. `split(ENV["..."], ':')`).

Generally, I haven't encountered the need for such complex inclusion/exclusion of tests yet, so I'm a bit skeptic whether the additional complexity (both for users and developers) is worth it. Having the possibility to run all tests (typically the default but not used in CI) and pre-defined groups of tests has been sufficient in all packages I have worked with. For running even smaller chunks of tests or subsets of test files in local development in my experience typically the most convenient approach is to use TestEnv anyway.

mhauru:
> Can we parse/extract test options in both TestEnv and ARGS at the start of runtests.jl and push them into an array of test configurations?

Yep, we could definitely have both available. Have e.g. the ARGS parsing implemented as in the current version of this PR and append to the list of included tests the value of an env var called `TURING_SELECT_TEST`. Would everyone be happy with that?

Regardless of ARGS/env vars choice, I would still like to stick to tests being selected by file path rather than by manually grouping them. Benefits are
* No code needed for introducing the groups. File folders should form natural groups anyway.
* Can't accidentally have a test that's not within a group, or a group that isn't included in CI.
* More granularity and flexibility in test selection.

Consequently calling the env var `GROUP` doesn't feel right, even though that seems to be what other SciML packages do.

torfjelde:
> For running even smaller chunks of tests or subsets of test files in local development in my experience typically the most convenient approach is to use TestEnv anyway.

So I've also done the same, _but_ having had experience with testing in other languages, I also do sometimes miss the ability to simply filter tests based on their name or whatever. It's particularly nice to do when you're working on bugfixes, since then there's a particular sub-test-suite you want to run, which then usually requires copy-pasting to another file, using TestEnv.jl, etc. 

> Can we parse/extract test options in both TestEnv and ARGS at the start of runtests.jl and push them into an array of test configurations? 

Both also seems non-ideal, no? :confused: I'd rather have to do that (push onto `ARGS`) manually than us having to support both.

torfjelde:
I'm also slowly coming to terms with the `ARGS` approach. In hindsight, I'd much prefer to have tried this out in another package, e.g. Bijectors or something, which is "less mission critical". But given the fantastic work @mhauru has done here, it might be worth just giving it a go in Turing.jl :thinking:

mhauru:
Given @yebai and @torfjelde have come around to args and @devmotion said earlier that he's not too bothered if others are happy with this, I'll go and merge. Everyone seems to agree that this is regardless an improvement over the current situation, and if people miss having a SciML-style `GROUP` env var we can easily add it by building on the mechanisms introduced here.

torfjelde:
Even better, in general you can do:
```julia
using Pkg
Pkg.test(Pkg.project().name, test_args=ARGS)
```
though `Pkg.project` is apparently an experimental feature.

torfjelde:
Question is whether we should just have a small script `test/testrunner.jl` in the actual repo that does this? Almost seems worth it, no?

mhauru:
This seems like a common problem for many Julia packages, is there no standard solution to this? It feels frustrating that `julia --project test/runtests.jl -- a b c --skip d e` so nearly works. The only reason it doesn't is that it doesn't create a TestEnv.

devmotion:
I'd suggest running
```shell
julia --project=. -e 'import Pkg; Pkg.test(; test_args=ARGS)' -- a b c --skip d e
```
if you don't work in a Julia REPL and want to execute the tests.

PR #2243, opened 2024-05-30
Drop support for ADTypes 0.2

yebai:
ADTypes 0.2 doesn't support AutoTapir yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.32.2"
+version = "0.32.3"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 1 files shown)


torfjelde (approved):

torfjelde (changes requested):
Wait. This doesn't really fix the issue though.

Even if we make a release now, whenever someone uses a package with ADTypes.jl v0.2, it will just install v0.32.2, which breaks. You need to pull the release @yebai

devmotion:
> The standard approach for fixing a bug is through a new release instead of reversing a release, right?

People seem to have quite diverse (and strong) opinions on this matter 😄 But some maintainers of the registry are definitely not happy about yanking releases. However, as [stated in the README](https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate)

> There is however, a special category of bugged releases that can not be resolved by having a patch release. These also may to be resolved by yanking. That special category is when the compat bounds have been set too wide. i.e. say v2.10.0 was released using a feature not on julia v1.6 but the compat entry for julia was not raised in the release. In this case releasing a v2.10.1 with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still resolve the broken v2.10.0, and as a minor bump, reverting the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement) or yank the offending release.

torfjelde:
Great! Approving now then.

torfjelde (approved):

PR #2250, opened 2024-06-04
Optimization compat fix

mhauru:
In #2237 we are running into test failures on Julia 1.7 because SciMLBase 1.92 is getting installed. This causes problems when we pass ADTypes to `OptimizationFunction`s, used in mode estimation, since SciMLBase 1.92.1 is the first version that made SciMLBase's AbstractADType be the same as that from ADTypes. This PR fixes our compat with Optimization.jl to be 3.21 or greater, since 3.21 is when Optimization.jl dropped support for SciMLBase 1.*.


mhauru:
Ah, should have seen that coming. Optimization.jl 3.21 only supports Julia 1.9 and up.

mhauru:
See #2251

Issue #2251, opened 2024-06-04
Compat issue


mhauru:
We have a problem with ADType/SciMLBase versions in #2237 where some tests are failing on Julia 1.7. I'll try to summarise my understanding of it below. This is partially to clarify my own thoughts, excuse me if this is a bit hard to follow, hopefully you can skip some of the details.

1. Optimization.jl gets its `OptimizationFunction` type from SciMLBase. One of its constructor arguments is an `AbstractADType`.
2. SciMLBase started using `ADType.AbstractADType` only from its v1.92.1 onwards, at v1.92.0 they still define their own `AbstractADType`. Until then they also had no compat entry for ADTypes.
3. Optimization.jl dropped support for Julia 1.6, 1.7, and 1.8 in its version v3.14.1.
4. Until v3.21 Optimization has a compat entry supporting SciMLBase all the way down to v1.79. At v3.21 the SciMLBase dependency is bumped to SciMLBase v2.16.3.
5. SciMLBase started supporting ADTypes v1 at its v2.34, until then it was ADTypes v0.2 or lower.
6. We dropped support for ADTypes v0.2 a few days ago.

So when installing on Julia 1.7
1. We get Optimization.jl v3.14, because that's the latest that supports 1.7.
2. We would need a version of SciMLBase that's compatible with both Optimization.jl v3.14, which means something in ^1.79, but also something that supports ADTypes v1, which means ^v2.34. These are clearly incompatible.
3. But fear not, for there's always SciMLBase 1.92.0 which provides no compat bounds whatsoever for ADTypes! So that's what we get.
4. But SciMLBase v1.92.0 doesn't even interface with ADTypes, and thus our tests fail with a "no matching method" when constructing an `OptimizationFunction`.

I'm not quite sure what to do about this. Can we restore support for ADTypes v0.2, which would then allow e.g. SciMLBase v1.95, and we would be fine? Or could we drop support for Julia <1.9? Something else?

devmotion:
Doesn't help but IMO it's pretty annoying that SciML packages bumped the Julia compat in this way (even though many packages would still work, sometimes with minor modifications, on older Julia versions). Turing is not the first project where this causes problems.

The ADTypes updates also broke many packages due to incorrect compat entries...

> But SciMLBase v1.92.0 doesn't even interface with ADTypes, and thus our tests fail with a "no matching method" when constructing an OptimizationFunction.

[...]

PR #2252, opened 2024-06-04
Restore compat with ADTypes v0.2. Make AutoTapir export conditional.

mhauru:
As discussed in #2251


torfjelde:
Lovely @mhauru :)

CI failure is a bit strange (though it seesm unrelated)

PR #2256, opened 2024-06-06
Add note about breaking v0.33.0 changes to HISTORY

mhauru:


mhauru:
Let's pretend that I meant to do that to test the new formatter action...

PR #2257, opened 2024-06-06
Add Aqua to tests

mhauru:
I tried out [Aqua.jl](https://juliatesting.github.io/Aqua.jl/stable/) and it caught some issues worth fixing, so I fixed them and made Aqua run as part of the test suite. Had to disable the method ambiguity checks though, because it was catching a lot of ambiguities not caused by us, but rather by our dependencies.

Ping @willtebbutt since he might be interested.


test/Aqua.jl
```diff
@@ -0,0 +1,10 @@
+module AquaTests
+
+using Aqua: Aqua
+using Turing
+
+# TODO(mhauru) We skip testing for method ambiguities because it catches a lot of problems
+# in dependencies. Would like to check it for just Turing.jl itself though.
+Aqua.test_all(Turing; ambiguities=false)
```
devmotion, line 8: If you want to test ambiguities, I suggest
# Test ambiguities separately without Base and Core
```suggestion
@testset "Aqua" begin
    # Test ambiguities separately without Base and Core
    # Ref: https://github.com/JuliaTesting/Aqua.jl/issues/77
    Aqua.test_all(Turing; ambiguities = false)
    Aqua.test_ambiguities(Turing)
end
```

mhauru: I won't get to this today, and next week I'm off this project, so I'll merge what's here now and open an issue to come back to method ambiguities.

PR #2267, opened 2024-06-18
Bump SciMLBase compat lower bound

mhauru:
Fixes SciMLBase compat issues that were partially fixed in #2251 and #2252.


mhauru:
One test failed due to happenstance, rerunning that, but I think we are good here.

torfjelde (approved):
Lovely:)

Issue #1417, opened 2020-09-28
Feature Request: Better Indexing for ModeResult


BlackWingedKing:
Based on the discussion [here](https://julialang.slack.com/archives/CCYDC34A0/p1601211459038100)
I was using Turing's MLE, MAP estimation. The  object returned by `optimize(model, MLE())` is a `ModeResult` object
Here is a sample example.
```
mle_estimate = optimize(model, MLE())
ModeResult with maximized lp of -141.20
36-element Named Array{Float64,1}
A                 │ 
──────────────────┼────────
:scale            │     1.4
Symbol("lsd[1]")  │ 7.23206
Symbol("lsd[2]")  │ 14.6571
Symbol("lsd[3]")  │ 9.90191
Symbol("lsd[4]")  │ 7.91973
Symbol("lsd[5]")  │ 5.67751
Symbol("lsd[6]")  │ 6.80444
Symbol("lsd[7]")  │ 5.82526
Symbol("lsd[8]")  │ 6.97793
Symbol("lsd[9]")  │  14.528
Symbol("lsd[10]") │  9.5812
Symbol("lsd[11]") │ 13.3083
⋮                         ⋮
Symbol("u[7]")    │ 73.7783
Symbol("u[8]")    │ 25.0389
Symbol("u[9]")    │ 30.4734
Symbol("u[10]")   │ 61.6176
Symbol("u[11]")   │  17.547
Symbol("u[12]")   │ 29.3112
Symbol("u[13]")   │ 56.5756
Symbol("u[14]")   │ 93.1553
Symbol("u[15]")   │ 29.0809
Symbol("u[16]")   │ 57.9938
Symbol("u[17]")   │ 27.6348
:σ                │ 2.27592
```
`mle_estimate.values` is a `NamedArray` and to get all the values corresponding to `u[i]` in the array I need to do `getindex(mle_estimate.values, [Symbol("u[$(i)]") for i in 1:17])`. For dynamic `u` array size we might have to follow the roundabout suggested in the slack thread.

It would be great to have a function `get` similar to the one for `chains` object
So that `get(mle_estimate, :u)` returns all the values corresponding to `u[i]` in an array

BlackWingedKing:
@cpfiffer

devmotion:
Maybe it would make sense to not flatten the values? E.g. as a ComponentArray (or ComponentMatrix, to be precise)?

devmotion:
[...]

PR #2269, opened 2024-06-20
Add Base.get method for ModeResult

mhauru:
Answers the feature request from #1417

The return type here is a bit questionable. It currently mimics the `NamedTuples` that MCMCChains returns for `Base.get`, but I also considered returning `NamedArrays` like the `ModeResult.values` field. Opinions on this are welcome.

### DIFF
--- a/src/optimisation/Optimisation.jl
+++ b/src/optimisation/Optimisation.jl
@@ -274,6 +274,35 @@ StatsBase.vcov(m::ModeResult) = inv(StatsBase.informationmatrix(m))
 StatsBase.loglikelihood(m::ModeResult) = m.lp
 
 """
+    Base.get(m::ModeResult, var_symbol::Symbol)
+    Base.get(m::ModeResult, var_symbols::Vector{Symbol})
+
+Return the values of all the variables with the symbol(s) `var_symbol` in the mode result
+`m`. The return value is a `NamedTuple` with `var_symbols` as the key(s).
+"""
+function Base.get(m::ModeResult, var_symbols::Vector{Symbol})
+    log_density = m.f
+    # Get all the variable names in the model. This is the same as the list of keys in
+    # m.values, but they are more convenient to filter when they are VarNames rather than
+    # Symbols.
+    varnames = collect(
+        map(first, Turing.Inference.getparams(log_density.model, log_density.varinfo))
+    )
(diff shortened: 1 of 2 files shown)


test/optimisation/Optimisation.jl
```diff
@@ -4,13 +4,14 @@ using ..Models: gdemo, gdemo_default
...
 using DynamicPPL: DynamicPPL
-using LinearAlgebra: I
+using LinearAlgebra: Diagonal, I
 using Random: Random
 using Optimization
 using Optimization: Optimization
 using OptimizationBBO: OptimizationBBO
 using OptimizationNLopt: OptimizationNLopt
 using OptimizationOptimJL: OptimizationOptimJL
+using ReverseDiff: ReverseDiff
```
mhauru, line 14: This is unrelated to the other changes in this PR. Not having this here before was just a plain bug.

src/optimisation/Optimisation.jl
```diff
@@ -273,6 +273,35 @@ StatsBase.params(m::ModeResult) = StatsBase.coefnames(m)
 StatsBase.vcov(m::ModeResult) = inv(StatsBase.informationmatrix(m))
 StatsBase.loglikelihood(m::ModeResult) = m.lp
 
+"""
+    Base.get(m::ModeResult, var_symbol::Symbol)
+    Base.get(m::ModeResult, var_symbols::Vector{Symbol})
+
+Return the values of all the variables with the symbol(s) `var_symbol` in the mode result
+`m`. The return value is a `NamedTuple` with `var_symbols` as the key(s).
```
devmotion, line 281: Intuitively I would assume that indexing with a `Tuple` might (possibly) return a `NamedTuple`. But returning a `NamedTuple` for `Vector` inputs seems quite surprising, and also means that not even the keys of the `NamedTuple` return type can be inferred.

mhauru: Good point. I changed this so that the second argument can be any iterator type, and made the single symbol version use a singleton tuple. I also stabilised the array types a bit. Let me know if you are happy with this.

mhauru:
As a fun side effect of this PR, I ran into this case of bizarrely slow type inference: https://github.com/JuliaLang/julia/issues/54879

PR #2278, opened 2024-07-02
Version bump 0.33+

mhauru:
Marking this as 0.33.1 for now, but have to still check that there isn't anything breaking here.


mhauru:
I can't see anything breaking here, so I think 0.33.1 is good.

mhauru:
Here's the relevant diff: https://github.com/TuringLang/Turing.jl/compare/v0.33.0...master

Issue #2279, opened 2024-07-04
Make MLE and MAP more robust


itsdfish:
Hello, 

For many models, MLE and MAP return results for local rather than global maxima. Sometimes switching optimizers, or tweaking optimizer options can help. In my experience, a simple solution is to run the optimizer multiple times and select the maximum from the various attempts. Would it be possible to add a second method for MLE and MAP, which allows users to specify the number of attemps? For example, 

```
function maximum_likelihood(model::DynamicPPL.Model, n_reps::Integer, args...; kwargs...)
    best_lp = -Inf 
    mle = estimate_mode(model, MLE(), args...; kwargs...)
    for i in 2:n_reps 
        _mle = estimate_mode(model, MLE(), args...; kwargs...)
        mle = _mle.lp > best_lp ? _mle : mle 
    end
    return mle 
end
```

Here is an example of the problem using a fairly simple model:

# Example 
```
using SequentialSamplingModels
using Random 
using Turing 

Random.seed!(100)

n_samples = 50
rts = rand(ShiftedLogNormal(ν=-1, σ=.8, τ=.3), n_samples)

@model function model(rts; min_rt = minimum(rts))
    ν ~ Normal(-1, 2)
    σ ~ truncated(Normal(.8, 2), 0, Inf)
    τ ~ Uniform(0, min_rt)
    rts ~ ShiftedLogNormal(ν, σ, τ)
    return (;ν, σ, τ)
end


lb = [-1,0,0]
ub = [10, 10, minimum(rts)]

# Generate a MLE estimate.
lps = map(_ -> maximum_likelihood(model(rts); lb, ub).lp, 1:10)

# Generate a MAP estimate.
map_estimate = maximum_a_posteriori(model(rts); lb, ub)
```

# Results 

```
10-element Vector{Float64}:
 -12.226295752323539
 -12.226295752292035
 -80.44912241263312
 -12.226295752293689
 -64.40215460329794
 -12.22629575228706
 -12.226295752286504
 -84.43334167030277
 -76.50164720237467
 -55.60874854749243
```

As you can see, half of the time the algorithm landed on a local maxima. 

If this is something you are willing to support, I can submit a PR.

itsdfish:
Thank you for your reply. If my proposal for `maximum_a_posterior` and `maximum_likelihood` is acceptable, I can submit a PR.

mhauru:
[...]

PR #2280, opened 2024-07-08
Add keyword n_tries to maximum_likelihood and maximum_a_posteriori

itsdfish:
This PR adds a keyword `n_tries` to run `maximum_likelihood` and `maximum_a_posteriori` from multiple starting points to mitigate problems with local maxima (see #2279). I added two unit tests and updated the doc strings. Please let me know if there are any edits you would like me to make.


devmotion:
Generally, if local minima/maxima are problematic - wouldn't it be better to adjust the optimization method (if possible) or turn to global optimization methods? Or use a less naive multi-start optimization method such as MultistartOptimization.jl (https://docs.sciml.ai/Optimization/stable/optimization_packages/multistartoptimization/)? I'm a bit worried that exposing the functionality in the PR in an official API might make it easier for users to miss these potentially better alternatives.

itsdfish:
I was not aware of `MultistartOptimization`. In that case, it would be better to use that and disregard the PR.

mhauru:
Thanks @devmotion, that looks handy; Sorry @itsdfish that I wasn't aware of that when we talked about this.

We could still use the test case you wrote just to check that MultistartOptimization solves it. Or alternatively use your example in the docs. I feel like this is not going to be the last time local maxima come up, and would be handy to have some code to point at and say "you could try doing this".

itsdfish:
@mhauru, no worries! It looks like at least 3 people were unaware of `MultistartOptimization`.  Let me know if you would like me to reporpose  this PR as a test for `MultistartOptimization`. I can update later this week.

mhauru:
If you have the time, I'd be happy to merge an extra test on a model that requires multistart to find the optimum. Might convert it into a paragraph in the docs later.

PR #2283, opened 2024-07-09
Re-enable coverage uploads

### LINKED ISSUE #2282: Fix code test coverage
Code test coverage is at 0% from the icons linked in README, probably due to some configuration issue; let's fix that.

devmotion:
Fixes #2282.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -69,6 +69,7 @@ jobs:
           echo "Test arguments: ${{ matrix.test-args }}"
       - name: (De)activate coverage analysis
         run: echo "COVERAGE=${{ matrix.version == '1' && matrix.os == 'ubuntu-latest' && matrix.num_threads == 2 }}" >> "$GITHUB_ENV"
+        shell: bash
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -67,6 +67,8 @@ jobs:
           echo "Julia version: ${{ matrix.version }}"
           echo "Number of threads: ${{ matrix.num_threads }}"
           echo "Test arguments: ${{ matrix.test-args }}"
+      - name: (De)activate coverage analysis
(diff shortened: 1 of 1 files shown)


.github/workflows/Tests.yml
```diff
@@ -72,16 +72,37 @@ jobs:
         with:
           version: '${{ matrix.version }}'
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v4
+      - uses: julia-actions/cache@v1
+      - uses: julia-actions/julia-buildpkg@v1
+      # TODO: Use julia-actions/julia-runtest when test_args are supported
```
devmotion, line 77: I think it's a quite unfortunate consequence of https://github.com/TuringLang/Turing.jl/pull/2237 that we have to work around the standard tooling in the Julia ecosystem.

.github/workflows/Tests.yml
```diff
@@ -72,16 +72,37 @@ jobs:
...
+      # Custom calls of Pkg.test tend to miss features such as e.g. adjustments for CompatHelper PRs
+      # Ref https://github.com/julia-actions/julia-runtest/pull/73  
+      - name: Call Pkg.test
+        run: julia --color=yes --inline=yes --depwarn=yes --check-bounds=yes --threads=${{ matrix.num_threads }} --project=@. -e 'import Pkg; Pkg.test(; coverage=parse(Bool, ENV["COVERAGE"]), test_args=ARGS)' -- ${{ matrix.test-args }}
         env:
-          cache-name: cache-artifacts
+          COVERAGE: ${{ matrix.version == '1' && matrix.os == 'ubuntu-latest' && matrix.num_threads == 2 }}
```
mhauru, line 83: I tiny detail, but `${{ matrix.version == '1' && matrix.os == 'ubuntu-latest' && matrix.num_threads == 2 }}` could be stored in a variable and thus defined once, making sure the different if-clauses always use the same condition.

mhauru (approved):
I didn't know of the coveralls parallel option, that's really neat and makes this much easier than I thought it would be. Thanks @devmotion!

I had one tiny style point to made, but I'll approve already since it's fine to merge as-is.

PR #2290, opened 2024-07-18
Enabling aqua ambiguity testing for Turing

abhinavsns:
We can test ambiguities only for Turing and not its dependencies.


devmotion:
Indeed, that's what I suggested in https://github.com/TuringLang/Turing.jl/pull/2257#discussion_r163024095. Maybe put it in a testset and maybe only test ambiguities on Julia >= 1.10 (or some other recent Julia versions) since it might not be worth it fixing ambiguities that only show up in old Julia versions (which sometimes also imply old dependencies)?

penelopeysm:
(Rebased against master to check if #2294 works)

PR #2291, opened 2024-07-18
Check that the correct AD backend is being used

### LINKED ISSUE #2235: Tests to check that the correct AD method is actually used
In #2221 I ran into a case where I wanted to write a test that checked if the ADType that the caller specified actually got used (e.g. to check that the argument isn't being accidentally ignored). Would be useful to have some utilities for this. See conversation in https://github.com/TuringLang/Turing.jl/pull/2221/files#r1608405934.

mhauru:
Introduces tests that check that the ADType we specify is actually the one being used. This alleviates my worry that we might mess up passing the ADType around and accidentally fall back on the default.

I've added tests for calling `sample` and `maximum_likelihood/maximum_a_posteriori`. Any other user-facing functions we should check to make sure they, too, use their `adtype` arguments correctly?

Closes #2235

### DIFF
--- a/test/test_utils/ad_utils.jl
+++ b/test/test_utils/ad_utils.jl
@@ -96,7 +96,7 @@ end
 """
     valid_eltypes(context::ADTypeCheckContext)
 
-Return the element types that are valid for the ADType of `context`.
+Return the element types that are valid for the ADType of `context` as a tuple.
 """
 function valid_eltypes(context::ADTypeCheckContext)
     context_at = adtype(context)
@@ -116,7 +116,7 @@ Check that the element types in `vi` are compatible with the ADType of `context`
 
 Throw an `IncompatibleADTypeError` if an incompatible element type is encountered.
 """
-function check_adtype(context::ADTypeCheckContext, vi::DynamicPPL.VarInfo)
+function check_adtype(context::ADTypeCheckContext, vi::DynamicPPL.AbstractVarInfo)
     valids = valid_eltypes(context)
     for val in vi[:]
         valtype = typeof(val)
@@ -182,6 +182,14 @@ end
 function DynamicPPL.dot_tilde_observe(context::ADTypeCheckContext, right, left, vi)
     logp, vi = DynamicPPL.dot_tilde_observe(
         DynamicPPL.childcontext(context), right, left, vi
+    )
+    check_adtype(context, vi)
+    return logp, vi
+end
+
+function DynamicPPL.dot_tilde_observe(context::ADTypeCheckContext, sampler, right, left, vi)
+    logp, vi = DynamicPPL.dot_tilde_observe(
+        DynamicPPL.childcontext(context), sampler, right, left, vi
     )
     check_adtype(context, vi)
     return logp, vi
--- a/test/test_utils/ad_utils.jl
+++ b/test/test_utils/ad_utils.jl
@@ -25,8 +25,8 @@ const eltypes_by_adtype = Dict(
         ReverseDiff.TrackedVecOrMat,
         ReverseDiff.TrackedVector,
     ),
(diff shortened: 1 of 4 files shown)


mhauru:
@torfjelde pointed out in https://github.com/TuringLang/Turing.jl/pull/2221/files#r1608405934 that there's a similarity between this and `DebugContext`, in that they both call their child context as usual and just add an extra call before/after. Thus we could rather implement a more general `CallbackContext`, and unify the implementation of `DebugContext` and `ADTypeCheckContext` under it.

I think this is a great idea, but I chose not to do it in this PR because
1) Would need to give a good think as to what powers the callbacks should have. Are they e.g. allowed to modify `varinfo`?
2) Would want to check whether any other contexts we currently have fit this `CallbackContext` format, and bring those in as well.

mhauru:
>I don't think comparing the types of values in varinfo is particularly elegant, but also I don't see other ways to do it with Contexts.

Agree on lack of elegance, better proposals welcome.

test/test_utils/ad_utils.jl
```diff
@@ -0,0 +1,213 @@
...
+        ReverseDiff.TrackedMatrix,
+        ReverseDiff.TrackedReal,
+        ReverseDiff.TrackedStyle,
+        ReverseDiff.TrackedType,
+        ReverseDiff.TrackedVecOrMat,
+        ReverseDiff.TrackedVector,
+    ),
+    # TODO(mhauru) Zygote.Dual is actually the same as ForwardDiff.Dual, so can't
+    # distinguish between the two.
+    Turing.AutoZygote => (Zygote.Dual,),
```
torfjelde, line 30: Yeah there's no way you can do this here unfortunately 😕 

Possibly crazy idea, but it might make sense to specifically overload the _adjoint_ computation for a given backend which doesn't use types, e.g.
```julia
Zygote.@adjoint function check_adtype(...)
    # Should only be hit if we're using Zygote.jl.
    ...
end
```
? Could do the same with Enzyme.jl and Tapir.jl, both of which would suffer from the same issue

test/test_utils/ad_utils.jl
```diff
@@ -0,0 +1,213 @@
+module ADUtils
+
+import ForwardDiff
+import ReverseDiff
+import Test
+import Tracker
+import Turing
+import Turing: DynamicPPL
+import Zygote
```
torfjelde, line 9: Isn't `using ForwardDiff: ForwardDiff` preferred to `import ForwardDiff`?

torfjelde (commented):
I like the idea very much:) Seems like something we should definitively test. And agree that we shouldn't do the `CallbackContext` in this PR.

I left a comment on a potentially way we can check Zygote.jl, Tapir.jl, and Enzyme.jl too.

test/test_utils/ad_utils.jl
mhauru: I like this, and implemented it. Turns out there's `Zygote.hook` for exactly these sorts of things.

I thought a bit about switching to doing all checks like this, and giving up on the element type approach, but that seemed like it would get more complicated than is worth at this point. We now catch all cases except if one uses ForwardDiff when Zygote is expected, which seems sufficient to me.

test/test_utils/ad_utils.jl
mhauru: Yes, fixed.

mhauru:
The only CI failure is a tiny, trivial loss of code coverage. @torfjelde, are you happy to have this merged as is?

test/test_utils/ad_utils.jl
torfjelde: Very, very nice @mhauru :)

torfjelde (approved):
Very, very nice @mhauru :)

PR #2294, opened 2024-07-29
Add GITHUB_TOKEN permissions for format job

### LINKED ISSUE #2293: JuliaFormatter GHA workflow fails due to insufficient permissions
See e.g. https://github.com/TuringLang/Turing.jl/actions/runs/9989440306/job/28060704792?pr=2290.

I'm fairly certain the answer is to add a permissions block to the job as described e.g. in the README https://github.com/reviewdog/action-suggester

penelopeysm:
Closes #2293


penelopeysm:
@yebai I just learnt that there is one more step for this. According to [these GitHub docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job), for PRs from public forks (such as #2290) the permissions are decreased to read-only unless the "Send write tokens to workflows from pull requests" option is selected in the repo settings.

You could either enable the setting, or if you trust me enough, make me an owner of the organisation and I can do it :)

PR #2299, opened 2024-08-09
Resolve ADTypeCheckContext method ambiguity

mhauru:
Should fix the issue in https://github.com/TuringLang/DynamicPPL.jl/pull/638. I'm a bit bothered by way the method ambiguity did not come up in Turing's own CI, but only in DPPL's. Also, I _think_ this will fix the issue, but might need to run DPPL's tests to see.

### DIFF
--- a/test/test_utils/ad_utils.jl
+++ b/test/test_utils/ad_utils.jl
@@ -1,6 +1,7 @@
 module ADUtils
 
 using ForwardDiff: ForwardDiff
+using Random: Random
 using ReverseDiff: ReverseDiff
 using Test: Test
 using Tracker: Tracker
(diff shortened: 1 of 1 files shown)


devmotion:
I guess these ambiguities arise from https://github.com/TuringLang/DynamicPPL.jl/pull/636 which has not been released yet (and therefore it shows only up in DynamicPPL CI but not in Turing).

mhauru:
Thanks @devmotion, that makes sense.

We seem to have an unfortunate case here where a PR that passed tests (#2291) was merged, and then immediately after merging tests on master started to fail (https://github.com/TuringLang/Turing.jl/actions/runs/10281463329/job/28451089388). That's still the failure in this PR too. It's a concurrency problem of some kind, only appears on multithreaded. Haven't dug into it yet.

mhauru:
Merging this since the test failure is on master too and the loss of coverage seems unrelated, and this is holding up the DPPL PR. Need to get to the bottom of the test failure and how it slipped through in the first place though.

PR #2328, opened 2024-09-23
Replace old Gibbs sampler with the experimental one.

mhauru:
Closes TuringLang/Turing.jl#2318.

Work in progress.

### DIFF
--- a/test/dynamicppl/compiler.jl
+++ b/test/dynamicppl/compiler.jl
@@ -177,7 +177,7 @@ const gdemo_default = gdemo_d()
     end
 
     @testset "sample" begin
-        alg = Gibbs(; m=HMC(0.2, 3), PG(10))
+        alg = Gibbs(; m=HMC(0.2, 3), s=PG(10))
         chn = sample(gdemo_default, alg, 1000)
     end
 


HISTORY.md
```diff
@@ -1,3 +1,19 @@
...
+that the internals of it having been completely rewritten in a very different manner, there
+may be accidental breakage that we haven't anticipated. Please report any you find.
+
+`GibbsConditional` has also been removed. It was never very user-facing, but it was exported, so technically this is breaking.
+
+The old Gibbs constructor relied on being called with several subsamplers, and each of the constructors of the subsamplers would take as arguments the symbols for the variables that they are to sample, e.g. `Gibbs(HMC(:x), MH(:y))`. This constructor has been deprecated, and will be removed in the future. The new constructor works by assigning samplers to either symbols or `VarNames`, e.g. `Gibbs(; x=HMC(), y=MH())` or `Gibbs(@varname(x) => HMC(), @varname(y) => MH())`. This allows more granular specification of which sampler to use for which variable.
+
+Likewise, the old constructor for calling one subsampler more often than another, `Gibbs((HMC(:x), 2), (MH(:y), 1))` has been deprecated. The new way to achieve this effect is to list the same sampler multiple times, e.g. as `hmc = HMC(); mh = MH(); Gibbs(@varname(x) => hmc, @varname(x) => hmc, @varname(y) => mh)`.
```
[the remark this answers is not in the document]
mhauru: Yes. We had a chat about a closely related issue with @torfjelde too, I'll rework the interface around this a bit.

mhauru:
@torfjelde, if you have a moment to take a look at the one remaining test failure, would be interested in your thoughts. We are sampling for a model with two vector variables, `m` and `z`, and we seem to somehow end up with a case where there's a `VarInfo` with only `z` in it, but the sampler is looking for `m` too. I wonder if it's something about the interaction between particle sampling with Libtask and how the new Gibbs does things with the local varinfos. The test that fails is this one:
```julia
    @testset "dynamic model" begin
        @model function imm(y, alpha, ::Type{M}=Vector{Float64}) where {M}
            N = length(y)
            rpm = DirichletProcess(alpha)

            z = zeros(Int, N)
            cluster_counts = zeros(Int, N)
            fill!(cluster_counts, 0)

            for i in 1:N
                z[i] ~ ChineseRestaurantProcess(rpm, cluster_counts)
                cluster_counts[z[i]] += 1
            end

            Kmax = findlast(!iszero, cluster_counts)
            m = M(undef, Kmax)
            for k in 1:Kmax
                m[k] ~ Normal(1.0, 1.0)
            end
        end
        model = imm(Random.randn(100), 1.0)
        # https://github.com/TuringLang/Turing.jl/issues/1725
        # sample(model, Gibbs(MH(:z), HMC(0.01, 4, :m)), 100);
        sample(model, Gibbs(; z=PG(10), m=HMC(0.01, 4; adtype=adbackend)), 100)
    end
```

torfjelde:
Will have a look at this in a bit @mhauru (just need to do some grocery shopping :grimacing: )

mhauru:
Collecting links to old relevant PRs so I don't have to look for them again: #2231, #2099

torfjelde:
Think I found the error: if the number of `m` increases, say, from `length(m) = 2` to `length(m) = 3` during the `PG` step, then the lines

https://github.com/TuringLang/Turing.jl/blob/6f9679ac659210276fd1cb9acdc2a728573699e3/src/mcmc/gibbs.jl#L57-L65

doesn't hit the `Gibbs` branch since `@varname(m[3])` is _not_ present in the `GibbsContext` :confused:

torfjelde:
> doesn't hit the Gibbs branch since @varname(m[3]) is not present in the GibbsContext

I'm a bit uncertain how we should best handle this @yebai @mhauru 

The first partially viable idea that comes to mind is to `subset` the `varinfo` to make sure that it only contains the correct variables. If we do this, then `m[3]` will just be "ignored" (in the varinfos) until we're actually sampling the `m` variables, in which case it would be captured correctly.

_But_ this would not quite be equivalent to the current implementation of `Gibbs`, which, AFAIK, keeps the very first occurence of `m` around rather than resampling everytime. And naively, I would expect this to be incorrect.

Another way is to explicitly add the `varinfos` to the `GibbsContext` itself, and then, when we encounter a value that should in fact go into a different `varinfo`, we add it there. But this has a few issues:
1. Requires the `VarInfo` to be mutable.
2. Requires the `VarInfo` to have a container that can keep the new incoming value `m[3]`.
3. Implementation of `Gibbs` does end up being more complicated than the current approach. However, it might be worth it.

Thoughts?

torfjelde:
This does however complicate the new Gibbs sampling procedure quite drastically :confused: 

And it makes me bring up a question I really didn't think I'd be asking: is it then actually preferable to the current `Gibbs` with keeping it all in a single `VarInfo` with a flag to specify whether it should be sampled or not? :grimacing:

I guess we should first have a go at implementing this for the new `Gibbs` and then we can see :+1: 

Another point to add to the conversation that @mhauru brought to my attention the other day: we also want to support stuff like `Gibbs(@varname(m) => NUTS(), @varname(m) => HMC())`, i.e. multiple samplers targeting the same variables. This adds a few "complications" (beyond addressing the growing model problem discussed above):
1. Need to determine which `varinfo` to pick from `varinfos` based on the varnames present / targeted.
2. A _naive_ implementation will result in duplicated entries in `varinfos`. We can however address this if we really feel like it's worth it, so probably a non-issue atm.

So all in all, immediate things we need to address with Gibbs:
1. Support changing dimensions.
2. Support picking a `varinfo` to condition on based on the varnames present rather than based on `===`.

mhauru:
I've been trying to think of a way to fix this, that would also fix the problem where different Gibbs subsamplers can't sample the same variables (e.g. you can't first sample `x` and `y` using one sampler, and then `y` and `z` with a different one). My best thought at the moment is the following design:

1. There is only one, global `VarInfo`, call it `vi`.
2. `make_conditional` takes that `vi` and a list of `VarName`s that the current subsampler samples. It hijacks the tilde pipeline to condition all other variables to their current values in `vi`.
3. `vi` may have some variables linked, some not.
4. Every time we call a subsampler we can hand it `vi` as the `VarInfo.` It won’t mess with any of the variables it’s not supposed to touch, because the tilde pipeline hijack from point 2.

Point 3. is maybe undesirable, but I think it’s minor compared to all the `Selector`/`gibbsid` stuff, which we would still get rid of.

The only problem I see with this is combining the local state from the previous iteration of the current subsampler with the global `vi`. Somehow we would need to join up-to-date information from the global `vi` with state-information from the previous iteration, specific to this subsampler. The right way to do this depends on the state, which is a different type of object for different subsamplers. EDIT: Actually, maybe this is okay, because we seem to already assume that every state object has a field called `state.vi` , we could just reset that.

The great benefit of sticking to one, global `VarInfo` is never having to worry about moving data between the local `VarInfo`s. That would have to happen in both cases, when a new variable is introduced by one sampler (the failing test in this PR) and when two samplers sample the same variable. It sounds like a pain to implement.

mhauru:
I can imagine two different philosophies to implementing a Gibbs sampler:
1. Every subsampler is doing its own sampling process on a low-dimensional model (a conditioned version of the full model), independent of the others. The logprobability function it's sampling from just keeps changing between iterations, because the other variables change and thus the conditioned model changes, but otherwise it's blind to the existence of the variables it isn't sampling. This is what the new Gibbs sampler does.
2. Every subsampler is working with the same, full model, with all the variables, but only makes the changes to a subset of those variables. It still "sees" the whole model. This is what the old Gibbs sampler did.

My above proposal would essentially be doing 2., but using code that's very much like the new sampler, where the information about which sampler modifies which variables is in the sampler/`GibbsContext`, and not in `VarInfo` like it was in the old Gibbs.

The reason I'm leaning towards 2. is that 1. seems to run to some fundamental issues in cases where either

* Variables appear and disappear based on values of other variables,
* Two samplers want to modify the value of the same variable.

Both of those situations quite deeply violate the idea that the different subsamplers can operate mostly independently of each other.

Any thoughts very welcome, I'm still very much trying to understand the landscape of the problem.

torfjelde:
> Every subsampler is working with the same, full model, with all the variables, but only makes the changes to a subset of those variables. It still "sees" the whole model. This is what the old Gibbs sampler did.

Overall, I'm also in favour of this @mhauru :+1: I think your reasoning is solid here.

The only other "option" I'm seeing is to keep track of which variables correpond to which varinfos (with each varinfo only containing the relevant information), but then we're effectively just re-implementing a lot of the functionality that is already provided in `varinfo` :confused:

The only "issue" is that this does mean we have to support this "link / transform only part of the `varinfo`, which does mean we need something "equivalent" to all the `getindex(varinfo, sampler)` stuff that we've been trying to move away from (since we need a way to extract the vectorized part relevant only for the specific sampler we're going to use in that particular step) :confused:

Doulby however, I think we can make this much nicer than the current approach by simply making all these `getindex(varinfo, sampler)` instead take the relevant varnames instead of the samplers themselves, which should make it all less painful.

But yeah, don't see how we can take approach (1) in a "nice" way, and so I'm also in favour of just trying to make (2) as painless as possible to maintain.

mhauru:
Thanks for the comments both, this is very helpful.

>Doulby however, I think we can make this much nicer than the current approach by simply making all these getindex(varinfo, sampler) instead take the relevant varnames instead of the samplers themselves, which should make it all less painful.

Yeah, I think this is the way to go.

mhauru:
I made some progress, but it's not done yet, maybe hold off reviewing for now. Also depends on https://github.com/TuringLang/DynamicPPL.jl/pull/694.

penelopeysm (commented):
I can't suggest changes because the file isn't changed, but I think these lines should be removed.

https://github.com/TuringLang/Turing.jl/blob/24e68701b01695bffe69eda9e948e910c1ae2996/src/mcmc/mh.jl#L23

https://github.com/TuringLang/Turing.jl/blob/24e68701b01695bffe69eda9e948e910c1ae2996/src/mcmc/mh.jl#L44-L51

(I also checked for the other samplers, but MH is the only one that suffers from the success of having very thorough documentation.)

mhauru:
Nice catch @penelopeysm, removed.

torfjelde:
Ping me when you want me to have a look @mhauru :)

mhauru:
All the tests now pass for ReverseDiff. Mooncake fails in some cases because of https://github.com/compintell/Mooncake.jl/issues/315. There are also some ForwardDiff specific failures, that I may have worked around now, let's see with the latest CI run.

@torfjelde, once you've recovered it could be a good time for you to take another look

One thing I probably still want to change is the names and arguments of `recompute_logprob!!` and `reset_state!!`. Would be interested in other people's opinions on what the right interface here is.

mhauru:
@torfjelde, I think I made all the changes we discussed on the call. I may still move the `setparams!!` methods to a different file, but that's details. I _hope_ tests will pass with ReverseDiff, but I didn't have time to run the whole suite locally before having to go, so let's see.

mhauru:
All the tests have passed on at least one platform, so I assume they'll all eventually pass. @torfjelde, wanna take another look?

mhauru:
Some optimisation is clearly required. The following runs in sub 10s with the Gibbs sampler, takes more than a minute with new one:
```julia
julia> module Benchmark1
       using Turing, Test, Random
       using Turing.RandomMeasures: ChineseRestaurantProcess, DirichletProcess
       import ReverseDiff

       adbackend = AutoReverseDiff()
       @testset "dynamic model" begin
           @model function imm(y, alpha, ::Type{M}=Vector{Float64}) where {M}
               N = length(y)
               rpm = DirichletProcess(alpha)

               z = zeros(Int, N)
               cluster_counts = zeros(Int, N)
               fill!(cluster_counts, 0)

               for i in 1:N
                   z[i] ~ ChineseRestaurantProcess(rpm, cluster_counts)
                   cluster_counts[z[i]] += 1
               end

               Kmax = findlast(!iszero, cluster_counts)
               m = M(undef, Kmax)
               for k in 1:Kmax
                   m[k] ~ Normal(1.0, 1.0)
               end
           end
           model = imm(Random.randn(100), 1.0)
           # https://github.com/TuringLang/Turing.jl/issues/1725
           # sample(model, Gibbs(MH(:z), HMC(0.01, 4, :m)), 100);
           sample(model, Gibbs(; z=PG(10), m=HMC(0.01, 4; adtype=adbackend)), 200)
       end

       end
```

Here's another one. With the new sampler:
```julia
julia> module Benchmark2
       using Turing, Random
       import ReverseDiff

       adbackend = AutoReverseDiff()

       @model function gdemo(x, y)
           s ~ InverseGamma(2, 3)
           m ~ Normal(0, sqrt(s))
           x ~ Normal(m, sqrt(s))
           y ~ Normal(m, sqrt(s))
           return s, m
       end

       Random.seed!(100)
       alg = Gibbs(; s=CSMC(15), m=HMC(0.2, 4; adtype=adbackend))
       @time sample(gdemo(1.5, 2.0), alg, 1_000)
       end
WARNING: replacing module Benchmark2.
 38.251275 seconds (13.16 M allocations: 965.908 MiB, 0.70% gc time, 7.68% compilation time)
Main.Benchmark2
```
With the old Gibbs sampler:
```julia
 16.980426 seconds (13.63 M allocations: 1.046 GiB, 1.01% gc time, 14.13% compilation time)
```

There are some obvious optimisations to do, I'll see where they get us.

mhauru:
@torfjelde and I had a conversation about performance and optimisation thereof, and concluded that we should do none of it now. Tor's argument was that the only significant cases are heavy models with lots of variables, in which case all the cost is going to be in the gradient/logprob evaluations anyway, and your component samplers will probably do lots of them. So as long as we have type stability in the tilde pipeline, everything else is details.

Based on that I
1) Added some explicit type stability tests to the Gibbs test suite.
2) Removed `recompute_logprob!!`. We call `setparams!!` at every iteration, and that one requires recomputing the logprob anyway.
3) Removed the small of piece of machinery that was in place to give the user the option to _not_ recompute logprob in some cases. This would have gotten significantly more complicated to do correctly for `setparams!!`, now that we can have multiple samplers affecting the same variables.

So in short, we are recomputing the logprob at every iteration of every component sampler in the `setparams!!` call, and we've decided that this doesn't matter enough at the moment to warrant optimising.

Tests pass, ready for review again @torfjelde.

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+            kwargs...,
+        )
+        new_vi_local = varinfo(new_state_local)
+        # TODO(mhauru) Remove the below loop once samplers no longer depend on selectors.
+        # For some reason not having this in place was causing trouble for ESS, but not for
+        # other samplers. I didn't get to the bottom of it.
+        for vn in keys(new_vi_local)
+            DynamicPPL.setgid!(new_vi_local, sampler_local.selector, vn)
         end
+        # This merges in any new variables that were introduced during the step, but that
```
torfjelde, line 383:
```suggestion
        # Merge in any new variables that were introduced during the step, but that
```

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+        # were not in the domain of the current sampler.
+        vi = merge(vi, context_local.global_varinfo[])
+        # This merges the new values for all the variables sampled by the current sampler.
```
torfjelde, line 386:
```suggestion
        # Merge the new values for all the variables sampled by the current sampler.
```

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+            # FIXME: This will cause issues if the sampler expects initial params in unconstrained space.
+            # This is not the case for any samplers in Turing.jl, but will be for external samplers, etc.
+            initial_params=initial_params_local,
+            kwargs...,
+        )
+        new_vi_local = varinfo(new_state_local)
+        # TODO(mhauru) Remove the below loop once samplers no longer depend on selectors.
+        # For some reason not having this in place was causing trouble for ESS, but not for
+        # other samplers. I didn't get to the bottom of it.
+        for vn in keys(new_vi_local)
```
torfjelde, line 380: Is this just because you haven't updated the `ESS` sampler yet? Looking at it's constructor, it has one that takes in a `Symbol`, thus it will have a `space` that is only that symbol(s). With the new `Gibbs`, none of the `InferenceAlgorithm`s should have a symbol attached, right?

torfjelde: Ah, I see that we aren't actually altering this 😕 

My suggestion here is to do the following:
1. Make a function called `drop_space(::InferenceAlgorithm)` which we implement for our existing samplers. This should a new instance of the alg but _without_ any `space`.
2. Call `drop_space` on the algs provided to `Gibbs`.

If we do this, then we can keep the old interface still functioning but without relying on `space` at all internally:)

torfjelde: For example, for `ESS`, we just do
```julia
drop_space(::ESS) = ESS()
```
which would then drop the `space` in `ESS{space}`.

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
-    # Compute initial transition and state.
-    transition = Transition(model, vi)
-    state = GibbsState(vi, samplers, states)
+function setparams_varinfo!!(
+    model::DynamicPPL.Model, sampler::Sampler{<:MH}, state::VarInfo, params::AbstractVarInfo
+)
+    # The state is already a VarInfo, so we can just return `params`, but first we need to
+    # update its logprob.
+    # TODO(mhauru) Is this the right context to use?
+    return last(DynamicPPL.evaluate!!(model, params, DynamicPPL.leafcontext(model.context)))
```
torfjelde, line 441: Yep, that's correct. But maybe add a comment as to why it's okay? E.g.
```suggestion
    # NOTE: Using `leafcontext(model.context)` here is a no-op, as it will be concatenated
    # with `model.context` before hitting `model.f`.
    return last(DynamicPPL.evaluate!!(model, params, DynamicPPL.leafcontext(model.context)))
```

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
-    return transition, state
+function setparams_varinfo!!(
+    model::DynamicPPL.Model,
+    sampler::Sampler{<:ESS},
+    state::VarInfo,
+    params::AbstractVarInfo,
+)
+    # The state is already a VarInfo, so we can just return `params`, but first we need to
+    # update its logprob.
+    # TODO(mhauru) Is this the right context to use?
```
torfjelde, line 452: Same here

torfjelde: But yes, using `LikelihoodContext` here is correct 👍

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
 )
-    # Iterate through each of the samplers.
-    vi = state.vi
-    samplers = state.samplers
-    states = map(samplers, spl.alg.iterations, state.states) do _sampler, iteration, _state
-        # Recompute `vi.logp` if needed.
-        if _sampler.selector.rerun
-            vi = last(DynamicPPL.evaluate!!(model, rng, vi, _sampler))
-        end
+    logdensity = DynamicPPL.setmodel(state.logdensity, model, sampler.alg.adtype)
```
torfjelde, line 459: Does `alg.adtype` always exist here?

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+    # update its logprob.
+    # TODO(mhauru) Is this the right context to use?
+    return last(DynamicPPL.evaluate!!(model, params, DynamicPPL.LikelihoodContext()))
 end
 
-# Subsequent steps
-function AbstractMCMC.step(
-    rng::AbstractRNG, model::Model, spl::Sampler{<:Gibbs}, state::GibbsState; kwargs...
+function setparams_varinfo!!(
+    model::DynamicPPL.Model, sampler::Sampler, state::TuringState, params::AbstractVarInfo
```
torfjelde, line 457: Is there a reaosn why we're not specializing on `Sampler{<:T}` here?

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+    new_inner_state = setparams_varinfo!!(
+        AbstractMCMC.LogDensityModel(logdensity), sampler, state.state, params
+    )
+    return TuringState(new_inner_state, logdensity)
+end
 
-        # Update state of current sampler with updated `VarInfo` object.
-        current_state = gibbs_state(model, _sampler, _state, vi)
+function setparams_varinfo!!(
+    model::DynamicPPL.Model, sampler::Sampler, state::HMCState, params::AbstractVarInfo
```
torfjelde, line 467: Same here; no specialization on `Sampler{<:T}`?

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+    z = state.z
+    resize!(z.θ, length(θ_new))
+    z.θ .= θ_new
+    return HMCState(params, state.i, state.kernel, hamiltonian, z, state.adaptor)
+end
 
-        # Update `VarInfo` object.
-        vi = gibbs_varinfo(model, _sampler, newstate)
+function setparams_varinfo!!(
+    model::DynamicPPL.Model, sampler::Sampler, state::PGState, params::AbstractVarInfo
```
torfjelde, line 481: Same here; no specialization?

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+    global_vi;
+    kwargs...,
+)
+    # Construct the conditional model and the varinfo that this sampler should use.
+    model_local, context_local = make_conditional(model, varnames_local, global_vi)
+    varinfo_local = subset(global_vi, varnames_local)
+    # TODO(mhauru) Remove the below loop once samplers no longer depend on selectors.
+    # For some reason not having this in place was causing trouble for ESS, but not for
+    # other samplers. I didn't get to the bottom of it.
+    for vn in keys(varinfo_local)
```
torfjelde, line 501: Should be fixed after `drop_space`

test/mcmc/gibbs.jl
```diff
@@ -1,66 +1,199 @@
...
+        @test sample(gdemo_default, g, N) isa MCMCChains.Chains
+    end
 
-        _, state = AbstractMCMC.step(Random.default_rng(), gdemo_default, g)
-        @test state.samplers[1].selector != g.selector
-        @test state.samplers[2].selector != g.selector
-        @test state.samplers[1].selector != state.samplers[2].selector
+    @testset "Gibbs constructors" begin
+        N = 10
+        s1 = Gibbs((@varname(s), @varname(m)) => HMC(0.1, 5, :s, :m; adtype=adbackend))
```
torfjelde, line 151: Maybe a comment on what these cases cover?

test/mcmc/gibbs.jl
```diff
@@ -135,10 +281,271 @@ using Turing.RandomMeasures: ChineseRestaurantProcess, DirichletProcess
...
+    end
+
+    @testset "dynamic model with dot tilde" begin
+        @model function dynamic_model_with_dot_tilde(num_zs=10)
+            z = Vector(undef, num_zs)
+            z .~ Exponential(1.0)
+            num_ms = Int(round(sum(z)))
+            m = Vector(undef, num_ms)
+            return m .~ Normal(1.0, 1.0)
+        end
```
torfjelde, line 297: Maybe add typing for this?
```suggestion
        @model function dynamic_model_with_dot_tilde(num_zs=10, ::Type{TV}=Vector{Float64}) where {TV}
            z = TV(undef, num_zs)
            z .~ Exponential(1.0)
            num_ms = Int(round(sum(z)))
            m = TV(undef, num_ms)
            return m .~ Normal(1.0, 1.0)
        end
```

torfjelde: Also, is tehre a reason why you're doing `Exponential` and converting rather than just doing something like `Poisson`, i.e. discrete distribution?

test/mcmc/gibbs.jl
```diff
@@ -135,10 +281,271 @@ using Turing.RandomMeasures: ChineseRestaurantProcess, DirichletProcess
...
+        model = dynamic_model_with_dot_tilde()
+        # TODO(mhauru) This is broken because of
+        # https://github.com/TuringLang/DynamicPPL.jl/issues/700.
+        @test_broken (
+            sample(
+                model,
+                Gibbs(; z=NUTS(; adtype=adbackend), m=HMC(0.01, 4; adtype=adbackend)),
```
torfjelde, line 304: If you do discrete `z`, then `NUTS` won't work here (but it also won't do the correct thing when you do continuous + `round`, though it'll probably still run)

torfjelde: The `round` ruins gradient info

torfjelde (commented):
Sorry for the delay! Took a bit to get through it properly.

A few "major" things:
1. I think the correct thing to do is to use the `GibbsContext` as the very last context before we hit the `leafcontext`, i.e. it has to be an immediate parent of a leaf. See comments on relevant code for more info.
2. I think I have a solution to the `setgid!!` stuff that you've commented on.

src/mcmc/gibbs.jl
mhauru: I like the idea of `drop_space`. Will have to see that that doesn't cause trouble with e.g. MH, which I think uses spaces to specify proposal distributions, or something like that. Not having to edit those sorts of internals of samplers in the same PR was the reason why I tried to just ignore, rather than suppress, the old space `Selector` mechanism.

src/mcmc/gibbs.jl
mhauru: Nope, fixed.

src/mcmc/gibbs.jl
mhauru: Now that the argument is type bound to `Sampler{<:ExternalSampler}`, yes.

src/mcmc/gibbs.jl
mhauru: Fixed.

src/mcmc/gibbs.jl
mhauru: Fixed.

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+function DynamicPPL.tilde_assume(context::GibbsContext, right, vn, vi)
+    if is_target_varname(context, vn)
+        # Fall back to the default behavior.
+        return DynamicPPL.tilde_assume(DynamicPPL.childcontext(context), right, vn, vi)
+    elseif has_conditioned_gibbs(context, vn)
+        # Short-circuit the tilde assume if `vn` is present in `context`.
+        value = get_conditioned_gibbs(context, vn)
+        # TODO(mhauru) Is the call to logpdf correct if context.context is not
+        # DefaultContext?
+        return value, logpdf(right, value), vi
```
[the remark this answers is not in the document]
mhauru: I'm trying to think of a way to avoid inserting contexts in the middle of the context stack. Could we instead do something like
```julia
child = childcontext(context)
while child isa SamplingContext
    child =  childcontext(child)
end
return DynamicPPL.tilde_assume(
            rng, child, right, vn, vi
        )
```
? So drop the `sampler` and any `SamplingContext`, and then call the usual `tilde_assume` pipeline. I'm not sure how the `sampler` argument and `SamplingContext` interact, but might also be necessary to go through the whole context stack and remove `SamplingContexts` anywhere they appear, not just at the top of the stack.

test/mcmc/gibbs.jl
```diff
@@ -42,6 +43,82 @@ const DEMO_MODELS_WITHOUT_DOT_ASSUME = Union{
...
+                    if !is_target && typed
+                        @inferred Turing.Inference.get_conditioned_gibbs(ctx, subkeys)
+                    end
+                end
+
+                # Check that evaluate!! and the result it returns are type stable.
+                conditioned_model = DynamicPPL.contextualize(model, ctx)
+                _, post_eval_varinfo = @inferred DynamicPPL.evaluate!!(
+                    conditioned_model, local_varinfo
+                )
```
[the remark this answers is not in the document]
mhauru: Good idea. With that change the `typed=false` test was no longer checking anything, so I removed it.

test/mcmc/gibbs.jl
mhauru: Done

test/mcmc/gibbs.jl
mhauru: Now using Poisson and PG

src/mcmc/abstractmcmc.jl
```diff
@@ -27,6 +27,9 @@ function varinfo(state::TuringState)
     # TODO: Do we need to link here first?
     return DynamicPPL.unflatten(varinfo_from_logdensityfn(state.logdensity), θ)
 end
+varinfo(state::AbstractVarInfo) = state
+# TODO(mhauru) Could we have a type bound on the argument below, for documentation purposes?
+varinfo(state) = state.vi
```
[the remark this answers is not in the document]
mhauru: Not quite, but I think
```julia
varinfo(state::HMCState) = state.vi
varinfo(state::PGState) = state.vi
varinfo(state::GibbsState) = state.vi
```
does. Using that now. When I originally left that comment I had just copied that line from the old code and didn't understand it.

src/mcmc/gibbs.jl
```diff
@@ -1,3 +1,23 @@
...
+isgibbscomponent(spl::Sampler) = isgibbscomponent(spl.alg)
+
+isgibbscomponent(::ESS) = true
+isgibbscomponent(::HMC) = true
+isgibbscomponent(::HMCDA) = true
+isgibbscomponent(::NUTS) = true
+isgibbscomponent(::MH) = true
+isgibbscomponent(::PG) = true
+isgibbscomponent(::AdvancedHMC.HMC) = true
+isgibbscomponent(::AdvancedMH.MetropolisHastings) = true
```
torfjelde, line 19: Maybe put these and the `ExternalSampler` definition next to each other since these are externalsamplers?

src/mcmc/gibbs.jl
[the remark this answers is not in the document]
torfjelde: Oh I was thinking more of doing the following:

```julia
function make_conditional(
    model::DynamicPPL.Model, target_variables::AbstractVector{<:VarName}, varinfo
)
    # Insert the `GibbsContext` just before the leaf.
    # 1. Extract the `leafcontext` from `model` and wrap in `GibbsContext`.
    gibbs_context_inner = GibbsContext(target_variables, Ref(varinfo), leafcontext(model.context))
    # 2. Set the leaf context to be the `GibbsContext` wrapping `leafcontext(model.context)`.
    gibbs_context = DynamicPPL.setleafcontext(model.context, gibbs_context_inner)
    return DynamicPPL.contextualize(model, gibbs_context), gibbs_context
end
```

You could also accompany this with something like

```julia
function GibbsContext(target_variables, varinfo_ref, context)
    # Check that the provided context is fact a leaf.
    DynamicPPL.NodeTrait(context) isa IsLeaf && throw(ArgumentError("`GibbsContext` can only wrap a leaf context"))
    ...
end
```

in the constructor.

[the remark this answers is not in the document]
mhauru: Is there a reason why that's preferable to what I have implemented, which is this:
```julia
    elseif has_conditioned_gibbs(context, vn)
        child = DynamicPPL.childcontext(context)
        if child isa SamplingContext
            # TODO(mhauru) Would it ever be valid to have a SamplingContext as the child?
            # We could just raise a warning, or optionally go down the stack of contexts
            # skipping all SamplingContexts. The erroring is being conservative.
            error("GibbsContext has a SamplingContext as its child.")
        end
        value, lp, _ = DynamicPPL.tilde_assume(
            child, right, vn, get_global_varinfo(context)
        )
        value, lp, vi
    else
```
What I like about this solution, is that it follows more the usual pattern of "do something and call the tilde pipeline on the childcontext". It also avoids inserting contexts in between other contexts, which feels less risky. For instance, if `GibbsContext` goes through the context stack and inserts it right above the leaf, then if any other context also needs to directly above the leaf for whatever reason, that's going to cause trouble, possibly the silent unexpected behavior sort of trouble. The above solution rather opts for the "do the simple, natural thing and error loudly if it might go wrong".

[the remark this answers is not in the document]
torfjelde: With your suggestion we'll have to do remove _all_ `SamplingContext`s in the context stack, not just the `child` as the `SamplingContext` might be further down the stack.

> then if any other context also needs to directly above the leaf for whatever reason, that's going to cause trouble, possibly the silent unexpected behavior sort of trouble.

Making the constructor of `GibbsContext` error if we give it a context whichi sn't a leaf would address this, no?

Assuming we address the above by issue (i.e. look at entire stack rather than just `child`), you'll have to put that check in every tilde-statement. I was thinking it would be better to just deal with it _once_ in the "beginning" and then you're good 🤷 I'm also a bit worried about passing the `global_varinfo` into a pipeline which expects the "local" varinfo (I can't think of any cases right now that would cause issues here, _but_ seems non-ideal to me).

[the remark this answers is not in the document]
mhauru: > With your suggestion we'll have to do remove all SamplingContexts in the context stack, not just the child as the SamplingContext might be further down the stack.

Yeah, that's true. Or rather I might error if any `SamplingContext`s are found in the stack below. There's no valid reason to have one, right?

> Making the constructor of GibbsContext error if we give it a context whichi sn't a leaf would address this, no?

You still couldn't combine `GibbsContext` with any other context that depends on particularly positioning in the context stack, such as another context that would also always need to be at the very bottom, right above the leaf.

> Assuming we address the above by issue (i.e. look at entire stack rather than just child), you'll have to put that check in every tilde-statement.

I could (and should) move the check to constructor time, when we insert the GibbsContext into the stack. Or are you worried that something might mess with the context stack after the GibbsContext has been created? Not sure if this could happen, given the `GibbsContext ` is created within the call to `step`.

> I'm also a bit worried about passing the global_varinfo into a pipeline which expects the "local" varinfo

On the other hand, I like the fact that in my solution we call `tilde_assume` with some varinfo and the child context. Manually putting in a call to `logpdf` feels a bit like its duplicating code from much lower in down in DPPL, but at a much higher level in the sampler implementation. There's a certain neatness to saying "we call `tilde_assume` with the child context, we only vary whether we do the call with the `sampler` and the local varinfo, or without the sampler and with the global varinfo".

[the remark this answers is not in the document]
torfjelde: > Manually putting in a call to logpdf feels a bit like its duplicating code from much lower in down in DPPL, but at a much higher level in the sampler implementation. 

So if we do the "parent to a leaf" approach, then you could still just defer evaluation to the child context:) In fact, that was part of the reason why I wanted to do this.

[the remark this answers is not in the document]
torfjelde: > Or rather I might error if any SamplingContexts are found in the stack below. There's no valid reason to have one, right?

Hmm, I _think_ so, but can't say 100%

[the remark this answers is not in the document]
mhauru: Seems this boils down to which one is more likely, that we'll need to have a SamplingContext under a GibbsContext, or that another context is going to be upset by GibbsContext insisting on inserting itself right above the leaf.

I'm parking this for a moment though, because I currently have a more pressing issue where I seem to have lost type stability, and I don't know why...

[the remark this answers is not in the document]
mhauru: I went with your solution in the end, because it seems more fail-safe: You don't need to watch for modifications to the whole context stack, just the childcontext. We can revisit if this ever becomes prohibitively strict.

Fun fact: Using the wrong context here, like we were originally doing by essentially hard-coding DefaultContext, was masking a bug in `setparams_varinfo!!` for ESS. Unmasking the bug caused ESS to go into an infinite loop. You didn't need to know this, but I needed to tell someone, for therapeutic reasons, because it took me a day to figure this out.

src/mcmc/gibbs.jl
mhauru: Done.

src/mcmc/gibbs.jl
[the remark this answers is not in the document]
mhauru: I think this should now be fixed, but leaving this thread open until a) you've had a look at the new implementation, b) I've added a test that would have caught this.

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+    rng::Random.AbstractRNG, context::GibbsContext, sampler, right, vn, vi
+)
+    # See comment in the above, rng-less version of this method for an explanation.
+    if is_target_varname(context, vn)
+        return DynamicPPL.tilde_assume(
+            rng, DynamicPPL.childcontext(context), sampler, right, vn, vi
+        )
+    elseif has_conditioned_gibbs(context, vn)
+        value = get_conditioned_gibbs(context, vn)
+        # TODO(mhauru) As above, is logpdf correct if context.context is not DefaultContext?
```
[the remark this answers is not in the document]
mhauru: Done

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+# Like the above tilde_assume methods, but with dot_tilde_assume and broadcasting of logpdf.
+# See comments there for more details.
+function DynamicPPL.dot_tilde_assume(context::GibbsContext, right, left, vns, vi)
+    if is_target_varname(context, vns)
+        return DynamicPPL.dot_tilde_assume(
+            DynamicPPL.childcontext(context), right, left, vns, vi
+        )
+    elseif has_conditioned_gibbs(context, vns)
+        value = reconstruct_getvalue(right, get_conditioned_gibbs(context, vns))
+        # TODO(mhauru) As above, is logpdf correct if context.context is not DefaultContext?
```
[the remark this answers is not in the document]
mhauru: Done

src/mcmc/gibbs.jl
```diff
@@ -1,251 +1,528 @@
...
+function DynamicPPL.dot_tilde_assume(
+    rng::Random.AbstractRNG, context::GibbsContext, sampler, right, left, vns, vi
+)
+    if is_target_varname(context, vns)
+        return DynamicPPL.dot_tilde_assume(
+            rng, DynamicPPL.childcontext(context), sampler, right, left, vns, vi
+        )
+    elseif has_conditioned_gibbs(context, vns)
+        value = reconstruct_getvalue(right, get_conditioned_gibbs(context, vns))
+        # TODO(mhauru) As above, is logpdf correct if context.context is not DefaultContext?
```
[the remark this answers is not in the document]
mhauru: Done

mhauru:
Tests are passing again, @torfjelde want to take another look? Only things left on my mind are more tests (see above) and Hong's comment about the clunky interface of repeat samplers.

HISTORY.md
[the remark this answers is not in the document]
mhauru: I introduced a wrapper sampler: `Gibbs(@varname(x) => RepeatSampler(HMC(0.01, 4), 2), @varname(y) => MH())`

torfjelde:
Will do buddio!

mhauru:
Maybe give it a moment to see that I get tests to pass again. I messed a bit with the constructors.

src/mcmc/gibbs.jl
```diff
@@ -11,22 +34,29 @@
...
+`VNs` is a a tuple of symbols for `VarName`s that the current component
+sampler is sampling. For those `VarName`s, `GibbsContext` will just pass `tilde_assume`
+calls to its child context. For other variables, their values will be fixed to the values
+they have in `global_varinfo`.
+
+The naive implementation of `GibbsContext` would simply have a field `target_varnames` that
+would be a collection of `VarName`s that the current component sampler is sampling. The
+reason we instead have a `Tuple` type parameter listing `Symbol`s is to allow
+`is_target_varname` to benefit from compile time constant propagation. This is important
+for type stability of `tilde_assume`.
```
torfjelde, line 52: Hmm. We sure we want to do this?

I'm happy with doing this for now, but this does seem a bit overly restrictive.

There's nothing stopping us from taking the same approach as we do in `VarInfo`, allowing both handling of the case where compile-time is possible, and also the case when it's not.

torfjelde: As in, this is technically breaking functionality of `Turing.Experimental.Gibbs`, which had as one of the motivation to allow more flexible conditioning, e.g. `@varname(x[1]) => ...`, similar to `condition` and `fix`.

src/mcmc/gibbs.jl
```diff
@@ -478,11 +547,54 @@ function setparams_varinfo!!(
...
+function match_linking!!(varinfo_local, prev_state_local, model)
+    prev_varinfo_local = varinfo(prev_state_local)
+    was_linked = DynamicPPL.istrans(prev_varinfo_local)
+    is_linked = DynamicPPL.istrans(varinfo_local)
+    if was_linked && !is_linked
+        varinfo_local = DynamicPPL.link!!(varinfo_local, model)
+    elseif !was_linked && is_linked
+        varinfo_local = DynamicPPL.invlink!!(varinfo_local, model)
+    end
+    # TODO(mhauru) The above might run into trouble if some variables are linked and others
```
torfjelde, line 571: We should probably have an issue accompany this 👍

src/mcmc/gibbs.jl
```diff
@@ -80,11 +130,11 @@ function get_conditioned_gibbs(context::GibbsContext, vns::AbstractArray{<:VarNa
 end
 
 function is_target_varname(context::GibbsContext, vn::VarName)
-    return Iterators.any(
-        Iterators.map(target -> subsumes(target, vn), context.target_varnames)
-    )
+    return is_target_varname(context, DynamicPPL.getsym(vn))
 end
 
+is_target_varname(::GibbsContext{T}, vn_symbol::Symbol) where {T} = vn_symbol in T
```
torfjelde, line 136: I'd highly recommend just making this one function that does
```julia
is_target_varname(context::GibbsContext{T}, vn::VarName{sym}) where {sym} = sym in T
```
Otherwise we're heavily relying on Julia's constant propagation hitting => less likely to get inlined as program size grows

src/mcmc/gibbs.jl
```diff
@@ -433,28 +491,36 @@ function setparams_varinfo!!(model, ::Sampler, state, params::AbstractVarInfo)
...
     params::AbstractVarInfo,
 )
     # The state is already a VarInfo, so we can just return `params`, but first we need to
-    # update its logprob.
-    # TODO(mhauru) Is this the right context to use?
-    return last(DynamicPPL.evaluate!!(model, params, DynamicPPL.LikelihoodContext()))
+    # update its logprob. To do this, we have to call evaluate!! with the sampler, rather
+    # than just a context, because ESS is peculiar in how it uses LikelihoodContext for
+    # some variables and DefaultContext for others.
+    return last(DynamicPPL.evaluate!!(model, params, sampler))
```
torfjelde, line 516:
```suggestion
    return last(DynamicPPL.evaluate!!(model, params, SamplingContext(sampler))
```
just in case we remove the other `evaluate!!` later

test/mcmc/gibbs.jl
```diff
@@ -281,56 +445,85 @@ end
...
+        num_ms = count(ismissing.(Array(chn[:, (num_zs + 1):end, 1])); dims=2)
+        # The below are regression tests. The values we are comparing against are from
+        # running the above model on the "old" Gibbs sampler that was in place still on
+        # 2024-11-20. The model was run 5 times with 10_000 samples each time. The values
+        # to compare to are the mean of those 5 runs, atol is roughly estimated from the
+        # standard deviation of those 5 runs.
+        # TODO(mhauru) Could we do something smarter here? Maybe a dynamic model for which
+        # the posterior is analytically known? Doing 10_000 samples to run the test suite
+        # is not ideal
+        @test isapprox(mean(num_ms), 8.6087; atol=0.5)
```
torfjelde, line 466: Maybe add an issue for future ref?

torfjelde (approved):
I've added some comments and whatnot, but, as we discused on slack, we'll defer these changes to later / another PR:)

Just make sure to add some issues so we can keep track of it 👍 

Awesome, awsome stuff @mhauru ! I imagine this one has been quite painful, so appreciate you stickiing with it until the end 👏

src/mcmc/gibbs.jl
mhauru: Yeah, agree that this is an unfortunate regression of the abilities of the experimental Gibbs, if not of the old main Gibbs sampler. Tor and I discussed this on Slack, and the reason the regression happened was that the parts that were handling all this `target_varnames` business had to be essentially rewritten to deal with the issues with dynamic models, and thus I first went for something that works and type stable when targets are just `Symbol`s, which is the vast majority of use cases.

I agree with that this should be improved, and as, again, discussed on Slack, shouldn't be impossible to do. I would like to merge this PR though, because it does not regress non-experimental stuff, and this has already gotten too big and long-lived as a PR.

src/mcmc/gibbs.jl
mhauru: Agreed, done.

src/mcmc/gibbs.jl
mhauru: Done.

src/mcmc/gibbs.jl
mhauru: Opened here: https://github.com/TuringLang/Turing.jl/issues/2401

test/mcmc/gibbs.jl
mhauru: Opened here: https://github.com/TuringLang/Turing.jl/issues/2402. Will also add comments in the code referencing this issue and the other one I just opened.

src/mcmc/gibbs.jl
mhauru: Issue to track: https://github.com/TuringLang/Turing.jl/issues/2403

mhauru:
I'm done making the changes I had in mind. I may still experiment with some performance improvements, but not sure if any will make it in here. I'll also try to reduce the iteration counts in some tests to make them faster, the only CI failure is because one job just timed out at 6h.

Since both Tor and I seem to be happy, I'm gonna ping others in case they want to take a look: @penelopeysm, @willtebbutt, @sunxd3, @yebai. I think we can rely on @torfjelde giving an expert review, everyone else can judge for themselves how thorough a look they want to take, but I think everyone should be at least aware that this, somewhat major, change is happening. If you want to give this PR a review but haven't yet had time, self-request a review and we'll make sure to wait before merging.

For help in reviewing: This PR does a few things:

1. Deletes the old `src/mcmc/gibbs.jl`, and the related `src/mcmc/gibbs_conditional.jl`.
2. Moves `src/experimental/gibbs.jl` to be the new `src/mcmc/gibbs.jl`, and merges `test/experimental/gibbs.jl` and `test/mcmc/gibbs.jl`.
3. Makes a lot of edits to the experimental/new Gibbs to accommodate dynamic models and some other things.
4. Adds more, new tests to `test/mcmc/gibbs.jl`.
5. Introduces `RepeatSampler` and its tests. This has to be done in the same PR because the old Gibbs had repeat functionality built-in, whereas the new Gibbs doesn't.
6. Makes a bunch of small changes to various samplers to accommodate the new Gibbs.

Points 4-6 one can reviewed like usual, as a diff of a few hundred lines. Points 2-3 I think are better viewed as a new Gibbs sampler from scratch. The changes in point 3 are so extensive that reading it as a diff doesn't make much sense unless you know the old code really well.

penelopeysm:
I'm happy to take a look next week, but doubt I'll get to it today as my head is already several layers deep in DynamicPPL stuff 😄

mhauru:
I managed to decrease the iteration counts on a lot of the heaviest tests, the total runtime should be reduced substantially now. They seem to still pass somewhat robustly, i.e. I tried at least two random seeds.

Also did some quick checks of performance overheads, and the previous large overheads are gone in my example cases. Now, rather than being e.g. 100-500% slower than the old Gibbs we are more like 0-50% slower. This for models dominated by overheads from outside model evaluation, i.e. fast models where performance is not a big deal.

penelopeysm:
@yebai Sure! Are we happy otherwise with the PR, i.e. if conflicts are fixed and CI passes we can merge?

penelopeysm:
@yebai CI pretty much passes fine, apart from:

- Some numerical tests fail on x86 by a fairly small amount. I can't quite tell why – as far as I can tell, everything has been seeded correctly.
- There's some weird behaviour in that Gibbs test suite runs much, much slower on 1.11 than 1.10. It doesn't affect the outcome though.

Personally I don't think that either of these are serious enough to prevent us from merging this PR. If you agree, feel free to hit the button 😄

sunxd3:
🎉🎉

PR #2341, opened 2024-09-25
CompatHelper: DynamicPPL -> 0.29

penelopeysm:


penelopeysm:
Oh, this is fun. Julia 1.7 tests are failing because:

 - the last version of SciMLBase to support it is 2.6.0
   - which in turn requires ADTypes 0.2
 - but DynamicPPL 0.28.3 requires ADTypes 1.0 (https://github.com/TuringLang/DynamicPPL.jl/pull/638)

penelopeysm:
Tests still aren't passing, there are still a few changes that need to be made to bring it up to date with https://github.com/TuringLang/DynamicPPL.jl/pull/575.

I don't mind handling this, but it will take me a long time, so if anybody else feels like expediting it please feel free to jump in:)

mhauru:
I'll ask around if anyone objects to dropping <1.9.

penelopeysm:
In the meantime I can work to get the PR ready to merge, we will need it at some point anyway whether it's sooner or later:)

.github/workflows/Tests.yml
```diff
@@ -8,25 +8,34 @@ on:
...
+          - name: "mcmc/Inference"
+            args: "mcmc/Inference.jl"
+          - name: "experimental/gibbs"
+            args: "experimental/gibbs.jl"
+          - name: "mcmc/ess"
+            args: "mcmc/ess.jl"
+          - name: "everything else"
+            args: "--skip essential/ad.jl mcmc/gibbs.jl mcmc/hmc.jl mcmc/abstractmcmc.jl mcmc/Inference.jl experimental/gibbs.jl mcmc/ess.jl"
         version:
-          - '1.7'
```
mhauru, line 29: Do we want to remove this, or rather replace with 1.9? My thinking would be that we would fix one to 1.9 and let the other one be 1 and have it float up as new Julia releases come.

penelopeysm: I was kind of unsure about this. I feel like it must be worth explicitly testing on 1.10, but that would lead to us testing all three of 1.9 (min bound), 1.10 (LTS), and 1 (newest version), which might be a bit much.

penelopeysm: But if we had to cut one from the three, I'd cut 1.10, and assume that if it works on 1.9 and 1 then it should work on 1.10 (kind of a sandwich theorem).

src/mcmc/mh.jl
```diff
@@ -285,10 +285,8 @@ end
 unvectorize(dists::AbstractVector) = length(dists) == 1 ? first(dists) : dists
 
 # possibly unpack and reshape samples according to the prior distribution
-reconstruct(dist::Distribution, val::AbstractVector) = DynamicPPL.reconstruct(dist, val)
-function reconstruct(dist::AbstractVector{<:UnivariateDistribution}, val::AbstractVector)
-    return val
-end
+reconstruct(dist::Distribution, val::AbstractVector) = DynamicPPL.from_vec_transform(dist)(val)
```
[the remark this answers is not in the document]
penelopeysm: fine, fine

penelopeysm:
As it turns out, Julia 1.9 won't work either, because SciMLBase.

penelopeysm:
I think everything almost works, except for a Mooncake bug which I'll report tomorrow.

penelopeysm:
@yebai / @mhauru Since we're making fairly sweeping changes here, is it an opportune time to remove Tracker support as well? https://github.com/TuringLang/Turing.jl/issues/2356

penelopeysm:
@torfjelde Could I get you to take a look at my changes to `src/mcmc` please? The tests all pass, but I'm a little anxious 😄

Also, this PR fixes the following bug, but not entirely sure what's the best way to write a test for it. (Other samplers already work correctly with this, it's just MH that fails.) Would it make sense to extend `test/test_utils/models.jl` and `test/test_utils/numerical_tests.jl`?

```julia
julia> using Turing

julia> @model f() = x ~ LKJCholesky(2, 1)

julia> sample(f(), MH(), 100)
ERROR: MethodError: no method matching length(::LinearAlgebra.Cholesky{Float64, Matrix{Float64}})
[...]
```

src/mcmc/mh.jl
```diff
@@ -212,43 +212,20 @@ end
...
-                end
-            else
-                error("Cannot assign `NamedTuple` to `VarInfo`")
-            end
+        if vals isa AbstractVector
+            vals = unvectorize(vals)
+        end
+        if length(vns) == 1
+            # Only one variable, assign the values to it
+            DynamicPPL.setindex!(vi, vals, vns[1])
```
torfjelde, line 222: Is this done for optimization purposes? If so, I say just remove it and use the for-loop.

torfjelde:
> (Other samplers already work correctly with this, it's just MH that fails.) Would it make sense to extend test/test_utils/models.jl and test/test_utils/numerical_tests.jl?

I'd just add it to the `MH` specific tests for now :+1

torfjelde:
https://github.com/TuringLang/AbstractMCMC.jl/issues/85

src/mcmc/mh.jl
penelopeysm: No, not for optimisation. I think I was just editing the code I removed and didn't think to collapse both cases 😄

penelopeysm: Actually no, this handles the multivariate case of e.g. `z ~ MvNormal`, where `vns = [z]` and `vals = [0, 1]` for example. I couldn't find a nice way to simplify it so have left it for now!

penelopeysm:
This upgrade is also blocked by https://github.com/TuringLang/docs/issues/521 – ideally we want to make sure any relevant docs build with Mooncake. Because we don't want to upgrade to 0.35.0, patch Mooncake to work for the docs, and then release 0.35.1 with a new compat entry for Mooncake just so that the docs work.

However, the Bayesian NN example errors with Mooncake. Will create specific upstrea issues etc when I get the time

penelopeysm:
Once tests pass, still TODO:

- [ ] Update API docs

penelopeysm:
CI is finally passing, so it's time to ask for you all to take a final look!

To be honest, there are no major code changes since the last round of reviews. Most of the CI errors were related to upstream issues which have since been fixed (mostly due to @willtebbutt's heroic efforts 🙌 ). There are a few outstanding TODOs which I've opened issues for:

 - #2369 
 - #2371 
 - #2372 

However, this PR makes some sweeping changes to the library and to avoid this PR from scope creep, I think we should address these in separate PRs.

mhauru (approved):
I'm happy. Great work @penelopeysm!

.github/workflows/Tests.yml
```diff
@@ -8,25 +8,35 @@ on:
...
```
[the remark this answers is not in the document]
penelopeysm: https://github.com/TuringLang/Turing.jl/issues/2373

sunxd3:
great work! thanks guys

PR #2347, opened 2024-09-26
Generate API docs

### LINKED ISSUE #2344: Regenerate API docs
There are no API docs for Turing, thus there is no documentation for things inside Turing (e.g. samplers) and also no way to find out what symbols are exported (without looking at the source code).

Should make a basic Documenter setup and fix everything that it complains about.

penelopeysm:
## Description

This PR generates API docs for Turing.jl itself. This exposes documentation for important things like samplers which are otherwise undocumented.

Many of the exported identifiers are defined in other packages (DynamicPPL, etc.) and re-exported and/or extended by Turing. Rather than inserting the actual docs (which can actually be quite difficult sometimes as it requires locating the right method), I've generally opted to put them in a table with links to the corresponding package documentation.

This PR also adds docstrings to exported identifiers where necessary, and fixes some inaccuracies in the `MH()` docstrings.

Closes #2344.

## Preview the docs

https://turinglang.org/Turing.jl/previews/PR2347/


docs/src/api.md
```diff
@@ -0,0 +1,168 @@
...
+| `unfix`                    | [`DynamicPPL.unfix`](@extref)                                                                                                     | Unfix the value of a variable                                   |
+| `OrderedDict`              | [`OrderedCollections.OrderedDict`](https://juliacollections.github.io/OrderedCollections.jl/dev/ordered_containers/#OrderedDicts) | An ordered dictionary                                           |
+
+### Extra re-exports from Bijectors
+
+Note that Bijectors itself does not export `ordered`.
+
+```@docs
+Bijectors.ordered
+```
```
penelopeysm, line 158: Is there a particular reason why Bijectors doesn't export `ordered`, and Turing does?

That seems counterintuitive to me because if it's important enough for Turing to export it, surely it must be important enough for Bijectors to export it.

If we export it from Bijectors, then we can add it to the docs there and then just link to it here.

penelopeysm:
I think this is finally ready :) Most of the changes can be seen via the preview docs (https://turinglang.org/Turing.jl/previews/PR2347/) but also please do check over the docstrings I added.

penelopeysm:
1.11 tests failing because of https://github.com/compintell/Mooncake.jl/issues/180

.github/workflows/Docs.yml
```diff
@@ -0,0 +1,37 @@
...
+permissions:
+  contents: write
+  pull-requests: read
+  statuses: write
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
sunxd3, line 27:
```suggestion
      - uses: actions/checkout@v4
```
maybe?

docs/src/api.md
```diff
@@ -0,0 +1,170 @@
...
+See the [AD guide](https://turinglang.org/docs/tutorials/docs-10-using-turing-autodiff/) for more information.
+
+| Exported symbol   | Documentation                        | Description                               |
+|:----------------- |:------------------------------------ |:----------------------------------------- |
+| `AutoForwardDiff` | [`ADTypes.AutoForwardDiff`](@extref) | ForwardDiff.jl backend                    |
+| `AutoReverseDiff` | [`ADTypes.AutoReverseDiff`](@extref) | ReverseDiff.jl backend                    |
+| `AutoZygote`      | [`ADTypes.AutoZygote`](@extref)      | Zygote.jl backend                         |
+| `AutoTracker`     | [`ADTypes.AutoTracker`](@extref)     | Tracker.jl backend                        |
+| `AutoTapir`       | [`ADTypes.AutoTapir`](@extref)       | Tapir.jl backend, only for ADTypes >= 1.0 |
+
```
sunxd3, line 95: should we add `mooncake` (and `enzyme`)?

docs/src/api/Inference.md
```diff
@@ -0,0 +1,6 @@
+# API: `Turing.Inference`
+
+```@autodocs
+Modules = [Turing.Inference]
```
sunxd3, line 4: ideally, this page is more structured. but I think we should merge this now  and fix it later.

src/Turing.jl
```diff
@@ -100,8 +100,8 @@ export @model,                 # modelling
     vi,                     # variational inference
     ADVI,
     sample,                 # inference
-    @logprob_str,
-    @prob_str,
+    @logprob_str,  # TODO: Remove, see https://github.com/TuringLang/DynamicPPL.jl/issues/356
```
sunxd3, line 103: maybe we can setup a `todo` workflow like https://github.com/marketplace/actions/track-todo-action? (I haven't used it, just a thought)

sunxd3 (approved):
Great work @penelopeysm

docs/src/api.md
penelopeysm: Mooncake depends on https://github.com/TuringLang/Turing.jl/pull/2341, so I guess it depends on which we merge first. But I think this one is easier to merge :D

src/Turing.jl
penelopeysm: That sounds useful! We definitely have a lot of todos. Many of them labelled with names too. I'll open an issue for it

torfjelde:
Just saw this; really like this @penelopeysm :)

Issue #2315, opened 2024-09-05
Adding new issues and PRs to Project Board automatically. 


yebai:
It might be possible to achieve this goal via an Action, see, e.g., https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions

This allows us to have a consolidated view of issues and PRs for all repos, which would save time for us to go through repos manually.

penelopeysm:
Issue templates weren't the solution :(

I will try and see if this can be done with a GitHub app: https://docs.github.com/en/apps/creating-github-apps

My hope is that we can make a single app and add it to the entire organisation, then it'll cover all repos in the organisation. (as opposed to e.g. adding a workflow to every repository individually)

PR #2355, opened 2024-09-29
Add issue templates

penelopeysm:
## Description

GitHub has a nice feature for issue templates, which basically mean that people submitting an issue do it by filling in a form rather than providing freeform text.

This has a couple of benefits: 

1. We can explicitly require people to provide an MWE as well as the info needed to reproduce their bug (i.e. manifest and version info). Not having this info is annoying.

2. ~~Issues added via this form _should_ be automatically added to our project board. (See #2315)~~ Only happens if the issue is opened by someone who already has write access to the project board.

This PR adds two templates - one for bug reports and one for 'anything else', which is mostly blank. In theory we could have a feature request one too, although I don't think that a feature request template would substantially differ from 'anything else'.

It also adds a config file which disables blank issues (i.e., it forces people to use the templates).

## Preview the changes

To have a peek at what this would look like, I've added the same issue template to a playground repository of mine: https://github.com/penelopeysm/Shaymin.jl/issues Feel free to create an issue there and see the output. **Note:** on this repository blank issues are still permitted (blue circle below). This is just to let you see what that looks like. I've disabled the option in this PR.

Screenshot 2024-09-29 at 02 37 25

## The rest of the organisation

If we like how this works, then I propose we add the same templates to other repos in TuringLang too, which (should) accomplish our aim of issues across the organisation being added to the project board.


torfjelde (approved):
I'm very much in favour of this:) Dopey stuff @penelopeysm

mhauru (commented):
I like it, especially getting issues on the board automatically. A bit unsure about whether to disable blank issues. I guess the Anything Else template is very simple, and explicitly says version info is optional.

penelopeysm:
Yeah, I'm kind of unsure about that too. I don't really have any problem with blank issues in and of themselves, it's just that it's less "complete" since blank issues won't get added to the project board.

penelopeysm:
@torfjelde [Without <details>](https://github.com/penelopeysm/Shaymin.jl/issues/new?assignees=&labels=bug&projects=TuringLang%2F3&template=01-bug-report.yml), [with <details>](https://github.com/penelopeysm/Shaymin.jl/issues/new?assignees=&labels=bug&projects=TuringLang%2F3&template=02-bug-report-details.yml) I do like the latter 🙂

penelopeysm:
Of course, it's never as good as one expects:

> Note: The person opening the issue must have write permissions for the specified projects.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

I'll look into a different way of making that happen. But I suppose it's still a useful way to guide people submitting bugs.

PR #2360, opened 2024-10-04
Update to tilde overloads in `mh.jl`

torfjelde:
@penelopeysm I was looking at your PR #2341 and found some bad bugs in the _existing_ codebase for the `MH`, so I figured we should just get these fixed too.

### DIFF
--- a/src/mcmc/mh.jl
+++ b/src/mcmc/mh.jl
@@ -442,42 +442,45 @@ end
 ####
 #### Compiler interface, i.e. tilde operators.
 ####
-function DynamicPPL.assume(rng, spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi)
+function DynamicPPL.assume(
+    rng::Random.AbstractRNG, spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi
+)
+    # Just defer to `SampleFromPrior`.
+    retval = DynamicPPL.assume(rng, SampleFromPrior(), dist, vn, vi)
+    # Update the Gibbs IDs because they might have been assigned in the `SampleFromPrior` call.
     DynamicPPL.updategid!(vi, vn, spl)
-    r = vi[vn]
-    return r, logpdf_with_trans(dist, r, istrans(vi, vn)), vi
+    # Return.
+    return retval
 end
 
 function DynamicPPL.dot_assume(
     rng,
     spl::Sampler{<:MH},
     dist::MultivariateDistribution,
-    vn::VarName,
+    vns::AbstractVector{<:VarName},
     var::AbstractMatrix,
-    vi,
+    vi::AbstractVarInfo,
 )
-    @assert dim(dist) == size(var, 1)
-    getvn = i -> VarName(vn, vn.indexing * "[:,$i]")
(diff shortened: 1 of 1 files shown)


src/mcmc/mh.jl
```diff
@@ -442,42 +442,45 @@ end
...
-    DynamicPPL.updategid!.(Ref(vi), vns, Ref(spl))
-    r = vi[vns]
-    var .= r
-    return var, sum(logpdf_with_trans(dist, r, istrans(vi, vns[1]))), vi
+    # Just defer to `SampleFromPrior`.
+    retval = DynamicPPL.dot_assume(rng, SampleFromPrior(), dist, vns[1], var, vi)
+    # Update the Gibbs IDs because they might have been assigned in the `SampleFromPrior` call.
+    DynamicPPL.updategid!.((vi,), vns, (spl,))
+    # Return.
+    return retval
```
torfjelde, line 469: I'm wondering if we maybe should just move this "default" impl which uses `SampleFromPrior` + `updategid!` to DynamicPPL.jl itself. Thoughts @penelopeysm @mhauru ?

torfjelde:
Also, Mooncake.jl tests are failiing; is this expected @penelopeysm @willtebbutt ?

penelopeysm:
Oh, I see, I'm just confused. The values are stored in varinfo from the previous iteration and so when evaluating the lp it just extracts those values. I had assumed it would call rand() on the prior, but now I observe that that's not the case ;)

src/mcmc/mh.jl
penelopeysm: Would this be useful for multiple samplers beyond MH?

penelopeysm:
Sorry, not trying to be annoying and get a final word in, but having spent a good amount of time figuring out the interplay between Turing and AdvancedMH I feel like this behaviour is a bit unexpected (and I actually now understand why it happens 😂):

```julia
@model function gdemo(x, y)
    s² ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s²))
    x ~ Normal(m, sqrt(s²))
    y ~ Normal(m, sqrt(s²))
end
chain = sample(
    gdemo(1.5, 2.0),
    MH(:m => AdvancedMH.RandomWalkProposal(Normal(0, 0.25))),
    10
)
```

Here `s²` is sampled once at the start of the chain and it's never touched again. I feel like people (perhaps just me 😄) would expect that it either defaults to a static proposal (the prior) or that it errors. If you agree I can open a separate issue :) this is obviously not part of this PR

torfjelde:
Very much agree with you that this is very confusing :confused: IMO we should be erroring if we some variables don't have a Gibbs sampler assigned to them

src/mcmc/mh.jl
torfjelde: HMC-related samplers could also make use of this.

mhauru:
Maybe worth checking if there is a similar issue with the new Gibbs sampler? If not, #2328 might fix this by accident.

src/mcmc/mh.jl
penelopeysm: In that case, that makes sense to move. I can open an issue!

PR #1887, opened 2022-09-28
Test Enzyme and reexport `ADTypes.AutoEnzyme`

devmotion:
Note: This does not work yet

---

I opened this PR to make it easier to debug (and possibly fix) issues with Enzyme.

Currently, the following example does ~~not~~ work (note that the snippet does not require the PR which solely reexports `AutoEnzyme` at this point):
```julia
using Turing
using Enzyme
using ADTypes
Enzyme.API.runtimeActivity!(true);
Enzyme.API.typeWarning!(false);

@model function model()
    m ~ Normal(0, 1)
    s ~ InverseGamma()
    x ~ Normal(m, s)
end

sample(model() | (; x=0.5), NUTS(; adtype = ADTypes.AutoEnzyme()), 10)
```

~~With Enzyme#main my Julia (1.8.1) segfaults. An incomplete (it filled my whole terminal) output: https://gist.github.com/devmotion/1352197f2354c6fecddd7b778ec4bcf7#file-log-txt~~

The example works (latest releases of Turing, Enzyme, and ADTypes on Julia 1.10.0) but the following warnings show up:
```julia
warning: didn't implement memmove, using memcpy as fallback which can result in errors
warning: didn't implement memmove, using memcpy as fallback which can result in errors
```

PR #2374, opened 2024-10-24
Run JuliaFormatter on more files, remove trailing whitespace

mhauru:
The only actual change is the change in `.JuliaFormatter.toml`. Everything else is a consequence of running autoformatting and auto-whitespace-removal.

This will clean-up the diff in #1887 a lot, since the formatter has already been run there.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -8,7 +8,4 @@ ignore = [
     # https://github.com/TuringLang/Turing.jl/pull/2328/files
     "src/experimental/gibbs.jl",
     "test/experimental/gibbs.jl",
-    # https://github.com/TuringLang/Turing.jl/pull/1887 # Enzyme PR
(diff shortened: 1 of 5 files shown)


penelopeysm (approved):
LGTM. I don't entirely feel comfortable with model functions being edited to include return statements (they aren't really functions), but that's probably just me.

mhauru:
The two syntaxes, with and without explicit `return`, are equivalent though, and different from having e.g. `return nothing`, because `generated_quantities` (and I think `@submodel`, which will soon be the same thing: https://github.com/TuringLang/DynamicPPL.jl/pull/696) makes use of the return value.

penelopeysm:
They are equivalent, yeah. I just think that the rationale for inserting `return` is more applicable to ordinary functions (i.e. be explicit about what you're returning) than it is to models (where the return function isn't always used); but as I said that's just me :)

PR #2381, opened 2024-10-29
Fix the global Random.seed in runtests.jl

### LINKED ISSUE #2371: AdvancedMH external-sampler tests fail intermittently with 2 threads
https://github.com/TuringLang/Turing.jl/blob/24e68701b01695bffe69eda9e948e910c1ae2996/test/mcmc/abstractmcmc.jl#L187-L211

These tests seem to fail intermittently on the runner using ubuntu-latest; Julia 1.10; x64; 2 threads. I've observed it happen twice now.

Compare e.g. [this failure](https://github.com/TuringLang/Turing.jl/actions/runs/11466651176/job/31907789466?pr=2341) with [this success](https://github.com/TuringLang/Turing.jl/actions/runs/11466651176/job/31917554907?pr=2341) (the codebase was exactly the same, the only thing I did to get it to pass was to rerun CI).

mhauru:
Based on local testing, this should close #2371.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,9 +1,13 @@
 include("test_utils/SelectiveTests.jl")
 using .SelectiveTests: isincluded, parse_args
 using Pkg
+using Random: seed!
 using Test
 using TimerOutputs: TimerOutputs, @timeit
 import Turing
+
+# Fix the global Random.seed for reproducibility.
+seed!(23)
 
 include(pkgdir(Turing) * "/test/test_utils/models.jl")
 include(pkgdir(Turing) * "/test/test_utils/numerical_tests.jl")


torfjelde:
> IIRC, we have a few test models whose posteriors have a large variance, which can cause challenges for MCMC samplers.

Specifically, all the `DynamicPPL.TestUtils.DEMO_MODELS` are examples the Inverse-Gamma-Normal model because it has a closed-from posterior so it's easy to compare samples to ground-truth. In hindsight, this was a bad idea. The reason for this is that the `InverseGamma` prior has infinite variance 😬 Once we condition on some observations, the variance should become finite, but because we only condition on a few observations (either 1 or 2 in all our "demo models"), the variance will still be huge. This in turn increases the probability of one of the samplers just completely falling on its butt 😕 

There are a few ways we can try to make this better, e.g. choose a different base model for testing, or just use more observations in the posterior (though this increases the computational burden of the tests).

torfjelde:
Also, regarding fixing the seeds: note that this won't be consistent across hardware nor Julia versions. For that we'd have to use StableRNGs.jl, but this in turn comes with additional computational overhead 😕 However, I do agree that maybe fixing the seeds can be helpful.

penelopeysm (changes requested):
Would it be possible to add the call to `seed!` slightly closer to the point of use? Specifically, I'm looking at `test/mcmc/abstractmcmc.jl` and inserting them after lines 120, 161, and 185.

We would still have the issue with `seed!` not being consistent across Julia versions, but at least we wouldn't get different results depending on which tests are run.

(I was pleasantly surprised to find that [the PRNG is actually consistent across operating systems!](https://github.com/penelopeysm/Shaymin.jl/actions/runs/11592058124/job/32273123156) I know the algorithm was changed from MersenneTwister to Xoshiro in Julia 1.7, not sure if that had an impact on it.)

mhauru:
`@testset` actually already resets the seed for us:
```julia
julia> module MWE

       using Test, Random

       @testset "A" begin
          @show rand()
       end
       @testset "B" begin
          @show rand()
       end

       end
rand() = 0.05361041735414662
Test Summary: |Time
A             | None  0.0s
rand() = 0.05361041735414662
Test Summary: |Time
B             | None  0.0s
Main.MWE
```
However, without fixing the seed at the start globally, the local seed set by `@testset` will be different in different Julia sessions.

@penelopeysm does that mitigate the issue you worry about?

mhauru:
Yup:
```
$ rg "seed\!" test | wc -l
      70
```

I would leave fixing that for later though, since it's mostly a style/clean-up question. Any other changes you want to see here @penelopeysm?

penelopeysm:
Nope, feel free to go ahead :) And thanks for diagnosing it!

penelopeysm:
GitHub's mobile app doesn't seem to let me approve so take this comment as an approval

PR #2383, opened 2024-10-29
Move ADTypeCheckContext tests to a separate module

mhauru:
They were being run when test utils was imported, which was not ideal, and hard to leave out. Now they are in a module of their own.


penelopeysm (approved):
Lovely, feel free to merge when CI passes :)

I wonder if we should have a merge queue for Turing as well?

Issue #2389, opened 2024-11-04
Sampling with HMC can hang if AD is bugged


penelopeysm:
### Minimal working example

```julia
using Turing

@model function model1()
    σ ~ InverseGamma(2, 3)
    V ~ truncated(Normal(0, σ), 0, Inf)
end

sample(model1(), NUTS(), 100)
```

### Description

Hangs.

This isn't because of a bug in Turing; it's actually a bug in ForwardDiff, which returns NaN's when calculating the gradient. (This is true of all other currently supported AD backends too, see https://discourse.julialang.org/t/help-cant-get-turing-to-work-on-a-simple-model/122107/2)

Because the gradient is always returned with `NaN`'s, `isfinite()` on it returns false, and this block goes into an infinite loop:

https://github.com/TuringLang/Turing.jl/blob/397d1a7c521f72ab18675f3d65c20925ed55f69f/src/mcmc/hmc.jl#L176-L194

It would probably make sense to just error after a sufficiently large number of attempts (not sure about the exact number, but 1000 seems reasonable perhaps?). Alternatively, or additionally, we could also check for NaN's and just directly error if logp or its gradient contains NaN's.

### Julia version info

<details><summary>versioninfo()</summary>

```
julia> versioninfo()
Julia Version 1.11.1
Commit 8f5b7ca12ad (2024-10-16 10:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 10 × Apple M1 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```
</details>

### Manifest

The relevant parts are:

```jullia
(ppl) pkg> st
Status `~/ppl/Project.toml`
  [f6369f11] ForwardDiff v0.10.37
  [fce5fe82] Turing v0.35.1
```

I did paste the whole thing here because I'm the one who made this issue template and I had better abide by it 😄 

<details><summary>]st --manifest</summary>

```
(ppl) pkg> st --manifest
Status `~/ppl/Manifest.toml`
  [47edcb42] ADTypes v1.9.0
  [621f4979] AbstractFFTs v1.5.0
  [80f14c24] AbstractMCMC v5.6.0
  [7a57a42e] AbstractPPL v0.9.0
[...]

PR #2392, opened 2024-11-06
Don't get stuck in an infinite loop if HMC can't find an initial point

penelopeysm:
Quick fix for #2389.

See also https://github.com/TuringLang/AdvancedHMC.jl/pull/381 for added context.


willtebbutt:
Also LGTM -- definitely a drastic improvement over hanging.

PR #2393, opened 2024-11-15
Add DynamicPPL integration tests

penelopeysm:
This PR adds the Turing/DynamicPPL integration tests (from the [`test/turing` folder in DynamicPPL](https://github.com/TuringLang/DynamicPPL.jl/tree/f5890a11ef4002026e8edeb355b6e8a50c68d115/test/turing)) to this repository instead.

The main motivation behind this PR is to ensure that the tests added here are always run whenever a new version of Turing is released, thus making it impossible to release a version of Turing that contains a buggy version of DynamicPPL. See https://github.com/TuringLang/DynamicPPL.jl/issues/703 for more thorough explanation.

Some implementation details:

 - I made each test file its own module. This mimics the structure of the other tests in Turing.jl.
 - I had to add AbstractPPL and BangBang as explicit test depedencies (previously they were transitive deps).
 - Because the DynamicPPL integration tests run very quickly, it's probably OK for to run them under the 'everything else' bucket in the CI setup. Hence this PR makes no change to the CI.
 - I fixed a couple of deprecation warnings (specifically `link!(vi, sampler)` -> `link!!(vi, sampler, model)`; the same with `invlink!`; and `AutoReverseDiff(false)` -> `AutoReverseDiff(; compile=false)`).

Otherwise, this is a straight copy paste.

### DIFF
--- a/test/dynamicppl/compiler.jl
+++ b/test/dynamicppl/compiler.jl
@@ -1,0 +1,370 @@
+module DynamicPPLCompilerTests
+
+using ..NumericalTests: check_numerical
+using LinearAlgebra: I
+using Test: @test, @testset, @test_throws
+using Turing
+
+# TODO(penelopeysm): Move this to a DynamicPPL Test Utils module
+# We use this a lot!
+@model function gdemo_d()
+    s ~ InverseGamma(2, 3)
+    m ~ Normal(0, sqrt(s))
+    1.5 ~ Normal(m, sqrt(s))
+    2.0 ~ Normal(m, sqrt(s))
+    return s, m
+end
+const gdemo_default = gdemo_d()
+
+@testset "compiler.jl" begin
+    @testset "assume" begin
+        @model function test_assume()
+            x ~ Bernoulli(1)
+            y ~ Bernoulli(x / 2)
+            return x, y
+        end
+
+        smc = SMC()
+        pg = PG(10)
+
+        res1 = sample(test_assume(), smc, 1000)
+        res2 = sample(test_assume(), pg, 1000)
+
+        check_numerical(res1, [:y], [0.5]; atol=0.1)
+        check_numerical(res2, [:y], [0.5]; atol=0.1)
+
+        # Check that all xs are 1.
+        @test all(isone, res1[:x])
+        @test all(isone, res2[:x])
+    end
+    @testset "beta binomial" begin
+        prior = Beta(2, 2)
+        obs = [0, 1, 0, 1, 1, 1, 1, 1, 1, 1]
+        exact = Beta(prior.α + sum(obs), prior.β + length(obs) - sum(obs))
+        meanp = exact.α / (exact.α + exact.β)
+
+        @model function testbb(obs)
+            p ~ Beta(2, 2)
+            x ~ Bernoulli(p)
+            for i in 1:length(obs)
+                obs[i] ~ Bernoulli(p)
+            end
+            return p, x
+        end
+
+        smc = SMC()
+        pg = PG(10)
+        gibbs = Gibbs(HMC(0.2, 3, :p), PG(10, :x))
+
+        chn_s = sample(testbb(obs), smc, 1000)
+        chn_p = sample(testbb(obs), pg, 2000)
+        chn_g = sample(testbb(obs), gibbs, 1500)
+
+        check_numerical(chn_s, [:p], [meanp]; atol=0.05)
+        check_numerical(chn_p, [:x], [meanp]; atol=0.1)
+        check_numerical(chn_g, [:x], [meanp]; atol=0.1)
+    end
+    @testset "forbid global" begin
+        xs = [1.5 2.0]
+        # xx = 1
+
+        @model function fggibbstest(xs)
+            s ~ InverseGamma(2, 3)
+            m ~ Normal(0, sqrt(s))
+            # xx ~ Normal(m, sqrt(s)) # this is illegal
+
+            for i in 1:length(xs)
+                xs[i] ~ Normal(m, sqrt(s))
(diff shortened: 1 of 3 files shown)


test/dynamicppl/varinfo.jl
```diff
@@ -0,0 +1,369 @@
...
+
+        # Transforming only a subset of the variables
+        DynamicPPL.link!(vi, spl, Val((:m,)))
+        @test all(x -> !DynamicPPL.istrans(vi, x), meta.s.vns)
+        @test all(x -> DynamicPPL.istrans(vi, x), meta.m.vns)
+        DynamicPPL.invlink!(vi, spl, Val((:m,)))
+        @test all(x -> !DynamicPPL.istrans(vi, x), meta.s.vns)
+        @test all(x -> !DynamicPPL.istrans(vi, x), meta.m.vns)
+        @test meta.s.vals == v_s
+        @test meta.m.vals == v_m
```
penelopeysm, line 84: These methods are deprecated: 

https://github.com/TuringLang/DynamicPPL.jl/blob/f5890a11ef4002026e8edeb355b6e8a50c68d115/src/varinfo.jl#L1158-L1164

And AFAIK there is no replacement. The same is true of `invlink!`. Is it OK to drop these tests?

penelopeysm: @mhauru – maybe related to your work on Gibbs?

mhauru: Without testing out whether this works, maybe you could use `link!!(vi, sampler, model)` but with a sampler that has a space set, so e.g.
```julia
alg = HMC(0.1, 5, :m)
spl = DynamicPPL.Sampler(alg, model)
```
instead of `Val((:m,))` for `space`. I think it should have the same effect.

Eventually we would like to replace these with ones where you give the `VarName` to link as an argument, but I think we aren't there yet.

penelopeysm: This works!

penelopeysm:
Should be mergeable, I think, then I'll get rid of them from DynamicPPL.

test/dynamicppl/compiler.jl
```diff
@@ -0,0 +1,370 @@
+module DynamicPPLCompilerTests
+
+using ..NumericalTests: check_numerical
+using LinearAlgebra: I
+using Test: @test, @testset, @test_throws
+using Turing
+
+# TODO(penelopeysm): Move this to a DynamicPPL Test Utils module
```
torfjelde, line 8: This is equivalent to one of the DynamicPPL.TestUtils models; I believe it's the same as `demo_assume_literal_observe`?

test/dynamicppl/loglikelihoods.jl
```diff
@@ -0,0 +1,48 @@
+module DynamicPPLLogLikelihoodsTest
```
torfjelde, line 1: I believe this can actually be moved to DynamicPPL.jl's test suite proper under the `test/ext` dir, by just replacing the specific sampler calls with explicit constructions of `Chains`:)

THis is from before we had DynamicPPLMCMCChainsExt.

test/dynamicppl/model.jl
```diff
@@ -0,0 +1,63 @@
+module DynamicPPLModelTests
+
+using Test: @testset, @test
+using Turing
+
+# TODO(penelopeysm): Move this to DynamicPPL Test Utils module
```
torfjelde, line 6: This method depends on `MCMCChains.jl` though, so need to find a way around it.

torfjelde (approved):
Some of these tests could deffo be moved to DynamicPPL.jl proper, but happy to merge them now if accompanied with an issue over at DPPL to ensure we don't loose track of the TODOs:)

test/dynamicppl/compiler.jl
penelopeysm: It's the same as this one https://github.com/TuringLang/DynamicPPL.jl/blob/ba6e3b8d5337b36b3ea680fa3de393d05bd85d34/src/test_utils/models.jl#L145-L152 but we don't actually have a copy in the demo models

test/dynamicppl/loglikelihoods.jl
penelopeysm: Okay, will delete from here and move to MCMCChainsExt in the followup PR

test/dynamicppl/model.jl
penelopeysm: Looking at this file more carefully, shouldn't we be able to move all of this to DynamicPPLMCMCChainsExt too?

test/dynamicppl/compiler.jl
torfjelde: Ah, okay, but then at least we could use that / make that parto f demo models

penelopeysm: Doing this as part of https://github.com/TuringLang/DynamicPPL.jl/pull/733

test/dynamicppl/loglikelihoods.jl
penelopeysm: Actually, aren't these tested here already? https://github.com/TuringLang/DynamicPPL.jl/blob/master/test/pointwise_logdensities.jl

I think this is ok to just delete.

torfjelde: Oh yeah, good catch 👍

test/dynamicppl/model.jl
penelopeysm: Moved all of these to DynamicPPL tests 😄

penelopeysm:
In line with the above, I'm going to close this PR and open a new one

Issue #2398, opened 2024-11-29
Integrating Turing and MarginalLogDensities


ElOceanografo:
Moving a discussion with @yebai from Slack to here. @PavanChaggar asked if there was a way to do Laplace approximation in Turing, and I gave this little example of how it could be accomplished with [MarginalLogDensities.jl](https://github.com/ElOceanografo/MarginalLogDensities.jl):

```julia
using MarginalLogDensities
using ReverseDiff, FiniteDiff
using Turing

# simple density function
f(x, p) = -sum(abs2, diff(x))
# define a MarginalLogDensity that integrates out parameters 5 thru 100
mld = MarginalLogDensity(f, rand(100), 5:100, (), LaplaceApprox(adtype=AutoReverseDiff()))

@model function MLDTest(mld)
    θ ~ MvNormal(zeros(njoint(mld)), 1.0)
    Turing.Turing.@addlogprob! mld(θ)
end

mod = MLDTest(mld)
chn = sample(mod, NUTS(adtype = AutoFiniteDiff()), 100)
```

This issue is to discuss how this capability might be integrated better into Turing, probably via a package extension. (See also https://github.com/TuringLang/Turing.jl/issues/1382, which I opened before I made MLD). From a user perspective, an interface like this makes sense to me:

```julia
@model function Example(y)
    a ~ SomeDist()
    b ~ AnotherDist()
    mu = somefunction(a, b)
    x ~ MvNormal(mu, 1.0)
    y ~ MvNormal(x, 1.0)
end

fullmodel = Example(ydata)
marginalmodel = marginalize(fullmodel, (:x,))
sample(marginalmodel, NUTS(), 1000)
maximum_a_posteriori(marginalmodel, LBFGS())
```
I think there are two basic ways to implement this:
* `marginalize` constructs a new, marginalized `DynamicPPL.Model`, or 
* It returns a `MarginalLogDensities.MarginalLogDensity`, with new methods for `sample`, `maximum_a_posteriori`, and `maximum_likelihood` defined for it.
I'm not very familiar with Turing's internals, so happy to be corrected if there are other approaches that make more sense....

The other current roadblock is making calls to `MarginalLogDensity` objects differentiable (https://github.com/ElOceanografo/MarginalLogDensities.jl/issues/34). This is doable, I just need to do it.

torfjelde:
This looks neat 👀 

Can you elaborate a bit on what mathematical model the following line should correspond to?
```julia
[...]

PR #2421, opened 2024-12-05
Extension for MarginalLogDensities.jl

torfjelde:
This PR adds an extension for MarginalLogDensities.jl, allowing one to call `marginalize` on a `Model`, resulting in a `MarginalLogDensity` from MarginalLogDensities.jl.

Example:

``` julia
julia> using Turing, MarginalLogDensities

julia> @model function demo()
           x ~ Normal(0, 1)
           y ~ Normal(x, 1)
       end

julia> model = demo();

julia> # Marginalize out `x`.
       marginalized = marginalize(model, [@varname(x)]);

julia> # Compute the marginal log-density of `y = 0.0`.
       abs(marginalized([0.0]) - logpdf(Normal(0, √2), 0.0)) < 1e-1
true
```

Ref: #2398


torfjelde:
A bit uncertain if we should `export` this `marginalize` or hide it for now.

torfjelde:
Note that this will fail until https://github.com/ElOceanografo/MarginalLogDensities.jl/pull/36 has made its way to the public (i.e. v0.3.6). Not entirely certain why this hasn't happened yet (it is in the registry :confused: )

torfjelde:
We need https://github.com/TuringLang/DynamicPPL.jl/pull/738 to go through before this will work :+1:

seabbs:
Found this after reading the recent Stan release and losing myself in a rabbit hole. This would be super handy for some modular fitting stuff as I think said elsewhere. What is the status? Anything aside from cheerleading to do?

penelopeysm:
@ElOceanografo Hello! Are you still interested in this? I'm happy to try to pick it up (although I will probably need a bit of time to read the previous discussions + wrap my head around the stats).

ElOceanografo:
Yes, I'm still interested in this! I have not tried running this code myself yet, but based on a quick skim it looks like it's doing what I'd expect. Let me know if I can help out in any way.

ElOceanografo:
Just finished digging into this and made a PR to this branch that should get thing working again: https://github.com/TuringLang/Turing.jl/pull/2662

If it's simpler, I could also just open it as a new PR.

ElOceanografo:
Opened a new version of this at #2664 based on the up-to-date main branch, so this PR can be closed in favor of that one.

PR #2433, opened 2024-12-06
Tweak sample counts and other details in tests

mhauru:
This PR makes a number of small changes to the tests
* Lowers the sample counts of many tests.
* Adds a few comments, including some TODO notes saying we should add an explanatory comment.
* Uses StableRNGs more. Still not using them everywhere, but I added them in many places where stability matters.
* Removes unnecessary calls to `Random.seed!`. Every `@testset` resets the seed anyway.
* Removes some dead or commented out tests that weren't doing anything.
* Adds `verbose = true` to many of the top-level `@testset`s, to get some timing numbers even when tests pass.
* Adds info prints in the beginning of some of the longer-running testsets, saying which test is starting.
* Adds some more subtestsets to some of the longer-running ones, to get more granular info about what fails.
* Fixes one commented-out test to actually do something.

This started with me thinking that since we do a lot of samples in many tests, and our test suite is quite slow, clearly we should reduce the sample counts if we can. What I learned in the process is that most of the tests that do sampling are negligble compared to the tests that use particle methods. Anything involving PG/CSMC tends to take 10-100 times longer than comparable tests with other samplers.

In the end I've tried to adjust iteration counts only when either the count truly doesn't matter (we are not checking stats of the chain), or when doing so actually has a noticable, though not necessarily huge, impact on run time. I may have slipped a few times, and reduced something that didn' really make a difference. Also, when reducing iteration counts I tried to check that they pass if I pick a few different seeds, to make sure I'm not making them very brittle to future changes.


test/mcmc/ess.jl
```diff
@@ -43,41 +45,49 @@ using Turing
...
-        # Gaussian priors, we restrict ourselves to this subspace by conditioning
-        # on the non-Gaussian variables in `DEMO_MODELS`.
-        models_conditioned = map(DynamicPPL.TestUtils.DEMO_MODELS) do model
-            # Condition on the non-Gaussian random variables.
-            model | (s=DynamicPPL.TestUtils.posterior_mean(model).s,)
+        @info "Starting ESS inference tests"
+        rng = StableRNG(23)
+
+        @testset "demo_default" begin
+            chain = sample(copy(rng), demo_default, ESS(), 5_000)
```
penelopeysm, line 52: Is there a reason you prefer using `copy(rng)` over instantiating a fresh one with `StableRNG(23)`?

I feel like I'd prefer the latter, because even though it's more verbose it keeps the state self-contained to a testset. Otherwise for example someone could accidentally mutate the rng between testsets and then subsequent tests would go wrong.

test/mcmc/hmc.jl
```diff
@@ -72,13 +76,15 @@ using Turing
...
         n_samples = 1_000
         vs = map(1:3) do _
-            chain = sample(rng, model_f, HMC(0.15, 7; adtype=adbackend), n_samples)
+            chain = sample(copy(rng), model_f, HMC(0.15, 7; adtype=adbackend), n_samples)
             r = reshape(Array(group(chain, :v)), n_samples, 2, 2)
             reshape(mean(r; dims=1), 2, 2)
         end
 
+        # TODO(mhauru) This test needs a comment explaining what is being tested.
         @test maximum(abs, mean(vs) - (7 * [1 0.5; 0.5 1])) <= 0.5
```
penelopeysm, line 85: The test is to make sure that the samples obey the mean of the Wishart distribution (if `x ~ Wishart(V, n)`, then `E[X] = V * n`).

I don't fully understand this testset, though. Why are we sampling 3 chains? It seems unnecessary. Especially if we're using StableRNGs, we shouldn't need to sample so much and then average

Could we replace the contents of the testset with

```julia
V = 7; n = [1 0.5; 0.5 1]
@model hmcmatrixsup() = v ~ Wishart(V, n)
model_f = hmcmatrixsup()
n_samples = 1_000

## or maybe StableRNG(23) if you agree with my other comment
chain = sample(copy(rng), model_f, HMC(0.15, 7; adtype=adbackend), n_samples)
r = reshape(Array(group(chain, :v)), n_samples, 2, 2)
r_mean = dropdims(mean(r; dims=1); dims=1)

# Check that the samples obey the mean of the Wishart distribution
@test maximum(abs.(r_mean - (V * n))) < 0.2
```

I tested locally with `StableRNG(23)` and the MAE on the mean was 0.215259. I'll leave it to you to choose an appropriate tolerance 😄

penelopeysm (changes requested):
Also, if you could bump Mooncake compat to 0.4.61, that would hopefully take care of CI 👍

test/mcmc/hmc.jl
penelopeysm: I suppose this whole test should really go once we have a better set of DPPL models sampling from a range of distribtions. Just thinking about how much time it'll take to go through the tests to understand their purpose is, ahh, I'll need 5 Christmases to do it.

test/mcmc/ess.jl
mhauru: The only reasons were the verbosity, and ability to change the seed globally if desired. I do see your point though.

Would you prefer having `sample(StableRNG(23)` or `sample(StableRNG(seed)`, where `seed` would be defined at the top? I think the latter could be a best-of-both-worlds.

penelopeysm: Hmm... in a purist sense `StableRNG(23)` appeals to me because the 23 in one testset isn't the same as the 23 in the other testset. The only case where the seed must be the same is if you're trying to prove that sampling is deterministic i.e. `seed = 23; sample(StableRNG(seed), ...) == sample(StableRNG(seed), ...)`.

Sprinkling that over the entire test suite would be ugly, though, so happy to go with the latter, and as you say that preserves our ability to change everything globally in the way we could with `Random.seed!`.

Also, is there a particular reason for 23? 😄

test/mcmc/hmc.jl
mhauru: That's much nicer, thanks. I simplified it a tiny bit more and made the change.

test/mcmc/ess.jl
mhauru: I agree that there's no need for the seeds to be exactly the same, but it's rather being able to have this cascade of changing a global seed and having everything be affected. Handy for instance if we want to test how brittle our whole test suite is to changes in randomisation.

> Also, is there a particular reason for 23? 😄

It is known to be the most synchronistic number, which means that the pseudorandom numbers generated will have more invisible correlations in them, making the tests more robust.

penelopeysm: 🤯 And here I am just choosing my favourite Pokemon's dex number

mhauru: Now doing `StableRNG(seed)` all over.

penelopeysm:
I can approve and merge once CI passes :)

penelopeysm:
Some Ubuntu tests are timing out, but I'm going to merge on the basis that the tests pass on Windows and macOS and the root cause for the timeout is known (https://github.com/TuringLang/DynamicPPL.jl/issues/750) and will presumably be fixed soon (https://github.com/TuringLang/DynamicPPL.jl/pull/751).

PR #2437, opened 2024-12-13
Pin DifferentiationInterface version

penelopeysm:
This PR pins DifferentiationInterface to < v0.6.24 to avoid Mooncake stack overflow bugs.

cf https://github.com/compintell/Mooncake.jl/pull/416#issuecomment-2540445449

Update: Will says he will make a quick patch for Mooncake to avoid needing this, so we can close this in favour of that when it's ready.


penelopeysm:
Closed in favour of Mooncake 0.4.61

PR #2456, opened 2025-01-07
Rework Gibbs constructors

### LINKED ISSUE #2442: Remove `Gibbs(; m=HMC(0.2, 3), s=PG(10))` in favour of `Gibbs(:m=>HMC(0.2, 3), :s=>PG(10))`
Markus and I also discussed keeping only the syntax `Gibbs(m=>HMC(0.2, 3), s=>PG(10))` for the new Gibbs sampler to keep the interface simple. The two ways are identical, so they are redundant. 

For parameters with complex variable names, we will do `Gibbs(@varname(m[1])=>HMC(0.2, 3), @varname(s[1])=>PG(10))`, for example.

mhauru:
In addition to removing the kwarg constructor as discussed in #2442, I also removed the `NamedTuple` and `Dict` constructors, to keep the interface simple. I don't see many people using them, and the user can trivially call `Gibbs(my_dict...)` or `Gibbs(pairs(my_namedtuple)...)` if need be.

The PR also simplifies a few unrelated lines and expands some docstrings.

Closes #2442


src/mcmc/gibbs.jl
```diff
@@ -292,11 +292,36 @@ function set_selector(x::RepeatSampler)
 end
 set_selector(x::InferenceAlgorithm) = DynamicPPL.Sampler(x, DynamicPPL.Selector(0))
 
+to_varname(vn::VarName) = vn
+to_varname(s::Symbol) = VarName{s}()
```
penelopeysm, line 296: I think you can use the `AbstractPPL.VarName` constructor directly:

```julia
julia> using AbstractPPL

julia> VarName(:x) == @varname(x)
true

julia> VarName(@varname(x)) == @varname(x)
true
```

src/mcmc/gibbs.jl
```diff
@@ -292,11 +292,36 @@ function set_selector(x::RepeatSampler)
...
+
+to_varname_list(x::Union{VarName,Symbol}) = [to_varname(x)]
+# Any other value is assumed to be an iterable of VarNames and Symbols.
+to_varname_list(t) = map(to_varname, collect(t))
```
penelopeysm, line 300: Is there a difference between this and `collect(map(to_varname, t))`? I don't know Julia well enough, but in most languages `map` can be lazy and `collect` should be used outside to force the generator elements to go into a list. Probably super minor anyway, more curious than anything.

HISTORY.md
```diff
@@ -4,13 +4,11 @@
...
 
-The old Gibbs constructor relied on being called with several subsamplers, and each of the constructors of the subsamplers would take as arguments the symbols for the variables that they are to sample, e.g. `Gibbs(HMC(:x), MH(:y))`. This constructor has been deprecated, and will be removed in the future. The new constructor works by assigning samplers to either symbols or `VarNames`, e.g. `Gibbs(; x=HMC(), y=MH())` or `Gibbs(@varname(x) => HMC(), @varname(y) => MH())`. This allows more granular specification of which sampler to use for which variable.
+The old Gibbs constructor relied on being called with several subsamplers, and each of the constructors of the subsamplers would take as arguments the symbols for the variables that they are to sample, e.g. `Gibbs(HMC(:x), MH(:y))`. This constructor has been deprecated, and will be removed in the future. The new constructor works by mapping symbols, `VarName`s, or iterables of `VarName`s to samplers, e.g. `Gibbs(x=>HMC(), y=>MH())`, `Gibbs(@varname(x) => HMC(), @varname(y) => MH())`, `Gibbs((:x, :y) => NUTS(), :z => MH())`. This allows more granular specification of which sampler to use for which variable.
```
penelopeysm, line 11:
```suggestion
The old Gibbs constructor relied on being called with several subsamplers, and each of the constructors of the subsamplers would take as arguments the symbols for the variables that they are to sample, e.g. `Gibbs(HMC(:x), MH(:y))`. This constructor has been deprecated, and will be removed in the future. The new constructor works by mapping symbols, `VarName`s, or iterables thereof to samplers, e.g. `Gibbs(x=>HMC(), y=>MH())`, `Gibbs(@varname(x) => HMC(), @varname(y) => MH())`, `Gibbs((:x, :y) => NUTS(), :z => MH())`. This allows more granular specification of which sampler to use for which variable.
```

Since it can be iterables of symbols or iterables of varnames

src/mcmc/gibbs.jl
```diff
@@ -365,6 +379,9 @@ function Gibbs(algs::InferenceAlgorithm...)
     return Gibbs(varnames, map(set_selector ∘ drop_space, algs))
 end
 
+# This disambiguates a method ambiguity. To be removed when the above deprecated one is.
+Gibbs() = Gibbs([], [])
```
penelopeysm, line 383: You could avoid having this stray definition by changing the deprecated one to

```julia
function Gibbs(alg1::InferenceAlgorithm, other_algs::InferenceAlgorithm...)
    algs = [alg1, other_algs...]
    # ...
```

src/mcmc/gibbs.jl
```diff
@@ -292,11 +292,36 @@ function set_selector(x::RepeatSampler)
...
+Gibbs(@varname(x) => NUTS(), @varname(y) => MH())
+Gibbs((@varname(x), :y) => NUTS(), :z => MH())
+```
+
+Currently only variable names without indexing are supported, so for instance
+`Gibbs(@varname(x[1]) => NUTS())` does not work. This will hopefully change in the future.
+
 # Fields
 $(TYPEDFIELDS)
 """
```
penelopeysm, line 327: I can't comment on the line below. However, is it possible to provide more specific bounds on the type parameters `V` and `A`? For example, could `V` be `AbstractVector{<:AbstractVector{VarName}}` or something like that, and likewise for `A`?

That way, I'd feel better about removing the `_maybevec` calls below. I presume that they aren't needed because we now know something about what `V` should be.

penelopeysm:
Otherwise, super happy with this, I like APIs that are precise as opposed to ones that let you do whatever 👍😄

src/mcmc/gibbs.jl
mhauru: Nice, done

src/mcmc/gibbs.jl
mhauru: `map` isn't lazy, you need `Iterators.map` for that, so I can't think of a case where these would be different, but I like yours better stylistically, so done.

src/mcmc/gibbs.jl
mhauru: That seems appropriate given that these are being enforced/set in the inner constructor. I made it be `Gibbs{V<:AbstractVector{<:AbstractVector{<:VarName}},A<:AbstractVector}`. The reason I didn't specify an element type for `A` is that the eltype I think would have to be `Union{InferenceAlgorithm,AbstractSampler}`, and I don't trust the robustness of that type hiearchy/system, i.e. I'm not confident no one would ever want to plug in something different. Let me know what you think.

src/mcmc/gibbs.jl
penelopeysm: ahh, good to know 😅

src/mcmc/gibbs.jl
mhauru: Done

src/mcmc/gibbs.jl
penelopeysm: I can't think of any situation where someone would want something different, so I probably wouldn't mind putting in `AbstractVector{<:Union{InferenceAlgorithm,AbstractSampler}}`, but not super fussed. InferenceAlgorithm might eventually be on its way out anyway

penelopeysm: > could `V` be `<: AbstractVector{<:AbstractVector{<:VarName}}`

Looks like maybe that was not a correct assumption, oops

mhauru: I was actually wondering whether I should be using tuples rather than vectors. Will give this another think. Is there a chance that using tuples would allow some compiler optimisation? Opinions welcome.

mhauru: Today I learned that Julia's tuple types are covariant, unlike as far as I know any other Julia type. See the new tuple-based type signature.

I'm still a bit unsure if we should in fact make the elements of `Gibbs.varnames` be tuples as well. We currently end up with types like `Tuple{Vector{AbstractPPL.VarName{sym, typeof(identity)} where sym}}` when doing `Gibbs((:a, :b) => sampler)`, which seems like an unfortunate loss of concreteness. Note sure if that concreteness has any impact though. The argument against tuples is the fear that someone will have a reason to do `Gibbs([@varname(x[i]) for i in 1:100] => sampler)`, which would result in a 100-element tuple, which would probably be bad. One way out would be to not force tuple vs array choice, remove the type constraint altogether and leave it up to the user.

penelopeysm: Is there a reason why a 100-element tuple would be worse than a 100-element vector?

If not, it seems to me that `V <: NTuple{N,Tuple{Vararg{<:AbstractPPL.VarName}}}` works quite nicely.

penelopeysm: Technically, I'd prefer it even more if the elements of `varnames` and `samplers` were bundled together, but that's an internal change so I don't think it needs to be done in this PR. Right now the NTuple length parameter ensures that the lengths are the same which I like, but it seems unidiomatic to me that we accept a list of Pairs (which already guarantee that the length is the same) only to split them up and then zip them back up again in `AbstractMCMC.step`.

penelopeysm: (If it's of any interest, I did some unscientific benchmarking based on the first code block of https://jkrumbiegel.com/pages/2020-10-31-tuples-and-vectors/; even for long vectors/tuples tuples are still faster)

```julia
julia> using Chairmarks

julia> difference_matrix(points) = [p1 .- p2 for p1 in points, p2 in points]
difference_matrix (generic function with 1 method)

julia> @be [rand() for _ in 1:100] _ difference_matrix _
Benchmark: 7524 samples with 1 evaluation
 min    4.833 μs (3 allocs: 78.203 KiB)
 median 7.125 μs (3 allocs: 78.203 KiB)
 mean   7.057 μs (3 allocs: 78.203 KiB, 0.51% gc time)
 max    188.292 μs (3 allocs: 78.203 KiB, 92.04% gc time)

julia> @be tuple([rand() for _ in 1:100]) _ difference_matrix _
Benchmark: 21225 samples with 30 evaluations
 min    43.033 ns (4 allocs: 1008 bytes)
 median 54.167 ns (4 allocs: 1008 bytes)
 mean   70.760 ns (4 allocs: 1008 bytes, 0.37% gc time)
 max    9.938 μs (4 allocs: 1008 bytes, 94.30% gc time)
```

mhauru: >Is there a reason why a 100-element tuple would be worse than a 100-element vector?

I worry about type inference getting very slow, or giving up and treating it as `NTuple{100, Any}`. I don't actually know if either of those would happen, but I feel uneasy around huge tuples for type inference reasons.

penelopeysm: (For what it's worth, I did some benchmarking with the procedure in https://jkrumbiegel.com/pages/2020-10-31-tuples-and-vectors/. As the length increases, the relative performance benefit of tuples becomes smaller, but even for 500-element arrays/tuples the tuples are still marginally faster:)

```julia
julia> using Chairmarks

julia> difference_matrix(points) = [p1 .- p2 for p1 in points, p2 in points]
difference_matrix (generic function with 1 method)

julia> @be tuple([rand() for _ in 1:100]...) _ difference_matrix _
Benchmark: 4365 samples with 3 evaluations
 min    4.486 μs (3 allocs: 78.203 KiB)
 median 6.722 μs (3 allocs: 78.203 KiB)
 mean   7.335 μs (3 allocs: 78.203 KiB, 2.89% gc time)
 max    361.889 μs (3 allocs: 78.203 KiB, 96.87% gc time)

julia> @be [rand() for _ in 1:100] _ difference_matrix _
Benchmark: 2909 samples with 2 evaluations
 min    4.896 μs (3 allocs: 78.203 KiB)
 median 7.208 μs (3 allocs: 78.203 KiB)
 mean   15.347 μs (3 allocs: 78.203 KiB, 1.58% gc time)
 max    17.437 ms (3 allocs: 78.203 KiB, 99.92% gc time)

julia> @be tuple([rand() for _ in 1:500]...) _ difference_matrix _
Benchmark: 440 samples with 1 evaluation
 min    174.042 μs (3 allocs: 1.907 MiB)
 median 178.583 μs (3 allocs: 1.907 MiB)
 mean   226.531 μs (3 allocs: 1.907 MiB, 12.47% gc time)
 max    2.603 ms (3 allocs: 1.907 MiB, 85.92% gc time)

julia> @be [rand() for _ in 1:500] _ difference_matrix _
Benchmark: 388 samples with 1 evaluation
 min    159.417 μs (3 allocs: 1.907 MiB)
 median 201.917 μs (3 allocs: 1.907 MiB)
 mean   240.440 μs (3 allocs: 1.907 MiB, 11.50% gc time)
 max    693.375 μs (3 allocs: 1.907 MiB, 58.45% gc time)
```

penelopeysm: > type inference

Ehhh, I see. 

Given that the types inside the Gibbs constructor aren't user-facing, maybe we should punt this to a later PR with some proper benchmarking to see if it makes any difference?

mhauru: Currently the only things that are done with the elements of `Gibbs.varnames` are that they are passed to `DynamicPPL.subset` and `make_conditional`, both of which strictly require an `AbstractVector`. Whether we should make those functions, too, be happy with `Tuple`s is a fair question, but that feels like out of scope for this PR. Consequently, I would stick with the elements being arrays rather than tuples for now.

penelopeysm: Yeah, happy to stick to that!

mhauru: I wrote my message without seeing yours, but seems like we reached the same conclusion independently. I think storing `Pair`s would also be an improvement over the current setup of two tuples, but maybe for another PR as well?

penelopeysm (approved):
No more comments on my part 🙂

PR #2458, opened 2025-01-08
Remove selector/space stuff

mhauru:
This PR aims to remove all use of the `Selector`/Gibbs ID/`space` mechanisms that the old Gibbs sampler relied on. Methods for functions like `getspace` will still be defined for compatibility with the current DPPL version, but will not do anything. The goal is that after this has been merged a new release of DPPL could be made that removes the same mechanisms from `VarInfo` without anything breaking on the Turing.jl side.


penelopeysm:
I feel like intuitively I'd have removed it from DPPL first, is there a particular reason you started here?

src/mcmc/particle_mcmc.jl
```diff
@@ -15,38 +15,36 @@ Sequential Monte Carlo sampler.
...
+Create a sequential Monte Carlo sampler of type [`SMC`](@ref).
 
 If the algorithm for the resampling step is not specified explicitly, systematic resampling
 is performed if the estimated effective sample size per particle drops below 0.5.
 """
-function SMC(resampler=AdvancedPS.ResampleWithESSThreshold(), space::Tuple=())
-    return SMC{space,typeof(resampler)}(resampler)
+function SMC(resampler=AdvancedPS.ResampleWithESSThreshold())
+    return SMC{typeof(resampler)}(resampler)
 end
```
penelopeysm, line 38:
```suggestion
SMC() = SMC(AdvancedPS.ResampleWithESSThreshold())
```

src/mcmc/particle_mcmc.jl
```diff
@@ -15,38 +15,36 @@ Sequential Monte Carlo sampler.
...
 $(TYPEDFIELDS)
 """
-struct SMC{space,R} <: ParticleInference
+struct SMC{R} <: ParticleInference
     resampler::R
+
+    # Specifying this constructor stops the creation of default constructors. We rather
+    # define our own outer constructors below.
+    SMC{R}(resampler::R) where {R} = new{R}(resampler)
 end
```
penelopeysm, line 24: Does this explicit inner constructor have any effect? As far as I can see, it has the same behaviour as the default inner constructor that Julia generates, so we could omit it and it wouldn't be any different.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,38 +15,36 @@ Sequential Monte Carlo sampler.
...
 
 """
-    SMC(space...)
-    SMC([resampler = AdvancedPS.ResampleWithESSThreshold(), space = ()])
-    SMC([resampler = AdvancedPS.resample_systematic, ]threshold[, space = ()])
+    SMC()
+    SMC([resampler = AdvancedPS.ResampleWithESSThreshold()])
```
penelopeysm, line 28:
```suggestion
    SMC([resampler = AdvancedPS.ResampleWithESSThreshold()])
```

The first line is a sub-case of the second.

src/mcmc/particle_mcmc.jl
```diff
@@ -184,46 +182,39 @@ Particle Gibbs sampler.
...
+    # define our own outer constructors below.
+    PG{R}(nparticles::Int, resampler::R) where {R} = new{R}(nparticles, resampler)
 end
 
 """
-    PG(n, space...)
-    PG(n, [resampler = AdvancedPS.ResampleWithESSThreshold(), space = ()])
-    PG(n, [resampler = AdvancedPS.resample_systematic, ]threshold[, space = ()])
+    PG(n)
+    PG(n, [resampler = AdvancedPS.ResampleWithESSThreshold()])
```
penelopeysm, line 198:
```suggestion
    PG(n, [resampler = AdvancedPS.ResampleWithESSThreshold()])
```

src/mcmc/particle_mcmc.jl
```diff
@@ -184,46 +182,39 @@ Particle Gibbs sampler.
...
-struct PG{space,R} <: ParticleInference
+struct PG{R} <: ParticleInference
     """Number of particles."""
     nparticles::Int
     """Resampling algorithm."""
     resampler::R
+
+    # Specifying this constructor stops the creation of default constructors. We rather
+    # define our own outer constructors below.
+    PG{R}(nparticles::Int, resampler::R) where {R} = new{R}(nparticles, resampler)
```
penelopeysm, line 193: Same here

src/mcmc/particle_mcmc.jl
```diff
@@ -384,31 +375,21 @@ function DynamicPPL.assume(
...
-        end
-        lp = logpdf_with_trans(dist, r, istrans(vi, vn))
+    if ~haskey(vi, vn)
+        r = rand(trng, dist)
+        push!!(vi, vn, r, dist, spl)
+    elseif is_flagged(vi, vn, "del")
+        unset_flag!(vi, vn, "del") # Reference particle parent
+        r = rand(trng, dist)
+        vi[vn] = DynamicPPL.tovec(r)
+        DynamicPPL.setgid!(vi, spl.selector, vn)
```
penelopeysm, line 385: Is this line needed?

src/mcmc/sghmc.jl
```diff
@@ -37,22 +36,17 @@ Tianqi Chen, Emily Fox, & Carlos Guestrin (2014). Stochastic Gradient Hamiltonia
...
+function SGHMC(;
     learning_rate::Real,
     momentum_decay::Real,
     adtype::ADTypes.AbstractADType=Turing.DEFAULT_ADTYPE,
 )
     _learning_rate, _momentum_decay = promote(learning_rate, momentum_decay)
-    return SGHMC{typeof(adtype),space,typeof(_learning_rate)}(
+    return SGHMC{typeof(adtype),typeof(_learning_rate)}(
         _learning_rate, _momentum_decay, adtype
     )
```
penelopeysm, line 47:
```suggestion
    return SGHMC(_learning_rate, _momentum_decay, adtype)
```

Are the type parameters needed here? It seems unnecessarily verbose

src/mcmc/sghmc.jl
```diff
@@ -193,12 +182,10 @@ Dynamics. In: Proceedings of the 28th International Conference on Machine Learni
...
 """
-function SGLD(
-    space::Symbol...;
-    stepsize=PolynomialStepsize(0.01),
-    adtype::ADTypes.AbstractADType=Turing.DEFAULT_ADTYPE,
+function SGLD(;
+    stepsize=PolynomialStepsize(0.01), adtype::ADTypes.AbstractADType=Turing.DEFAULT_ADTYPE
 )
-    return SGLD{typeof(adtype),space,typeof(stepsize)}(stepsize, adtype)
+    return SGLD{typeof(adtype),typeof(stepsize)}(stepsize, adtype)
```
penelopeysm, line 188:
```suggestion
    return SGLD(stepsize, adtype)
```

penelopeysm (commented):
I suppose that the selector features in DPPL are almost purely motivated by this, so from that perspective it does make sense to me to remove this first. It's just a bit awkward with the testing machinery because you'd have to manually run the Turing testsuite with a dev'd version of DPPL..(?) Also, CI won't run the Turing integration tests.

Anyway, just some tiny tidyup suggestions below.

src/mcmc/mh.jl
```diff
@@ -108,45 +104,41 @@ mean(chain)
...
+                push!(prop_syms, s[1])
+                push!(props, proposal(s[2]))
+            elseif length(proposals) == 1
+                # If we hit this block, check to see if it's
+                # a run-of-the-mill proposal or covariance
+                # matrix.
+                prop = proposal(s)
+
+                # Return early, we got a covariance matrix.
+                return new{typeof(prop)}(prop)
```
penelopeysm, line 127: This feels confusing to me, it could probably be handled better with a new method, but maybe a case for another time too.

mhauru: Is it the early return that confuses you? It confuses me. It's been there for at least ~5 years though, and I just left it be for now.

src/mcmc/particle_mcmc.jl
mhauru: It was there to deal with a conflict between the default constructor that takes no type parameters and one of the explicit outer constructors. With your improved way of writing the default value for `resampler` it's no longer needed, so I removed it.

src/mcmc/particle_mcmc.jl
mhauru: Ditto, removed.

src/mcmc/particle_mcmc.jl
mhauru: Hmm, nice catch. I _hope_ not. It shouldn't? I tried removing it, and a bunch of other calls to `setgid!` and `updategid!`, let's see what the test suite says.

src/mcmc/particle_mcmc.jl
penelopeysm: Oh, ... I'm confused. I'll have to think about it, or you'll have to explain it to me some time, but I'm relieved that it's not needed!

src/mcmc/mh.jl
penelopeysm: Yeah, that and the call to `proposal()`, which is a bit opaque. But it's definitely out of scope for this PR.

src/mcmc/particle_mcmc.jl
mhauru: The default constructors include `SMC(resampler)`. Trying to then later define the outer constructor `SMC(resampler=default_value)` resulted in the default inner constructor being overshadowed, and Aqua (rightfully) complained. Defining the singular inner constructor `SMC{R}(resampler::R)` stopped the default constructor from being defined. In retrospect this was quite a silly solution, but that was the logic of it.

penelopeysm (approved):
Happy to merge!

mhauru:
Thanks @penelopeysm! Let's hold off for a while, since 0.36 hasn't even been released yet.

mhauru:
Changed the base branch to `breaking` to collect multiple breaking changes into a single branch/PR.

PR #2459, opened 2025-01-10
Update for DynamicPPL 0.33 and 0.34

penelopeysm:
This PR:

- removes the implementation of `predict`, which has been moved to DynamicPPL (https://github.com/TuringLang/DynamicPPL.jl/pull/716)
- modifies the signature of `values_as_in_model` as needed

DPPL 0.34 removed certain methods of `values_as_in_model`, but those weren't used by Turing, so this release will be compatible with both 0.33 and 0.34.


penelopeysm:
@sunxd3 Pinging you since the main breaking change (apart from me messing with `values_as_in_model`) was `predict`. Is it ok to just remove the function here entirely? :)

sunxd3:
>  Is it ok to just remove the function here entirely?

I think so. The tests for `predict` are [here](https://github.com/TuringLang/Turing.jl/blob/py/dppl-0.33/test/mcmc/utilities.jl). Not certain we want to keep a subset of them just to test the interface (probably not necessary).

sunxd3:
Another question: currently [Turing's `predict` is imported from `StatsBase`](https://github.com/TuringLang/Turing.jl/blob/a84c9a1b08eb0ce8d68defcb31c92c1753f36de1/src/Turing.jl#L75), while DPPL's `predict` is from `AbstractPPL`. Because `AbstractPPL` is implementing `StatsBase.predict`, so Turing's `predict` should be same as PDDL's. 

But should be start using AbstractPPL's `predict`?

penelopeysm:
> tests

Oh, good catch - the tests should definitely be only in DynamicPPL, I don't think there's a need to keep them here. Will remove

> imports

That's a good point! I was wondering the same about `sample`, actually, which is from AbstractMCMC but that's in turn also from StatsBase (I think). I'll investigate this properly on Monday if I get the time:)

penelopeysm:
I looked into it a bit and I feel like the most sensible option is this:

- in DynamicPPL, overload and export AbstractPPL.predict
- in Turing, import and re-export DynamicPPL.predict (I _think_ it should all resolve to the same function?).

Since we fixed the immediate issue with `:=`, would you feel comfortable re-exporting predict from DPPL?

sunxd3:
On the importing/exporting front, I don't have more to add to this PR.

But maybe delete the tests for `predict`?

penelopeysm:
Yup, I got rid of the tests already!

Re the exports I just realised AbstractPPL doesn't export predict either, so that's why we probably have to import it from StatsBase. let's keep it that way 😬

PR #2464, opened 2025-01-14
Replace Gibbs inner loop with recursion

### LINKED ISSUE #2445: Gibbs performance regression on Julia v1.11
Since #2328 was merged, the time it takes to run `mcmc/gibbs` is far longer than the time it used to take to run `mcmc/gibbs` plus `experimental/gibbs`.

Need to investigate the cause for this.

mhauru:
Fixes #2445

This seems to help with the type inference regression in Julia v1.11. #2445 should be fixed anyway in v1.11.3, but never hurts to help the type inference machinery, let the compiler unroll the loop if it feels like it.

Also made the same change in `initialstep`, for consistency, and to neatly ensure that `states` is always a tuple. Also, in the process, removed some initialisation of the varinfo in `initialstep`, that was repeating work done in https://github.com/TuringLang/DynamicPPL.jl/blob/e673b69210e85b60199f2ccd8165226cb03cf040/src/sampler.jl#L113.

### DIFF
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -654,14 +654,10 @@ function gibbs_step_recursive(
     # going to be a significant expense anyway.
     # Set the state of the current sampler, accounting for any changes made by other
     # samplers.
-    state = setparams_varinfo!!(
-        conditioned_model, sampler, state, vi
-    )
+    state = setparams_varinfo!!(conditioned_model, sampler, state, vi)
 
     # Take a step with the local sampler.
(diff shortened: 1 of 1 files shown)


mhauru:
The discrepancy in test suite runtimes between 1.10 and 1.11 is now gone, as expected.

penelopeysm (approved):

penelopeysm:
I supposeeeee we should wait for CI to run again, because I don't know if my renaming broke something, but otherwise happy

PR #2465, opened 2025-01-14
variable naming / destructuring

penelopeysm:
I thought I'd put some code suggestions in a PR rather than review comments, because to suggest all these changes would have been fiddly and error-prone.

Basically, I'm proposing these changes:

- the `_local` variable names are a bit repetitive and have lost their meaning since the loop was extracted into a function, so I got rid of most of them.
- I prefer to call the top thing `varname_tuples`, as that better reflects the fact that there are two nested levels of containers (as opposed to `varnames` which suggests only one level)
- using `h, t... = container` as opposed to explicit indexing `[2:end]`, as I think it's cleaner and better reflects what we're trying to do (pluck off the first element and keep the rest for later)

### DIFF
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -429,14 +429,14 @@ recursively on the remaining samplers, until no samplers remain. Return the glob
 and a tuple of initial states for all component samplers.
 """
 function gibbs_initialstep_recursive(
-    rng, model, varname_tuples, samplers, vi, states=(); initial_params=nothing, kwargs...
+    rng, model, varname_vecs, samplers, vi, states=(); initial_params=nothing, kwargs...
 )
     # End recursion
-    if isempty(varname_tuples) && isempty(samplers)
+    if isempty(varname_vecs) && isempty(samplers)
         return vi, states
     end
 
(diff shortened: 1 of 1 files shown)


penelopeysm:
Of course, feel free to push back on any of it like normal review comments:)

src/mcmc/gibbs.jl
```diff
@@ -429,49 +429,49 @@ recursively on the remaining samplers, until no samplers remain. Return the glob
 and a tuple of initial states for all component samplers.
 """
 function gibbs_initialstep_recursive(
-    rng, model, varnames, samplers, vi, states=(); initial_params=nothing, kwargs...
+    rng, model, varname_tuples, samplers, vi, states=(); initial_params=nothing, kwargs...
```
mhauru, line 432: I agree that the old name is not good, but should the new one be `varname_vecs`, since it's a tuple of vectors?

mhauru (changes requested):
Happy with the other changes, but see comment above for `varname_tuples`.

src/mcmc/gibbs.jl
penelopeysm: Oops, good point. For some reason I had in my mind it was a vec of tuples

PR #2471, opened 2025-01-20
Return NaN for negative ModeResult variance estimates

frankier:
Here's a modified example that gets negative estimates for variance of some parameters (`coefficients_versicolor[3]`):

```
using Turing
using RDatasets
using StatsPlots
using MLDataUtils: shuffleobs, splitobs, rescale!
using NNlib: softmax
using FillArrays
using LinearAlgebra
using Random
Random.seed!(0);

using Optim
using StatsBase

data = RDatasets.dataset("datasets", "iris");
data[rand(1:size(data, 1), 20), :]
species = ["setosa", "versicolor", "virginica"]
data[!, :Species_index] = indexin(data[!, :Species], species)
data[rand(1:size(data, 1), 20), [:Species, :Species_index]]
trainset, testset = splitobs(shuffleobs(data), 0.5)
features = [:SepalLength, :SepalWidth, :PetalLength, :PetalWidth]
target = :Species_index

train_features = Matrix(trainset[!, features])
test_features = Matrix(testset[!, features])
train_target = trainset[!, target]
test_target = testset[!, target]

μ, σ = rescale!(train_features; obsdim=1)
rescale!(test_features, μ, σ; obsdim=1);

@model function logistic_regression(x, y, σ)
    n = size(x, 1)
    length(y) == n ||
        throw(DimensionMismatch("number of observations in `x` and `y` is not equal"))

    # Priors of intercepts and coefficients.
    intercept_versicolor ~ Normal(0, σ)
    intercept_virginica ~ Normal(0, σ)
    coefficients_versicolor ~ MvNormal(Zeros(4), σ^2 * I)
    coefficients_virginica ~ MvNormal(Zeros(4), σ^2 * I)

    # Compute the likelihood of the observations.
    values_versicolor = intercept_versicolor .+ x * coefficients_versicolor
    values_virginica = intercept_virginica .+ x * coefficients_virginica
    for i in 1:n
        # the 0 corresponds to the base category `setosa`
        v = softmax([0, values_versicolor[i], values_virginica[i]])
        y[i] ~ Categorical(v)
    end
end;

model = logistic_regression(train_features, train_target, 1)
mle_estimate = Optim.optimize(model, MLE())
println(coeftable(mle_estimate))
```

Without this PR, this will throw a `DomainError` in `coeftable` when calling getting the stderr of `coefficients_versicolor[3]`.


mhauru:
Thanks @frankier, I agree that the current situation where `coeftable` fails with `DomainError` is not optimal, and this is an improvement. I wonder if we could be even more explicit though, and in our method for `coeftable`, catch the `DomainError`, print out a warning explaining that the solution seems to have negative variance and thus you should be very suspicious of your result, and then return a table with stderr as NaN. @frankier, as a user, do you think that would be helpful? This would also save us introducing a new dependency that we only use on a single line.

frankier:
Yes, I think on balance that would be better.

I think there is also the possibility of getting a SingularException in `inv`, which I guess also indicates model identifiability (and thus optimization) problems. So I guess it's better to catch these, aggregate them and report how it failed alongside the table.

I'll update this PR to work this way soon.

src/optimisation/Optimisation.jl
```diff
@@ -228,11 +229,47 @@ end
...
+    local stderrors
+    if numerrors_warnonly
+        infmat = StatsBase.informationmatrix(m)
+        local vcov
+        try
+            vcov = inv(infmat)
+        catch e
+            if isa(e, LinearAlgebra.SingularException)
+                stderrors = fill(NaN, length(m.values))
+                notes = ["Info. matrix is singular" for _ in 1:length(m.values)]
```
mhauru, line 246: Is there a reason to print this out `length(m.values)` times?

src/optimisation/Optimisation.jl
```diff
@@ -228,11 +229,47 @@ end
 
 # Various StatsBase methods for ModeResult
 
-function StatsBase.coeftable(m::ModeResult; level::Real=0.95)
+function StatsBase.coeftable(m::ModeResult; level::Real=0.95, numerrors_warnonly::Bool=true)
```
mhauru, line 232: Since we introduce a new argument compared to the usual `StatsBase.coeftable` method, could this have a docstring?

src/optimisation/Optimisation.jl
mhauru, line 246:
```suggestion
                notes = fill("Information matrix is singular", length(m.values))
```

src/optimisation/Optimisation.jl
```diff
@@ -228,11 +229,47 @@ end
...
+        catch e
+            if isa(e, LinearAlgebra.SingularException)
+                stderrors = fill(NaN, length(m.values))
+                notes = ["Info. matrix is singular" for _ in 1:length(m.values)]
+            else
+                rethrow(e)
+            end
+        else
+            stderrors = []
+            vars = LinearAlgebra.diag(vcov)
```
mhauru, line 252:
```suggestion
            vars = LinearAlgebra.diag(vcov)
            stderrors = eltype(vars)[]
```

src/optimisation/Optimisation.jl
```diff
@@ -228,11 +229,47 @@ end
 
 # Various StatsBase methods for ModeResult
 
-function StatsBase.coeftable(m::ModeResult; level::Real=0.95)
+function StatsBase.coeftable(m::ModeResult; level::Real=0.95, numerrors_warnonly::Bool=true)
     # Get columns for coeftable.
     terms = string.(StatsBase.coefnames(m))
     estimates = m.values.array[:, 1]
-    stderrors = StatsBase.stderror(m)
+    notes = nothing
```
mhauru, line 236:
```suggestion
    # If numerrors_warnonly is true, and if either the information matrix is singular or has
    # negative entries on its diagonal, then `notes` will be a list of strings for each
    # value in `m.values`, explaining why the standard error is NaN.
    notes = nothing
```

test/optimisation/Optimisation.jl
```diff
@@ -635,6 +635,24 @@ using Turing
...
+            return y ~ MvNormal(a .* x .+ b .* x, 1)
+        end
+
+        model = collinear(xs, ys)
+        mle_estimate = Turing.Optimisation.estimate_mode(model, MLE())
+        tab = coeftable(mle_estimate)
+        @assert isnan(tab.cols[2][1])
+        @assert tab.colnms[end] == "Error notes"
+        @assert occursin("singular", tab.cols[end][1])
+    end
```
mhauru, line 655: Could we also add a test where we check that for some harmless model where everything works out fine, `coeftable` returns the same result with and without `numerrors_warnonly`?

Also, a test case for negative variance would be great, to make sure tests cover all code paths.

mhauru (commented):
This is looking really nice @frankier! I had a few small proposals. Let me know once you're done making edits and I'm happy to merge. Also, I realise you're helping out on a volunteer basis, so if you don't have time to attend to my comments that's fine too, we can merge as is and I can add a couple of tests myself and call it done.

test/optimisation/Optimisation.jl
frankier: Yeah that's a good idea. Added! Agreed that a test with negative variance would be nice, but I'm not really sure how to construct a model which ends up with a negative variance estimate, and the original example where I got it is probably a bit heavy to add just for this. Since you offered to help finish off, maybe you could write this test? Thanks!

mhauru: Sure, I added a test for the negative variance case.

frankier:
Learnt something new from the test. LGTM!

mhauru:
The failing tests are the known x86 OOM issues we are working on, nothing to do with this PR. Merging.

Thanks so much @frankier! Always a pleasure to receive a community PR, and especially so when it's so well written.

PR #2472, opened 2025-01-20
Fix a Gibbs bug with models where linking changes variable dimension

mhauru:
Currently this PR only adds a test, the actual fix is to be implemented. The test reproduces the issue encountered in https://github.com/TuringLang/docs/pull/579.

I think the fix is to just remove this assertion in DPPL: https://github.com/TuringLang/DynamicPPL.jl/blob/938a69dfabb7089f52700ef01fa3a9f2d667b7d1/src/varinfo.jl#L532
I don't see a reason why, when merging two `VarInfo`s, all the variables should have the same dimension in both VIs being merged. However, due to [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) I'm uncomfortable making that change before I've understood why the assertion is there in the first place.


mhauru:
I'll update the compat bound for DynamicPPL once the release is out, and add the new test that prevents regression.

penelopeysm (approved):

PR #2474, opened 2025-01-21
Clean up exports

### LINKED ISSUE #2468: Clean up exports
https://github.com/TuringLang/Turing.jl/blob/24d5556786a835fd6fc3bdf11866cab93ff4cce3/src/Turing.jl#L80-L84

Turing re-exports the entire DynamicPPL module, as shown above. This means that any breaking change in DynamicPPL is also technically a breaking change in Turing.

However, we haven't really obeyed this at all: many breaking changes in DynamicPPL have been released as patch level bumps in Turing. This means that technically, a patch version of Turing could break any code that looks like

```julia
using Turing
Turing.DynamicPPL.foo(...)
```

where `foo` was something that was changed in a minor release of DynamicPPL.

*'But, Penny, who would write such horrible code?'* I hear you ask. Well, if nobody would do that, then there's no reason to re-export DynamicPPL ;)

### LINKED ISSUE #2481: `prefix` should be exported
### Minimal working example

```julia
julia> using Turing

julia> prefix
ERROR: UndefVarError: `prefix` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Hint: a global variable of this name also exists in DynamicPPL.
```

### Description

Unless there's a reason for us not to (clashes with other packages?), it seems that we forgot to add this to the export list when switching over the submodel stuff.

penelopeysm:
Closes #2481 
Closes #2468 

As mentioned in #2468, I don't think we need to export so many things. I've done a bunch of cuts here, with the intention that if we don't feel comfortable cutting something, we can easily revert specific commits. In my opinion, though, even though it looks like many sweeping changes, I don't think any of these are hugely controversial on their own.

Concretely, this PR does the following:

1. Stops re-exporting the DynamicPPL module.

2. Stops re-exporting `@logprob_str` and `@prob_str`, which were deprecated several months ago (we've killed off deprecated stuff in far shorter timespans)

3. ~~Stops re-exporting `DynamicPPL.LogDensityFunction`, this is an implementation detail that users probably don't need to construct themselves.~~

4. Bumps the Distributions.jl compat bound (from 0.23 to 0.25.77) and removes code that was required to support old versions of Distributions. Note that DynamicPPL already requires Distributions 0.25.

5. Removes the blanket re-export of Bijectors and Libtask, as well as the export of `Bijectors.ordered`. `ordered` is used in the [Gaussian mixture model tutorial](https://turinglang.org/docs/tutorials/gaussian-mixture-models), but even there it's imported directly from Bijectors, and I don't think it's unfair to expect users to manually import Bijectors if they want to use this function.

6. Removes the blanket re-export of AbstractMCMC, replacing it with an explicit export list of `sample`, `MCMCThreads`, `MCMCDistributed`, and `MCMCSerial`. Note that these are the only four identifiers which AbstractMCMC itself exports (https://github.com/TuringLang/AbstractMCMC.jl/blob/master/src/AbstractMCMC.jl) so the only change is that the module name itself (AbstractMCMC) is no longer accessible as `Turing.AbstractMCMC`.

7. For convenience, re-exports `LinearAlgebra.I`.

8. Exports `DynamicPPL.returned` and `DynamicPPL.prefix` (we forgot to re-export this when doing the submodel stuff).

Will let CI run before requesting a review.


mhauru:
I think I agree with all the changes listed in OP, although `LogDensityFunction` makes me pause and wonder how people might be using it. I don't think I should be the main reviewer here, the more senior members of the team understand user needs better.

Rather than merging straight to master, should we bundle this with #2458 to form a new breaking release to go out when we feel like is the right moment?

penelopeysm:
Happy to release both PRs together!

I can't say I feel particularly bad about LogDensityFunction tbh, it's still accessible via DynamicPPL with the same functionality, so we're really just asking people to import it from where it should be imported.

penelopeysm:
There's still the wholesale export of MCMCChains which I don't dare to touch. MCMCChains exports _a lot_ of stuff. https://github.com/TuringLang/MCMCChains.jl/blob/master/src/MCMCChains.jl

penelopeysm:
I'm just going to let CI run, but self-reminder, we should fix this too https://github.com/TuringLang/Turing.jl/issues/2481

penelopeysm:
I changed my mind on LogDensityFunction, it can stay exported. I suspect that over the course of AbstractMCMC refactoring it'll also start to play a larger role.
Nevertheless, I changed all the `Turing.LogDensityFunction` calls in the library to `DynamicPPL.LogDensityFunction` which feels a bit cleaner.

I think it's almost ready for review, will let CI run before requesting one though, as I'm not sure if there are still some imports I need to fix in the test suite.

src/essential/Essential.jl
```diff
@@ -18,11 +18,6 @@ using AdvancedPS: AdvancedPS
 include("container.jl")
 
 export @model,
-    @varname,
-    AutoForwardDiff,
-    AutoReverseDiff,
-    AutoMooncake,
-    @logprob_str,
-    @prob_str
+    @varname, AutoForwardDiff, AutoReverseDiff, AutoMooncake, @logprob_str, @prob_str
```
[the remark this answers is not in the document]
penelopeysm: tldr is no: https://github.com/domluna/JuliaFormatter.jl/issues/358

We could try doing
```julia
export a
export b
export c
```
instead. Part of me might actually prefer this (even though the keyword export is spammed) because it's more resistant towards moving lines up/down (you can't have a trailing comma at the end of the export list, so
```julia
export
    a,
    b,
    c,
```
doesn't work)

docs/src/api.md
```diff
@@ -108,52 +112,37 @@ OrderedLogistic
...
+| `decondition`              | [`AbstractPPL.decondition`](@extref)                                                                                              | Remove conditioning on data                        |
+| `conditioned`              | [`DynamicPPL.conditioned`](@extref)                                                                                               | Return the conditioned values of a model           |
+| `fix`                      | [`DynamicPPL.fix`](@extref)                                                                                                       | Fix the value of a variable                        |
+| `unfix`                    | [`DynamicPPL.unfix`](@extref)                                                                                                     | Unfix the value of a variable                      |
+| `OrderedDict`              | [`OrderedCollections.OrderedDict`](https://juliacollections.github.io/OrderedCollections.jl/dev/ordered_containers/#OrderedDicts) | An ordered dictionary                              |
```
mhauru, line 145:
```suggestion
| `OrderedDict`              | [`OrderedCollections.OrderedDict`](https://juliacollections.github.io/OrderedCollections.jl/stable/#OrderedDicts) | An ordered dictionary                              |
```
The old one is broken, and also linked to dev rather than stable.

mhauru (changes requested):
Built docs look good except for one link.

docs/src/api.md
penelopeysm: Good catch, thanks 👀 I'll actually replace it with an `extref` as well to protect against link fragility

src/essential/Essential.jl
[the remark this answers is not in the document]
mhauru: I don't mind spamming `export` either, or having a few per line that relate to each other if they fit on one line.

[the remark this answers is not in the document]
penelopeysm: I changed it to

```julia
export @model
export @varname
export AutoForwardDiff, AutoReverseDiff, AutoMooncake
```

and JuliaFormatter seems happy ☺️

(Also removed that stray logprob_str and prob_str, oops.)

HISTORY.md
```diff
@@ -23,6 +23,22 @@ Turing.jl v0.37 uses DynamicPPL v0.35, which brings with it several breaking cha
...
 For more details about all of the above, see the changelog of DynamicPPL [here](https://github.com/TuringLang/DynamicPPL.jl/releases/tag/v0.35.0).
 
+### Export list
+
+Turing.jl's export list has been cleaned up a fair bit. You may need to import things more explicitly than before.
+
+  - The `DynamicPPL` and `AbstractMCMC` modules: You will need to `import` or `using Module` yourself, which in turn means that they have to be made available in your project environment.
+
+  - `@logprob_str` and `@prob_str` have been removed following a long deprecation period.
+  - We no longer re-export everything from `Bijectors`. To get around this, add `using Bijectors` at the top of your script (but we prefer if you selectively import).
```
mhauru, line 33:
```suggestion
  - We no longer re-export everything from `Bijectors` and `Libtask`. To get around this, add `using Bijectors` or `using Libtask` at the top of your script (but we recommend using more selective imports).
```

HISTORY.md
```diff
@@ -23,6 +23,22 @@ Turing.jl v0.37 uses DynamicPPL v0.35, which brings with it several breaking cha
 
 For more details about all of the above, see the changelog of DynamicPPL [here](https://github.com/TuringLang/DynamicPPL.jl/releases/tag/v0.35.0).
 
+### Export list
+
+Turing.jl's export list has been cleaned up a fair bit. You may need to import things more explicitly than before.
+
+  - The `DynamicPPL` and `AbstractMCMC` modules: You will need to `import` or `using Module` yourself, which in turn means that they have to be made available in your project environment.
```
mhauru, line 30:
```suggestion
  - The `DynamicPPL` and `AbstractMCMC` modules are no longer exported. You will need to `import DynamicPPL` or `using DynamicPPL: DynamicPPL` (likewise `AbstractMCMC`) yourself, which in turn means that they have to be made available in your project environment.
```

HISTORY.md
```diff
@@ -23,6 +23,22 @@ Turing.jl v0.37 uses DynamicPPL v0.35, which brings with it several breaking cha
...
+
+  - `@logprob_str` and `@prob_str` have been removed following a long deprecation period.
+  - We no longer re-export everything from `Bijectors`. To get around this, add `using Bijectors` at the top of your script (but we prefer if you selectively import).
+    
+      + If you were using `Bijectors.ordered`, even Bijectors does not (currently) export this. You will have to manually import it with `using Bijectors: ordered`.
```
mhauru, line 35:
```suggestion
      + We no longer export `Bijectors.ordered`. If you were using `ordered`, even Bijectors does not (currently) export this. You will have to manually import it with `using Bijectors: ordered`.
```

mhauru (changes requested):
Some proposed changes to HISTORY.md, feel free to reject or edit.

mhauru (approved):
Thanks Penny!

PR #2489, opened 2025-02-18
Fixing doc import by prefixing `DynamicPPL` to `predict` 

sunxd3:


sunxd3:
unfortunately, I don't think fixed the issue (https://turinglang.org/Turing.jl/dev/api/#Predictions) , will start another one to address this.

PR #2490, opened 2025-02-20
remove LogDensityProblemsAD

penelopeysm:
this PR removes LogDensityProblemsAD in favour of using the new interface in https://github.com/TuringLang/DynamicPPL.jl/pull/806


test/essential/ad.jl
```diff
@@ -1,250 +0,0 @@
...
-        likelihood = make_logjoint(model, DynamicPPL.LikelihoodContext())
-        target(x) = likelihood(x; unlinked=true)
-
-        H_f = ForwardDiff.hessian(target, zeros(2))
-        H_r = ReverseDiff.hessian(target, zeros(2))
-        @test H_f == [1.0 0.0; 0.0 1.0]
-        @test H_f == H_r
-    end
-
-    @testset "memoization: issue #1393" begin
```
penelopeysm, line 214: i don't think this testset is relevant anymore because the memoisation code has been long removed https://github.com/TuringLang/Turing.jl/pull/1414

test/essential/ad.jl
```diff
@@ -1,250 +0,0 @@
...
-            d = Normal(0.0, i)
-            data = rand(d, N)
-            chn = sample(
-                demo(data), NUTS(0.65; adtype=AutoReverseDiff(; compile=true)), 1000
-            )
-            @test mean(Array(chn[:sigma])) ≈ std(data) atol = 0.5
-        end
-    end
-
-    @testset "ReverseDiff compiled without linking" begin
```
penelopeysm, line 231: this testset is fully contained in DynamicPPL already

test/essential/ad.jl
```diff
@@ -1,250 +0,0 @@
...
-        else
-            @test zygoteℓ isa
-                LogDensityProblemsAD.LogDensityProblemsADZygoteExt.ZygoteGradientLogDensity
-        end
-        @test zygoteℓ.ℓ === ℓ
-        ∇E2 = LogDensityProblems.logdensity_and_gradient(zygoteℓ, x)[2]
-        @test sort(∇E2) ≈ grad_FWAD atol = 1e-9
-    end
-
-    @testset "general AD tests" begin
```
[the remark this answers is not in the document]
mhauru: It's there already, in `test/compat/ad.jl`.

ext/TuringOptimExt.jl
```diff
@@ -162,17 +156,20 @@ Estimate a mode, i.e., compute a MLE or MAP estimate.
...
 )
     # Convert the initial values, since it is assumed that users provide them
     # in the constrained space.
-    f = Accessors.@set f.varinfo = DynamicPPL.unflatten(f.varinfo, init_vals)
-    f = Accessors.@set f.varinfo = DynamicPPL.link(f.varinfo, model)
-    init_vals = DynamicPPL.getparams(f)
+    # TODO(penelopeysm): As with in src/optimisation/Optimisation.jl, unclear
+    # whether initialisation is really necessary at all
+    vi = DynamicPPL.unflatten(f.ldf.varinfo, init_vals)
+    vi = DynamicPPL.link(vi, f.ldf.model)
```
mhauru, line 170: Is there a difference between `model` and `f.ldf.model`? If not, is there a reason to keep the first argument at all?

src/mcmc/Inference.jl
```diff
@@ -48,7 +48,6 @@ import AdvancedPS
 import Accessors
 import EllipticalSliceSampling
 import LogDensityProblems
-import LogDensityProblemsAD
```
mhauru, line 51: Can we also remove LDPAD from Project.toml?

src/optimisation/Optimisation.jl
```diff
@@ -95,22 +94,40 @@ function DynamicPPL.tilde_observe(
 end
 
 """
-    OptimLogDensity{M<:DynamicPPL.Model,C<:Context,V<:DynamicPPL.VarInfo}
+    OptimLogDensity{M<:DynamicPPL.Model,V<:DynamicPPL.VarInfo,C<:OptimizationContext,AD<:ADTypes.AbstractADType}
 
 A struct that stores the negative log density function of a `DynamicPPL` model.
+
+TODO(penelopeysm): It _doesn't_ really store the negative, does it? It's more like we
```
mhauru, line 101: Do we want the TODO note to be user visible in the docstring, or rather a comment above?

src/optimisation/Optimisation.jl
```diff
@@ -95,22 +94,40 @@ function DynamicPPL.tilde_observe(
...
+overrode logdensity to give the negative logdensity.
 """
-const OptimLogDensity{M<:DynamicPPL.Model,C<:OptimizationContext,V<:DynamicPPL.VarInfo,AD} = Turing.LogDensityFunction{
-    M,V,C,AD
+struct OptimLogDensity{
```
mhauru, line 104: What's the benefit of wrapping in a struct rather than type aliasing?

src/optimisation/Optimisation.jl
```diff
@@ -454,28 +468,34 @@ function estimate_mode(
     end
 
     # Create an OptimLogDensity object that can be used to evaluate the objective function,
-    # i.e. the negative log density. Set its VarInfo to the initial parameters.
-    log_density = let
```
mhauru, line 458: Any particular reason to remove the `let`. I kinda liked how it marked `inner_context` and `ctx` as being strictly for defining `log_density` and making them not be in scope beyond. Or does it just not flow as nicely with how `varinfo` is modified now?

test/ext/OptimInterface.jl
```diff
@@ -143,7 +143,6 @@ using Turing
         DynamicPPL.TestUtils.demo_assume_multivariate_observe_literal,
         DynamicPPL.TestUtils.demo_dot_assume_observe_submodel,
         DynamicPPL.TestUtils.demo_dot_assume_observe_matrix_index,
-        DynamicPPL.TestUtils.demo_dot_assume_matrix_dot_observe_matrix,
```
mhauru, line 146: Why no longer test on this?

test/optimisation/Optimisation.jl
```diff
@@ -545,7 +545,6 @@ using Turing
         DynamicPPL.TestUtils.demo_assume_multivariate_observe_literal,
         DynamicPPL.TestUtils.demo_dot_assume_observe_submodel,
         DynamicPPL.TestUtils.demo_dot_assume_observe_matrix_index,
-        DynamicPPL.TestUtils.demo_dot_assume_matrix_dot_observe_matrix,
```
mhauru, line 548: Why no longer test on this?

src/optimisation/Optimisation.jl
penelopeysm: More the latter. Because we can't modify LDF with Accessors.set anymore we have to only construct it at the bottom of the function, so the context has to be in scope for quite a while. We could still wrap the whole thing in a let if you think that's better!

src/optimisation/Optimisation.jl
penelopeysm: I think I shall write a better docstring and solve this properly

test/ext/OptimInterface.jl
mhauru: 🤦 my bad

src/optimisation/Optimisation.jl
penelopeysm: Avoids things like inadvertently making LogDensityFunction callable 🙃 basically OptimLogDensity has to obey a different interface to LDF (such as being callable) so I feel that a wrapper keeps this clean

ext/TuringOptimExt.jl
penelopeysm: Good point, I didn't notice that. It does look redundant at first glance - I'll check more carefully later.

src/optimisation/Optimisation.jl
mhauru: Nah, it's fine, just curious if you disagreed with the original decision to use `let` on some grounds.

src/optimisation/Optimisation.jl
mhauru: Happy with that

ext/TuringOptimExt.jl
penelopeysm: Removed the argument now

test/essential/ad.jl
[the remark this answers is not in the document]
penelopeysm: Okay what that's so weird. Doubly happy to delete it then

test/test_utils/ad_utils.jl
```diff
@@ -202,7 +202,7 @@ end
 All the ADTypes on which we want to run the tests.
 """
 adbackends = [
-    Turing.AutoForwardDiff(; chunksize=0),
+    Turing.AutoForwardDiff(),
```
mhauru, line 205: What's the significance of not setting `chunksize`?

PR #2491, opened 2025-02-20
more test fixes

penelopeysm:
some more fixes to get more tests to pass, i commented on a couple of bits & the rest should be fairly self explanatory


src/mcmc/gibbs.jl
```diff
@@ -438,7 +438,9 @@ function setparams_varinfo!!(
     state::TuringState,
     params::AbstractVarInfo,
 )
-    logdensity = DynamicPPL.setmodel(state.ldf, model, sampler.alg.adtype)
+    logdensity = DynamicPPL.LogDensityFunction(
+        model, state.ldf.varinfo, state.ldf.context; adtype=sampler.alg.adtype
+    )
```
penelopeysm, line 443: `setmodel` technically still exists, but it doesn't take the `adtype` argument anymore, and i think it's clearer to reconstruct the LDF (which is what `setmodel` does anyway)

test/mcmc/Inference.jl
```diff
@@ -560,7 +560,7 @@ using Turing
 
         @model function vdemo7()
             x = Array{Real}(undef, N, N)
-            return x .~ [InverseGamma(2, 3) for i in 1:N]
+            return x ~ filldist(InverseGamma(2, 3), N, N)
```
penelopeysm, line 563: I'm kind of unsure as to whether we really need to keep the `vdemo` models around. As far as I can tell, they're just legacy tests that try out different 'forms' of vectorisation and presumably the timings that are printed are meant to be some marker of their performance. But surely this is covered by DynamicPPL demo models already? I personally wouldn't mind if we just deleted the entire block

mhauru (approved):

Issue #1308, opened 2020-06-04
Incorrect handling of type parameter in HMC


smonsays:
As discussed on [discourse.julialang.org](https://discourse.julialang.org/t/inference-using-hmc-on-a-hidden-markov-model/40719), the type parameter is incorrectly handled by `HMC()` in the following example:
```julia
using LinearAlgebra
using Turing

# Linear Model
Δt = 0.1
A = [1.0 Δt; 0.0 1.0]
Q = [0.25 * Δt^4 0.5 * Δt^3; 0.5 * Δt^3 Δt^2] + 1e-6 * I

H = [1.0 0.0; 0.0 1.0;]
R = 0.1 * I

@model function hmm(states, ::Type{T} = Array{Float64}) where {T}
    # Initialize arrays for the time-series
    if states === missing
        states = T(undef, 2, 100)
    end
    observations = T(undef, 2, 100)

    # Prior for first step
    states[:,1] ~ MvNormal([0.0, 0.5], 0.001 * I)
    observations[:,1] ~ MvNormal(states[:,1], R)

    for t in 2:size(states, 2)
        states[:,t] ~ MvNormal(A * states[:,t-1], Q)
        observations[:,t] ~ MvNormal(states[:,t], R)
    end

    return states
end

hmm_prior = hmm(missing)
states = hmm_prior()
hmm_conditioned = hmm(states)

chain = sample(hmm_conditioned, HMC(0.1, 5), 1000)
``` 
This will throw the following error while trying to sample `observations`:
```
TypeError: in typeassert, expected Float64, got ForwardDiff.Dual{Nothing,Float64,10}
```

PR #2494, opened 2025-02-25
More fixes for DynamicPPL 0.35

penelopeysm:
I think these should be the final fixes.

There's one more Aqua test for persistent tasks that fails. I'm not 100% sure, but I think this is because of the fact that DynamicPPL 0.35 hasn't yet been released, and hence it has to grab the DynamicPPL code from a github repo + install it. My suspicion is that once DPPL is released and we update the Project.toml accordingly this will start to pass.

Comments on individual tests to follow


test/mcmc/hmc.jl
```diff
@@ -201,28 +201,6 @@ using Turing
...
-            m = Matrix{T}(undef, 2, 3)
-            return m .~ MvNormal(zeros(2), I)
-        end
-        @test sample(StableRNG(seed), mwe1(), HMC(0.2, 4; adtype=adbackend), 100) isa Chains
-
-        @model function mwe2(::Type{T}=Matrix{Float64}) where {T}
-            m = T(undef, 2, 3)
-            return m .~ MvNormal(zeros(2), I)
-        end
-        @test sample(StableRNG(seed), mwe2(), HMC(0.2, 4; adtype=adbackend), 100) isa Chains
```
penelopeysm, line 216: These tests point to an issue that says `.~ is broken for MvNormal`. https://github.com/TuringLang/DynamicPPL.jl/issues/27

Since `.~` is no longer meant to work with MvNormal, I think it is fine to remove them.

We could of course change them to filldist or a loop, but that would be testing a different thing and not really needed.

test/mcmc/hmc.jl
```diff
@@ -201,28 +201,6 @@ using Turing
...
-
-        # https://github.com/TuringLang/Turing.jl/issues/1308
-        @model function mwe3(::Type{T}=Array{Float64}) where {T}
-            m = T(undef, 2, 3)
-            return m ~ filldist(MvNormal(zeros(2), I), 3)
-        end
-        @test sample(StableRNG(seed), mwe3(), HMC(0.2, 4; adtype=adbackend), 100) isa Chains
```
penelopeysm, line 223: This test is trying to catch ForwardDiff bugs when using `~ MvNormal`, see the MWE in https://github.com/TuringLang/Turing.jl/issues/1308, however

1. The DynamicPPL demo models contain `~ MvNormal`
2. The demo models are already tested with ForwardDiff (plus other AD backends) in the DynamicPPL test suite

So I think this is also safe to remove

test/mcmc/Inference.jl
```diff
@@ -440,15 +440,6 @@ using Turing
...
 
-        @model function vdemo2(x)
-            μ ~ MvNormal(zeros(size(x, 1)), I)
-            @. x ~ $(MvNormal(μ, I))
-        end
-
-        D = 2
-        alg = HMC(0.01, 5; adtype=adbackend)
-        res = sample(StableRNG(seed), vdemo2(randn(D, 100)), alg, 10)
-
```
penelopeysm, line 451: I think vdemo2 can just be removed, since it just directly expands to `x .~ MvNormal` which is now forbidden, and if the point is to test interpolation inside `@.`, then that should be added to the compiler tests in DynamicPPL.

If you'd like me to add that test over there, let me know and I can open an issue

test/mcmc/hmc.jl
```diff
@@ -201,28 +201,6 @@ using Turing
...
-    end
-
```
mhauru, line 225: Why remove these? The last one seems unrelated to `.~`, and I'm not sure if the particular use of `T=Matrix` from the first two is tested elsewhere. Sorry to nitpick, but I'm a bit paranoid about removing tests, especially regression tests, unless they are clearly pointless.

penelopeysm:
We can let CI run to confirm, but I believe that this should fix all the failing tests (except for the Aqua one which I noted in my original comment).
Even if we aren't ready to merge this into Turing main, I think it suggests that DPPL 0.35 is 'good enough' to be merged into DPPL main? 😄

mhauru:
> Even if we aren't ready to merge this into Turing main, I think it suggests that DPPL 0.35 is 'good enough' to be merged into DPPL main? 😄

Agreed.

test/mcmc/hmc.jl
mhauru: Didn't refresh before sending, reading your comments now.

penelopeysm:
Spoke too soon about tests anyway :/

test/mcmc/hmc.jl
mhauru: Is the `::Type{T}=Array{Float64}` significant? It's in the MWE of #1308 and isn't done anywhere else in the test suite of Turing.jl or DPP (nor is `::Type{T}=Matrix`).

penelopeysm: Possibly, I didn't think about that one. I guess the point of having the type parameter is to allow someone to specify the eltype, and although we don't have a Matrix example in DPPL we do have plenty of Vector examples in the demo models. I can't necessarily see that it would make a difference but ...

I understand the hesitance to delete old tests but also I would really like to get rid of them – even if they are still relevant, at the very least they should be tested in DPPL not here. Maybe we should do this in a systematic way, like having a big issue in DPPL where we list all the syntax features that we want to test in models and then we can slowly add them into demo models?

test/mcmc/Inference.jl
```diff
@@ -492,7 +483,7 @@ using Turing
         N = 3
         @model function vdemo7()
             x = Array{Real}(undef, N, N)
-            @. x ~ [InverseGamma(2, 3) for i in 1:N]
+            x ~ filldist(InverseGamma(2, 3), N, N)
         end
```
penelopeysm, line 487: I think this is a straightforward replacement

We could also do `x ~ product_distribution([product_distribution([InverseGamma(2, 3) for i in 1:N]) for j in 1:N])` if you prefer which seems a bit closer to the original

test/mcmc/hmc.jl
penelopeysm: (To clarify, I don't mind leaving the tests in Turing.jl for this PR, but I'd like us to set up a mechanism that allows us to start removing them with more confidence.)

mhauru: > I guess the point of having the type parameter is to allow someone to specify the eltype

This is necessary for AD:
```julia
julia> module MWE
       using DynamicPPL, LogDensityProblems, LogDensityProblemsAD, ADTypes, Distributions
       import ForwardDiff, ReverseDiff

       @model function f()
           x = Vector{Float64}(undef, 2)
           x .~ Normal()
       end

       m = f()
       ldf = DynamicPPL.LogDensityFunction(m)
       vi = VarInfo(m)
       adtype = ADTypes.AutoForwardDiff()
       g = LogDensityProblemsAD.ADgradient(adtype, ldf)
       @show LogDensityProblems.logdensity_and_gradient(g, randn(2))
       end
WARNING: replacing module MWE.
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{DynamicPPL.DynamicPPLTag, Float64}, Float64, 2})
```

test/mcmc/Inference.jl
mhauru: The nested `product_distribution` feels like it's more a test for Distributions.jl. Happy with the change you've made here.

Issue #2493, opened 2025-02-25
`step_warmup` for Gibbs (and RepeatSampler)


mhauru:
@joelkandiah pointed out that we are ignoring warm-up in Gibbs. We don't have an implementation of `step_warmup` for it, and it just falls back on the default implementation in AbstractMCMC, which calls `step`.

At the very least we should put in place an implementation that's a carbon copy of `step` but it calls `step_warmup` on the component samplers.

However, if you have `Gibbs(@varname(x) => Sampler1, @varname(y) => Sampler2)`, you might also imagine wanting to do different numbers of warm-up steps for `Sampler1` and `Sampler2`. Should we support that? Also, what if one does `Gibbs(@varname(x) => Sampler1, @varname(y) => RepeatSampler(Sampler2, 2))`, how should we count warm-up steps for `RepeatSampler`?

One solution would be to say that the number of warm-up steps should be a property of the sampler and not just an argument for the call to `sample`, though that would be a larger change to implement.

joelkandiah:
Some brief motivation for the issue (a bit more terse than above):

A quick example about why the number of warmup steps might matter (using adaptive samplers). 
It is frequently important in adaptive samplers to perform steps where adaptation doesn't occur. As such, it makes sense to implement these as warm_up steps (where elsewhere they may be considered the "burn-in" steps. However this isn't universal across all adaptive samplers: 
Cf. Algorithm 4 in https://link.springer.com/article/10.1007/s11222-008-9110-y (Andrieu and Thoms) and https://onlinelibrary.wiley.com/doi/10.1111/anzs.12344 (Spencer) (which requires no non-adaptive steps). 

Some initial testing of these algorithms suggests the Spencer samplermay perform better when "warmup" steps are used. So using these samplers (Spencer, and Andrieu and Thoms) together in a gibbs sampler would suggest a preference for each sampler to have different num_warmup steps.

In practice, it seems to have previously been the case in the literature that when using a single sampler rolled individually by users such as in https://gitlab.com/pjbirrell/real-time-mcmc the number of warmup steps is over the steps of the component sampler (equivalent to 1 gibbs step = 1 warmup/burnin step and so every component of the component sampler performs a warmup step during a warmup "gibbs" step). This comes with the note that when people write their own custom samplers they are frequently only using one type of sampler within the component sampler....

Perhaps this means syntax for both could be considered?

joelkandiah:
[...]

PR #2502, opened 2025-03-06
Make Gibbs work with step_warmup

mhauru:
Gibbs used to just ignore warm-up as a concept, and always called `step` rather than `step_warmup`. This was discussed in #2493. This PR makes Gibbs respect `num_warmup` by calling `step_warmup` on the component samplers whenever Gibbs itself is called with `step_warmup`. It also does the same thing for `RepeatSampler`. Note that this means that e.g. if you use `RepeatSampler(spl, 3)` and `num_warmup=100` `spl` will take 300 warm-up steps.

#2493 should be left open for now, because we might want to improve this so that one could specify different amounts of warm-up for different component samplers. This is just a stop-gap bug fix to at least do the minimal sensible thing.

### DIFF
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -417,7 +417,7 @@ function initial_varinfo(rng, model, spl, initial_params)
 
     # Update the parameters if provided.
     if initial_params !== nothing
-        vi = DynamicPPL.initialize_parameters!!(vi, initial_params, model)
+        vi = DynamicPPL.initialize_parameters!!(vi, initial_params, spl, model)
 
         # Update joint log probability.
         # This is a quick fix for https://github.com/TuringLang/Turing.jl/issues/1588
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -405,20 +405,75 @@ end
 
 varinfo(state::GibbsState) = state.vi
 
-function DynamicPPL.initialstep(
+"""
+Initialise a VarInfo for the Gibbs sampler.
+
+This is straight up copypasta from DynamicPPL's src/sampler.jl. It is repeated here to
+support calling both step and step_warmup as the initial step. DynamicPPL initialstep is
+incompatible with step_warmup.
+"""
+function initial_varinfo(rng, model, spl, initial_params)
+    vi = DynamicPPL.default_varinfo(rng, model, spl)
+
+    # Update the parameters if provided.
+    if initial_params !== nothing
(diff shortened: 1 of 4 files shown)


src/mcmc/gibbs.jl
```diff
@@ -620,10 +709,14 @@ end
...
 Run a Gibbs step for the first varname/sampler/state tuple, and recursively call the same
 function on the tail, until there are no more samplers left.
+
+The `step_function` argument should always be either AbstractMCMC.step or
+AbstractMCMC.step_warmup.
 """
 function gibbs_step_recursive(
     rng::Random.AbstractRNG,
     model::DynamicPPL.Model,
+    step_function::Function,
```
sunxd3, line 719: would this create some kind of type instability? my first order thought is that `Function` is abstract type

sunxd3 (commented):
Generally looks very good. One concern: from the coverage, `repeat_sampler` is not tested, is it worthwhile to add a test?

src/mcmc/gibbs.jl
mhauru: I think it should be fine, since each function has its own type. E.g.
```julia
julia> typeof(identity)
typeof(identity) (singleton type of function identity, subtype of Function)

julia> typeof(sin)
typeof(sin) (singleton type of function sin, subtype of Function)
```
The `::Function` bit just enforces that you can't pass as the `step_function` argument anything that isn't of type `Function`, but the compiler will still see the concrete type of the argument. I also checked that making this change didn't have a substantial impact on runtime.

mhauru:
Hmm, I'm confused by the coverage complaint, I thought these lines from `test/mcmc/gibbs.jl` would have tested those lines:
```julia
    num_reps = 2
    wuc = WarmupCounter()
    sample(m, Gibbs(:x => RepeatSampler(wuc, num_reps)), num_samples; num_warmup=num_warmup)
    @test wuc.warmup_init_count == 1
    @test wuc.non_warmup_init_count == 0
    @test wuc.warmup_count == num_warmup * num_reps
    @test wuc.non_warmup_count == (num_samples - 1) * num_reps
```
Need to check this next week.

mhauru:
@sunxd3, I think those comments are out of date, you can see coverage hitting those lines here: https://app.codecov.io/gh/TuringLang/Turing.jl/pull/2502?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=TuringLang

sunxd3 (approved):
LGTM! Thanks

PR #2529, opened 2025-04-07
test #2528

penelopeysm:
Testing that #2528 works correctly.

(I already tested it on my own repo. I just want to be triply sure.)


penelopeysm:
Look mum, it works.

PR #2535, opened 2025-04-27
DynamicPPL 0.36

### LINKED ISSUE #2403: Allow Gibbs sampler to have non-identity lenses for target variables
See conversation at https://github.com/TuringLang/Turing.jl/pull/2328#discussion_r1862506933

penelopeysm:
This PR adds compatibility with DynamicPPL 0.36.

Given that there are some reasonably breaking changes in the way that submodels behave, I've decided to increment the minor version number for Turing.

Because submodels now also prefix VarNames in a 'proper' way, submodel variables now have non-identity lenses. In order to allow Gibbs to sample models with submodels, this PR also fixes #2403 by letting GibbsContext store a tuple of VarNames rather than just Symbols. Performance is compared below.

There's some nasty lack of reproducibility in HMCDA sampling, which is causing the failures in `test/mcmc/Inference`. I haven't yet been able to repro this locally. However, I would like to deal with this in a separate PR, as the failure isn't introduced by this PR (the last CI run on main was already failing).

This PR supersedes the CompatHelper ones at #2501 #2532 #2533 

## Benchmarks

I'm not 100% convinced of these, because my benchmarks have returned wildly variable results upon being run multiple times / when I restart my Julia session. Sometimes when I run `sample(rng, model, spl, 10000)` it runs in < 1 second and sometimes it takes 4 seconds. (This is after running it multiple times in the same session, so it's not an issue of JIT compilation.) I think I would probably feel a little bit more reassured if somebody else could test them out.

However, broadly speaking, I think it's safe to say that I don't have any evidence of performance getting _worse_. Another noteworthy point is that CI for `test/mcmc/gibbs` (on ubuntu-latest, 1 thread) for this PR ran in 1 hr 25 min, which is the same time it took for the same job on `main`. So it seems unlikely that there are any serious performance implications (good or bad).

```julia
using AbstractMCMC: step
using DynamicPPL, Turing
using StableRNGs: StableRNG
using Chairmarks

@model function f()
    x ~ Normal()
    y ~ Normal()
end
rng = StableRNG(468)
model = f()
spl = Gibbs(@varname(x) => MH(), @varname(y) => MH())
dspl = DynamicPPL.Sampler(spl)

# This benchmark compares the initial step of MCMC
@be step($rng, $model, $dspl)

# Before: 91.250 μs
# After: 91.208 µs

# This compares the 2nd step onwards
_, st = step(rng, model, dspl)
@be step($rng, $model, $dspl, $st)

# Before: 66.500 μs
# After: 66.417 μs
```


test/mcmc/gibbs.jl
```diff
@@ -727,6 +723,39 @@ end
         end
     end
 
+    @testset "non-identity varnames" begin
+        struct Wrap{T}
+            a::T
+        end
+        @model function model1(::Type{T}=Float64) where {T}
```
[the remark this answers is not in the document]
penelopeysm: this is a JuliaFormatter v2 change, I've now pinned the format action to v1 so this shouldn't happen again

src/mcmc/gibbs.jl
```diff
@@ -67,26 +77,14 @@ struct GibbsContext{VNs,GVI<:Ref{<:AbstractVarInfo},Ctx<:DynamicPPL.AbstractCont
...
-        if !(DynamicPPL.NodeTrait(context) isa DynamicPPL.IsLeaf)
-            error("GibbsContext can only wrap a leaf context, not a $(context).")
-        end
-        return new{VNs,typeof(global_varinfo),typeof(context)}(global_varinfo, context)
-    end
-
     function GibbsContext(target_varnames, global_varinfo, context)
-        if !(DynamicPPL.NodeTrait(context) isa DynamicPPL.IsLeaf)
+        if !can_be_wrapped(context)
             error("GibbsContext can only wrap a leaf context, not a $(context).")
```
mhauru, line 82:
```suggestion
            error("GibbsContext can only wrap a leaf or prefix context, not a $(context).")
```

src/mcmc/gibbs.jl
```diff
@@ -129,7 +129,9 @@ function get_conditioned_gibbs(context::GibbsContext, vns::AbstractArray{<:VarNa
     return map(Base.Fix1(get_conditioned_gibbs, context), vns)
 end
 
-is_target_varname(::GibbsContext{VNs}, ::VarName{sym}) where {VNs,sym} = sym in VNs
+function is_target_varname(ctx::GibbsContext, vn::VarName)
+    return any(Base.Fix2(subsumes, vn), ctx.target_varnames)
+end
```
mhauru, line 134: Have you checked what impact this has on compilation and run times? I think the previous version got constant propagated away, so that `tilde_obssume` could compile-time optimise away the big if-statement of "is this a target variable or not". I think that was necessary for said `tilde_obssume` to be type stable. Unsure whether the same would happen with this version of `is_target_varname`. The necessary type information is there, in the case where only property or identity lenses are in play, but not sure if it's too much for the compiler to fold away. If the compiler can't do it, we could turn `is_target_varname` into a generated function and do it ourselves.

HISTORY.md
```diff
@@ -1,3 +1,23 @@
...
+@model outer() = a ~ to_submodel(inner())
+```
+
+`keys(VarInfo(outer()))` now returns `[@varname(a.x)]` instead of `[@varname(var"a.x")]`
+
+Furthermore, you can now either condition on the outer model like `outer() | (@varname(a.x) => 1.0)`, or the inner model like `inner() | (@varname(x) => 1.0)`.
+If you use the conditioned inner model as a submodel, the conditioning will still apply correctly.
+
+Please see [the DynamicPPL release notes](https://github.com/TuringLang/DynamicPPL.jl/releases/tag/v0.36.0) for fuller details.
+
```
mhauru, line 20: Shall we advertise here that Gibbs now works with non-identity lenses, and say a bit about performance implications, which I'm guessing are nill when using property lenses and possibly significant if using index lenses.

mhauru (commented):
This is great! Leaving it unapproved just because I want to confirm that nothing bad has happened to type stability and hence runtimes.

Might it also make sense to throw in some non-trivially lensed varnames in the test models of the `"type stability"` and `"Sampler call order"` testsets in `test/mcmc/gibbs.jl`?

src/mcmc/gibbs.jl
penelopeysm: Regarding constant propagation, it still works as long as there is no need to check `IndexLens`, as we were guessing:

```julia
using Turing

# helper function so that you don't have to construct a full GibbsContext
function is_target_varname(vn_tuple::Tuple{Vararg{VarName}}, vn::VarName)
    return any(Base.Fix2(subsumes, vn), vn_tuple)
end

julia> @code_warntype is_target_varname((@varname(a), @varname(b)), @varname(a))
MethodInstance for Turing.Inference.is_target_varname(::Tuple{AbstractPPL.VarName{:a, typeof(identity)}, AbstractPPL.VarName{:b, typeof(identity)}}, ::AbstractPPL.VarName{:a, typeof(identity)})
  from is_target_varname(vn_tuple::Tuple{Vararg{AbstractPPL.VarName}}, vn::AbstractPPL.VarName) @ Turing.Inference ~/ppl/lib/src/mcmc/gibbs.jl:132
Arguments
  #self#::Core.Const(Turing.Inference.is_target_varname)
  vn_tuple::Core.Const((a, b))
  vn::Core.Const(a)
Body::Bool
1 ─ %1 = Turing.Inference.any::Core.Const(any)
│   %2 = Base.Fix2::Core.Const(Base.Fix2)
│   %3 = Turing.Inference.subsumes::Core.Const(AbstractPPL.subsumes)
│   %4 = (%2)(%3, vn)::Core.Const(Base.Fix2{typeof(AbstractPPL.subsumes), AbstractPPL.VarName{:a, typeof(identity)}}(AbstractPPL.subsumes, a))
│   %5 = (%1)(%4, vn_tuple)::Core.Const(true)
└──      return %5

julia> @code_warntype is_target_varname((@varname(a.x), @varname(b)), @varname(a.x[1]))
MethodInstance for Turing.Inference.is_target_varname(::Tuple{AbstractPPL.VarName{:a, Accessors.PropertyLens{:x}}, AbstractPPL.VarName{:b, typeof(identity)}}, ::AbstractPPL.VarName{:a, ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.PropertyLens{:x}}})
  from is_target_varname(vn_tuple::Tuple{Vararg{AbstractPPL.VarName}}, vn::AbstractPPL.VarName) @ Turing.Inference ~/ppl/lib/src/mcmc/gibbs.jl:132
Arguments
  #self#::Core.Const(Turing.Inference.is_target_varname)
  vn_tuple::Core.Const((a.x, b))
  vn::AbstractPPL.VarName{:a, ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.PropertyLens{:x}}}
Body::Bool
1 ─ %1 = Turing.Inference.any::Core.Const(any)
│   %2 = Base.Fix2::Core.Const(Base.Fix2)
│   %3 = Turing.Inference.subsumes::Core.Const(AbstractPPL.subsumes)
│   %4 = (%2)(%3, vn)::Base.Fix2{typeof(AbstractPPL.subsumes), AbstractPPL.VarName{:a, ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.PropertyLens{:x}}}}
│   %5 = (%1)(%4, vn_tuple)::Core.Const(true)
└──      return %5

julia> @code_warntype is_target_varname((@varname(a[1]), @varname(b)), @varname(a[1]))
MethodInstance for Turing.Inference.is_target_varname(::Tuple{AbstractPPL.VarName{:a, Accessors.IndexLens{Tuple{Int64}}}, AbstractPPL.VarName{:b, typeof(identity)}}, ::AbstractPPL.VarName{:a, Accessors.IndexLens{Tuple{Int64}}})
  from is_target_varname(vn_tuple::Tuple{Vararg{AbstractPPL.VarName}}, vn::AbstractPPL.VarName) @ Turing.Inference ~/ppl/lib/src/mcmc/gibbs.jl:132
Arguments
  #self#::Core.Const(Turing.Inference.is_target_varname)
  vn_tuple::Tuple{AbstractPPL.VarName{:a, Accessors.IndexLens{Tuple{Int64}}}, AbstractPPL.VarName{:b, typeof(identity)}}
  vn::AbstractPPL.VarName{:a, Accessors.IndexLens{Tuple{Int64}}}
Body::Bool
1 ─ %1 = Turing.Inference.any::Core.Const(any)
│   %2 = Base.Fix2::Core.Const(Base.Fix2)
│   %3 = Turing.Inference.subsumes::Core.Const(AbstractPPL.subsumes)
│   %4 = (%2)(%3, vn)::Base.Fix2{typeof(AbstractPPL.subsumes), AbstractPPL.VarName{:a, Accessors.IndexLens{Tuple{Int64}}}}
│   %5 = (%1)(%4, vn_tuple)::Bool
└──      return %5
```

I'll see if I can find an example where this causes any notable slowdown for variables with IndexLens.

penelopeysm: Somewhat ironically, the one that uses IndexLens is faster

```julia
using Turing, DynamicPPL, Chairmarks, Random
using AbstractMCMC: step
rng = Random.default_rng()

@model function f()
    x = Vector{Float64}(undef, 2)
    x[1] ~ Normal()
    x[2] ~ Normal()
end
model_f = f()
spl_f = Gibbs(@varname(x[1]) => MH(), @varname(x[2]) => MH())
dspl_f = DynamicPPL.Sampler(spl_f)
_, st_f = step(rng, model_f, dspl_f)

@be step($rng, $model_f, $dspl_f) # median 83.375 μs

@be step($rng, $model_f, $dspl_f, $st_f) # median 60.708 μs


@model function g()
    x ~ Normal()
    y ~ Normal()
end
model_g = g()
spl_g = Gibbs(@varname(x) => MH(), @varname(y) => MH())
dspl_g = DynamicPPL.Sampler(spl_g)
_, st_g = step(rng, model_g, dspl_g)

@be step($rng, $model_g, $dspl_g) # median 98.708 μs

@be step($rng, $model_g, $dspl_g, $st_g) # median 66.708 μs
```

mhauru: Thanks for checking!

mhauru (approved):

PR #2546, opened 2025-05-08
Update docstrings to account for non-identity varnames

penelopeysm:
I'm silly and forgot to change these.


mhauru (approved):
Should have caught this in my review too, sorry.

Issue #2542, opened 2025-05-06
Gibbs Compilation Performance


joelkandiah:
The runtime of the compilation of the Gibbs sampler appears to be O(N_component_samplers), however some tests show that the cost to compile any given call to sample is high, and the cost of each individual component sampler is high.

I have attached a test file, which allows a user to pass in an argument for the number of components to use (between 3 and 9) where 17 Normal(0,1) variables are sampled. These are sampled using a default Metropolis Hastings sampler (i.e. from the prior) and no data is provided. We run this for two samples to allow the code to compile the sample call with and without varinfo being passed.

(Bad practice: On my local machine) This estimates a runtime of ~25-30 seconds per extra component sampler, where for 3 components the time taken is ~150 seconds and for 8 components the time taken is ~290 seconds.
This is significantly longer than using just an MH sampler without Gibbs ~10 seconds.

```julia
using Turing # v0.37.0
using Distributions

# --- Probabilistic Model Definition ---
@model function test_gibbs_individual()
    # Define the model parameters with Normal priors
    p1 ~ Normal(0, 1)
    p2 ~ Normal(0, 1)
    p3 ~ Normal(0, 1)
    p4 ~ Normal(0, 1)
    p5 ~ Normal(0, 1)
    p6 ~ Normal(0, 1)
    p7 ~ Normal(0, 1)
    p8 ~ Normal(0, 1)

    vp1 ~ Normal(0, 1)
    vp2 ~ Normal(0, 1)
    vp3 ~ Normal(0, 1)
    vp4 ~ Normal(0, 1)
    vp5 ~ Normal(0, 1)
    vp6 ~ Normal(0, 1)
    vp7 ~ Normal(0, 1)
    vp8 ~ Normal(0, 1)

    global_p1 ~ Normal(0, 1)

    return
end


# --- Gibbs Construction --- #
# Define a simple gibbs construction for 9 compartments (dynamic construction with the same number of parameters can be found in the attached .txt file
gibbs_MH = Gibbs(
    (@varname(global_p1),) => MH(),
    (@varname(p1), @varname(vp1)) => MH(),
    (@varname(p2), @varname(vp2)) => MH(),
    (@varname(p3), @varname(vp3)) => MH(),
    (@varname(p4), @varname(vp4)) => MH(),
    (@varname(p5), @varname(vp5)) => MH(),
    (@varname(p6), @varname(vp6)) => MH(),
    (@varname(p7), @varname(vp7)) => MH(),
[...]

Issue #2413, opened 2024-12-03
Remove (most) samplers being defined explicitly in Turing.jl


torfjelde:
We're duplicating a lot of code and a lot of effort by having a bunch of sampler (or rather, `InferenceAlgorithm`) implementations in Turing.jl itself.

There are a few reasons for this is / was the case:
1. The old approach of doing Gibbs sampling took an approach that required hooking into the `assume` and `observe` statements for samplers _and_ to mutate the varinfo in a particular, even if the functionality of the sampler itself (when used outside of Gibbs) didn't require it.
2. The samplers in Turing.jl would often offer more convenient constructors while the sampler packages themselves, e.g. AdvancedHMC.jl, would offer a more flexible but also more complicated interfaces.
3. `InferenceAlgorithm` allows us to overload the `sample` call explicitly to do some "non-standard" things, e.g. use `chain_type=MCMCChains.Chains` as the default, instead of `chain_type=Vector` as is default in AbstractMCMC.jl.

Everything but (3) is "easily" addressable (i.e. only requires dev-time, not necessarily any discussion on _how_ to do it):
- [ ] (1) is being addressed in #2328 (issue ref: #2318). This should therefore be addressed very soon.
- [ ] (2) should be addressed by simply moving any convenience constructors from Turing.jl itself into the respective package. There's no reason why we should keep convenient constructors in a different package (Turing.jl in this case) than the package implementing the samplers. Effort has been made towards this, e.g. https://github.com/TuringLang/AdvancedHMC.jl/pull/325, but we need to through all the samplers and check which have missing "convenience" constructors.
- [ ] (3) is somewhat tricky. There are a few aspects to this that we need to handle: a) how to default to `chain_type=Chains` for Turing.jl models (ref: https://github.com/TuringLang/AbstractMCMC.jl/pull/120, https://github.com/TuringLang/AbstractMCMC.jl/issues/118), b) how to allow extraction of other interesting information than _just_ the realizations for the variables from `sample` calls, and c) extraction of parameter names used to construct the chain. See the section below for more extensive discussion of this issue.

### Problem

Currently, all the samplers in Turing.jl have most of their code living outside of Turing.jl + inside Turing.jl we define a "duplicate" which is _not_ an `AbstractMCMC.AbstractSampler` (as typically expected by `AbstractMCMC.sample`), but instead a subtype of `Turing.Infernece.InferenceAlgorithm`:

[...]

PR #2550, opened 2025-05-15
Support DPPL 0.37

mhauru:
Currently in a very unfinished state.


yebai:
Lots of interface code changes here are unnecessary and are planned to be removed in https://github.com/TuringLang/Turing.jl/issues/2413 

Perhaps we should address https://github.com/TuringLang/Turing.jl/issues/2413 first.

mhauru:
That is tempting, to not waste time fixing code that's on its way out. I worry though that removing the samplers will take a while still, and in the meanwhile all the accumulator stuff, and other DPPL changes that build on it, would be held back from Turing.jl. For instance, introducing ValuesAsInModelAccumulator would cut our inference time in #2542 by half.

penelopeysm:
IMO, it goes both ways. Reducing sampler complexity would make this PR easier. On the other hand, merging this PR would also make it easier to remove the duplicate samplers. (To be precise, DPPL 0.37 would make it easier.)

I think DPPL 0.37 has taken a long time and we should prioritise this, rather than trying to squeeze in the changes to samplers.

I'm going to fix the merge conflicts and add a `[sources]` bit to Project.toml to point to the unreleased DPPL branch, so that CI can at least run on 1.11. (See https://pkgdocs.julialang.org/v1/toml-files/#The-%5Bsources%5D-section)

Note that 1.10 CI will always fail as `[sources]` isn't understood on 1.10.

test/mcmc/hmc.jl
```diff
@@ -171,6 +171,8 @@ using Turing
         @test Array(res1) == Array(res2) == Array(res3)
     end
 
+    # TODO(mhauru) Do we give up being able to sample from only prior/likelihood like this,
+    # or do we implement some way to pass `whichlogprob=:LogPrior` through `sample`?
```
penelopeysm: `sample(::LogDensityFunction)` would solve that, since we could set `getlogprior` in the LDF, but I guess that'll be the next release.

src/mcmc/hmc.jl
```diff
@@ -190,16 +190,7 @@ function DynamicPPL.initialstep(
...
     metric = metricT(length(theta))
-    ldf = DynamicPPL.LogDensityFunction(
-        model,
-        vi,
-        # TODO(penelopeysm): Can we just use leafcontext(model.context)? Do we
-        # need to pass in the sampler? (In fact LogDensityFunction defaults to
-        # using leafcontext(model.context) so could we just remove the argument
-        # entirely?)
-        DynamicPPL.SamplingContext(rng, spl, DynamicPPL.leafcontext(model.context));
-        adtype=spl.alg.adtype,
```
penelopeysm, line 201: As established in (e.g.) https://github.com/TuringLang/DynamicPPL.jl/issues/955#issuecomment-2990775251 SamplingContext for Hamiltonians was never overloaded so it is equivalent to just use DefaultContext in the LDF.

src/mcmc/hmc.jl
```diff
@@ -242,10 +233,12 @@ function DynamicPPL.initialstep(
...
     if t.stat.is_accept
         vi = DynamicPPL.unflatten(vi, t.z.θ)
-        vi = setlogp!!(vi, t.stat.log_density)
+        # TODO(mhauru) Is setloglikelihood! the right thing here?
+        vi = setloglikelihood!!(vi, t.stat.log_density)
     else
         vi = DynamicPPL.unflatten(vi, theta)
-        vi = setlogp!!(vi, log_density_old)
+        # TODO(mhauru) Is setloglikelihood! the right thing here?
+        vi = setloglikelihood!!(vi, log_density_old)
```
penelopeysm: Not fully sure what to do here, `t.stat.log_density` is a joint and not sure how to decompose it into likelihood + prior.

There is an argument that we should re-evaluate the model anyway because `t.stat.log_density` provides a logpdf in linked space whereas for user-facing purposes we want the invlinked logpdf https://github.com/TuringLang/Turing.jl/issues/2617. And I don't immediately see how to calculate this without re-evaluating the model anyway

src/optimisation/Optimisation.jl
```diff
@@ -43,75 +43,129 @@ Concrete type for maximum a posteriori estimation. Only used for the Optim.jl in
...
-    OptimizationContext{C<:AbstractContext} <: AbstractContext
+    LogPriorWithoutJacobianAccumulator{T} <: DynamicPPL.AbstractAccumulator
 
-The `OptimizationContext` transforms variables to their constrained space, but
-does not use the density with respect to the transformation. This context is
-intended to allow an optimizer to sample in R^n freely.
+Exactly like DynamicPPL.LogPriorAccumulator, but does not include the log determinant of the
+Jacobian of any variable transformations.
+
+Used for MAP optimisation.
```
penelopeysm: I am actually quite confused as to why this is needed (and OptimizationContext before it) because if the VarInfo is linked, calling `getlogp` on it won't include the logjac. (This is the source of the problem in #2617)

This would only be needed if we wanted to unconditionally ignore logjac even if the VarInfo is unlinked.

penelopeysm: Oh, I think there's a terminology mismatch here. I must be thinking of logjac of the invlink transform (and that's also what #2617 refers to), whereas this is talking about logjac of the link transform.

i.e. you want to plug in values in linked space and get the invlinked logp as the optimisation target, ignoring the logjac required to transform values to linked space.

whereas the way I've been seeing it is that we can sample in linked space but we need to make sure to include the logjac of the invlink transform so that we get the invlinked logp.

penelopeysm:
Because Turing re-exports some things that were changed in DPPL 0.37 (for example, `LogDensityFunction`), this has to go into `breaking`. (It probably philosophically should anyway, since I think this might be the biggest makeover that DPPL has had in a while.)

src/mcmc/hmc.jl
penelopeysm: Fixed by re-evaluating.

penelopeysm:
This is getting too unwieldy, unfortunately.

Project.toml
```diff
@@ -90,3 +90,6 @@ julia = "1.10.2"
 [extras]
 DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: Must not forget to remove this.

ext/TuringDynamicHMCExt.jl
```diff
@@ -78,7 +75,9 @@ function DynamicPPL.initialstep(
 
     # Update the variables.
     vi = DynamicPPL.unflatten(vi, Q.q)
-    vi = DynamicPPL.setlogp!!(vi, Q.ℓq)
+    # TODO(DPPL0.37/penelopeysm): This is obviously incorrect. Fix this.
+    vi = DynamicPPL.setloglikelihood!!(vi, Q.ℓq)
+    vi = DynamicPPL.setlogprior!!(vi, 0.0)
```
mhauru: Must not forget to fix this.

src/mcmc/external_sampler.jl
```diff
@@ -126,7 +126,13 @@ function make_updated_varinfo(
...
         last(DynamicPPL.evaluate!!(f.model, new_varinfo, f.context))
     else
-        DynamicPPL.setlogp!!(new_varinfo, new_logp)
+        # TODO(DPPL0.37/penelopeysm) This is obviously wrong. Note that we
+        # have the same problem here as in HMC in that the sampler doesn't
+        # tell us about how logp is broken down into prior and likelihood.
+        # We should probably just re-evaluate unconditionally. A bit
+        # unfortunate.
+        DynamicPPL.setlogprior!!(new_varinfo, 0.0)
+        DynamicPPL.setloglikelihood!!(new_varinfo, new_logp)
```
mhauru: Must not forget to fix this.

src/mcmc/gibbs.jl
```diff
@@ -178,24 +178,26 @@ function DynamicPPL.tilde_assume(context::GibbsContext, right, vn, vi)
         DynamicPPL.tilde_assume(child_context, right, vn, vi)
     elseif has_conditioned_gibbs(context, vn)
         # Short-circuit the tilde assume if `vn` is present in `context`.
-        value, lp, _ = DynamicPPL.tilde_assume(
+        # TODO(mhauru) Fix accumulation here. In this branch anything that gets
+        # accumulated just gets discarded with `_`.
+        value, _ = DynamicPPL.tilde_assume(
```
mhauru: Must not forget to fix this.

src/mcmc/hmc.jl
```diff
@@ -239,13 +232,21 @@ function DynamicPPL.initialstep(
...
         vi = DynamicPPL.unflatten(vi, t.z.θ)
-        vi = setlogp!!(vi, t.stat.log_density)
+        # Re-evaluate to calculate log probability density.
+        # TODO(penelopeysm): This seems a little bit wasteful. Unfortunately,
+        # even though `t.stat.log_density` contains some kind of logp, this
+        # doesn't track prior and likelihood separately but rather a single
+        # log-joint (and in linked space), so which we have no way to decompose
+        # this back into prior and likelihood. I don't immediately see how to
+        # solve this without re-evaluating the model.
+        _, vi = DynamicPPL.evaluate!!(model, vi)
```
mhauru: Must not forget to attend to this and make a call on whether to pay the cost of re-evaluating or live without a logp.

src/mcmc/mh.jl
```diff
@@ -304,25 +304,27 @@ function propose!!(
...
     trans, _ = AbstractMCMC.step(rng, densitymodel, mh_sampler, prev_trans)
 
     # TODO: Make this compatible with immutable `VarInfo`.
     # Update the values in the VarInfo.
+    # TODO(DPPL0.37/penelopeysm): This is obviously incorrect. We need to
+    # re-evaluate the model.
     set_namedtuple!(vi, trans.params)
-    return setlogp!!(vi, trans.lp)
+    vi = DynamicPPL.setloglikelihood!!(vi, trans.lp)
+    return DynamicPPL.setlogprior!!(vi, 0.0)
```
mhauru: Must not forget to fix this.

src/mcmc/mh.jl
```diff
@@ -339,22 +341,25 @@ function propose!!(
...
         ),
     )
     trans, _ = AbstractMCMC.step(rng, densitymodel, mh_sampler, prev_trans)
 
-    return setlogp!!(DynamicPPL.unflatten(vi, trans.params), trans.lp)
+    # TODO(DPPL0.37/penelopeysm): This is obviously incorrect. We need to
+    # re-evaluate the model.
+    vi = DynamicPPL.unflatten(vi, trans.params)
+    vi = DynamicPPL.setloglikelihood!!(vi, trans.lp)
+    return DynamicPPL.setlogprior!!(vi, 0.0)
```
mhauru: Must not forget to fix this.

src/mcmc/particle_mcmc.jl
```diff
@@ -428,6 +429,8 @@ function trace_local_rng_maybe(rng::Random.AbstractRNG)
     end
 end
 
+# TODO(DPPL0.37/penelopeysm) The whole tilde pipeline for particle MCMC needs to be
+# thoroughly fixed.
```
mhauru: Must not forget to fix this.

test/Project.toml
```diff
@@ -77,3 +77,6 @@ StatsBase = "0.33, 0.34"
 StatsFuns = "0.9.5, 1"
 TimerOutputs = "0.5"
 julia = "1.10"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: Must not forget to remove this.

mhauru (commented):
I went through the code and highlighted with comments any bits that seemed like they weren't done.

With @penelopeysm we agreed that this PR is getting unwieldy. All the obvious interface changes should now be done. The things that don't work are getting logp from samplers, Gibbs, and particle MCMC. We should make individual PRs to fix those, merge them into this one, and then check the comments I left to make sure everything is done. Except for the comments raised, I consider the code in here currently to be ready to merge once tests pass.

@penelopeysm feel free to likewise flag any parts of the code that we must not forget to attend to before merging this. Hopefully we can then not do another review of what's here now.

penelopeysm:
Seems to me that there are several stages:

- [x] Introduce logjac acc in DPPL
   TuringLang/DynamicPPL.jl#998
- [x] Use that here
   #2630
- [x] Fix sampler logp issue here
   #2630
- [x] Fix Gibbs
   #2628
- [x] Fix pMCMC
   #2625 
   #2629

I'll go off and get started on the logjac acc, shouldn't be too hard since it's all the same code as before.

src/optimisation/Optimisation.jl
penelopeysm: simplified in #2630 now

src/mcmc/mh.jl
penelopeysm: fixed #2630

src/mcmc/mh.jl
penelopeysm: fixed #2630

src/mcmc/hmc.jl
penelopeysm: fixed #2630 (albeit not as cleanly as I would like, but I'll open another issue for that)

src/mcmc/gibbs.jl
penelopeysm: fixed #2628

src/mcmc/external_sampler.jl
penelopeysm: fixed #2630 -- the re-evaluation happens in Transition and happily is now also tested because of #2640

ext/TuringDynamicHMCExt.jl
penelopeysm: Fixed in #2630

src/mcmc/particle_mcmc.jl
penelopeysm: fixed in #2625, #2629

test/mcmc/gibbs.jl
```diff
@@ -565,40 +565,98 @@ end
...
+            @model function dynamic_bernoulli()
+                b ~ Bernoulli()
+                if b
+                    dists = [Normal(1.0)]
+                else
+                    dists = [Normal(1.0), Normal(2.0)]
+                end
+                return x ~ product_distribution(dists)
+            end
+            model = dynamic_bernoulli()
```
penelopeysm, line 619: To be discussed @ 4 Aug meeting

penelopeysm:
Apologies for force push, I did it on the wrong branch. I undid it (force pushed back to the old commit) so it shouldn't affect any local git history. CI is now failing because DynamicPPL's breaking branch doesn't exist... hmmmm

penelopeysm:
I'd be happy to merge https://github.com/TuringLang/DynamicPPL.jl/pull/1005, release DPPL, then remove the `[sources]` thing here and maybe merge this into breaking (I don't think there are any other breaking changes that this will hold up).

mhauru:
I agree on that plan.

penelopeysm (approved):
I'm very pleased with where we are on this, would be happy to merge into breaking and then main. Was there anything else you wanted to add to 0.40?

HISTORY.md
```diff
@@ -1,6 +1,57 @@
...
+**Particle Gibbs**
+
+Previously, only 'true' observations (i.e., `x ~ dist` where `x` is a model argument or conditioned upon) would trigger resampling of particles.
+Specifically, there were two cases where resampling would not be triggered:
+
+  - Calls to `@addlogprob!`
+  - Gibbs-conditioned variables: e.g. `y` in `Gibbs(:x => PG(20), :y => MH())`
+
+Turing 0.40 changes this such that both of the above cause resampling.
+(The second case follows from the changes to the Gibbs sampler, see above.)
```
mhauru, line 45: So this breaks PG-within-Gibbs for some dynamic models, right? Or does it just mean that things that produced incorrect results before now crash? I've already lost the plot on that conversation... Wondering if we should say something about it here.

mhauru: Maybe a note saying something like "Note that, when using PG within Gibbs, since `~` statements for variables handled by other component samplers now trigger resampling for PG, and since all particles in PG must have the same number of resampling events, some dynamic models that used to run will no longer do so with Turing.jl v0.40 when using PG within Gibbs. If you need to sample from such a model, we recommend using Turing.jl v0.39, but also thinking very carefully about what the expected behaviour of the model is, and checking that Turing is sampling from it correctly. We would also welcome any GitHub issues highlighting such problems. Our support for dynamic models is incomplete and likely to through changes."

penelopeysm: Sure. I think we should be careful about the meaning of dynamic, because I think there are at least three distinct uses of the word which get conflated: what we mean is that the dimensions of the model are not fixed. Otherwise happy to include a note.

> I've already lost the plot on that conversation

I'm still not fully convinced that the 'correct' result (i.e. sample from the prior) is correct, but even if we put that aside, it was only being obtained for the wrong reason (because no final resampling was triggered if there were Gibbs-conditioned-variables or addlogprob occurring after the last observe statement). Now it crashes and the naive workarounds (e.g. `@addlogprob! 0.0`) produce the 'incorrect' result.

mhauru: Agree on the word "dynamic". Generally feel free to rephrase, but I think it would be good to have something in the changelist that's helpful for anyone whose model changes behaviour on 0.40 when sampled with PG-within-Gibbs.

penelopeysm: Added a note with an example model.

mhauru:
Thanks, very happy with that. I'm good to merge and release once tests pass. You wanna do it?

penelopeysm:
Okay 😄

penelopeysm:
I'll merge into breaking then let CI run on that branch.

PR #2561, opened 2025-05-22
Fix various methods for on empty varinfo

penelopeysm:
Followup from https://github.com/TuringLang/DynamicPPL.jl/pull/926.

MH still doesn't quite work because AdvancedMH needs some changes too. But at this point I'm tempted to just let it be.


test/mcmc/Inference.jl
```diff
@@ -631,6 +631,39 @@ using Turing
             StableRNG(seed), demo_incorrect_missing([missing]), NUTS(), 10; check_model=true
         )
     end
+
+    @testset "getparams" begin
+        @model function e(x=1.0)
+            return x ~ Normal()
+        end
+        evi = Turing.VarInfo(e())
```
penelopeysm, line 639: I'm aware that these tests are being run once-per-adtype, and there's really no need to do so. But this is true of quite a number of tests in this file and so I thought that was a bigger problem that I didn't want to deal with in this PR.

PR #2564, opened 2025-05-23
[WIP] Remove AD backend loops

### LINKED ISSUE #2307: Refactoring AD Tests
Currently, there are many situation in which we run an entire sampler in order to test AD, such as many cases in https://github.com/TuringLang/Turing.jl/blob/master/test/mcmc/abstractmcmc.jl .

This is quite compute intensive. @yebai pointed out that perhaps we should just check that the gradient runs and is correct, rather than thousants of mcmc chains.

penelopeysm:
Closes #2307 

Blocked upstream by https://github.com/TuringLang/DynamicPPL.jl/pull/935

### DIFF
--- a/test/mcmc/abstractmcmc.jl
+++ b/test/mcmc/abstractmcmc.jl
@@ -112,36 +112,36 @@ end
 
 @testset "External samplers" begin
     @testset "AdvancedHMC.jl" begin
-        @testset "adtype=$adtype" for adtype in ADUtils.adbackends
-            @testset "$(model.f)" for model in DynamicPPL.TestUtils.DEMO_MODELS
-                # Need some functionality to initialize the sampler.
-                # TODO: Remove this once the constructors in the respective packages become "lazy".
-                sampler = initialize_nuts(model)
-                sampler_ext = DynamicPPL.Sampler(
-                    externalsampler(sampler; adtype, unconstrained=true)
-                )
-                # FIXME: Once https://github.com/TuringLang/AdvancedHMC.jl/pull/366 goes through, uncomment.
-                # @testset "initial_params" begin
-                #     test_initial_params(model, sampler_ext; n_adapts=0)
-                # end
+        @testset "$(model.f)" for model in DynamicPPL.TestUtils.DEMO_MODELS
+            adtype = Turing.DEFAULT_ADTYPE
 
-                sample_kwargs = (
-                    n_adapts=1_000,
-                    discard_initial=1_000,
-                    # FIXME: Remove this once we can run `test_initial_params` above.
-                    initial_params=DynamicPPL.VarInfo(model)[:],
-                )
+            # Need some functionality to initialize the sampler.
+            # TODO: Remove this once the constructors in the respective packages become "lazy".
+            sampler = initialize_nuts(model)
+            sampler_ext = DynamicPPL.Sampler(
+                externalsampler(sampler; adtype, unconstrained=true)
+            )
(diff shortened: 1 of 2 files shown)


test/mcmc/abstractmcmc.jl
```diff
@@ -112,36 +112,36 @@ end
 
 @testset "External samplers" begin
     @testset "AdvancedHMC.jl" begin
-        @testset "adtype=$adtype" for adtype in ADUtils.adbackends
-            @testset "$(model.f)" for model in DynamicPPL.TestUtils.DEMO_MODELS
```
penelopeysm, line 116: externalsampler for some reason doesn't use SamplingContext, it just uses DefaultContext.

AD testing on DEMO_MODELS with DefaultContext is already done in DPPL CI, so no need to redo here.

test/runtests.jl
```diff
@@ -9,9 +9,8 @@ import Turing
 # Fix the global Random.seed for reproducibility.
 seed!(23)
 
-include(pkgdir(Turing) * "/test/test_utils/models.jl")
-include(pkgdir(Turing) * "/test/test_utils/numerical_tests.jl")
-include(pkgdir(Turing) * "/test/test_utils/ad_utils.jl")
+include("test_utils/models.jl")
+include("test_utils/numerical_tests.jl")
```
penelopeysm, line 13: Was there a reason for using the `pkgdir()` thing rather than the standard include path? Happy to revert if there was.

mhauru: 🤷

mhauru (commented):
I, too, am a bit nervous about how much this reduces our test coverage (and I don't mean lines of code but functionality), but it definitely seems like the right way to do things long term.

One concern I have is if some bits of code that aren't within `logdensity_and_gradient` calls might still somehow fail, due to e.g. not handling tracer types. For instance, Gibbs does a bunch of resetting of sampler states, and nasty things could happen if some vector element types get messed up. `run_ad` wouldn't catch that, since it evaluates the LDF without calling `step`. Maybe in the loop over algorithms in `ad.jl` we could also do a call to `sample` with each AD backend and model, just to ensure that it doesn't crash?

test/ad.jl
```diff
@@ -280,4 +279,14 @@ end
...
 end
 
+@testset verbose = true "AD / Gibbs sampling" begin
+    # Make sure that Gibbs sampling doesn't fall over when using AD.
+    spl = Gibbs(@varname(s) => HMC(0.1, 10), @varname(m) => HMC(0.1, 10))
+    @testset "adtype=$adtype" for adtype in ADTYPES
+        @testset "model=$(model.f)" for model in DEMO_MODELS
+            @test sample(model, spl, 2) isa Any
+        end
+    end
```
mhauru, line 289: Does this use the `adtype`?

penelopeysm: Agh.

sunxd3:
Also, we can probably get rid of the `adbackend` and just use `Turing.DEFAULT_ADTYPE`? Not essential.

https://github.com/TuringLang/Turing.jl/blob/5e7ed7dcfa7ade904b100c475a792b8d618cbe6c/test/mcmc/gibbs.jl#L376
https://github.com/TuringLang/Turing.jl/blob/5e7ed7dcfa7ade904b100c475a792b8d618cbe6c/test/mcmc/hmc.jl#L22
https://github.com/TuringLang/Turing.jl/blob/5e7ed7dcfa7ade904b100c475a792b8d618cbe6c/test/mcmc/Inference.jl#L23

PR #2566, opened 2025-05-25
Rework sample() call stack, part 1: HMC

### LINKED ISSUE #2555: Rework `sample()` call stack to use LogDensityFunction
# Overview of progress

- [x] Hamiltonians: `hmc.jl`, `sghmc.jl`, `DynamicHMCExt` #2588
- [ ] ESS: `ess.jl` #2590
- [ ] MH: `mh.jl` #2590
- [ ] IS: `is.jl`
- [ ] Prior: `prior.jl`
- [ ] Emcee: `emcee.jl`
- [ ] Particle samplers: `particle_mcmc.jl`
- [ ] `ExternalSampler`
- [ ] `RepeatSampler`
- [ ] Add method warning to `AbstractMCMC.step(rng, model, spl[, state])`

## Current situation

Right now, if you call `sample(::Model, ::InferenceAlgorithm, ::Int)` this first goes to `src/mcmc/Inference.jl` where the `InferenceAlgorithm` gets wrapped in `DynamicPPL.Sampler`, e.g.

https://github.com/TuringLang/Turing.jl/blob/5acc97f8c6f004d44359337d3c7f0fdf38c70315/src/mcmc/Inference.jl#L268-L278

This then goes to `src/mcmc/$sampler.jl` which defines the methods `sample(::Model, ::Sampler{<:InferenceAlgorithm}, ::Int)`, e.g.

https://github.com/TuringLang/Turing.jl/blob/5acc97f8c6f004d44359337d3c7f0fdf38c70315/src/mcmc/hmc.jl#L82-L104

This then goes to AbstractMCMC's `sample`:

https://github.com/TuringLang/AbstractMCMC.jl/blob/fdaa0ebce22ce227b068e847415cd9ee0e15c004/src/sample.jl#L255-L259

Which then calls `step(::AbstractRNG, ::Model, ::Sampler{<:InferenceAlgorithm})`, which is defined in DynamicPPL:

https://github.com/TuringLang/DynamicPPL.jl/blob/072234d094d1d68064bf259d3c3e815a87c18c8e/src/sampler.jl#L108-L126

Which then calls `initialstep`, which goes back to being defined in `src/mcmc/$sampler.jl`:

https://github.com/TuringLang/Turing.jl/blob/5acc97f8c6f004d44359337d3c7f0fdf38c70315/src/mcmc/hmc.jl#L141-L149

(this signature claims to work on `AbstractModel`, it only really works for `DynamicPPL.Model`)

Inside here, we finally construct a `LogDensityFunction` from the model. So, there are very many steps between the time that `sample()` is called, and the time where a `LogDensityFunction` is actually constructed.

## Proposal

Rework everything below the very first call to accept `LogDensityFunction` rather than `Model`. That is to say, the method `sample(::Model, ::InferenceAlgorithm, ::Int)` should look something like this:

```julia
function sample(rng::Random.AbstracRNG, model::Model, alg::InferenceAlgorithm, N::Int, kwargs...)
    adtype = get_adtype(alg) # returns nothing (e.g. for MH) or AbstractADType (e.g. NUTS)
    ldf = DynamicPPL.LogDensityFunction(model; adtype=adtype)
    spl = DynamicPPL.Sampler(alg)
    sample(rng, ldf, spl, N; kwargs...)
end

# similar methods for sample(::Random.AbstractRNG, ::Model, ::Sampler{<:InferenceAlgorithm})
# as well as sample(:Random.AbstractRNG, ::LogDensityFunction, ::InferenceAlgorithm)

function sample(
    rng::Random.AbstractRNG,
    ldf::DynamicPPL.LogDensityFunction,
    spl::DynamicPPL.Sampler{<:InferenceAlgorithm}, N::Int;
    chain_type=MCMCChains.Chains, check=true, ...
)
    # All of Turing's magic behaviour, e.g. checking the model, setting the chain_type,
    # should happen in this method.
    check && check_model(ldf.model)
    ... (whatever else)
    # Then we can call mcmcsample, and that means that everything from mcmcsample
    # onwards _knows_ that it's dealing with a LogDensityFunction and a Sampler{...}.
    AbstractMCMC.mcmcsample(...)
end

# handle rng-less methods (ugly boilerplate, because somebody decided that the *optional*
# rng argument should be the first argument and not a keyword argument, ugh!)
function sample(model::Model, alg::InferenceAlgorithm, N::Int, kwargs...)
    sample(Random.default_rng(), model, alg, N; kwargs...)
end
# similar for sample(::Model, ::Sampler{<:InferenceAlgorithm})
#  as well as sample(::LogDensityFunction, ::InferenceAlgorithm)
#  as well as sample(::LogDensityFunction, ::Sampler{<:InferenceAlgorithm})

# oh don't forget to handle the methods with MCMCEnsemble... argh...
# so that's 16 methods to make sure that everything works correctly
# - 2x with/without rng
# - 2x for Model and LDF
# - 2x for InferenceAlgorithm and Sampler{...}
# - 2x for standard and parallel
```

This would require making several changes across DynamicPPL and Turing. It (thankfully) probably does not need to touch AbstractMCMC, as long as we make `LogDensityFunction` a subtype of `AbstractMCMC.AbstractModel` (so that `mcmcsample` can work). That should be fine, because `AbstractModel` has no interface.

## Why?

For one, this is probably the best way to let people have greater control over their sampling process. For example:

- If you want to use a particular type of VarInfo, this interface would allow you to construct it, make the LogDensityFunction, _then_ pass it to `sample()`. Right now, this is actually very difficult to do. (Just try it!!) Note that this also provides a natural interface for opting into ThreadSafeVarInfo (cf. https://github.com/TuringLang/DynamicPPL.jl/issues/924)
- Allows constructing the LDF with a particular adtype (this was part of the point behind making adtype a field of the LDF, cf. https://github.com/TuringLang/DynamicPPL.jl/pull/806)

More philosophically, it's IMO the first step that's necessary towards encapsulating Turing's "magic behaviour" at the very top level of the call stack. We know that a `DynamicPPL.Model` on its own does not actually give enough information about how to evaluate it — it's only `LogDensityFunction` that contains the necessary information. Thus, it shouldn't be the job of the low-level functions like `step` to make this decision — they should just 'receive' objects that are already complete.

## I'm still not sure about

How this will work with Gibbs. I haven't looked at it deeply enough.

penelopeysm:
This PR moves in the general direction of #2555.

It will take a long time to get _everything_ to work, so I am trying to do this incrementally.

## Summary

The fundamental idea (see #2555) is that we want

`sample(::Union{Model, LogDensityFunction}, ::Union{InferenceAlgorithm, Sampler{<:InferenceAlgorithm}}, N)`

to always forward to

`sample(::LogDensityFunction, ::Sampler{<:InferenceAlgorithm}, N)`

(along the way, we construct the LDF if we need to, and also construct the sampler if we need to).

Then, the concrete AbstractMCMC interface functions (i.e., `mcmcsample`, `step`) do not ever see a Model, they only see an LDF.

Note that this allows us to sidestep (and eventually, completely remove) `DynamicPPL.initialstep`. The reason why that function exists is because [`AbstractMCMC.step` would do two things](https://github.com/TuringLang/DynamicPPL.jl/blob/bb0c8574814ed4c8bc95cae5fa4238f628ade5ab/src/sampler.jl#L107-L126): first, generate the VarInfo that would eventually go into the LDF, and secondly, call `initialstep` (which was sampler-specific). Since the VarInfo generation bit is now handled in the LDF construction, it means that instead of having an extra function, we can just go back to implementing the two basic `AbstractMCMC.step` methods, which is a nice bonus.

## Incremental changes

Changing this all at once is bound to be not only impossible to do but also impossible to review. Thus, I've decided to (try to) implement this in a few stages. This PR is probably the one that makes the most sweeping changes, and also establishes the interface required. It:

1. Moves `ExternalSampler` and `Prior` into their own files, because Inference.jl is really a huge mess and we could do with less of that. The code is directly copied.

2. Establishes the desired method dispatch behaviour for `sample` (see `src/mcmc/sample.jl` -- up until line 271). Because we aren't ready to extend this to every sampler and inference algorithm yet, these methods dispatch only on `LDFCompatibleAlgorithm` or `LDFCompatibleSampler`, which are defined at the top of the file. The idea is that we'll add samplers as we go along, and one day we'll eventually be ready to remove this type and just use `InferenceAlgorithm` etc.

3. When automatically constructing the LDF, there are a few things that we need to know to construct it properly:
   - Does the VarInfo need to be linked?
   - Does the LDF need to be constructed with an adtype?
  
   This PR therefore also introduces interface functions that all (LDF-compatible) samplers must conform to, namely `requires_unconstrained_space(::AbstractSampler)` and `get_adtype(::AbstractSampler)`. Sensible defaults of `true` and `nothing` are given. Note that these functions were already floating around the Turing codebase, so all I've really done is to bring it together and actually write docstrings for them.

4. Finally, there is an `update_sample_kwargs` function which samplers can use as a hook to modify the keyword arguments sent to `sample()`. This used to be handled in a specialised `sample()` method (e.g. the one for HMC has just been removed in this PR). I recognise that this is kind of just shifting complexity around, but I think we will benefit from having a unified interface that `InferenceAlgorithm` (or `Sampler{<:InferenceAlgorithm}`) needs to implement.

5. To this end, I moved `InferenceAlgorithm` and the methods it needs to implement to a new file, `algorithm.jl`.

## Fortunately

This doesn't actually require any changes to DynamicPPL, which I found to be a huge relief!

It's likely that some of the code in this PR will eventually be moved to DynamicPPL, as they don't have any non-DynamicPPL dependencies. But that can be handled very easily at a later stage, once we're confident that this all works.

## Unfortunately

Changing the interface one sampler at a time completely breaks Gibbs, because for Gibbs to work, it requires all of its component samplers to be updated. So we may have to live with the Gibbs tests being broken for a while, and rely on me promising that I'll fix it at some point in time.

Because I don't know how long this will take me, I don't even want to merge this into `breaking`, as I don't want to have a new release held up by the fact that only half the changes have been done. I've created a new base branch, `sample-ldf`, to collect all the work on this. When we're happy with it, we can merge that into `breaking`.

## TODO

- [ ] Split non-code changes into a separate PR
- [ ] Add tests for sampling LDF


mhauru:
The description makes sense to me. Do I understand correctly that the idea with `LDFCompatibleAlgorithm` is to have the new style sample pipeline and the old style sample pipeline live in parallel until `LDFCompatibleAlgorithm == InferenceAlgorithm`?

penelopeysm:
Yupyup, exactly. But I don't want to actually do a release with this until it's done, so an end-user would never see `LDFCompatibleAlgorithm`

penelopeysm:
Reopening a new PR as there have been quite a few code restructuring bits that will mess up the diff

PR #2567, opened 2025-05-28
Refactor HMC initialisation code

penelopeysm:
I'm really wary of #2566 becoming very big, so I'm trying to break off smaller bits.

This is a really basic bit of refactoring that I think just makes things easier to read.

Anyway, I've tested before and after with

```julia
using Turing, Random
@model f() = x ~ Normal()
sample(Xoshiro(468), f(), NUTS(), 1000)
```

and the exact numerical result is unchanged, so I'm pretty sure that there isn't actually any code change — hence the lack of a version bump.

```
└   ϵ = 3.2
Sampling 100%|██████████████████████████████████████████████████████████████████| Time: 0:00:03
Chains MCMC chain (1000×13×1 Array{Float64, 3}):

Iterations        = 501:1:1500
Number of chains  = 1
Samples per chain = 1000
Wall duration     = 4.59 seconds
Compute duration  = 4.59 seconds
parameters        = x
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters      mean       std      mcse   ess_bulk   ess_tail      rhat   ess_per_sec
      Symbol   Float64   Float64   Float64    Float64    Float64   Float64       Float64

           x   -0.0338    0.9883    0.0494   396.8876   699.6500    1.0029       86.5433

Quantiles

           x   -1.9956   -0.6683   -0.0578    0.6140    1.9267
```

### DIFF
--- a/src/mcmc/hmc.jl
+++ b/src/mcmc/hmc.jl
@@ -138,6 +138,36 @@ function AbstractMCMC.sample(
     end
 end
 
+function find_initial_params(
+    rng::Random.AbstractRNG,
+    model::DynamicPPL.Model,
+    varinfo::DynamicPPL.AbstractVarInfo,
+    hamiltonian::AHMC.Hamiltonian;
+    max_attempts::Int=1000,
+)
+    varinfo = deepcopy(varinfo)  # Don't mutate
+
+    for attempts in 1:max_attempts
+        theta = varinfo[:]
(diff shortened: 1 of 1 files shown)


mhauru (approved):
LGTM

Issue #2570, opened 2025-05-29
Specifically, sampling seems inordinately slow on 1.10


penelopeysm:
```julia
@testset verbose = true "AD / Gibbs sampling" begin
    # Make sure that Gibbs sampling doesn't fall over when using AD.
    @testset "adtype=$adtype" for adtype in ADTYPES
        spl = Gibbs(
            @varname(s) => HMC(0.1, 10; adtype=adtype),
            @varname(m) => HMC(0.1, 10; adtype=adtype),
        )
        @testset "model=$(model.f)" for model in DEMO_MODELS
            @test sample(model, spl, 2) isa Any
        end
    end
end
```

[This testset on 1.10:](https://github.com/TuringLang/Turing.jl/actions/runs/15329784139/job/43133473049?pr=2569)

```
    AD / Gibbs sampling                             |   39     39   93m39.7s
      adtype=ADTypes.AutoForwardDiff()              |   13     13   13m57.6s
      adtype=ADTypes.AutoReverseDiff()              |   13     13   36m59.2s
      adtype=ADTypes.AutoMooncake{Nothing}(nothing) |   13     13   42m42.9s
```

[This testset on 1.11:](https://github.com/TuringLang/Turing.jl/actions/runs/15329784139/job/43133472195?pr=2569)

```
     AD / Gibbs sampling                             |   39     39  12m58.3s
      adtype=ADTypes.AutoForwardDiff()              |   13     13   2m33.9s
      adtype=ADTypes.AutoReverseDiff()              |   13     13   2m13.8s
      adtype=ADTypes.AutoMooncake{Nothing}(nothing) |   13     13   8m10.6s
```

devmotion:
Maybe https://github.com/JuliaLang/julia/issues/53371 - a fix was backported to 1.10.2 but the tests were run with 1.10.0 so could be affected by it. Or it could be some other issue that was fixed in 1.10.x so it would be interesting to rerun tests with `lts` (latest 1.10 release).

penelopeysm:
It is indeed fixed on 1.10.2 (and slow on 1.10.1) so might very well be that bug specifically.

We could either specifically run CI on >= 1.10.2, or we could change the minimum Julia version to 1.10.2 -- thoughts? I have a mild preference for the latter because it's nice to be able to specify `min` in CI.

mhauru:
[...]

PR #2572, opened 2025-05-30
bisect slow 1.10 sampling

penelopeysm:
see #2570


penelopeysm:
it's ok on 1.10.2, slow on 1.10.1

penelopeysm:
letting CI run fully once to get new timings so that i can rebelance them again

Issue #2288, opened 2024-07-15
Import style and modularisation


mhauru:
Often when reading Turing.jl code (and many other Julia projects) I'm struck by how hard it is to see where names (functions, variables, types) are defined. There are two parts to this problem:
* The `using X` syntax brings things into a namespace without explicitly naming them.
* Having multiple files `include`d in one module means that things defined in one file are in scope in another, without either file explicitly referencing the other.

To counter this issue, I'd propose two conventions:
1. All imports should be explicit, so either `using X: a, b, c`, `using X: X`, `import X: a, b, c`, or `import X`.
2. Each file should be its own module.

I wouldn't be surprised if 2. turned out to be a bit overly strict in some cases, and I'm happy to have some flexibility there. Maybe 2. should just say "use more submodules".

I don't propose anyone goes and writes a single massive PR for this. Rather I'm asking for comments on this as a style convention. If there's support for this, I'd slowly start applying the new conventions as I edit code.

I raise this as a Turing.jl issue, but I'd eventually like for this to apply to all of TuringLang (and preferably the rest of the universe too).

Thoughts?

torfjelde:
> The using X syntax brings things into a namespace without explicitly naming them.

I don't have super-strong opionions regarding exactly where we define the different namespaces, but I very much agree with _not_ doing this 👍

mhauru:
>This might be overly restrictive. A more moderate proposal is that each subfolder under src should be a module, and all its included files should be in that subfolder too.

That could be a good middle ground if one-module-one-file is a bit too much. Happy to be flexible with this, but I do think more submodules would help structure our namespaces and provide some clear interfaces, defining what is an implementation detail and what is a key part of functionality. This would also help readability of the code, to understand what are the "external facing" functions of some subpart of a package.

To take an example, in DynamicPPL I would make everything related to VarInfo a module. Could be that VarInfo, SimpleVarInfo, and AbstractVarInfo are also all their own modules, or could be that they are all together, but I think there's a natural interface between those types and the rest of DynamicPPL, and making the code structure reflect that separation of concerns would, in my opinion, clarify the code.

>But vscode offers some tools for navigating code. There is also @which macro.

Yeah, these help, language servers help, but I do still find myself grepping the code to find where things are defined. Also, when browsing code on GitHub this comes up a lot.

mhauru:
Interesting. Seems like not using `using X` is about to become an official recommendation: https://github.com/JuliaLang/julia/pull/42080

[...]

PR #2573, opened 2025-05-30
Reorganise code in `src/mcmc/Inference.jl`

penelopeysm:
This PR does some code reshuffling in `src/mcmc/Inference.jl`.

In particular:

- `sample()` dispatch methods are moved to `sample.jl`
- `Prior()` is moved to its own file, `prior.jl` (it's an inference algorithm after all)
- `ExternalSampler` is moved to its own file, `external_sampler.jl`. The existing implementation of AbstractMCMC methods on ExternalSampler, which used to be `abstractmcmc.jl`, has been moved into this file as well, so that the naming is more coherent.
- `InferenceAlgorithm` is moved to its own file, `algorithm.jl`. This might seem excessive, as it's just one abstract type. However, as part of the work on #2555 I am going to introduce a few interface functions that InferenceAlgorithms should implement. Note that we actually already have some of these, such as `requires_unconstrained_space` or `getADtype`, that are scattered around this file / implemented for some samplers but not all. I think it would be good to unify these, and my eventual vision is that `algorithm.jl` will be where the interface + default implementations live. Some of this work is already in #2566.

The remainder is still quite messy, but it can be for another day...


src/mcmc/Inference.jl
```diff
@@ -79,85 +79,18 @@ export InferenceAlgorithm,
     predict,
     externalsampler
 
-#######################
-# Sampler abstraction #
-#######################
-abstract type AbstractAdapter end
```
penelopeysm, line 85: Deleted as this wasn't being used anywhere

mhauru: I would say it's fine since it wasn't exported.

src/mcmc/prior.jl
```diff
@@ -0,0 +1,25 @@
...
+        DynamicPPL.evaluate!!(
+            model,
+            VarInfo(),
+            SamplingContext(rng, DynamicPPL.SampleFromPrior(), DynamicPPL.PriorContext()),
+        ),
+    )
+    return vi, nothing
+end
+
+DynamicPPL.default_chain_type(sampler::Prior) = MCMCChains.Chains
```
mhauru, line 25: Is this needed given the fallback for `InferenceAlgorithm`? Also, is this function to be defined for the `Sampler` or the `InferenceAlgorithm`?

src/mcmc/sample.jl
```diff
@@ -0,0 +1,92 @@
+module Sample
```
mhauru: Any particular reason to have this as a submodule but the others not? Also, do we have a convention that files that define a module have their names CamelCased?

penelopeysm: > Any particular reason to have this as a submodule but the others not?

I would like everything to be a submodule, but I thought changing it now would be unnecessarily destructive.

> Also, do we have a convention that files that define a module have their names CamelCased?

As in, this should be `src/mcmc/Sample.jl`? I'm not sure I recall hearing about such a convention.

src/mcmc/prior.jl
penelopeysm: No, it's not needed. Purely from the perspective of the `sample()` interface, this function should be defined for AbstractSampler, not InferenceAlgorithm -- although in practice, I think that the interface functions for InferenceAlgorithm are likely to be of the form

```julia
foo(::AbstractSampler) = foo_default
foo(s::DynamicPPL.Sampler{<:InferenceAlgorithm}) = foo(s.alg)
foo(alg::InferenceAlgorithm) = # each individual algorithm implements this
```

so the answer might well be 'both'.

penelopeysm: More on topic: I think the `default_chain_type` function is likely to be completely removed in a subsequent PR, so I'm not fussed about leaving it in for this one. There are two reasons:

1. The chain type depends on both the sampler and the model (I don't think you can _really_ construct MCMCChains from any random model, at least not with the code we have) so the signature is not really right to begin with
2. The correct implementation of this is `default_chain_type(::DynamicPPL.Model, ::DynamicPPL.Sampler) = MCMCChains.Chains`, which is a constant, so the function doesn't really need to exist

src/mcmc/sample.jl
mhauru: I'm not sure it's documented anywhere, but I've noticed a correlation between things that are submodules and things that are CamesCased (as in `src/mcmc/Sample.jl`). I wonder if there are any Julia/community guidelines for this.

src/mcmc/prior.jl
mhauru: Ok, not bothered by keeping this.

src/mcmc/sample.jl
yebai: I suggest avoiding a module here, since it mostly overloads `AbstractMCMC.sample` functions. 

Let's use a submodule only for a coherent set of types and APIs. I appreciate that this might differ from other languages, but staying close to Julia's practices is better.

yebai: We can merge `sample.jl` into the existing file [`abstratmcmc.jl` ](https://github.com/TuringLang/Turing.jl/blob/01506d6e54bd8921a27a1d7a0d21274d87c654aa/src/mcmc/abstractmcmc.jl#L34).

penelopeysm: Yes, although it's not just about what I'm used to from a different language -- the main benefit of having code organised in modules is to prevent imports from leaking into / out of files. We discussed this previously here https://github.com/TuringLang/Turing.jl/issues/2288 -- and I still believe it's a good thing to put code into modules, and my opinion is independent of what the Julia community or any other language's community does.

Of course, maybe it's overkill to create a module just for these methods, since by definition it's not possible to leak code _out_ of new method overloads.

penelopeysm: And in fact, abstractmcmc.jl should be put in external_sampler.jl, since the bulk of its code deals with ExternalSampler.

(Specifically, it implements the AbstractMCMC interface for `ExternalSampler`, and thus it should be in the same code file as the definition of `ExternalSampler` --- just like how the implementation of the AbstractMCMC interface for `NUTS` is in `hmc.jl`.)

penelopeysm: Removed the module now, so I assume this settles the question of file naming (I did have a very, very brief look on Julia Discourse, and couldn't find anything).

penelopeysm:
Will merge now, but since this is a no-op in terms of code changes, happy to address any followup issues in a new PR

PR #2574, opened 2025-06-02
Clean up old code

### LINKED ISSUE #2512: Get rid of src/essential
It looks to me that `src/essential/Essential.jl` is no longer needed and `src/essential/container.jl` can be merged into `src/mcmc/particle_mcmc.jl`?

penelopeysm:
- Removes `src/deprecated.jl`, which contained no usable functionality, just some nice warnings for some very old AD backend manipulation utilities.

- Removes `src/essential` either through deletion or by moving particle MCMC code to `src/mcmc/particle_mcmc.jl`, as described in #2512 

Closes #2512


sunxd3:
Worth tagging @yebai for the export of `@addlogprob!`. I think this is the right way to go.

PR #2576, opened 2025-06-03
Rename `mcmc/sample.jl` to `mcmc/abstractmcmc.jl`

yebai:
`abstractmcmc.jl` is clearer than `sample.jl` and more consistent with other TuringLang packages (like AdvancedHMC, JuliaBUGS).


src/mcmc/Inference.jl
```diff
@@ -368,11 +368,11 @@ include("sghmc.jl")
 include("emcee.jl")
 include("prior.jl")
 
-####################################
-# Generic sample() method dispatch #
-####################################
+#################################################
+# Generic AbstractMCMC.sample() method dispatch #
```
yebai:
```suggestion
# Generic AbstractMCMC methods dispatch #
```

penelopeysm:
Happy with that!

PR #2581, opened 2025-06-04
pin ForwardDiff=1 in test

penelopeysm:
compathelper PR ran CI with ForwardDiff 0.10, so just want to check that it's really alright

don't merge this


penelopeysm:
1.10 test fails because FD=1 requires new Bijectors and new Bijectors requires 1.10.8. we don't need to push Turing to 1.10.8 on main, but we can at least test on CI with it in this pr to check if FD=1 introduces any issue on 1.10.8.

penelopeysm:
random OptimizationOptimJL precompile failure reported upstream https://github.com/SciML/Optimization.jl/issues/926

penelopeysm:
tests passed with FD@1 so we should be fine!

PR #2585, opened 2025-06-06
AdvancedPS v0.7 (and thus Libtask v0.9) support

mhauru:
The complement PR of https://github.com/TuringLang/AdvancedPS.jl/pull/114, which adds support for the newly rewritten Libtask.

~~Work in progress, currently blocked by https://github.com/TuringLang/Libtask.jl/issues/186~~


mhauru:
The tests that I had the patience to run locally now pass. Waiting for the AdvancedPS release to be able to run the full test suite on CI.

Some indicators of speed:

```julia
julia> module MWE

       using Turing

       @model function gdemo(x, y)
           s ~ InverseGamma(2, 3)
           m ~ Normal(0, sqrt(s))
           x ~ Normal(m, sqrt(s))
           y ~ Normal(m, sqrt(s))
           return s, m
       end

       @time chn = sample(gdemo(2.5, 1.0), PG(10), 10_000)
       describe(chn)

       end
```
On main:
```
104.715858 seconds (58.48 M allocations: 13.259 GiB, 1.10% gc time, 1.20% compilation time)
```
On this branch:
```
 16.612050 seconds (116.52 M allocations: 16.296 GiB, 8.27% gc time, 5.91% compilation time)
```

```julia
julia> module MWE

       using Turing

       @model function f(dim=20, ::Type{T}=Float64) where T
           s = Vector{Bool}(undef, dim)
           x = Vector{T}(undef, dim)
           for i in 1:dim
               s[i] ~ Bernoulli()
               if s[i]
                   x[i] ~ Normal()
                else
                   x[i] ~ Beta()
                end
                0.0 ~ Normal(x[i])
           end
           return nothing
       end

       alg = Gibbs(
           @varname(s)=>PG(10),
           @varname(x)=>HMC(0.1, 5),
       )
       @time chn = sample(f(), alg, 1_000)

       end
```
On main:
```
 49.682945 seconds (65.43 M allocations: 9.463 GiB, 1.79% gc time, 8.23% compilation time)
```
On this branch:
```
  9.180071 seconds (61.58 M allocations: 4.028 GiB, 4.04% gc time, 55.49% compilation time)
```

Obviously the speed gains are all due to @willtebbutt's fantastic work on Libtask, everything else is just wrapping that work.

Project.toml
```diff
@@ -85,7 +85,7 @@ Statistics = "1.6"
 StatsAPI = "1.6"
 StatsBase = "0.32, 0.33, 0.34"
 StatsFuns = "0.8, 0.9, 1"
-julia = "1.10.2"
+julia = "1.10.8"
```
yebai, line 88: Libtask requires 1.10.8 at a minimum.

src/mcmc/particle_mcmc.jl
```diff
@@ -402,11 +391,11 @@ end
 
 function trace_local_varinfo_maybe(varinfo)
     try
-        trace = AdvancedPS.current_trace()
-        return trace.model.f.varinfo
+        trace = Libtask.get_taped_globals(Any).other
```
yebai, line 394: If we change `Libtask.get_taped_globals` to return `nothing` if not inside a running `TapedTask`, the following `try .. catch ... end` can be removed.

src/mcmc/particle_mcmc.jl
```diff
@@ -416,11 +405,10 @@ end
 
 function trace_local_rng_maybe(rng::Random.AbstractRNG)
     try
-        trace = AdvancedPS.current_trace()
-        return trace.rng
+        return Libtask.get_taped_globals(Any).rng
```
yebai, line 408: Same with above.

penelopeysm:
Is this reviewable? The tests are failing, there's a method ambiguity that Aqua complains about, there's a Gibbs failure on 1.12 which should be disabled with `@test_broken`, and the sampling in `mcmc/Inference` is returning numerically inaccurate values:

```
beta binomial: Test Failed at /home/runner/work/Turing.jl/Turing.jl/test/test_utils/numerical_tests.jl:55
  Expression: ≈(E, val, atol = atol, rtol = rtol)
   Evaluated: 0.5430087089005213 ≈ 0.7142857142857143 (atol=0.05, rtol=0.0)

beta binomial: Test Failed at /home/runner/work/Turing.jl/Turing.jl/test/test_utils/numerical_tests.jl:55
  Expression: ≈(E, val, atol = atol, rtol = rtol)
   Evaluated: 0.4 ≈ 0.7142857142857143 (atol=0.1, rtol=0.0)
```

I don't want to speak for @mhauru in his absence but last time we spoke about this PR, it was clear that there were still a few gaps to bridge. If I were to review it at this stage, my sole comment would be to fix the tests.

mhauru:
The remaining test failures on v1.10 and v1.11 should be fixed once https://github.com/TuringLang/Libtask.jl/pull/192 and https://github.com/TuringLang/Libtask.jl/pull/191 are merged and released (I see the tests passing locally). Will then have a look at what's going on with v1.12, and check code quality, and hopefully this could then be done.

Project.toml
```diff
@@ -67,7 +67,7 @@ DynamicHMC = "3.4"
 DynamicPPL = "0.36.3"
 EllipticalSliceSampling = "0.5, 1, 2"
 ForwardDiff = "0.10.3, 1"
-Libtask = "0.8.8"
+Libtask = "0.9.2"
```
yebai:
```suggestion
Libtask = "0.9.3"
```

mhauru:
Seems I spoke too soon. All tests now pass on _some_ version of Julia. But the beta binomial test still fails on v1.11, whereas the ESS test fails on v1.10. I'll try to reproduce locally, although no immediate success.

Project.toml
mhauru: Done

src/mcmc/particle_mcmc.jl
```diff
@@ -59,20 +58,10 @@ function AdvancedPS.reset_logprob!(trace::TracedModel)
...
-    trace::AdvancedPS.Trace{<:AdvancedPS.LibtaskModel{<:TracedModel}}
-)
-    # Extract the `args`.
-    args = trace.model.ctask.args
-    # From `args`, extract the `SamplingContext`, which contains the RNG.
-    sampling_context = args[3]
-    rng = sampling_context.rng
-    trace.rng = rng
-    return trace
-end
```
mhauru, line 72: This was used to keep the internal state of `trace` consistent, and should now be taken care of in AdvancedPS.

src/mcmc/particle_mcmc.jl
mhauru: Libtask would then lose a distinction between having `nothing` as the taped global within a task, and just not being within a task at all. I wonder if that distinction could be useful in some situations.

src/mcmc/particle_mcmc.jl
```diff
@@ -481,6 +469,25 @@ function AdvancedPS.Trace(
 
     tmodel = TracedModel(model, sampler, newvarinfo, rng)
     newtrace = AdvancedPS.Trace(tmodel, rng)
-    AdvancedPS.addreference!(newtrace.model.ctask.task, newtrace)
```
mhauru, line 484: This was about keeping internal state of `newtrace` consistent, and is now dealt with in the `AdvancedPS.Trace` constructor.

test/mcmc/Inference.jl
```diff
@@ -44,13 +44,13 @@ using Turing
             end
 
             # Should also be stable with an explicit RNG
-            seed = 5
-            rng = Random.MersenneTwister(seed)
+            local_seed = 5
```
mhauru, line 47: This is unrelated to the other changes in this PR, but was discovered as a bug: We were accidentally setting the global seed in a local test.

test/mcmc/Inference.jl
```diff
@@ -256,9 +256,9 @@ using Turing
         pg = PG(10)
         gibbs = Gibbs(:p => HMC(0.2, 3), :x => PG(10))
 
-        chn_s = sample(StableRNG(seed), testbb(obs), smc, 200)
-        chn_p = sample(StableRNG(seed), testbb(obs), pg, 200)
-        chn_g = sample(StableRNG(seed), testbb(obs), gibbs, 200)
+        chn_s = sample(StableRNG(seed), testbb(obs), smc, 2_000)
+        chn_p = sample(StableRNG(seed), testbb(obs), pg, 2_000)
+        chn_g = sample(StableRNG(seed), testbb(obs), gibbs, 2_000)
```
mhauru, line 261: Had to bump these sample counts to get the tests to pass. I assume that's just random happenstance, but small risk that I'm masking a subtle bug here.

test/mcmc/ess.jl
```diff
@@ -64,7 +64,7 @@ using Turing
                 @varname(mu1) => ESS(),
                 @varname(mu2) => ESS(),
             )
-            chain = sample(StableRNG(seed), MoGtest_default, alg, 2000)
+            chain = sample(StableRNG(seed), MoGtest_default, alg, 5000)
```
mhauru, line 67: Same thing, had to increase samples for this to pass.

mhauru:
@penelopeysm, this is ready for attention now. The only test failures are Libtask v1.12 issues.

src/mcmc/particle_mcmc.jl
penelopeysm: For my edification: AdvancedPS uses its own RNG, right? Is that distinct from the RNG in the SamplingContext (that was used here) / how are they related?

test/mcmc/ess.jl
penelopeysm: I'm a bit worried about this because it might be hiding a correctness problem. I think the only good way to test, though, is very boring:

- Check out `main` and run this test with a number of different seeds, compare something like the mean MCSE, or the stdev of the parameter means, in the results.
- Repeat with this PR and check that the mean MCSE (or stdev of means) is unchanged.

I'm happy to do this test once and see where it goes.

penelopeysm: (I agree with your other comment that the chance of it being a correctness issue is small, but IMO the risk of correctness issues is approximately `+Inf`, and thus I am very motivated to rule it out.)

penelopeysm: Update: I checked quite extensively and am reasonably confident that there isn't a correctness issue, see main thread.

src/mcmc/particle_mcmc.jl
mhauru: Don't know, but I don't think they are always the same. AdvancedPS does all sorts of seed splitting stuff to make sure each trace has its own source of randomness. For this purpose it also uses a particular RNG type called `Random123`. Frankly not sure how that will relate to the one in the context.

mhauru:
Thanks for testing that, I find that reassuring. I'll try to see if Libtask on v1.12 is an easy fix, and then either get that working or merge this as-is.

Waaaait, what, one of the v1.10 CI runs just segfaulted. Ewww.

mhauru:
Passed after rerunning CI. Do we have an indeterministic test, or did we get unlucky and GHA bugged out?

penelopeysm:
I would be happy to ignore it, unless and until it shows up again.

mhauru:
Fixing Libtask for v1.12 doesn't look like it's going to happen in a few short days' time. Merging this as-is now, to be able to move on here and e.g. avoid further merge conflicts with #2550.

PR #2588, opened 2025-06-08
sample with LogDensityFunction: part 1 - HMC

penelopeysm:


ext/TuringDynamicHMCExt.jl
penelopeysm, line 1: Many of the changes in sghmc.jl are quite similar to the ones in this file, so I added some comments explaining.

ext/TuringDynamicHMCExt.jl
```diff
@@ -35,8 +35,7 @@ State of the [`DynamicNUTS`](@ref) sampler.
 # Fields
 $(TYPEDFIELDS)
 """
-struct DynamicNUTSState{L,V<:DynamicPPL.AbstractVarInfo,C,M,S}
-    logdensity::L
+struct DynamicNUTSState{V<:DynamicPPL.AbstractVarInfo,C,M,S}
```
penelopeysm, line 38: The sampler state, traditionally, has included the `LogDensityFunction` as a field so that it doesn't need to be re-constructed on each iteration from the model + varinfo. This is no longer necessary because the LDF is itself an argument to `AbstractMCMC.step`.

ext/TuringDynamicHMCExt.jl
```diff
@@ -48,30 +47,17 @@ function DynamicPPL.initialsampler(::DynamicPPL.Sampler{<:DynamicNUTS})
...
-        vi = last(DynamicPPL.evaluate!!(model, vi, DynamicPPL.SamplingContext(rng, spl)))
-    end
-
-    # Define log-density function.
-    ℓ = DynamicPPL.LogDensityFunction(
-        model,
-        vi,
-        DynamicPPL.SamplingContext(spl, DynamicPPL.DefaultContext());
-        adtype=spl.alg.adtype,
-    )
```
penelopeysm, line 70: All of this stuff is now handled inside `AbstractMCMC.sample()`, so there's no longer a need to duplicate this code inside every `initialstep` method.

src/mcmc/Inference.jl
```diff
@@ -244,7 +244,7 @@ getlogevidence(transitions, sampler, state) = missing
 # This is type piracy (at least for SampleFromPrior).
 function AbstractMCMC.bundle_samples(
     ts::Vector{<:Union{AbstractTransition,AbstractVarInfo}},
-    model::AbstractModel,
+    model_or_ldf::Union{DynamicPPL.Model,DynamicPPL.LogDensityFunction},
     spl::Union{Sampler{<:InferenceAlgorithm},SampleFromPrior,RepeatSampler},
     state,
     chain_type::Type{MCMCChains.Chains};
```
penelopeysm: This signature is not super ideal, but it minimises breakage right now. Eventually when everything is fixed we can change the Union to just LDF.

src/mcmc/hmc.jl
```diff
@@ -80,68 +117,6 @@ function HMC(
...
-    N::Integer;
-    chain_type=DynamicPPL.default_chain_type(sampler),
-    resume_from=nothing,
-    initial_state=DynamicPPL.loadstate(resume_from),
-    progress=PROGRESS[],
-    nadapts=sampler.alg.n_adapts,
-    discard_adapt=true,
-    discard_initial=-1,
-    kwargs...,
-)
```
penelopeysm, line 99: The purpose of this overload was purely to modify the kwargs to sample. I ditched it in favour of adding a new hook, `update_sample_kwargs`, which does the same thing without abusing multiple dispatch. I think that function does the same thing. It's of course quite hard to prove this, _although_ separating it into a different function _does_ allow us to write unit tests for it to make sure that it's doing the right thing (which are now in `test/mcmc/hmc.jl`), so that's another benefit.

(Overloading sample() for individual samplers like this is quite precarious because we can't recursively call `AbstractMCMC.sample` or we will end up with infinite recursion -- it has to call `mcmcsample`. So, there's no way to 'extend' this with extra behaviour by e.g. calling another method of `sample` before calling `mcmcsample`.)

src/mcmc/hmc.jl
```diff
@@ -467,9 +431,9 @@ function NUTS(; kwargs...)
     return NUTS(-1, 0.65; kwargs...)
 end
 
-for alg in (:HMC, :HMCDA, :NUTS)
-    @eval getmetricT(::$alg{<:Any,metricT}) where {metricT} = metricT
-end
+getmetricT(::HMC{<:Any,metricT}) where {metricT} = metricT
+getmetricT(::HMCDA{<:Any,metricT}) where {metricT} = metricT
+getmetricT(::NUTS{<:Any,metricT}) where {metricT} = metricT
```
penelopeysm, line 438: Metaprogramming is cool and all, but this wasn't really necessary, imo.

test/mcmc/hmc.jl
```diff
@@ -291,10 +297,11 @@ using Turing
...
         @testset "$(alg)" for alg in algs
             # Construct a HMC state by taking a single step
+            vi = DynamicPPL.VarInfo(gdemo_default)
+            vi = DynamicPPL.link(vi, gdemo_default)
+            ldf = LogDensityFunction(gdemo_default, vi; adtype=Turing.DEFAULT_ADTYPE)
             spl = Sampler(alg)
-            hmc_state = DynamicPPL.initialstep(
-                Random.default_rng(), gdemo_default, spl, DynamicPPL.VarInfo(gdemo_default)
-            )[2]
+            _, hmc_state = AbstractMCMC.step(Random.default_rng(), ldf, spl)
```
penelopeysm: Finally, I think this test reveals one drawback of the current proposal: it becomes more annoying to directly call the AbstractMCMC interface. Let's say we want to benchmark the first step of a given sampler (for example, we were doing this the other day on the Gibbs sampler). Previously, we'd do:

```julia
rng = Random.default_rng()
model = ...
spl = ...
@be AbstractMCMC.step(rng, model, spl)
```

Now, we have to do:

```julia
rng = Random.default_rng()
model = ...
vi = link(VarInfo(model), model)
ldf = LogDensityFunction(model, vi; adtype=AutoForwardDiff())
spl = ...
@be AbstractMCMC.step(rng, ldf, spl)
```

I think this is a fairly small price to pay because the occasions where we reach directly for AbstractMCMC interface are quite few, and the code simplification is more important than this. But I thought this was probably something just worth noting.

This would be less problematic if we introduced more convenient constructors for LDF: https://github.com/TuringLang/DynamicPPL.jl/issues/863 so it might be worth keeping that in mind.

penelopeysm:
@sunxd3 I removed the `Sampler` wrapper for HMC as you suggested, and I think adding the method for `AbstractMCMC.step` will have to come at the end (and we'll put it in DynamicPPL) but I've now added a note to #2555

src/mcmc/Inference.jl
```diff
@@ -1,9 +1,13 @@
 module Inference
 
 using DynamicPPL:
+    DynamicPPL,
     @model,
     Metadata,
     VarInfo,
+    LogDensityFunction,
+    SimpleVarInfo,
+    AbstractVarInfo,
```
sunxd3, line 10: Being pedantic and annoying, I think it'll be better if we reorder so that the `*VarInfo`s stay together.

src/mcmc/algorithm.jl
```diff
@@ -11,4 +12,35 @@ this wrapping occurs automatically.
...
 DynamicPPL.default_chain_type(sampler::Sampler{<:InferenceAlgorithm}) = MCMCChains.Chains
+
+"""
+    update_sample_kwargs(spl::AbstractSampler, N::Integer, kwargs)
+
+Some samplers carry additional information about the keyword arguments that
+should be passed to `AbstractMCMC.sample`. This function provides a hook for
+them to update the default keyword arguments. The default implementation is for
+no changes to be made to `kwargs`.
+"""
```
sunxd3, line 25: I think it's worth adding that this function returns a `NamedTuple` (if I understand correctly).

src/mcmc/algorithm.jl
```diff
@@ -11,4 +12,35 @@ this wrapping occurs automatically.
...
+Return the automatic differentiation (AD) backend to use for the sampler. This
+is needed for constructing a LogDensityFunction. By default, returns nothing,
+i.e. the LogDensityFunction that is constructed will not know how to calculate
+its gradients. If the sampler requires gradient information, then this function
+must return an `ADTypes.AbstractADType`.
+"""
+get_adtype(::AbstractSampler) = nothing
+
+"""
+    requires_unconstrained_space(sampler::AbstractSampler)
```
sunxd3, line 40: Maybe this is a unimportant point: for something like MH, both constrained and unconstrained spaces might be fine.

On a high level, I do think "**require**s_unconstrained_space"-ness is a sampler property, but it's not sufficient condition to control linking.

Don't know if what I wrote makes sense.

src/mcmc/hmc.jl
```diff
@@ -1,7 +1,46 @@
-abstract type Hamiltonian <: InferenceAlgorithm end
+# AbstractSampler interface for Turing
+
+abstract type Hamiltonian <: AbstractMCMC.AbstractSampler end
```
sunxd3, line 3: I want to tag @yebai for visibility.

I can see the motivation for `InferenceAlgorithm`, but I think this makes things a lot cleaner.

src/mcmc/algorithm.jl
sunxd3: also, it's prob worth mentioning what is `N`?

src/mcmc/hmc.jl
```diff
@@ -172,68 +149,58 @@ function find_initial_params(
...
-    vi_original::AbstractVarInfo;
+    ldf::LogDensityFunction,
+    spl::Hamiltonian;
     initial_params=nothing,
     nadapts=0,
     kwargs...,
 )
-    # Transform the samples to unconstrained space and compute the joint log probability.
-    vi = DynamicPPL.link(vi_original, model)
+    ldf.adtype === nothing &&
```
sunxd3, line 160: this is a good sanity check, does it make sense to force ldf and spl has the same adtype?

src/mcmc/abstractmcmc.jl
```diff
@@ -1,15 +1,352 @@
-# TODO: Implement additional checks for certain samplers, e.g.
-# HMC not supporting discrete parameters.
-function _check_model(model::DynamicPPL.Model)
-    return DynamicPPL.check_model(model; error_on_failure=true)
+# This file contains the basic methods for `AbstractMCMC.sample`.
+# The overall aim is that users can call
+#
+#    sample(::Model, ::InferenceAlgorithm, N)
```
sunxd3, line 4: still valid?

src/mcmc/abstractmcmc.jl
```diff
@@ -1,15 +1,352 @@
...
+#
+# and have it be (eventually) forwarded to
+#
+#    sample(::LogDensityFunction, ::Sampler{InferenceAlgorithm}, N)
```
sunxd3, line 8: if not need to change this

src/mcmc/abstractmcmc.jl
```diff
@@ -1,15 +1,352 @@
...
+        kwargs...
+    )
+    sample(
+        [rng::Random.AbstractRNG, ]
+        ldf::DynamicPPL.LogDensityFunction,
+        alg::InferenceAlgorithm,
+        N::Integer;
+        kwargs...
+    )
+
```
sunxd3, line 42: also these

src/mcmc/abstractmcmc.jl
sunxd3, line 4: I assume `InferenceAlgorithm` here is still needed until we update all the interface?

penelopeysm:
Thanks for all the reviews & comments. I'm probably going to have to close this since it's too out of date and also too big, but its spirit will live on; I'll try to break it up into smaller steps 🙂

PR #2589, opened 2025-06-11
clean up some more imports

penelopeysm:
This PR removes the blanket `using DynamicPPL` in `Turing.Inference`.


sunxd3 (approved):
Like it a lot

PR #2590, opened 2025-06-16
sample with LogDensityFunction: part 2 - `ess.jl` + `mh.jl`

penelopeysm:
See #2555

Pending #2588

### DIFF
--- a/test/mcmc/repeat_sampler.jl
+++ b/test/mcmc/repeat_sampler.jl
@@ -14,18 +14,23 @@ using Turing
     num_chains = 2
 
     rng = StableRNG(0)
+    # TODO(penelopeysm): sample on both model and LDF for both samplers.
+    # Right now it can handle LDF but not model (because RepeatSampler
+    # needs to be added to LDFCompatibleSampler)
     for sampler in [MH(), HMC(0.01, 4)]
-        model_or_ldf = if sampler isa MH
-            gdemo_default
+        ctx = DynamicPPL.SamplingContext(rng, sampler)
+        vi = if sampler isa MH
+            DynamicPPL.VarInfo(gdemo_default)
         else
             vi = DynamicPPL.VarInfo(gdemo_default)
             vi = DynamicPPL.link(vi, gdemo_default)
-            LogDensityFunction(gdemo_default, vi; adtype=Turing.DEFAULT_ADTYPE)
+            vi
         end
+        ldf = LogDensityFunction(gdemo_default, vi, ctx; adtype=Turing.DEFAULT_ADTYPE)
 
         chn1 = sample(
             copy(rng),
-            model_or_ldf,
+            ldf,
             sampler,
             MCMCThreads(),
             num_samples,
@@ -34,7 +39,7 @@ using Turing
         )
         repeat_sampler = RepeatSampler(sampler, num_repeats)
         chn2 = sample(
-            copy(rng), model_or_ldf, repeat_sampler, MCMCThreads(), num_samples, num_chains
+            copy(rng), ldf, repeat_sampler, MCMCThreads(), num_samples, num_chains
         )
         @test chn1.value == chn2.value
     end


src/mcmc/abstractmcmc.jl
```diff
@@ -80,7 +79,7 @@ function AbstractMCMC.sample(
     ctx = if ldf.context isa SamplingContext
         ldf.context
     else
-        SamplingContext(rng, spl)
+        SamplingContext(rng, spl, ldf.context)
     end
```
penelopeysm, line 83: otherwise the existing context won't be obeyed

src/mcmc/mh.jl
```diff
@@ -139,18 +139,23 @@ struct MH{P} <: InferenceAlgorithm
...
-function transform_maybe(proposal::AMH.RandomWalkProposal)
-    return AMH.RandomWalkProposal(Bijectors.transformed(proposal.proposal))
-end
-
-function MH(model::Model; proposal_type=AMH.StaticProposal)
-    priors = DynamicPPL.extract_priors(model)
-    props = Tuple([proposal_type(prop) for prop in values(priors)])
-    vars = Tuple(map(Symbol, collect(keys(priors))))
-    priors = map(transform_maybe, NamedTuple{vars}(props))
-    return AMH.MetropolisHastings(priors)
```
penelopeysm, line 153: this code was broken for a long time and nobody bothered to update it. the idea would be that this would return an `AMH.MetropolisHastings` which then needed to be wrapped in an `ExternalSampler`. however the external sampler would break (see tests below) I just removed it because it wasn't documented and wasn't working.

test/mcmc/mh.jl
```diff
@@ -21,31 +20,72 @@ GKernel(var) = (x) -> Normal(x, sqrt.(var))
...
-        s5 = Gibbs(:m => MH(), :s => MH())
-        c5 = sample(gdemo_default, s5, N)
-
-        # s6 = externalsampler(MH(gdemo_default, proposal_type=AdvancedMH.RandomWalkProposal))
-        # c6 = sample(gdemo_default, s6, N)
-
-        # NOTE: Broken because MH doesn't really follow the `logdensity` interface, but calls
-        # it with `NamedTuple` instead of `AbstractVector`.
-        # s7 = externalsampler(MH(gdemo_default, proposal_type=AdvancedMH.StaticProposal))
-        # c7 = sample(gdemo_default, s7, N)
```
penelopeysm, line 48: these were the corresponding tests

test/mcmc/mh.jl
```diff
@@ -225,62 +271,17 @@ GKernel(var) = (x) -> Normal(x, sqrt.(var))
         end
     end
 
-    @testset "MH link/invlink" begin
-        vi_base = DynamicPPL.VarInfo(gdemo_default)
```
penelopeysm, line 229: these tests covered by `requires_unconstrained_space`

test/mcmc/mh.jl
```diff
@@ -225,62 +271,17 @@ GKernel(var) = (x) -> Normal(x, sqrt.(var))
...
-            StableRNG(seed),
-            gdemo_default_prior,
-            alg,
-            n;
-            discard_initial=burnin,
-            thinning=10,
+            StableRNG(seed), model, MH(), 10_000; discard_initial=10_000, thinning=10
         )
-        check_numerical(chain, [:s, :m], [mean(InverseGamma(2, 3)), 0]; atol=0.3)
+        @test mean(chain[:a]) ≈ 0.0 atol = 0.1
```
penelopeysm, line 284: this test was annoyingly difficult to get right. with the gdemo model i had to bump this up to 50000 samples to get it to work reasonably often. I replaced it with a simpler model for which it's obvious when it's being sampled from the prior (a = 0) rather than the posterior (a = 2.5).

penelopeysm:
Closing as it's too out of date; but the code will probably be recycled into a new PR at some point.

Issue #2602, opened 2025-06-25
Minimum of OrderedLogistic is 0, while we can't sample 0 from the distribution?


Noricc:
### Minimal working example

```julia
using Distributions, Turing

d = OrderedLogistic(0, [1])

println(pdf(d, 0)) # Prints 0.2689...
println(pdf(d, 1)) # Prints 0.73105...
println(pdf(d, 2)) # Pritns 0.2689...

# These sum to more than 1

# The support of the distribution also includes 0
println(support(d)) # Prints 0:2

# But we can't sample 0
println(mean(rand(d, 1000) .== 0)) # Prints 0.0
```

### Description

When we look at the support (or the pdf) of the `OrderedLogistic` distribution, we see that it's possible to get an outcome of 0, while the documentation says that the support of the distribution is 1..K. 

I suspect the problem is that the `minimum` of the distribution is defined as 0, while it should be 1. 

### Julia version info

<details><summary>versioninfo()</summary>

```
Julia Version 1.11.5
Commit 760b2e5b739 (2025-04-14 06:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, skylake)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```
</details>

### Manifest

<details><summary>]st --manifest</summary>

```
Status `~/.julia/environments/v1.11/Manifest.toml`
  [47edcb42] ADTypes v1.15.0
  [621f4979] AbstractFFTs v1.5.0
  [80f14c24] AbstractMCMC v5.6.3
  [7a57a42e] AbstractPPL v0.11.0
  [1520ce14] AbstractTrees v0.4.5
  [7d9f7c33] Accessors v0.1.42
  [79e6a3ab] Adapt v4.3.0
  [0bf59076] AdvancedHMC v0.8.0
  [5b7e9947] AdvancedMH v0.8.8
⌅ [576499cb] AdvancedPS v0.6.2
  [b5ca4192] AdvancedVI v0.4.1
  [66dad0bd] AliasTables v1.1.3
  [dce04be8] ArgCheck v2.5.0
  [7d9fca2a] Arpack v0.5.4
  [4fba245c] ArrayInterface v7.19.0
[...]

PR #2603, opened 2025-06-25
Set minimum of OrderedLogistic distribution to 1

Noricc:
The minimum of the OrderedLogistic distribution is supposed to be 1, not 0. This is consistent with the documentation, and the behavior of the distribution when using rand().

Fixes bug #2602

### DIFF
--- a/src/stdlib/distributions.jl
+++ b/src/stdlib/distributions.jl
@@ -126,7 +126,7 @@ function OrderedLogistic(η, cutpoints::AbstractVector)
     return OrderedLogistic{typeof(η),typeof(cutpoints)}(η, cutpoints)
 end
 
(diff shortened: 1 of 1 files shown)


penelopeysm:
That's great, thanks! Could I trouble you to add a test to `test/stdlib/distributions.jl` (not sure exactly what to test, maybe that logpdf of the support adds up to 1?)

Noricc:
I can try later today!

Noricc:
Added a test now.

penelopeysm:
Wonderful, thank you for the great work @Noricc! I'll just bump the patch so that we can release, then will merge once CI is happy.

penelopeysm (approved):

PR #2608, opened 2025-07-04
Replace 2018 AISTATS conference paper with the latest 2025 ACM paper

sunxd3:


penelopeysm:
I wonder if we want to keep both, or just one? cc @yebai

penelopeysm (approved):
thanks:)

PR #2625, opened 2025-07-22
DPPL 0.37 compat for particle MCMC

mhauru:
This may or may not be done functionality-wise, except VariableOrderAccumulator seems to go wrong when using Gibbs. (This is probably the same problem as before in merging accumulators.) Let's see how CI does. The code has some pretty ugly bits though, that should be improved.

FYI @penelopeysm

### DIFF
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -177,11 +177,14 @@ function DynamicPPL.tilde_assume(context::GibbsContext, right, vn, vi)
         # Fall back to the default behavior.
         DynamicPPL.tilde_assume(child_context, right, vn, vi)
     elseif has_conditioned_gibbs(context, vn)
-        # TODO(DPPL0.37/penelopeysm): Unsure if this is bad for SMC as it
-        # will trigger resampling. We may need to do a special kind of observe
-        # that does not trigger resampling.
-        global_vi = get_global_varinfo(context)
-        val = global_vi[vn]
+        # This branch means that a different sampler is supposed to handle this
+        # variable. From the perspective of this sampler, this variable is
+        # conditioned on, so we can just treat it as an observation.
+        # The only catch is that the value that we need is to be obtained from
+        # the global VarInfo (since the local VarInfo has no knowledge of it).
+        # Note that tilde_observe!! will trigger resampling in particle methods
+        # for variables that are handled by other Gibbs component samplers.
+        val = get_conditioned_gibbs(context, vn)
         DynamicPPL.tilde_observe!!(child_context, right, val, vn, vi)
     else
         # If the varname has not been conditioned on, nor is it a target variable, its
@@ -219,11 +222,9 @@ function DynamicPPL.tilde_assume(
         # conditioned on, so we can just treat it as an observation.
         # The only catch is that the value that we need is to be obtained from
         # the global VarInfo (since the local VarInfo has no knowledge of it).
-        # TODO(DPPL0.37/penelopeysm): Unsure if this is bad for SMC as it
-        # will trigger resampling. We may need to do a special kind of observe
-        # that does not trigger resampling.
-        global_vi = get_global_varinfo(context)
-        val = global_vi[vn]
+        # Note that tilde_observe!! will trigger resampling in particle methods
+        # for variables that are handled by other Gibbs component samplers.
+        val = get_conditioned_gibbs(context, vn)
         DynamicPPL.tilde_observe!!(child_context, right, val, vn, vi)
     else
         # If the varname has not been conditioned on, nor is it a target variable, its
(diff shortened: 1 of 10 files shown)


penelopeysm:
As discussed earlier it's looking good. Finer details we talked about:

- Having a specific accumulator for the particle weight (i.e., the log-likelihood of the most recent observation). Adding a quantity to this accumulator would be responsible for triggering the resampling (via Libtask.produce).
- This would allow us to accumulate logprior / loglikelihood ( / logjac) as usual, so that they don't have a special meaning inside the pMCMC file (the current code hijacks `vi.logp[]` for its own purposes).
- Unsure whether `@addlogprob!` should also trigger resampling. We think it probably should (if it is incrementing the likelihood by a nonzero amount).
- We'll probably leave `set_retained_vns_del!` there for now.

penelopeysm:
https://github.com/TuringLang/Turing.jl/blob/b0df6a66a0b5ce28462fabc4380e3c36647235cb/src/mcmc/particle_mcmc.jl#L45-L51

I realised the call to `getlogjoint` catches not only `@addlogprob!` but also Gibbs-conditioned variables (I think).

mhauru:
Most, though not quite all, of the remaining test failures are because `merge` and `subset` don't handle `VariableOrderAccumulator`. Do you want to merge this first and make that a separate PR?

penelopeysm:
Quite happy to!

src/mcmc/particle_mcmc.jl
```diff
@@ -478,13 +505,128 @@ function AdvancedPS.Trace(
...
+)
+    return ProduceLogLikelihoodAccumulator(acc.logp + acc2.logp)
+end
+
+function Base.:+(
+    acc1::ProduceLogLikelihoodAccumulator, acc2::DynamicPPL.LogLikelihoodAccumulator
+)
+    Libtask.produce(acc2.logp)
+    return ProduceLogLikelihoodAccumulator(acc1.logp + acc2.logp)
+end
```
penelopeysm: I wonder if this might be a bit unintuitive. I'd normally expect a function called `+` to be commutative, but here the order of arguments matters.

mhauru: I've actually been thinking of getting rid of this whole function for all of the accs, and calling it `accumulate` and having it take a number rather than an accumulator.

penelopeysm: I'd be quite on board with that. Would that be part of the larger simple-float-accumulator refactor?

src/mcmc/particle_mcmc.jl
```diff
@@ -316,6 +305,7 @@ function DynamicPPL.initialstep(
     vi::AbstractVarInfo;
     kwargs...,
 )
+    vi = DynamicPPL.setacc!!(vi, ProduceLogLikelihoodAccumulator())
```
penelopeysm, line 308: line 337 needs the same fix as per discussion in https://github.com/TuringLang/DynamicPPL.jl/pull/999.

penelopeysm (changes requested):

src/mcmc/particle_mcmc.jl
penelopeysm: (the fix for this is in #2629)

src/mcmc/gibbs.jl
```diff
@@ -177,13 +177,12 @@ function DynamicPPL.tilde_assume(context::GibbsContext, right, vn, vi)
...
-        # accumulated just gets discarded with `_`.
-        value, _ = DynamicPPL.tilde_assume(
-            child_context, right, vn, get_global_varinfo(context)
-        )
-        value, vi
+        # TODO(DPPL0.37/penelopeysm): Unsure if this is bad for SMC as it
+        # will trigger resampling. We may need to do a special kind of observe
+        # that does not trigger resampling.
+        global_vi = get_global_varinfo(context)
+        val = global_vi[vn]
```
penelopeysm:
```suggestion
        # This branch means that a different sampler is supposed to handle this
        # variable. From the perspective of this sampler, this variable is
        # conditioned on, so we can just treat it as an observation.
        # The only catch is that the value that we need is to be obtained from
        # the global VarInfo (since the local VarInfo has no knowledge of it).
        # Note that tilde_observe!! will trigger resampling in particle methods
        # for variables that are handled by other Gibbs component samplers.
        val = get_conditioned_gibbs(context, vn)
```

src/mcmc/gibbs.jl
```diff
@@ -210,13 +209,27 @@ function DynamicPPL.tilde_assume(
...
+        # variable. From the perspective of this sampler, this variable is
+        # conditioned on, so we can just treat it as an observation.
+        # The only catch is that the value that we need is to be obtained from
+        # the global VarInfo (since the local VarInfo has no knowledge of it).
+        # TODO(DPPL0.37/penelopeysm): Unsure if this is bad for SMC as it
+        # will trigger resampling. We may need to do a special kind of observe
+        # that does not trigger resampling.
+        global_vi = get_global_varinfo(context)
+        val = global_vi[vn]
+        DynamicPPL.tilde_observe!!(child_context, right, val, vn, vi)
```
penelopeysm:
```suggestion
        # Note that tilde_observe!! will trigger resampling in particle methods
        # for variables that are handled by other Gibbs component samplers.
        val = get_conditioned_gibbs(context, vn)
        DynamicPPL.tilde_observe!!(child_context, right, val, vn, vi)
```

PR #2629, opened 2025-07-25
"Fixes" for PG-in-Gibbs

penelopeysm:
This PR:

(1) Adds `ProduceLogLikelihoodAccumulator` on steps n ≥ 2 of PG, otherwise `Libtask.produce` is never called.

(2) Removes all uses of `set_retained_vns_del!`, to avoid errors with PG-within-Gibbs (issue is that it would get a VariableOrderAccumulator from a different component sampler that didn't have the variables that PG cared about).

The role of `set_retained_vns_del!` was to mark specific variables for resampling by comparing `order` against `num_produce`. The idea was that we would only mark variables for resampling if they were 'in the future' i.e. we hadn't reached their tilde-statements in the model yet.

Instead of doing that, we now simply mark **everything** as to be resampled. This is fine because models can only be propagated in the forward direction i.e. **there is really no harm in marking previously seen variables for resampling** because we won't see them again! There are two cases where this could lead to different behaviour:

 - if the same variable appears on the lhs of tilde twice, which is generally invalid anyway.
 - if the same varinfo (with the `del` flags set) is then re-used for evaluating the model, e.g. to calculate logp. To circumvent this problem, I introduce an inverse function which marks all variables as **not** being for resampling.

**Collectively, this completely removes the need to track which variables have already been seen.** With these proposed changes there is simply no need to track the `order` of a variable or the `num_produce` of the varinfo, meaning that we can remove VariableOrderAccumulator entirely (with concomitant positive implications for performance).

I have not exhaustively tested these changes on `breaking`, but I have tried making exactly the same changes to the `del` flag on `main` and they lead to no changes in the sampled results if the PRNG seed is set.

----------------

> [!NOTE]  
> While we aren't quite there yet, this change also paves the way for eventually getting rid of the `del` flag entirely (https://github.com/TuringLang/DynamicPPL.jl/issues/982).
>
> Notice that the changes in this PR already remove the need for a per-variable `del` flag; instead we can have a single boolean for the entire VarInfo which marks whether something should be resampled or not. This would already represent a substantial simplification.
>
> However, ultimately we should like to shift this information outside of the VarInfo because then it means that VarInfo doesn't need to track information that only PG needs. The way I would propose to do this is with contexts: one which never resamples variables (which would correspond to all `del` flags unset), and one which always resamples variables (corresponding to all `del` flags set). Notice further that these roles are very similar to `DefaultContext` and the upcoming `InitContext`. There are probably a ton of subtleties (e.g., in that we have to use the traced rng rather than just naively resampling), but I think there is light at the end of the tunnel.

### DIFF
--- a/src/mcmc/particle_mcmc.jl
+++ b/src/mcmc/particle_mcmc.jl
@@ -580,3 +580,97 @@ function Libtask.might_produce(
     return true
 end
 Libtask.might_produce(::Type{<:Tuple{<:DynamicPPL.Model,Vararg}}) = true
+
+"""
+    ProduceLogLikelihoodAccumulator{T<:Real} <: AbstractAccumulator
+
+Exactly like `LogLikelihoodAccumulator`, but calls `Libtask.produce` on every increase.
+
+# Fields
+$(TYPEDFIELDS)
+"""
+struct ProduceLogLikelihoodAccumulator{T<:Real} <: DynamicPPL.AbstractAccumulator
+    "the scalar log likelihood value"
+    logp::T
+end
+
+"""
+    ProduceLogLikelihoodAccumulator{T}()
+
+Create a new `ProduceLogLikelihoodAccumulator` accumulator with the log likelihood of zero.
+"""
+ProduceLogLikelihoodAccumulator{T}() where {T<:Real} =
+    ProduceLogLikelihoodAccumulator(zero(T))
+function ProduceLogLikelihoodAccumulator()
+    return ProduceLogLikelihoodAccumulator{DynamicPPL.LogProbType}()
+end
+
+Base.copy(acc::ProduceLogLikelihoodAccumulator) = acc
+
+function Base.show(io::IO, acc::ProduceLogLikelihoodAccumulator)
+    return print(io, "ProduceLogLikelihoodAccumulator($(repr(acc.logp)))")
+end
+
+function Base.:(==)(
+    acc1::ProduceLogLikelihoodAccumulator, acc2::ProduceLogLikelihoodAccumulator
+)
+    return acc1.logp == acc2.logp
+end
+
+function Base.isequal(
+    acc1::ProduceLogLikelihoodAccumulator, acc2::ProduceLogLikelihoodAccumulator
+)
+    return isequal(acc1.logp, acc2.logp)
+end
+
+function Base.hash(acc::ProduceLogLikelihoodAccumulator, h::UInt)
+    return hash((ProduceLogLikelihoodAccumulator, acc.logp), h)
+end
+
+# Note that this uses the same name as `LogLikelihoodAccumulator`. Thus only one of the two
+# can be used in a given VarInfo.
+DynamicPPL.accumulator_name(::Type{<:ProduceLogLikelihoodAccumulator}) = :LogLikelihood
+
+function DynamicPPL.split(::ProduceLogLikelihoodAccumulator{T}) where {T}
+    return ProduceLogLikelihoodAccumulator(zero(T))
+end
+
+function DynamicPPL.combine(
+    acc::ProduceLogLikelihoodAccumulator, acc2::ProduceLogLikelihoodAccumulator
+)
+    return ProduceLogLikelihoodAccumulator(acc.logp + acc2.logp)
+end
+
+function Base.:+(
+    acc1::ProduceLogLikelihoodAccumulator, acc2::DynamicPPL.LogLikelihoodAccumulator
+)
+    Libtask.produce(acc2.logp)
+    return ProduceLogLikelihoodAccumulator(acc1.logp + acc2.logp)
+end
+
+function Base.zero(acc::ProduceLogLikelihoodAccumulator)
+    return ProduceLogLikelihoodAccumulator(zero(acc.logp))
+end
+
+function DynamicPPL.accumulate_assume!!(
+    acc::ProduceLogLikelihoodAccumulator, val, logjac, vn, right
+)
+    return acc
+end
+function DynamicPPL.accumulate_observe!!(
+    acc::ProduceLogLikelihoodAccumulator, right, left, vn
+)
+    return acc +
+           DynamicPPL.LogLikelihoodAccumulator(Distributions.loglikelihood(right, left))
+end
+
+function Base.convert(
+    ::Type{ProduceLogLikelihoodAccumulator{T}}, acc::ProduceLogLikelihoodAccumulator
+) where {T}
+    return ProduceLogLikelihoodAccumulator(convert(T, acc.logp))
+end
+function DynamicPPL.convert_eltype(
+    ::Type{T}, acc::ProduceLogLikelihoodAccumulator
+) where {T}
+    return ProduceLogLikelihoodAccumulator(convert(T, acc.logp))
+end
--- a/src/mcmc/particle_mcmc.jl
+++ b/src/mcmc/particle_mcmc.jl
@@ -11,6 +11,12 @@ Set the "del" flag for all variables in the VarInfo `vi`, thus marking them for
 resampling.
 """
 function set_all_del!(vi::AbstractVarInfo)
+    # TODO(penelopeysm): Instead of being a 'del' flag on the VarInfo, we
+    # could either:
+    # - keep a boolean 'resample' flag on the trace, or
+    # - modify the model context appropriately.
+    # However, this refactoring will have to wait until InitContext is
+    # merged into DPPL.
     for vn in keys(vi)
         DynamicPPL.set_flag!(vi, vn, "del")
     end
@@ -59,13 +65,6 @@ end
 function AdvancedPS.advance!(
     trace::AdvancedPS.Trace{<:AdvancedPS.LibtaskModel{<:TracedModel}}, isref::Bool=false
 )
-    # We want to increment num produce for the VarInfo stored in the trace. The trace is
-    # mutable, so we create a new model with the incremented VarInfo and set it in the trace
-    model = trace.model
-    model = Accessors.@set model.f.varinfo = DynamicPPL.increment_num_produce!!(
-        model.f.varinfo
-    )
-    trace.model = model
     # Make sure we load/reset the rng in the new replaying mechanism
     isref ? AdvancedPS.load_state!(trace.rng) : AdvancedPS.save_state!(trace.rng)
     score = consume(trace.model.ctask)
@@ -73,23 +72,23 @@ function AdvancedPS.advance!(
 end
 
 function AdvancedPS.delete_retained!(trace::TracedModel)
-    # TODO(DPPL0.37/penelopeysm): Explain this a bit better.
-    #
     # This method is called if, during a CSMC update, we perform a resampling
     # and choose the reference particle as the trajectory to carry on from.
     # In such a case, we need to ensure that when we continue sampling (i.e.
     # the next time we hit tilde_assume), we don't use the values in the 
     # reference particle but rather sample new values.
-    # In this implementation, we indiscriminately set the 'del' flag for all
-    # variables in the VarInfo. This is slightly overkill: it is not necessary
-    # to set the 'del' flag for variables that were already sampled. However,
-    # it allows us to avoid using DynamicPPL.set_retained_vns_del!.
+    #
+    # Here, we indiscriminately set the 'del' flag for all variables in the
+    # VarInfo. This is slightly overkill: it is not necessary to set the 'del'
+    # flag for variables that were already sampled. However, it allows us to
+    # avoid keeping track of which variables were sampled, which leads to many
+    # simplifications in the VarInfo data structure.
     set_all_del!(trace.varinfo)
     return trace
 end
 
 function AdvancedPS.reset_model(trace::TracedModel)
-    return Accessors.@set trace.varinfo = DynamicPPL.reset_num_produce!!(trace.varinfo)
+    return trace
 end
 
 function Libtask.TapedTask(taped_globals, model::TracedModel; kwargs...)
@@ -213,7 +212,6 @@ function DynamicPPL.initialstep(
 )
     # Reset the VarInfo.
     vi = DynamicPPL.setacc!!(vi, ProduceLogLikelihoodAccumulator())
-    vi = DynamicPPL.reset_num_produce!!(vi)
     set_all_del!(vi)
     vi = DynamicPPL.resetlogp!!(vi)
     vi = DynamicPPL.empty!!(vi)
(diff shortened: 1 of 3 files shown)


test/essential/container.jl
```diff
@@ -9,19 +9,26 @@ using Turing
...
         b ~ Gamma(2, 3)
+        println("b")
         1 ~ Bernoulli(x / 2)
+        println("1")
         c ~ Beta()
+        println("c")
         0 ~ Bernoulli(x / 2)
+        println("0")
         return x
     end
```
penelopeysm: remove

src/mcmc/particle_mcmc.jl
```diff
@@ -33,26 +65,30 @@ end
 function AdvancedPS.advance!(
     trace::AdvancedPS.Trace{<:AdvancedPS.LibtaskModel{<:TracedModel}}, isref::Bool=false
 )
-    # We want to increment num produce for the VarInfo stored in the trace. The trace is
-    # mutable, so we create a new model with the incremented VarInfo and set it in the trace
-    model = trace.model
-    model = Accessors.@set model.f.varinfo = DynamicPPL.increment_num_produce!!(
-        model.f.varinfo
-    )
-    trace.model = model
```
penelopeysm, line 42: Since we don't ever use num_produce, we don't need to modify it either.

test/mcmc/gibbs.jl
penelopeysm, line 1: Changes in this file are all due to the issue with Gibbs-conditioned-variables now triggering resampling. Thus now it's forbidden to have a dynamic number of Gibbs-conditioned-variables.

test/mcmc/gibbs.jl
```diff
@@ -565,42 +565,6 @@ end
...
-
-        m = dynamic_model_for_ess()
-        chain = sample(m, Gibbs(:b => PG(10), :x => ESS()), 2000; discard_initial=100)
-        means = Dict(:b => 0.5, "x[1]" => 1.0, "x[2]" => 2.0)
-        stds = Dict(:b => 0.5, "x[1]" => 1.0, "x[2]" => 1.0)
-        for vn in keys(means)
-            @test isapprox(mean(skipmissing(chain[:, vn, 1])), means[vn]; atol=0.1)
-            @test isapprox(std(skipmissing(chain[:, vn, 1])), stds[vn]; atol=0.1)
-        end
-    end
```
penelopeysm: _In general_, we now can't handle the different number of Gibbs-conditioned-variables. However, in the specific case of these tests, we can change the model to something equivalent which has the same number of observations:

```julia
@model function dynamic_model_for_ess()
    b ~ Bernoulli()
    if b
        x ~ MvNormal([1.0], I)
    else
        x ~ MvNormal([1.0, 2.0], I)
    end
end
```

The one below this can similarly use `filldist`.

Do you think it's worth keeping these tests? i.e. was the point to make sure that the sampling results are numerically correct, or was the point of the test to make sure that you could have different numbers of Gibbs-conditioned variables (in which case rewriting it would defeat the purpose)?

penelopeysm: (If it's the former, I think we should include a likelihood term, I think being able to sample from the prior correctly is a bit of a low bar to clear for an inference algorithm implementation.)

mhauru: It's a regression test, and I don't know what the original issue was that caused it to error. We don't have too many dynamic model tests, so I would keep it with the `MvNormal` because it doesn't cost us much and might still catch something.

An alternative, though: Any interest in turning this into a test to check whether you can still do dynamic models with loops if you balance out the branches with `@addlogprob!!(0.0)` calls? Hacky, but I kinda want to know that it works as a workaround if need be. So that would be something like
```julia
@model function dynamic_model_for_ess()
    b ~ Bernoulli()
    x_length = b ? 1 : 2
    max_x = 2
    x = Vector{Float64}(undef, x_length)
    for i in 1:x_length
        x[i] ~ Normal(i, 1.0)
    end
    for _ in (x_length+1):max_x:
        @addlogprob!!(0.0)
    end
end
```

penelopeysm: Okay, happy to do that, will add a comment with the intent too

penelopeysm: Currently unsure what to do with the tests.

```julia
@model function dynamic_bernoulli()
    b ~ Bernoulli()
    if b
        x ~ MvNormal([1.0], I)
    else
        x ~ MvNormal([1.0, 2.0], I)
    end
end

chain = sample(dynamic_bernoulli(), Gibbs(:b => PG(10), :x => ESS()), 2000; discard_initial=100)
```

This errors because if `x` in the global varinfo is a length-2 vector and PG selects `b = 1`, then `logpdf(MvNormal([1.0], I), x)` will fail.

```julia
@model function dynamic_bernoulli_2()
    b ~ Bernoulli()
    x_length = b ? 1 : 2
    max_x = 2
    x = Vector{Float64}(undef, x_length)
    for i in 1:x_length
        x[i] ~ Normal(i, 1.0)
    end
    for _ in (x_length+1):max_x
        @addlogprob!(0.0)
    end
end

chain = sample(dynamic_bernoulli_2(), Gibbs(:b => PG(10), :x => ESS()), 2000; discard_initial=100)
```

This runs but gives a rather incorrect value of `b = 0.75` (ish, depends on number of iterations and stuff).

mhauru: Isn't the latter then doing its job in catching a bug for us?

penelopeysm: Sort of; but I just found out that if you run `dynamic_bernoulli_2()` with `StableRNG(468)` it runs fine (and gives a wrong mean), but with `StableRNG(469)` it crashes with

```julia
julia> mean(sample(StableRNG(469), model, Gibbs(:b => PG(10), :x => ESS()), 2000; discard_initial=100))
ERROR: mis-aligned execution traces: # particles = 10 # completed trajectories = 1. Please make sure the number of observations is NOT random.
```

penelopeysm: I actually wonder if this is a Libtask thing. Let's say that we have a particle with `b == false`, so `x_length = 2`. After it sees `x[1]`, it has to resample. Suppose it then resamples (transforms itself) into a particle with `b == true` and `x_length = 1`. What happens then, does it go on to sample `x[2]` or does it exit the loop because the termination condition `x_length == 1` is now hit?

penelopeysm: I suppose it should exit the loop because in the new (forked) particle `b` was always `true` to begin with, so `x_length` was always `1`. I really didn't expect to have to think about many-world theories today.

mhauru: I think I don't understand PG well enough. I thought if anything would ever change the value of `b`, that would also be paired with re-executing the model from `b ~ Bernoulli()` onwards. If the value of `b` can change "retroactively" e.g. in the middle of the loop then I have to rethink things quite a lot.

penelopeysm: Oh, no, it doesn't rewind the model and re-execute. `AdvancedPS.reweight!` calls `AdvancedPS.advance!` which calls `Libtask.consume` which only ever moves the model forward. That's all in https://github.com/TuringLang/AdvancedPS.jl/blob/main/src/container.jl

The resampling step is in `resample_propagate!` (same file) and it is kind of like this: let's say you have

- particle 1 `b = 1`
- particle 2 `b = 2`

then you hit an observe statement. It uses the likelihood of that observe statement as the unnormalised weights (let's say it's 1.5 and 2.5), then it generates 2 new particles by drawing from `rand(Categorical(weights ./ norm(weights)))` = `rand(Categorical([0.4, 0.6])`.

Suppose the `rand(Categorical)` returns `[1, 1]`. In that case it has to create two new copies of the first particle, and it does that with `AdvancedPS.fork` which just copies/deepcopies the entire Libtask trace https://github.com/TuringLang/AdvancedPS.jl/blob/ee1052dc22c1115cc2acba0f2db1cf885a346cfb/ext/AdvancedPSLibtaskExt.jl#L101-L108 and sets a new rng seed so that they don't both generate the same variables on the next assume statement.

So it doesn't rewind anything, rather it makes a copy of the trace and continues executing it from there onwards. You can see this if you insert a bunch of prints throughout the model.

But in the new copy of the trace there's never any question about what `b` was; it was always `1` as far as that trace is concerned.

penelopeysm: I think I actually understand why the mean is wrong now (at least when it runs; I don't know why it crashes). Suppose that `x[1] = 1.0` and `x[2] = 2.0` (on average this is true).

For particles with `b = true` (i.e. only observe `x[1]`), `Libtask.produce` gets called twice: once with `logpdf(Normal(1.0), 1.0) = -0.919` and once with a manual logpdf of `0`.

For particles with `b = false` (i.e. observe `x[1]` and `x[2]`), `Libtask.produce` gets called twice, once with
`logpdf(Normal(1.0), 1.0) = -0.919` and once with `logpdf(Normal(2.0), 2.0) = -0.919`.

When PG starts its thing, it just samples `b` from the prior, so there should be 50% `true` and 50% `false`.

On the first reweighting, the logpdf's (i.e., weights) are equal and thus one would expect that the resampling step generates an equal amounts of `true` and `false`.

On the second reweighting however the weights are different. The unnormalised weights are `exp(0) = 1` and `exp(-0.919) = 0.3989`, which is a ratio of about 2.5, so when resampling you're approximately 2.5x more likely to get a particle with `b = true` rather than `b = false`.

So `mean(b)` would be expected to be `2.5 / (2.5 + 1) = 0.715`, which is sort of in the right region.

Furthermore, in general `x[2]` will not be equal to 2.0, meaning that `logpdf(Normal(2.0), x[2])` will be smaller and the ratio of weights larger than 2.5, and `mean(b)` will be even more biased towards 1. This is exactly what we observe with `mean(b)` coming out at 0.772.

penelopeysm: Indeed if you make this change:

```diff
  for _ in (x_length+1):max_x
-     @addlogprob!(0.0)
+     # same as @addlogprob!(logpdf(Normal(2.0), 2.0))
+     @addlogprob!(logpdf(Normal(), 0.0))
  end
```

you get back a value of `mean(b) = 0.592` which is closer to 0.5 (albeit still larger than 0.5 because of the last paragraph in the previous comment).

And to completely remove the bias, I think you'd have to add a term that is the expectation value of `logpdf(Normal(), x)` where `x ~ Normal()`. That's a slightly nasty integral, but we can approximate it with Monte Carlo ([or alternatively use WolframAlpha](https://www.wolframalpha.com/input?i=integral+from+-inf+to+inf+of+%28-0.5+*+%28log%282*pi%29+%2B+x%5E2%29%29+*+%281%2Fsqrt%282*pi%29%29+*+exp%28-x%5E2%2F2%29+dx)):

```julia
julia> mean(logpdf.(Normal(), randn(10_000_000)))
-1.418849291536427
```

Calling `@addlogprob!` with this gives... `mean(b) = 0.5050`!

penelopeysm: Finally, none of this actually explains why it used to "work" with the previous non-tilde-observe Gibbs sampler. The current behaviour makes complete sense to me given our tilde-observe-Gibbs implementation, but it still seems to me that the exact same thing should have happened before this PR. To understand that I might have to take another long walk.

mhauru: Thanks. That makes sense.

I think I know why it used to work: The problem was masked by another bug. Namely, we would accumulate the logp contributions from the Gibbs-conditioned `tilde_assume` statements into the VarInfo, and then add it to the score/log likelihood at the next produce/resampling, with the line `return score + DynamicPPL.getlogp(varinfo)` in `advance!`. But in the above model there is no next produce/resampling, because there is no likelihood term. Observe instead this, running on latest release:
```julia
julia> module MWE

       using Turing

       @model function dynamic_bernoulli()
           b ~ Bernoulli()
           x_length = b ? 1 : 2
           x = Vector{Float64}(undef, x_length)
           for i in 1:x_length
               x[i] ~ Normal(i, 1.0)
           end
           0.0 ~ Normal()
       end

       chain = sample(dynamic_bernoulli(), Gibbs(:b => PG(10), :x => ESS()), 2000; discard_initial=100)
       describe(chain)
       end
WARNING: replacing module MWE.
Sampling 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:04
Chains MCMC chain (2000×4×1 reshape(::Matrix{Union{Missing, Float64}}, 2000, 4, 1) with eltype Union{Missing, Float64}):

Iterations        = 101:1:2100
Number of chains  = 1
Samples per chain = 2000
Wall duration     = 4.23 seconds
Compute duration  = 4.23 seconds
parameters        = b, x[1], x[2]
internals         = lp

Summary Statistics
  parameters      mean       std       mcse    ess_bulk    ess_tail       rhat   ess_per_sec
      Symbol   Float64   Float64   Float64?    Float64?    Float64?   Float64?      Float64?

           b    0.7975    0.4020     0.0098   1692.0019         NaN     0.9995      399.7170
        x[1]    0.9588    1.0110     0.0232   1888.4870   1014.7637     0.9999      446.1344
        x[2]    2.0290    0.9868    missing     missing     missing    missing       missing

Quantiles

           b    0.0000    1.0000    1.0000    1.0000    1.0000
        x[1]   -1.0495    0.2851    0.9589    1.6325    2.9982
        x[2]    0.1122    1.2627    2.0212    2.6988    3.9777
```

mhauru: On your earlier explanation of how PG works, that's helpful, and matches what I imagined it would be. In that light, I just didn't understand how this could come to pass:
>I actually wonder if this is a Libtask thing. Let's say that we have a particle with b == false, so x_length = 2. After it sees x[1], it has to resample. Suppose it then resamples (transforms itself) into a particle with b == true and x_length = 1. What happens then, does it go on to sample x[2] or does it exit the loop because the termination condition x_length == 1 is now hit?

penelopeysm:
Pretty sure that the diff of this PR matches when you reviewed so will merge. Dynamic models to be discussed at next Monday's team meeting...

penelopeysm:
VNT "broke" this all over again, although it was already broken before.

A large part of the bad behaviour with models like these

```julia
@model function dynamic_bernoulli()
   b ~ Bernoulli()
   x_length = b ? 1 : 2
   x = Vector{Float64}(undef, x_length)
   for i in 1:x_length
       x[i] ~ Normal(i, 1.0)
   end
end
```

was due to the fact that the value of `x[2]` from previous iterations would be cached and reused in subsequent iterations, because of the behaviour of VarInfo merging. This problem is still there, but in a different form, and for reasons I don't know, the behaviour is now different, in that things can still sample, but the mean is no longer the same. I think it's to do with values being merged in a different way now that we have PartialArrays.

I think until we figure that out for good, it's meaningless to keep pretending that we promise any particular results when you run these models. It's overpromising and I don't think it does us any good to attempt to support behaviour that existed for the wrong reasons.

PR #2632, opened 2025-07-27
Avoid splitting up varnames until absolutely necessary

penelopeysm:
MCMCChains.jl doesn't like vector-valued or matrix-valued variables. Consequently, in Turing we use `DynamicPPL.varname_and_value_leaves` to split things like `x` up into `x[1]`, `x[2]`, ...

This is in general a lossy operation since `x` cannot be reconstructed from its individual elements.

This conversion used to be done when constructing a `Transition`. However, there was no need to do it so early. This PR changes it such that it is only done within the `_params_to_array` method, which is called inside `bundle_samples`.


sunxd3 (approved):
No objection in principle. 

Failed tests look unrelated, but may still worth a look?

penelopeysm:
Yupppp, failures are due to Libtask not working on 1.12 yet. Thanks!

PR #2637, opened 2025-07-29
Improve error message for initialization failures with troubleshootin…

AoifeHughes:
…g link


src/mcmc/hmc.jl
```diff
@@ -168,7 +168,7 @@ function find_initial_params(
 
     # if we failed to find valid initial parameters, error
     return error(
-        "failed to find valid initial parameters in $(max_attempts) tries. This may indicate an error with the model or AD backend; please open an issue at https://github.com/TuringLang/Turing.jl/issues",
+        "failed to find valid initial parameters in $(max_attempts) tries. This may indicate an error with the model or AD backend. See https://turinglang.org/docs/usage/troubleshooting/#initial-parameters for common causes and solutions. If the issue persists, please open an issue at https://github.com/TuringLang/Turing.jl/issues",
```
mhauru: Maybe the sentence "This may indicate an error with the model or AD backend" can go? The message is quite long, and I think the link gives a more nuanced explanation of what might be wrong.

test/mcmc/hmc.jl
```diff
@@ -276,6 +276,29 @@ using Turing
...
+        catch e
+            err = e
+        end
+
+        @test err isa ErrorException
+        @test contains(
+            string(err),
+            "https://turinglang.org/docs/usage/troubleshooting/#initial-parameters",
+        )
+    end
```
mhauru: This works, but check out `Test.@test_throws`, which can do this for for you.

test/mcmc/hmc.jl
```diff
@@ -276,6 +276,29 @@ using Turing
             @test Turing.Inference.getstepsize(spl, hmc_state) isa Float64
         end
     end
+
+    @testset "improved error message for initialization failures" begin
+        # Model that fails to initialize due to VonMises AD issue
+        @model function failing_model()
+            μ ~ Uniform(-π, π)
+            κ ~ InverseGamma(2, 3)
+            return x ~ VonMises(μ, κ)
```
mhauru: This model might get fixed at some point, independently of us (the issue is in AD packages/Distributions.jl). Would something like any super simple model but with `@addlogprob!!(-Inf)` thrown in at the end trigger the same error more reliably? `@addlogprob!!` is a thing you can call within a model to add any number you want to the log likelihood, and I think `@addlogprob!!(-Inf)` should make all samples be rejected. `@addlogprob!!(NaN)` might work too.

penelopeysm: line 208 through 215 already contains the same test.

penelopeysm: with `@addlogprob! -Inf`

test/mcmc/hmc.jl
```diff
@@ -276,6 +276,29 @@ using Turing
...
+        err = try
+            sample(failing_model(), NUTS(), 10; progress=false)
+        catch e
+            e
+        end
+
+        @test contains(
+            string(err),
+            "https://turinglang.org/docs/usage/troubleshooting/#initial-parameters",
+        )
```
mhauru: This, too, can in fact be done with `@test_throws`, since its first argument can be a string that must be contained in the error message. E.g.
```julia
julia> f() = throw(ErrorException("this is an error message that contains the phrase 'this must be in the error'"))

julia> @test_throws "this must be in the error" f()
Test Passed
     Message: "this is an error message that contains the phrase 'this must be in the error'"
```
I'm not sure if you can check for both the substring _and_ the exception type in one call to `@test_throws`. There may be a way, or you may need to call `@test_throws` twice.

src/mcmc/hmc.jl
```diff
@@ -168,7 +168,7 @@ function find_initial_params(
 
     # if we failed to find valid initial parameters, error
     return error(
-        "failed to find valid initial parameters in $(max_attempts) tries. This may indicate an error with the model or AD backend; please open an issue at https://github.com/TuringLang/Turing.jl/issues",
+        "failed to find valid initial parameters in $(max_attempts) tries. See https://turinglang.org/docs/usage/troubleshooting/#initial-parameters for common causes and solutions. If the issue persists, please open an issue at https://github.com/TuringLang/Turing.jl/issues",
```
yebai: Can we set up a more robust URI for this link, e.g. via `https://turinglang.org/docs/uri/initial-parameters`? This URI redirects to the corresponding docs link, and would never change if we restructure the docs. 

Quarto supports this via `redirects`: https://quarto.org/docs/websites/website-navigation.html#redirects

Cc @shravanngoswamii

shravanngoswamii: Sure, let me add a redirects template in docs repo so we can easily create permanent redirects for docs.

shravanngoswamii: https://github.com/TuringLang/docs/pull/632

AoifeHughes: resolved in last commit

test/mcmc/hmc.jl
```diff
@@ -259,6 +259,20 @@ using Turing
...
+    @testset "improved error message for initialization failures" begin
+        # Model that always fails to initialize
+        @model function failing_model()
+            x ~ Normal()
+            @addlogprob! -Inf
+        end
+
+        # Test that error message includes troubleshooting link
+        @test_throws ErrorException sample(failing_model(), NUTS(), 10; progress=false)
+        @test_throws "https://turinglang.org/docs/usage/troubleshooting/#initial-parameters" sample(
```
mhauru: Sorry, I approved too soon. The URL here needs to match.
```suggestion
        @test_throws "https://turinglang.org/docs/uri/initial-parameters" sample(
```

PR #2616, opened 2025-07-14
Update logp in varinfo when external samplers are used

penelopeysm:
Closes #2583, where the logp from `externalsampler` was not being correctly set in `varinfo` and hence the chain.

Plots shown below:  
Before = logp was never updated, so logp is just a flat line.  
After = logp changes with the parameters.

These were run using AdvancedMH. I've tested with AdvancedHMC and SliceSampling (cf. https://github.com/TuringLang/SliceSampling.jl/issues/15) and similar improvements are observed.

```julia
using Turing, Plots, AdvancedMH

@model function test_model(y)
    a ~ Normal(0, 1)
    y ~ Normal(a, 1)
end

y = rand(Normal(0.5, 1), 100)
model = test_model(y)

chain_1 = sample(model, externalsampler(AdvancedMH.RWMH(1)), 100);

plot(chain_1, [:a, :lp], seriestype = :traceplot)
```

## Before

Screenshot 2025-07-14 at 15 39 09

## After

Screenshot 2025-07-14 at 15 39 48

PR #2640, opened 2025-08-01
Fix externalsampler interface

penelopeysm:
v0.39.5 #2616 inadvertently changed the required interface for external samplers to include 

```julia
AbstractMCMC.getparams(::DynamicPPL.Model, ::SamplerState)
```

rather than 

```julia
Turing.Inference.getparams(::DynamicPPL.Model, ::SamplerTransition)
```

which is what came before it. Note that `Turing.Inference.getparams` serves a different purpose compared to `AbstractMCMC.getparams` and the external sampler interface **should** really use `AbstractMCMC.getparams`.

Furthermore, `Turing.Inference.getparams` is an internal function so in principle changing it should not require a breaking release --- but unfortunately I didn't catch that this internal function was part of the external sampler interface. So it was a mistake to make this change in a patch release. (Sorry.) This PR therefore reverts the change.

On top of that, the existing externalsampler tests did not catch this because they only tested on AdvancedMH and AdvancedHMC, for which _both_ methods are overloaded.

So this PR also adds a test with a external sampler that satisfies the minimal interface as specified in the `ExternalSampler` docstring. This ensures that if we change the interface this test will break. See e.g. 'The Test Fake' in https://invenia.github.io/blog/2020/11/06/interfacetesting/.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,22 @@
+# 0.39.9
+
+Revert a bug introduced in 0.39.5 in the external sampler interface.
+For Turing 0.39, external samplers should define
+
+```
+Turing.Inference.getparams(::DynamicPPL.Model, ::MySamplerTransition)`
+```
+
+rather than
+
+```
+AbstractMCMC.getparams(::DynamicPPL.Model, ::MySamplerState)
+```
+
+to obtain a vector of parameters from the model.
+
+Note that this may change in future breaking releases.
+
 # 0.39.8
 
 MCMCChains.jl doesn't understand vector- or matrix-valued variables, and in Turing we split up such values into their individual components.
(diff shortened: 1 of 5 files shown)


penelopeysm:
Yeahh, just noticed that too. It's been happening a few times and we're not sure what triggered it although rerunning CI tends to make it pass, so it's not fully deterministic.

penelopeysm:
It might be this? https://github.com/JuliaLang/julia/pull/54840

penelopeysm:
I think that's the offending bug. The stack trace shows `speccache_eq`, and the segfault only happens with 2 threads, and it only happens sometimes, which all lines up with the fact that the issue is about data races. I checked the Julia source code and this bugfix doesn't seem to have been backported to 1.10.

In fact it wasn't backported to 1.11 either but I suppose it's possible that other changes were made to 1.11 that made the race conditions less likely to happen.

FYI @mhauru @AoifeHughes

PR #2644, opened 2025-08-04
Do not re-evaluate model for Prior

### LINKED ISSUE #2641: Skip extra re-evaluation with `Prior`
On each MCMC iteration, Prior executes the model twice: once to get the varinfo, and once when later constructing the Transition from the varinfo.

https://github.com/TuringLang/Turing.jl/blob/937ea9c198187ff4261e4387441b21019a94bec1/src/mcmc/prior.jl#L8-L21

The second re-evaluation is not necessary because the first one should be able to obtain all the necessary info already as long as it has all the right accumulators.

This should definitely be fixed. I'm not entirely sure how: maybe `Transition` needs an opt-in flag to indicate "do not re-evaluate" (feels safer to me), or maybe it should not re-evaluate if the varinfo already has the right accumulators (feels a bit dangerous because maybe the accumulators have the wrong value).

penelopeysm:
Fixes #2641.

### DIFF
--- a/src/mcmc/prior.jl
+++ b/src/mcmc/prior.jl
@@ -16,8 +16,17 @@ function AbstractMCMC.step(
     sampling_model = DynamicPPL.contextualize(
         model, DynamicPPL.SamplingContext(rng, DynamicPPL.SampleFromPrior(), model.context)
     )
-    _, vi = DynamicPPL.evaluate!!(sampling_model, VarInfo())
-    return Transition(model, vi, nothing), nothing
+    vi = VarInfo()
+    vi = DynamicPPL.setaccs!!(
+        vi,
+        (
+            DynamicPPL.ValuesAsInModelAccumulator(true),
+            DynamicPPL.LogPriorAccumulator(),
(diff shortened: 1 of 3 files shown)


src/mcmc/prior.jl
```diff
@@ -16,8 +16,15 @@ function AbstractMCMC.step(
...
+            DynamicPPL.ValuesAsInModelAccumulator(true),
+            DynamicPPL.LogPriorAccumulator(),
+            DynamicPPL.LogLikelihoodAccumulator(),
+        ),
+    )
+    _, vi = DynamicPPL.evaluate!!(sampling_model, vi)
+    return Transition(model, vi, nothing; reevaluate=false), nothing
 end
-
-DynamicPPL.default_chain_type(sampler::Prior) = MCMCChains.Chains
```
penelopeysm, line 23: these lines are obsolete, it's actually the following method that gets called: https://github.com/TuringLang/Turing.jl/blob/6d548642f0367eddbe50a574d8902c308454b90c/src/mcmc/algorithm.jl#L13-L14

Issue #2547, opened 2025-05-08
Gibbs with closed-form conditions


Red-Portal:
The new version for the Gibbs sampler is missing a feature that was present in the previous [version](https://github.com/TuringLang/Turing.jl/blob/22028aa3840d274e257c124cac31a6f459d00f96/src/mcmc/gibbs_conditional.jl#L49). It would be nice to implement an equivalent (or improved!) feature.

## Interface

    GibbsConditional(sym, conditional)

A "pseudo-sampler" to manually provide analytical Gibbs conditionals to `Gibbs`.
`GibbsConditional(:x, cond)` will sample the variable `x` according to the conditional `cond`, which
must therefore be a function from a `NamedTuple` of the conditioned variables to a `Distribution`.

## Examples

```julia
α_0 = 2.0
θ_0 = inv(3.0)
x = [1.5, 2.0]
N = length(x)

@model function inverse_gdemo(x)
    λ ~ Gamma(α_0, θ_0)
    σ = sqrt(1 / λ)
    m ~ Normal(0, σ)
    @. x ~ \$(Normal(m, σ))
end

# The conditionals can be formulated in terms of the following statistics:
x_bar = mean(x) # sample mean
s2 = var(x; mean=x_bar, corrected=false) # sample variance
m_n = N * x_bar / (N + 1)

function cond_m(c)
    λ_n = c.λ * (N + 1)
    σ_n = sqrt(1 / λ_n)
    return Normal(m_n, σ_n)
end

function cond_λ(c)
    α_n = α_0 + (N - 1) / 2 + 1
    β_n = s2 * N / 2 + c.m^2 / 2 + inv(θ_0)
    return Gamma(α_n, inv(β_n))
end

m = inverse_gdemo(x)

sample(m, Gibbs(GibbsConditional(:λ, cond_λ), GibbsConditional(:m, cond_m)), 10)
```

mhauru:
I frankly don't know why this got removed at the Gibbs makeover, might have been me being careless. Should be easy to add back.

penelopeysm:
This is another place where NamedTuple won't quite cut it, unless we restrict ourselves to identity-optic-VarNames...

https://github.com/TuringLang/DynamicPPL.jl/issues/900

yebai:
@mhauru / @AoifeHughes, can you help take a look after you finish ongoing PRs?

AoifeHughes:
Yep, I'm happy to add this to my list

[...]

PR #2647, opened 2025-08-07
Implement GibbsConditional

AoifeHughes:
WRT: https://github.com/TuringLang/Turing.jl/issues/2547

### DIFF
--- a/src/mcmc/gibbs_conditional.jl
+++ b/src/mcmc/gibbs_conditional.jl
@@ -46,8 +46,12 @@ end
 function cond_precision(c)
     a = 2.0
     b = 3.0
-    m = c[@varname(m)]
-    x = c[@varname(x)]
+    # We use AbstractPPL.getvalue instead of indexing into `c` directly to guard against
+    # issues where e.g. you try to get `c[@varname(x[1])]` but only `@varname(x)` is present
+    # in `c`. `getvalue` handles that gracefully, `getindex` doesn't. In this case
+    # `getindex` would suffice, but `getvalue` is good practice.
+    m = AbstractPPL.getvalue(c, @varname(m))
+    x = AbstractPPL.getvalue(c, @varname(x))
     n = length(x)
     a_new = a + (n + 1) / 2
     b_new = b + sum(abs2, x .- m) / 2 + m^2 / 2
@@ -55,8 +59,8 @@ function cond_precision(c)
 end
 
 function cond_m(c)
-    precision = c[@varname(precision)]
-    x = c[@varname(x)]
+    precision = AbstractPPL.getvalue(c, @varname(precision))
+    x = AbstractPPL.getvalue(c, @varname(x))
     n = length(x)
     m_mean = sum(x) / (n + 1)
     m_var = 1 / (precision * (n + 1))


src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+
+# Define analytical conditionals
+function cond_λ(c::NamedTuple)
+    a = 2.0
+    b = 3.0
+    m = c.m
+    x = c.x
+    n = length(x)
+    a_new = a + (n + 1) / 2
+    b_new = b + sum((x[i] - m)^2 for i in 1:n) / 2 + m^2 / 2
```
mhauru: Likewise comparing to https://github.com/TuringLang/Turing.jl/blob/v0.35.5/src/mcmc/gibbs_conditional.jl, should the `m` in the variance term rather be the mean of `x`?

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+    model::DynamicPPL.Model,
+    sampler::DynamicPPL.Sampler{<:GibbsConditional{S}},
+    state::DynamicPPL.AbstractVarInfo;
+    kwargs...,
+) where {S}
+    alg = sampler.alg
+
+    # For GibbsConditional within Gibbs, we need to get all variable values
+    # Check if we're in a Gibbs context
+    global_vi = if hasproperty(model, :context) && model.context isa GibbsContext
```
mhauru: The core idea here of finding the possible `GibbsContext` and getting the global varinfo from it is good. However, GibbsContext is a bit weird, in that it's always inserted at the bottom of the context stack. By the context stack I mean the fact that contexts often have child contexts, and thus `model.context` may in fact be many nested contexts. See e.g. how the GibbsContext is set here, by calling `setleafcontext` rather than `setcontext`: https://github.com/TuringLang/Turing.jl/blob/d75e6f269b18800e366b383b5059ed08e10ad7a7/src/mcmc/gibbs.jl#L258

So rather than check whether `model.context isa GibbsContext`, I think you'll need to traverse the whole context stack, and check if any of them are a `GibbsContext`, until you hit a leaf context and the stack ends.

Moreover, I think you'll need to check not just for GibbsContext, but also for `ConditionContext` and `FixedContext`, which condition/fix the values of some variables. So all in all, if you go through the whole stack, starting with `model.context` and going through its child contexts, and collect any variables set in `ConditionContext`, `FixedContext`, and `GibbsContext`, that should give you all of the variable values you need. See here for more details on condition and fix: https://github.com/TuringLang/DynamicPPL.jl/blob/1ed8cc8d9f013f46806c88a83e93f7a4c5b891dd/src/contexts.jl#L258

As mentioned on Slack a week or two ago, all this context stack business is likely changing Soon (TM), since @penelopeysm is overhauling `condition` and `fix` over here, https://github.com/TuringLang/DynamicPPL.jl/issues/1010, and as a result we may be able to overhaul `GibbsContext` as well. You could wait for that to be finished first, at least if it looks like getting this to work would be a lot of work.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+    condvals = merge(condvals_vars, condvals_obs)
+
+    # Get the conditional distribution
+    conddist = alg.conditional(condvals)
+
+    # Sample from the conditional distribution
+    updated = rand(rng, conddist)
+
+    # Update the variable in state
+    # We need to get the actual VarName for this variable
```
mhauru: The operating principle of the new(ish) Gibbs sampler is that every component sampler only ever sees a VarInfo with the variables that that component sampler is supposed to sample. Thus, you should be able to assume that `updated` includes values for all the variables in `state`, and for nothing else. Hence the below checks and loops I think shouldn't be necessary. The solution be might be as simple as `new_state = unflatten(state, updated)`, though there may be details there that I'm not thinking of right now. (What if `state` is linked? But maybe we can guarantee that it's never linked, because the sampler can control it.) Happy to discuss details more if `unflatten` by itself doesn't seem to cut it.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+
+# Sample using GibbsConditional
+model = inverse_gdemo([1.0, 2.0, 3.0])
+chain = sample(model, Gibbs(
+    :λ => GibbsConditional(:λ, cond_λ),
+    :m => GibbsConditional(:m, cond_m)
+), 1000)
+```
+"""
+struct GibbsConditional{S,C} <: InferenceAlgorithm
```
mhauru: I think the type parameter `S` shouldn't be necessary once we don't use it any more to construct the VarName that is being sampled. See below comments for more details.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+            end
+        end
+        if !found
+            error("Could not find variable $S in VarInfo")
+        end
+        updated_vi
+    end
+
+    # Update log joint probability
+    new_vi = last(DynamicPPL.evaluate!!(model, new_vi, DynamicPPL.DefaultContext()))
```
mhauru: I think you shouldn't need this, because the log joint is going to be recomputed anyway by the Gibbs sampler once it's looped over all component samplers. Saves one model evaluation.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+end
+
+"""
+    gibbs_step_recursive(
+        rng, model, sampler::GibbsConditional, target_varnames, global_vi, state
+    )
+
+Perform a single step of GibbsConditional sampling.
+"""
+function gibbs_step_recursive(
```
mhauru: I would hope you wouldn't need to overload this or `gibbs_initialstep_recursive`. Also, the below implementation seems to be just a repeat of `step`.

test_gibbs_conditional.jl
```diff
@@ -0,0 +1,78 @@
...
+
+# Generate some observed data
+Random.seed!(42)
+x_obs = [1.0, 2.0, 3.0, 2.5, 1.5]
+
+# Create the model
+model = inverse_gdemo(x_obs)
+
+# Sample using GibbsConditional
+println("Testing GibbsConditional sampler...")
```
mhauru: You probably want various `@test` calls rather than a lot of prints.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,245 @@
...
+
+The `conditional` function should take a `NamedTuple` of conditioned variables and return
+a `Distribution` from which to sample the variable `sym`.
+
+# Examples
+
+```julia
+# Define a model
+@model function inverse_gdemo(x)
+    λ ~ Gamma(2, 3)
```
mhauru: Comparing to https://github.com/TuringLang/Turing.jl/blob/v0.35.5/src/mcmc/gibbs_conditional.jl, should the distribution be `Gamma(2, inv(3))`?

mhauru (changes requested):
As discussed on Slack, rather than doing a full review, I'm just going to give some high level comments and pointers for where to find more details on some of the relevant context.

src/mcmc/gibbs_conditional.jl
```diff
@@ -54,17 +54,48 @@ chain = sample(model, Gibbs(
...
-        return new{sym,C}(conditional)
+        return new{C}(conditional)
     end
 end
 
 # Mark GibbsConditional as a valid Gibbs component
 isgibbscomponent(::GibbsConditional) = true
 
+# Required methods for Gibbs constructor
+Base.length(::GibbsConditional) = 1  # Each GibbsConditional handles one variable
```
mhauru: Is this being called somewhere? Might be, but I don't remember having a need for `length` of samplers.

src/mcmc/gibbs_conditional.jl
```diff
@@ -54,17 +54,48 @@ chain = sample(model, Gibbs(
...
+function find_global_varinfo(context, fallback_vi)
+    # Start with the given context and traverse down
+    current_context = context
+
+    while current_context !== nothing
+        if current_context isa GibbsContext
+            # Found GibbsContext, return its global varinfo
+            return get_global_varinfo(current_context)
+        elseif hasproperty(current_context, :childcontext) &&
+            isdefined(DynamicPPL, :childcontext)
```
mhauru: This check can be done more robustly with the `DynamicPPL.NodeTrait` function, that is meant to check exactly this. If you write this as a recursion rather than as a loop, you can use multiple dispatch on the `IsLeaf` and `IsParent` types to implement this check. This is best explained by an example, see for instance the implementation of `hassampler` here: https://github.com/TuringLang/DynamicPPL.jl/blob/0cf3440549c7634c9b810e1f804eb2cd195f7e47/src/contexts.jl#L174

To explain a bit: `DynamicPPL.NodeTrait(ctx)` should return either `IsLeaf` or `IsParent` for all contexts `ctx`. Crucially those are _types_, not values, which means you can dispatch methods on them. The contract is that if `DynamicPPL.NodeTrait(ctx)` returns `IsParent` then there should be an implementation of `childcontext(ctx)` which returns the child context.

This design pattern is often called Holy Traits (after Tim Holy who presumably came up with them). It's kinda funky but cool, and worth learning, since it comes up with some frequency in Julia code. The benefit of Holy Traits is that the type of information of e.g. `IsLeaf` vs `IsParent` is known at compile time, and thus results in more efficient code.

src/mcmc/gibbs_conditional.jl
```diff
@@ -97,12 +128,10 @@ function AbstractMCMC.step(
     alg = sampler.alg
 
     # For GibbsConditional within Gibbs, we need to get all variable values
-    # Check if we're in a Gibbs context
-    global_vi = if hasproperty(model, :context) && model.context isa GibbsContext
-        # We're in a Gibbs context, get the global varinfo
-        get_global_varinfo(model.context)
+    # Traverse the context stack to find all conditioned/fixed/Gibbs variables
+    global_vi = if hasproperty(model, :context)
```
mhauru: Since `model` is a `DynamicPPL.Model`, it has to always have the field `context` (that's part of the definition of `DynamicPPL.Model`), and thus this check is unnecessary.

Red-Portal:
Hi, I think this PR needs to be worked on as quickly as possible given that lots of people (#2547 and in slack) have been requesting this feature to be added back. I could take this over if nobody has bandwidth to work on this. However, I am a bit less familiar with the relevant parts in Turing. @mhauru @penelopeysm could you take a look if you have any suggestions?

mhauru:
Aoife is no longer working on Turing, and I said I would take over this one, but just haven't found time. Let me try to see what the state of this is and report back.

mhauru:
I fixed and improved the implementation and added a lot of tests. I think this may be done now, but I'll reread it myself tomorrow before requesting reviews.

mhauru:
Julia package registry seems to be having trouble, will rerun CI in a moment. This is ready for review though. @Red-Portal, would you be happy to take an overall look? @penelopeysm could I ask you to take a look at the implementation and HISTORY.md entry at the least? Up to you if you want to also read the tests.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+
+```julia
+# Define a model
+@model function inverse_gdemo(x)
+    precision ~ Gamma(2, inv(3))
+    std = sqrt(1 / precision)
+    m ~ Normal(0, std)
+    for i in 1:length(x)
+        x[i] ~ Normal(m, std)
+    end
```
penelopeysm:
```suggestion
    for i in eachindex(x)
        x[i] ~ Normal(m, std)
    end
```

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+end
+
+# Define analytical conditionals
+function cond_precision(c)
+    a = 2.0
+    b = inv(3)
+    m = c[@varname(m)]
+    x = c[@varname(x)]
```
penelopeysm: I feel that the `c[@varname(x)]` here is deceptively simple... it will work under the following scenarios

1. `x` is passed as a single vector to the model arguments
2. `x` is not in the arguments, but is conditioned on as a vector i.e. `model() | (; x = vec)`

it will fail in the case where

3. `x` is not in the arguments, but is conditioned on as individual elements i.e. `model() | Dict(@varname(x[1]) => 1.0, @varname(x[2]) => 2.0)`

Notice that because the model has `x[i] ~ Normal`, all three cases will work correctly with plain model evaluation. But one of the cases will fail with GibbsConditional.

I don't really know how to fix this, and I don't know whether it should even be fixed, but it makes me quite uncomfortable.

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+    n = length(x)
+    a_new = a + (n + 1) / 2
+    b_new = b + sum((x[i] - m)^2 for i in 1:n) / 2 + m^2 / 2
```
penelopeysm:
```suggestion
    b_new = b + sum(abs2, x .- m) / 2 + m^2 / 2
```

penelopeysm: Where are these analytical conditionals from? Is there a known formula somewhere on Wikipedia or something?

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+
+Traverse the context stack of `model` and build a `Dict` of all the variable values that are
+set in GibbsContext, ConditionContext, or FixedContext.
+"""
+function build_variable_dict(model::DynamicPPL.Model)
+    context = model.context
+    cond_nt = DynamicPPL.conditioned(context)
+    fixed_nt = DynamicPPL.fixed(context)
+    # TODO(mhauru) Can we avoid invlinking all the time? Note that this causes a model
+    # evaluation, which may be expensive.
```
penelopeysm: I think for typed VarInfo it shouldn't need to evaluate the model. Obviously it still has a cost, just not as much as model evaluation.

```julia
julia> @model function f()
           @info "hi"
           x ~ Normal()
       end

julia> model = f(); v = VarInfo(model);
[ Info: hi

julia> v2 = DynamicPPL.link!!(v, model); v3 = DynamicPPL.invlink!!(v, model);
```

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+    global_vi = DynamicPPL.invlink(get_gibbs_global_varinfo(context), model)
+    return merge(
+        DynamicPPL.values_as(global_vi, Dict),
+        Dict(
+            (DynamicPPL.VarName{sym}() => val for (sym, val) in pairs(cond_nt))...,
+            (DynamicPPL.VarName{sym}() => val for (sym, val) in pairs(fixed_nt))...,
+            (DynamicPPL.VarName{sym}() => val for (sym, val) in pairs(model.args))...,
```
penelopeysm: `cond_nt` and `fixed_nt` might not be NamedTuples, they might be dicts, in which case this will fail in a very weird manner.

```julia
julia> VarName{@varname(x)}()
x

julia> VarName{@varname(x)}() == @varname(x) # not the same thing
false
```

I think you have to convert them to dicts first. `DynamicPPL.to_varname_dict` will do this (albeit inefficiently https://github.com/TuringLang/DynamicPPL.jl/issues/1134).

test/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,277 @@
...
+            :precision => GibbsConditional(cond_precision), :m => GibbsConditional(cond_m)
+        )
+        chain = sample(rng, model, sampler, 1_000)
+        @test size(chain, 1) == 1_000
+        @test mean(chain, :precision) ≈ mean(reference_chain, :precision) atol = 0.1
+        @test mean(chain, :m) ≈ mean(reference_chain, :m) atol = 0.1
+
+        # Mix GibbsConditional with an MCMC sampler
+        sampler = Gibbs(:precision => GibbsConditional(cond_precision), :m => MH())
+        chain = sample(rng, model, sampler, 1_000)
```
penelopeysm: I think there should be separate instances of `StableRNG()` per sample call so that reordering the tests or inserting new ones doesn't affect reproducibility of other tests.

test/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,277 @@
...
+        function cond_mean(c)
+            var = c[@varname(var)]
+            x = c[@varname(x)]
+            n = length(x)
+            # Prior: mean ~ Normal(0, 10)
+            # Likelihood: x[i] ~ Normal(mean, σ)
+            # Posterior: mean ~ Normal(μ_post, σ_post)
+            prior_var = 100.0  # 10^2
+            post_var = 1 / (1 / prior_var + n / var)
+            post_mean = post_var * (0 / prior_var + sum(x) / var)
```
penelopeysm, line 105:
```suggestion
            prior_var = 100.0  # 10^2
            post_var = 1 / (1 / prior_var + n / var)
            post_mean = post_var * (sum(x) / var)
```

again - are there formulas for this somewhere? Maybe a statistician looks at this and goes 'ah yes of course', but I feel quite uncomfortable.

test/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,277 @@
...
+        end
+
+        function cond_mean(var, x, prior_std)
+            n = length(x)
+            # Prior: mean ~ Normal(0, prior_std)
+            # Likelihood: x[i] ~ Normal(mean, σ)
+            # Posterior: mean ~ Normal(μ_post, σ_post)
+            prior_var = prior_std^2
+            post_var = 1 / (1 / prior_var + n / var)
+            post_mean = post_var * (0 / prior_var + sum(x) / var)
```
penelopeysm, line 142:
```suggestion
            post_mean = post_var * (sum(x) / var)
```

src/mcmc/gibbs_conditional.jl
penelopeysm: I'm guessing it's somewhere in here https://en.wikipedia.org/wiki/Normal_distribution#Bayesian_analysis_of_the_normal_distribution

src/mcmc/gibbs_conditional.jl
```diff
@@ -0,0 +1,166 @@
...
+    m ~ Normal(0, std)
+    for i in 1:length(x)
+        x[i] ~ Normal(m, std)
+    end
+end
+
+# Define analytical conditionals
+function cond_precision(c)
+    a = 2.0
+    b = inv(3)
```
penelopeysm:
```suggestion
    a = 2.0
    b = 3.0
```

mhauru: Thanks for spotting this, I forgot to check whether the example actually worked.

src/mcmc/gibbs_conditional.jl
mhauru: I'll add a link to the wikipedia page I was using.

test/mcmc/gibbs_conditional.jl
mhauru: Will add the link. I didn't write this part, but I think it's deliberately in this format to match the formula I found e.g. on wikipedia. The 0 I think is the mean of the prior.

penelopeysm: Ah okay in that case it can stay!

test/mcmc/gibbs_conditional.jl
mhauru: Fixed.

src/mcmc/gibbs_conditional.jl
mhauru: Nice catch. I fixed this and added a test that would have caught this.

Making so many Dicts is sad, but for now I just want to get this working, and worry about performance later.

src/mcmc/gibbs_conditional.jl
mhauru: Adjusted the comment to reflect this.

src/mcmc/gibbs_conditional.jl
mhauru: I've fixed all the other points you raised. This will have to wait untill tomorrow.

penelopeysm: As a simple enough solution, could we maybe take the model arguments out of the dictionary, and instead give `c` a field called `c.model_args = deepcopy(model.args)`? Then people can use `c.model_args.x`. Deepcopy would be needed to avoid accidental aliasing. (Or we could not copy, and leave it to the user to copy if they need it.)

penelopeysm: The outcome of this is:

> 1. `x` is passed as a single vector to the model arguments

OK, people can use `c.model_args.x`, clear.

> 2. `x` is not in the arguments, but is conditioned on as a vector i.e. `model() | (; x = vec)`

OK, people can use `c[@varname(x)]` because that's what they conditioned on.

> 3. `x` is not in the arguments, but is conditioned on as individual elements i.e. `model() | Dict(@varname(x[1]) => 1.0, @varname(x[2]) => 2.0)`

This will still fail but at least there's a good explanation for it, it's because they conditioned on `x[1]` and `x[2]` rather than `x`.

penelopeysm: Looking at my previous comment, actually the original implementation is also explainable along the same lines. I guess the tldr is basically, if you supply `x` as a single thing, that's fine. So actually I'd be OK with leaving it as is.

mhauru: The root cause of this is essentially the same as why you can't do
```julia
@model function f()
    x ~ MvNormal()
end

m = condition(f(), Dict(@varname(x[1]) => 1.0))
```
it's just that in that case you request the value of the conditioned `x` in the model, whereas here you request it in the `conditionals` function. So in some sense this is a bigger problem that needs to be fixed lower down (https://github.com/TuringLang/DynamicPPL.jl/issues/11480). So I'd be okay with leaving this be too.

The one thing that I was wondering about is that `ConditionContext` internally uses `getvalue`. If the user did the same in the `conditionals` function then at least they could get `x1 = getvalue(c, @varname(x[1]))` to work even if they conditioned on `x` as a whole. Thus I'm now thinking that maybe the best thing to do here would be to set a good example in the docstring and use `getvalue` rather than `getindex`, and leave it at that.

As a silver lining, if this goes wrong, at least the user gets a clear "key not found" error, and the error comes from code they wrote, in quite a traceable way.

penelopeysm: Yeah, that sounds good to me! Agree that getvalue is better, albeit a tiny bit more annoying since it has to be imported from AbstractPPL

mhauru (approved):

PR #2649, opened 2025-08-07
Use typed varinfo in Prior

penelopeysm:
```julia
using Turing 

@model function f(x)
    θ ~ Normal()
    for i in eachindex(x)
        x[i] ~ Normal(θ)
    end
end

model = f(fill(missing, 100))

@time sample(model, Prior(), 1000; progress=false);
```

Base branch:

```julia
julia> @time sample(model, Prior(), 1000; progress=false);
  4.703977 seconds (19.72 M allocations: 718.042 MiB, 1.54% gc time)
```

This PR:

```julia
julia> @time sample(model, Prior(), 1000; progress=false);
  0.411107 seconds (5.65 M allocations: 242.001 MiB, 7.42% gc time)
```

### DIFF
--- a/src/mcmc/prior.jl
+++ b/src/mcmc/prior.jl
@@ -26,5 +26,36 @@ function AbstractMCMC.step(
         ),
     )
     _, vi = DynamicPPL.evaluate!!(sampling_model, vi)
-    return Transition(model, vi, nothing; reevaluate=false), nothing
+    vi = DynamicPPL.typed_varinfo(vi)
+    return Transition(model, vi, nothing; reevaluate=false), vi
+end
+
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG,
+    model::DynamicPPL.Model,
+    sampler::DynamicPPL.Sampler{<:Prior},
+    vi::AbstractVarInfo;
+    kwargs...,
+)
+    # TODO(DPPL0.38/penelopeysm): replace this entire thing with init!!
+    #
+    # `vi` is a VarInfo from the previous step so already has all the
+    # right accumulators and stuff. The only thing we need to change is to make
+    # sure that the old values are overwritten when we resample.
+    #
+    # Note also that the values in the Transition (and hence the chain) are not
+    # obtained from the VarInfo's metadata itself, but are instead obtained
+    # from the ValuesAsInModelAccumulator, which is cleared in the evaluate!!
+    # call. Thus, the actual values in the VarInfo's metadata don't matter:
+    # we only set the del flag here to make sure that new values are sampled
+    # (and thus new values enter VAIMAcc), rather than the old ones being
+    # reused during the evaluation. Yes, SampleFromPrior really sucks.
+    for vn in keys(vi)
+        DynamicPPL.set_flag!(vi, vn, "del")
+    end
+    sampling_model = DynamicPPL.contextualize(
+        model, DynamicPPL.SamplingContext(rng, DynamicPPL.SampleFromPrior(), model.context)
+    )
+    _, vi = DynamicPPL.evaluate!!(sampling_model, vi)
+    return Transition(model, vi, nothing; reevaluate=false), vi
 end


penelopeysm:
Actually, both are broken, for different reasons. The first one seems to never change, the second just doesn't run. I'll have to investigate this further, I'm a bit confused as to why either of them are broken.

penelopeysm:
This works:

```julia
@model function f()
    x ~ Normal()
    if x > 0
        y ~ Normal()
    else
        z ~ Normal()
    end
end
```

It's because `push!!(vi, ...)` adds a new key to the NamedTuple-of-Metadata if it doesn't exist yet. https://github.com/TuringLang/DynamicPPL.jl/blob/f0ac1099c87326df95eada0aeb099029dbd09aee/src/varinfo.jl#L1723-L1735

MH and NUTS don't work on this model, only Prior does.

-----

This doesn't work:

```julia
@model function f()
    x ~ Normal()
    y = (; a = 1.0, b = 1.0) # need this line so that it can assign to y.a or y.b
    if x > 0
        y.a ~ Normal()
    else
        y.b ~ Normal()
    end
end
```

The key `y` already exists in the NamedTuple, so it descends into `push!(metadata, ...)` which errors because `metadata.idcs` is a Dict which only has `y.a` and you can't push `y.b` into that dict unless you broaden the key type.

https://github.com/TuringLang/DynamicPPL.jl/blob/f0ac1099c87326df95eada0aeb099029dbd09aee/src/varinfo.jl#L1757-L1769

MH and NUTS also don't work on this model.

mhauru:
I confirmed that both example models work on current release, when sampling with `Prior`.

I don't really mind losing
```julia
@model function f()
    x ~ Normal()
    if x > 0
        y ~ Normal()
    else
        z ~ Normal()
    end
end
```
I think that's syntax that we mostly don't support anyway.

The one mostly I care about is actually
```julia
@model function f()
    x ~ Normal()
    y = Vector{Float64}(undef, x > 0 ? 1 : 2)
    if x > 0
        y[1] ~ Normal()
    else
        y[1] ~ Normal()
        y[2] ~ Normal()
    end
end
```
That might be fine on this PR as-is?

I'm not sure if I care about
```julia
@model function f()
    x ~ Normal()
    y = (; a = 1.0, b = 1.0) # need this line so that it can assign to y.a or y.b
    if x > 0
        y.a ~ Normal()
    else
        y.b ~ Normal()
    end
end
```
I just didn't realise that it would behave differently from the Vector version when I proposed it as a test case.

penelopeysm:
> model with x, y[1], y[2] ... That might be fine on this PR as-is?

Yup, that still works on this PR. The x/y/z model also works on this PR, but x/y.a/y.b doesn't.

I think in terms of the code, this PR is mostly good to go. (I believe CI failures will be fixed by #2650.) But I'm a bit hesitant to merge into the upcoming minor version, because it does technically remove support for a few models.

In general, for arbitrary dynamic models like these, I think the solution to make sampling work is to use untyped VarInfo. That's what Prior used to do, which is also the only reason why it 'supported' all these things.

The long term solution would be to let people sample with whatever kind of varinfo they want. That was (is?) the purpose of the sample-with-LDF PRs I still have sitting around. But right now I'm kind of leaning towards not merging this until those are in, just in the interests of not gratuitously breaking things.

The flip side of course is that this PR does improve performance by _a lot_ and it will probably be months until sample-with-LDF is merged.

mhauru:
Without thinking this through carefully, I wonder if the performance gains are more important that losing support for some pretty niche models.

Actually, now that I said that, maybe those niche models include all models with dynamic use of submodels? Still kinda niche though.

Regardless, happy to leave it out of this upcoming release with DPPL 0.37.

mhauru:
Is it time to revisit this with InitContext?

penelopeysm:
> Is it time to revisit this with InitContext?

I think InitContext makes things prettier but it doesn't solve the issue of same-symbol-but-different-type-VarName (`y.a` and `y.b`) model, and in particular it doesn't fix it when they crop up in different iterations, so

```julia
@model function f()
    if randn() > 0
        y.a ~ Normal()
    else
        y.b ~ Normal()
    end
end
```

There are a couple of ways of solving this properly:

1. Declare that we don't support such models (the assumption being that such models can be 'rescued' by renaming variables).
2. Fix push!! for typed VarInfo so that you can push `y.b` to a typed VarInfo that already has `y.a`.

I don't have a strong opinion on (1); _but_ it does seem to me that independently of (1) we _should_ still fix (2). To this end, I'll open an issue on DynamicPPL and also cheekily stick this on the list of topics for the JuliaCon hackathon.

penelopeysm:
This will be completely irrelevant after https://github.com/TuringLang/DynamicPPL.jl/pull/1125, the solution would be to use neither typed nor untyped VarInfo but rather to skip metadata entirely, since for Prior sampling the only thing we care about are VAIMAcc and the logp accs.

penelopeysm:
Using that DynamicPPL PR, plus the following patch, gives similar speedups to that in the original post, plus a 20% reduction in allocations compared to using typed VarInfo.

And as a bonus, it also works on every single weird dynamic model in the comments above, because it doesn't need Metadata anymore.

It should be noted that of the remaining 0.5 second runtime, most of it is actually spent constructing the `MCMCChains.Chains`. If you don't do that it goes right down to 0.07 seconds. (FlexiChains is mildly faster, it runs in 0.3 seconds.)

```diff
diff --git a/src/mcmc/prior.jl b/src/mcmc/prior.jl
index c4ec6c6f..f0973dff 100644
--- a/src/mcmc/prior.jl
+++ b/src/mcmc/prior.jl
@@ -12,14 +12,12 @@ function AbstractMCMC.step(
     state=nothing;
     kwargs...,
 )
-    vi = DynamicPPL.setaccs!!(
-        DynamicPPL.VarInfo(),
-        (
-            DynamicPPL.ValuesAsInModelAccumulator(true),
-            DynamicPPL.LogPriorAccumulator(),
-            DynamicPPL.LogLikelihoodAccumulator(),
-        ),
-    )
+    accs = DynamicPPL.AccumulatorTuple((
+        DynamicPPL.ValuesAsInModelAccumulator(true),
+        DynamicPPL.LogPriorAccumulator(),
+        DynamicPPL.LogLikelihoodAccumulator(),
+    ))
+    vi = DynamicPPL.Experimental.OnlyAccsVarInfo(accs)
     _, vi = DynamicPPL.init!!(model, vi, DynamicPPL.InitFromPrior())
     return Transition(model, vi, nothing; reevaluate=false), nothing
 end
```

PR #2650, opened 2025-08-10
Remove calls to resetlogp!!

penelopeysm:
Previously, pMCMC used to hijack the varinfo's logp field for its own purposes. For example, contributions from Gibbs-conditioned-variables and addlogprob used to be added there, and then after each observation the logp would be reset to 0.

Since we now use `ProduceLogLikelihoodAccumulator` to control particle weights and resampling, we no longer need to call resetlogp on the varinfo explicitly (indeed the varinfo's logp could simply be used in a normal way, i.e., to accumulate the logp during model evaluation).


penelopeysm:
Since this PR was so boring, I wrote up a changelog as well to make it more exciting. Happy to edit as needed.

HISTORY.md
```diff
@@ -1,6 +1,56 @@
...
+For example, if the model is `@model f() = x ~ LogNormal()`, `chain[:lp]` would always contain the value of `logpdf(LogNormal(), x)` for each sampled value of `x`.
+Previously these values could be incorrect if linking had occurred: some samplers would return `logpdf(Normal(), log(x))` i.e. the log-density with respect to the transformed distribution.
+
+**Gibbs sampler**
+
+When using Turing's Gibbs sampler, e.g. `Gibbs(:x => MH(), :y => HMC(0.1, 20))`, the conditioned variables (for example `y` during the MH step, or `x` during the HMC step) are treated as true observations.
+Thus the log-density associated with them is added to the likelihood.
+Previously these would effectively be added to the prior (in the sense that if `LikelihoodContext` was used they would be ignored).
+This is unlikely to affect users but we mention it here to be explicit.
+
```
mhauru, line 35:
```suggestion
This change only affects the log probabilities as the Gibbs component samplers see them; the resulting chain will include the usual log prior, likelihood, and joint, as described above.

```

Issue #2631, opened 2025-07-26
Unify `src/mcmc/Inference.jl` methods


penelopeysm:
There are several kinds of 'transitions' that can come out of samplers:

- `Turing.Inference.Transition` is the most common one.
- There are some other special transitions, `SGLDTransition`, `SMCTransition`, and `PGTransition`. These, together with `Turing.Inference.Transition`, get lumped under `Turing.Inference.AbstractTransition`.
- Some samplers also spit out varinfos (*cough cough* SampleFromPrior *cough cough*, and I think Prior also does this).

We should get rid of all of these conceptually similar things and just use a single Transition object.

I think this will have to wait for further changes in DynamicPPL because SampleFromPrior is not yet removed, but once that is done we can definitely get rid of the VarInfo methods.

The rest should also not be overly difficult to do.

penelopeysm:
The joys of multiple dispatch... 🙄 

https://github.com/TuringLang/Turing.jl/blob/23b92ebaadb754f8ca6823bd5344d8d477813bd5/src/mcmc/Inference.jl#L139-L144

If you called `Transition(model, vi)` you would get the method on line 139 even though that's clearly not what it was intended to do. I suppose that by coincidence it all works out...

mhauru:
Would we rather solve this with https://github.com/TuringLang/AbstractMCMC.jl/issues/149?

penelopeysm:
Would be nice to, just don't know how close at hand that is.

penelopeysm:
I think this is pretty much done, there's only one `Turing.Inference.Transition` now and `SampleFrom...` have been removed in DPPL 0.38.

PR #2630, opened 2025-07-25
Use accumulators to fix all logp calculations when sampling

penelopeysm:
Following on from https://github.com/TuringLang/DynamicPPL.jl/pull/998, this PR:

- in the optimisation code, removes `LogPriorWithoutJacobianAccumulator` as DynamicPPL's `LogPriorAccumulator` now serves this purpose;

- replaces calls to `getlogjoint` with `getlogjoint_internal` for all samplers (as that's what they want);

- closes #2617 by re-calculating logp together with values_as_in_model when constructing a Transition. This also allows us to stick `logprior` and `loglikelihood` terms into the chain (`lp` is still there and refers to the joint).

- does a little tidy-up on MH: previously the sampler state was just a VarInfo, now we bundle the VarInfo together with its internal-logjoint so that the next iteration doesn't need to recalculate the internal-logjoint.

Between #2629 and this PR all the failing tests should be fixed. 🥳  The remaining issues are 1.10 (doesn't understand `[sources]` in Project.toml) and 1.12 (Libtask).

Closes #2617

PR #2651, opened 2025-08-11
Unify Turing `Transition`s, fix some tests

penelopeysm:
This PR removes `SGLDTransition`, `SMCTransition`, and `PGTransition` in favour of just using plain old `Transition`.

This is pretty much the last thing I want to add to 0.40. The reason why I'd like to squeeze it in now is because it will correctly fix all the `chain[:lp]`, `chain[:logprior]`, etc. for all of the samplers we have, which feels like a nice reward after all the upfront work on DynamicPPL accumulators 😄.

It also adds a deepcopy to the `Transition` constructor as otherwise it would mutate the varinfo passed in which messed things up for SGLD (I'm not sure why it didn't mess anything up for the other samplers). This deepcopy used to be there previously

https://github.com/TuringLang/Turing.jl/blob/d75e6f269b18800e366b383b5059ed08e10ad7a7/src/mcmc/Inference.jl#L192-L194

so I think maybe removing it in #2630 was too ambitious.

Fixes most of #2631, although properly fixing it will have to wait for the removal of SampleFromPrior.

### DIFF
--- a/src/mcmc/Inference.jl
+++ b/src/mcmc/Inference.jl
@@ -124,19 +124,16 @@ end
 # Default Transition #
 ######################
 getstats(::Any) = NamedTuple()
+getstats(nt::NamedTuple) = nt
 
-# TODO(penelopeysm): Remove this abstract type by converting SGLDTransition,
-# SMCTransition, and PGTransition to Turing.Inference.Transition instead.
-abstract type AbstractTransition end
-
-struct Transition{T,F<:AbstractFloat,N<:NamedTuple} <: AbstractTransition
+struct Transition{T,F<:AbstractFloat,N<:NamedTuple}
     θ::T
     logprior::F
     loglikelihood::F
     stat::N
 
     """
-        Transition(model::Model, vi::AbstractVarInfo, sampler_transition; reevaluate=true)
+        Transition(model::Model, vi::AbstractVarInfo, stats; reevaluate=true)
 
     Construct a new `Turing.Inference.Transition` object using the outputs of a
     sampler step.
@@ -146,8 +143,10 @@ struct Transition{T,F<:AbstractFloat,N<:NamedTuple} <: AbstractTransition
     have junk contents. The role of this method is to re-evaluate `model` and
     thus set the accumulators to the correct values.
 
-    `sampler_transition` is the transition object returned by the sampler
-    itself and is only used to extract statistics of interest.
+    `stats` is any object on which `Turing.Inference.getstats` can be called to
+    return a NamedTuple of statistics. This could be, for example, the transition
+    returned by an (unwrapped) external sampler. Or alternatively, it could
+    simply be a NamedTuple itself (for which `getstats` acts as the identity).
 
     By default, the model is re-evaluated in order to obtain values of:
       - the values of the parameters as per user parameterisation (`vals_as_in_model`)
@@ -167,8 +166,11 @@ struct Transition{T,F<:AbstractFloat,N<:NamedTuple} <: AbstractTransition
     must be set up to track `x := y` statements.
     """
     function Transition(
-        model::DynamicPPL.Model, vi::AbstractVarInfo, sampler_transition; reevaluate=true
+        model::DynamicPPL.Model, vi::AbstractVarInfo, stats; reevaluate=true
     )
+        # Avoid mutating vi as it may be used later e.g. when constructing
+        # sampler states.
+        vi = deepcopy(vi)
         if reevaluate
             vi = DynamicPPL.setaccs!!(
                 vi,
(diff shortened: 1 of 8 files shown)


test/mcmc/particle_mcmc.jl
```diff
@@ -114,6 +127,24 @@ end
         @test length(unique(c[:s])) == 1
     end
 
+    @testset "addlogprob leads to reweighting" begin
```
penelopeysm, line 130: this was sort of tested in PG-within-Gibbs, but we didn't have a PG-only test

test/mcmc/gibbs.jl
```diff
@@ -651,7 +651,7 @@ end
             chain = sample(
                 StableRNG(468),
                 model,
-                Gibbs(:b => PG(10), :x => ESS()),
+                Gibbs(:b => PG(20), :x => ESS()),
```
penelopeysm, line 654: This test was failing on 1.10 due to numerical inaccuracy. Kind of unsure why it was only failing on 1.10 but not 1.11 given that we were using StableRNGs. My first guess would be the rng splitting in AdvancedPS.

I just bumped the atol up anyway because this test is so wonky (really we're mostly checking that it samples at all, since the results are incorrect depending on interpretation of model).

src/mcmc/Inference.jl
```diff
@@ -167,8 +166,11 @@ struct Transition{T,F<:AbstractFloat,N<:NamedTuple} <: AbstractTransition
     must be set up to track `x := y` statements.
     """
     function Transition(
-        model::DynamicPPL.Model, vi::AbstractVarInfo, sampler_transition; reevaluate=true
+        model::DynamicPPL.Model, vi::AbstractVarInfo, stats; reevaluate=true
     )
+        # Avoid mutating vi as it may be used later e.g. when constructing
+        # sampler states.
+        vi = deepcopy(vi)
```
mhauru, line 173: I wonder if this could be a significant time cost. In which case we could make sure we have a proper `copy` method for varinfos in DPPL. Would probably be good to have that anyway.

test/mcmc/particle_mcmc.jl
```diff
@@ -114,6 +127,24 @@ end
...
+                @addlogprob! -2.0
+            else
+                # Need a balanced number of addlogprobs in all branches, or
+                # else PG will error
+                @addlogprob! 0.0
+            end
+        end
+        c = sample(addlogprob_demo(), PG(10), 100)
+        # Result should be biased towards x > 0.
+        @test mean(c[:x]) > 0.5
```
mhauru:
```suggestion
        @test mean(c[:x]) > 0.7
```
Could this get a bit more clearance from 0.5, with, if necessary, the `@addloprob!` value being increased? Otherwise there's 50% chance this will pass just by luck.

mhauru (changes requested):
Looks good, just one request for improving one of the tests.

src/mcmc/Inference.jl
penelopeysm: Does `copy` tend to be more performant than `deepcopy`?

test/mcmc/particle_mcmc.jl
penelopeysm: Technically, x is `~ Normal()` so the chance of it being larger than 0.5 is probably lower than 50%. That does assume that the chain has fully mixed though, which might not be valid with 100 iterations, and there's no harm to making the addlogprob value larger, so I'll do that.

penelopeysm:
Remaining CI failures are due to Libtask on 1.12 (3x pre), the Julia segfault bug https://github.com/JuliaLang/julia/pull/54840 (1x Inference / min), and https://github.com/SciML/RecursiveArrayTools.jl/issues/477 (2x everything else / min).

test/mcmc/particle_mcmc.jl
penelopeysm: Also, I don't know why I didn't put in StableRNGs here. Will do so too.

src/mcmc/Inference.jl
mhauru: Yeah, it can be. Depends on your data structures. `deepcopy` can be slow because it plays it safe with aliasing and such, whereas `copy` does whatever you make it do.

test/mcmc/particle_mcmc.jl
mhauru: Sorry, I wasn't thinking. I mixed up the cases of `x ~ Normal()` and `x ~ Bernoulli()`.

test/mcmc/particle_mcmc.jl
```diff
@@ -133,16 +133,17 @@ end
...
+                @addlogprob! -10.0
             else
                 # Need a balanced number of addlogprobs in all branches, or
                 # else PG will error
                 @addlogprob! 0.0
             end
         end
-        c = sample(addlogprob_demo(), PG(10), 100)
+        c = sample(StableRNG(468), addlogprob_demo(), PG(10), 100)
+        @show mean(c[:x])
```
mhauru:
```suggestion
```

mhauru (approved):
Would like to remove the `@show`, but I'll just trust that you'll do it and approve already, because otherwise I'm happy.

test/mcmc/particle_mcmc.jl
penelopeysm: Oh, ooops.

PR #2657, opened 2025-08-15
Suppress info message with verbose=false

### LINKED ISSUE #1398: Option to suppress "Warning" and "Info" statements
Currently, sampling from a model always prints warnings for numerical errors and info statements on the initial step size, even with the option `verbose=false`.  This can really fill up the REPL, especially when a model run generates frequent numerical errors, or when fitting models to multiple datasets in parallel.  Would it be possible to optionally suppress these messages too?

penelopeysm:
Closes #1398.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,8 +1,8 @@
-# 0.42.2
+# 0.40.2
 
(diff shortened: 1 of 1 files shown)


penelopeysm:
Test failures are known and can only really be fixed with Julia patches :((

PR #2662, opened 2025-08-22
Get MarginalLogDensities extension working with Turing v0.40

ElOceanografo:
Fixes some compatibility issues and uses the new `getlogjoint` function instead of `OptimizationContext`.


penelopeysm:
Oh wow, that PR was very old. 😄 Would you mind opening a new PR to main? I think that'll be easier for us to handle too. (To be upfront: I'm mostly away for the next 1.5 weeks, so it may be a couple of weeks before I get to this seriously, if that's alright.)

ElOceanografo:
Sure, I can open a fresh PR. No particular rush, it will be ready when you get back.

PR #2664, opened 2025-08-25
Extension for MarginalLogDensities.jl, take 2

ElOceanografo:
Opening a new version of #2421, as discussed in https://github.com/TuringLang/Turing.jl/pull/2662. Adds an extension for MarginalLogDensities.jl, along with a new exported  `marginalize` function.

### DIFF
--- a/ext/TuringMarginalLogDensitiesExt.jl
+++ b/ext/TuringMarginalLogDensitiesExt.jl
@@ -1,0 +1,46 @@
+module TuringMarginalLogDensitiesExt
+
+using Turing: Turing, DynamicPPL
+using Turing.Inference: LogDensityProblems
+using MarginalLogDensities: MarginalLogDensities
+
+
+# Use a struct for this to avoid closure overhead.
+struct Drop2ndArgAndFlipSign{F}
+    f::F
+end
+
+(f::Drop2ndArgAndFlipSign)(x, _) = -f.f(x)
+
+_to_varname(n::Symbol) = DynamicPPL.@varname($n)
+_to_varname(n::DynamicPPL.AbstractPPL.VarName) = n
+
+function Turing.marginalize(
+    model::DynamicPPL.Model,
+    varnames::Vector,
+    method::MarginalLogDensities.AbstractMarginalizer = MarginalLogDensities.LaplaceApprox(),
+)
+    # Determine the indices for the variables to marginalise out.
+    varinfo = DynamicPPL.typed_varinfo(model)
+    vns = _to_varname.(varnames)
+    varindices = reduce(vcat, DynamicPPL.vector_getranges(varinfo, vns))
+    # Construct the marginal log-density model.
+    # Use linked `varinfo` to that we're working in unconstrained space
+    varinfo_linked = DynamicPPL.link(varinfo, model)
+
+    f = Turing.Optimisation.OptimLogDensity(
+        model,
+        Turing.DynamicPPL.getlogjoint,
+        # Turing.DynamicPPL.typed_varinfo(model)
+        varinfo_linked
+    )
+
+    # HACK: need the sign-flip here because `OptimizationContext` is a hacky impl which
+    # represent the _negative_ log-density.
+    mdl = MarginalLogDensities.MarginalLogDensity(
+        Drop2ndArgAndFlipSign(f), varinfo_linked[:], varindices, (), method
+    )
+    return mdl
+end
+
+end
--- a/src/extensions.jl
+++ b/src/extensions.jl
@@ -1,0 +1,3 @@
+function marginalize(model, varnames, method)
+    error("This function is available after importing MarginalLogDensities.")
+end
--- a/Project.toml
+++ b/Project.toml
@@ -41,10 +41,12 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [weakdeps]
 DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
+MarginalLogDensities = "f0c3360a-fb8d-11e9-1194-5521fd7ee392"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 
 [extensions]
 TuringDynamicHMCExt = "DynamicHMC"
+TuringMarginalLogDensitiesExt = "MarginalLogDensities"
 TuringOptimExt = "Optim"
 
 [compat]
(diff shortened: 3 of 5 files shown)


ext/TuringMarginalLogDensitiesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    # Construct the marginal log-density model.
+    # Use linked `varinfo` to that we're working in unconstrained space
+    varinfo_linked = DynamicPPL.link(varinfo, model)
+
+    f = Turing.Optimisation.OptimLogDensity(
+        model,
+        Turing.DynamicPPL.getlogjoint,
+        # Turing.DynamicPPL.typed_varinfo(model)
+        varinfo_linked
+    )
```
penelopeysm: Just making a note here that I am pretty sure that there's a double negative sign somewhere here with `OptimLogDensity` and the `FlipSign` thing, so I think we will be able to simplify this. (Pre-DynamicPPL 0.37, it used to be that `OptimLogDensity` was the only way to get the log-joint _without_ the Jacobian term, so I'm guessing that's why it got used here, but now we can do that more easily with `getlogjoint`.)

src/extensions.jl
```diff
@@ -0,0 +1,3 @@
+function marginalize(model, varnames, method)
+    error("This function is available after importing MarginalLogDensities.")
+end
```
penelopeysm: I looked through MLD's and Turing's deps, and the only new dep would be HCubature which is a pretty small dep. So I might be in favour to just making this part of the source code itself instead of an extension. (I think Julia extensions are a really good thing in general but there are a few downsides to them, like having to do this kind of 'define in main package and extend in extension' stuff.)

ext/TuringMarginalLogDensitiesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    varinfo = DynamicPPL.typed_varinfo(model)
+    vns = _to_varname.(varnames)
+    varindices = reduce(vcat, DynamicPPL.vector_getranges(varinfo, vns))
+    # Construct the marginal log-density model.
+    # Use linked `varinfo` to that we're working in unconstrained space
+    varinfo_linked = DynamicPPL.link(varinfo, model)
+
+    f = Turing.Optimisation.OptimLogDensity(
+        model,
+        Turing.DynamicPPL.getlogjoint,
```
penelopeysm: I think it could also be nice to make this function `getlogjoint` an optional argument to `marginalize`, because if you substitute this out with `getloglikelihood` or `getlogprior` you can calculate other marginal densities. Again this was something that DPPL 0.37 made possible!

ext/TuringMarginalLogDensitiesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+        varinfo_linked
+    )
+
+    # HACK: need the sign-flip here because `OptimizationContext` is a hacky impl which
+    # represent the _negative_ log-density.
+    mdl = MarginalLogDensities.MarginalLogDensity(
+        Drop2ndArgAndFlipSign(f), varinfo_linked[:], varindices, (), method
+    )
+    return mdl
+end
```
penelopeysm: From reading the MLD readme, I gather that you can do a couple of things with this:

1. Evaluate the marginal log density by calling it
2. Perform optimisation to get point estimates

Would it also be useful to be able to perform MCMC sampling with this? In principle it's not too difficult: a function that takes a vector of parameters, and returns a float, is pretty much what NUTS needs.

To do it properly will probably take a bit of time: I think we would have to either change the interface of `DynamicPPL.LogDensityFunction` or (perhaps easier) do a wrapper around it that keeps track of which variables are marginalised out. So it wouldn't have to be part of this PR. But if this is worth doing then I could have a think about how to do it.

penelopeysm (commented):
Thanks for this! Just some thoughts (no actual code).

ext/TuringMarginalLogDensitiesExt.jl
ElOceanografo: This should already work in theory, since MLD objects implement the LogDensityProblems interface ([here](https://github.com/ElOceanografo/MarginalLogDensities.jl/blob/173be897692f67111ac058c065876ac46c1d93bb/src/MarginalLogDensities.jl#L250)). Would be good to add a test for it though.

src/extensions.jl
ElOceanografo: Sure, this one's up to you guys. I don't mind either way.

ext/TuringMarginalLogDensitiesExt.jl
ElOceanografo: I think it's still needed, `getlogjoint` appears to return a negative log-density:
```julia
@model function demo()
       x ~ Normal()
end
model = demo()
f = Turing.Optimisation.OptimLogDensity(
        model,
        DynamicPPL.getlogjoint,
        DynamicPPL.typed_varinfo(model)
)
f([0]) # 0.9189385332046728
logpdf(Normal(), 0) #-0.9189385332046728
```

ext/TuringMarginalLogDensitiesExt.jl
ElOceanografo: Never mind, I thought it was "just working" but it isn't. Shouldn't be too hard to with some kind of simple wrapper though.

ext/TuringMarginalLogDensitiesExt.jl
ElOceanografo: Done.

ext/TuringMarginalLogDensitiesExt.jl
penelopeysm: Yuppp, the samplers right now only work with DynamicPPL.Model. We should broaden those types and make an interface but it's quite a large undertaking.

Do you know if AD would be able to differentiate the marginal log-density?

ext/TuringMarginalLogDensitiesExt.jl
penelopeysm: Ah, but that's the fault of OptimLogDensity not getlogjoint :)

penelopeysm: Not asking you to fix it btw- happy to do it when I'm back to writing code!

ext/TuringMarginalLogDensitiesExt.jl
ElOceanografo: Not yet, it's the last major feature I need to add to MLD. It's a bit tricky since you need to differentiate through the Hessian of an optimization result, but it should be possible.

src/extensions.jl
penelopeysm: Alright, I'm back at work. I started trying to move some code, then I realised that nothing in this actually needs Turing. So I'm fairly certain that this should be in DynamicPPL (as an extension) rather than in Turing. I'm going to make a PR to DynamicPPL with the code shuffled around (and with the log-density sign thing fixed), will ping you over there :)

penelopeysm:
Transferred to https://github.com/TuringLang/DynamicPPL.jl/pull/1036!

PR #2667, opened 2025-09-01
[breaking] v0.41

penelopeysm:
The usual target branch for aggregating breaking changes.


penelopeysm:
OK, it's a tad annoying when a PR with any commit marked [no ci] is merged, the squash commit message also contains it, which means that CI doesn't run. :/

PR #2670, opened 2025-09-05
Fix multiple-chain method ambiguity

penelopeysm:
https://github.com/TuringLang/DynamicPPL.jl/pull/1035 introduces a method ambiguity with a stray method in Turing.

This PR removes that method.


penelopeysm:
Honestly no, it's just that `resume_from` was broken for multiple chains and I fixed it, and in the process of fixing it there was this incompatibility between latest DynamicPPL and latest Turing over the weekend. This isn't reflective of any larger underlying problem imo

penelopeysm:
I guess there is a larger problem that the codebase is not perfect, but that's my full-time job so I can't snap a finger and fix it ;)

seabbs:
I was wondering if there was any CI checking that could be setup for downstream deps or anything like that.

Issue #2655, opened 2025-08-13
Segfault on test mcmc/Inference


AoifeHughes:
### Description

Regarding the workflows: https://github.com/TuringLang/Turing.jl/actions/runs/16930815371/job/47975728305?pr=2637

@penelopeysm has linked this to: https://github.com/JuliaLang/julia/pull/54840

penelopeysm:
Still broken on 1.10.11, might just have to accept and deal with it.

PR #2676, opened 2025-09-24
Compatibility with DynamicPPL 0.38 + InitContext

penelopeysm:
It should be noted that due to the changes in DynamicPPL's `src/sampler.jl`, the results of running MCMC sampling on this branch will pretty much always differ from that on the main branch. Thus there is no (easy) way to test full reproducibility of MCMC results (we have to rely instead on statistics for converged chains).

TODO:

- [x] pMCMC (it at least runs and gives sensible results on simple models, proper tests will have to wait for CI to run)
- [x] Gibbs (same as above)
- [x] fix initial_params argument for most samplers to require AbstractInitStrategy
- [x] fix tests
- [x] changelog

Separate PRs:

- [ ] use InitStrategy for optimisation as well

    Note that the three pre-existing InitStrategies can be used directly with optimisation. However, to handle constraints properly, it seems necessary to introduce a new subtype of AbstractInitStrategy. I think this should be a separate PR because it's a fair bit of work.

- [x] fix docs for that argument, wherever it is (there's probably some in AbstractMCMC but it should probably be documented on the main site) EDIT: https://turinglang.org/docs/usage/sampling-options/#specifying-initial-parameters

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,10 @@ Please see [the DynamicPPL changelog](https://github.com/TuringLang/DynamicPPL.j
 
 A number of functions such as `returned` and `predict` will have substantially better performance in this release.
 
+### `ProductNamedTupleDistribution`
+
+`Distributions.ProductNamedTupleDistribution` can now be used on the right-hand side of `~` in Turing models.
+
 ### Initial parameters
 
 **Initial parameters for MCMC sampling must now be specified in a different form.**


src/mcmc/gibbs.jl
```diff
@@ -445,13 +397,6 @@ function gibbs_initialstep_recursive(
     varnames, varname_vecs_tail... = varname_vecs
     sampler, samplers_tail... = samplers
 
-    # Get the initial values for this component sampler.
-    initial_params_local = if initial_params === nothing
-        nothing
-    else
-        DynamicPPL.subset(vi, varnames)[:]
-    end
```
penelopeysm, line 453: I was quite pleased with this discovery. Previously the initial params had to be subsetted to be the correct length for the conditioned model. That's not only a faff, but also I get a bit scared whenever there's direct VarInfo manipulation like this.

Now, if you use InitFromParams with a NamedTuple/Dict that has extra params, the extra params are just ignored. So no need to subset it at all, just pass it through directly!

src/mcmc/mh.jl
```diff
@@ -178,8 +178,6 @@ get_varinfo(s::MHState) = s.varinfo
 # Utility functions #
 #####################
 
-# TODO(DPPL0.38/penelopeysm): This function should no longer be needed
-# once InitContext is merged.
```
penelopeysm, line 182: unfortunately `set_namedtuple!` is used elsewhere in this file (though it won't appear in this diff) so we can't delete it (yet)

src/mcmc/mh.jl
```diff
@@ -410,13 +398,20 @@ function AbstractMCMC.step(
...
-    retval = DynamicPPL.assume(rng, SampleFromPrior(), dist, vn, vi)
-    return retval
+    # Allow MH to sample new variables from the prior if it's not already present in the
+    # VarInfo.
+    dispatch_ctx = if haskey(vi, vn)
+        DynamicPPL.DefaultContext()
+    else
+        DynamicPPL.InitContext(context.rng, DynamicPPL.InitFromPrior())
+    end
+    return DynamicPPL.tilde_assume!!(dispatch_ctx, right, vn, vi)
```
penelopeysm, line 425: The behaviour of `SampleFromPrior` used to be: if the key is present, don't actually sample, and if it was absent, sample. This if/else replicates the old behaviour.

src/mcmc/particle_mcmc.jl
```diff
@@ -36,30 +36,28 @@ function unset_all_del!(vi::AbstractVarInfo)
...
+    rng::R
+end
+DynamicPPL.NodeTrait(::ParticleMCMCContext) = DynamicPPL.IsLeaf()
+
+struct TracedModel{V<:AbstractVarInfo,M<:Model,E<:Tuple} <: AdvancedPS.AbstractGenericModel
     model::M
-    sampler::S
     varinfo::V
     evaluator::E
+    resample::Bool
```
penelopeysm, line 16: For pMCMC, this Boolean field essentially replaces the del flag. Instead of `set_all_del` and `unset_all_del` we construct new `TracedModel` with this set to true and false respectively.

test/mcmc/ess.jl
```diff
@@ -108,8 +108,12 @@ using Turing
...
         spl_xy = Gibbs(@varname(z) => NUTS(), (@varname(x), @varname(y)) => ESS())
 
-        @test sample(StableRNG(23), xy(), spl_xy, num_samples).value ≈
-            sample(StableRNG(23), x12(), spl_x, num_samples).value
+        chn1 = sample(StableRNG(23), xy(), spl_xy, num_samples)
+        chn2 = sample(StableRNG(23), x12(), spl_x, num_samples)
+
+        @test mean(chn1[:z]) ≈ mean(chn2[:z]) atol = 0.05
+        @test mean(chn1[:x]) ≈ mean(chn2["x[1]"]) atol = 0.05
+        @test mean(chn1[:y]) ≈ mean(chn2["x[2]"]) atol = 0.05
```
penelopeysm: The values are no longer exactly the same (it has something to do with initialisation behaviour which is different for the two models). But we can still check that the results are sensibly similar, which is probably also more meaningful anyway as it means that ESS not only works on both models but also consistently converges regardless of how the model is specified.

test/stdlib/distributions.jl
```diff
@@ -130,7 +130,14 @@ using Turing
...
 
-                            chn = sample(StableRNG(468), m(), HMC(0.05, 20), n_samples)
+                            seed = if dist isa GeneralizedExtremeValue
+                                # GEV is prone to giving really wacky results that are quite
+                                # seed-dependent.
+                                StableRNG(469)
+                            else
+                                StableRNG(468)
+                            end
+                            chn = sample(seed, m(), HMC(0.05, 20), n_samples)
```
penelopeysm, line 140: Case in point:

```julia
julia> using Turing, StableRNGs

julia> dist = GeneralizedExtremeValue(0, 1, 0.5); @model m() = x ~ dist
m (generic function with 2 methods)

julia> mean(dist)
1.5449077018110322

julia> mean(sample(StableRNG(468), m(), HMC(0.05, 20), 10000; progress=false))
Mean
  parameters      mean
      Symbol   Float64

           x    3.9024


julia> mean(sample(StableRNG(469), m(), HMC(0.05, 20), 10000; progress=false))
Mean
  parameters      mean
      Symbol   Float64

           x    1.5868
```

penelopeysm:
For the record, 11 failing CI jobs is the expected number:

- 8x failing jobs because `[sources]` is not understood on 1.10
- 3x failing jobs because Libtask 1.12

There is also the failing job caused by base Julia segfault (#2655), but that's on 1.10 so overlaps with the first category.

penelopeysm:
@mhauru, I haven't run CI against the latest revisions like removal of the del flag, but I think this might be meaty enough as it stands and also any adjustments arising from that PR (like renaming islinked) should be quite trivial.

src/mcmc/abstractmcmc.jl
```diff
@@ -1,9 +1,9 @@
 # TODO: Implement additional checks for certain samplers, e.g.
 # HMC not supporting discrete parameters.
 function _check_model(model::DynamicPPL.Model)
-    # TODO(DPPL0.38/penelopeysm): use InitContext
-    spl_model = DynamicPPL.contextualize(model, DynamicPPL.SamplingContext(model.context))
-    return DynamicPPL.check_model(spl_model, VarInfo(); error_on_failure=true)
+    new_context = DynamicPPL.setleafcontext(model.context, DynamicPPL.InitContext())
+    new_model = DynamicPPL.contextualize(model, new_context)
```
mhauru: Isn't there a short-hand for these two lines now?

src/mcmc/mh.jl
```diff
@@ -410,13 +413,25 @@ function AbstractMCMC.step(
...
-#### Compiler interface, i.e. tilde operators.
-####
-function DynamicPPL.assume(
-    rng::Random.AbstractRNG, spl::Sampler{<:MH}, dist::Distribution, vn::VarName, vi
+struct MHContext{R<:AbstractRNG} <: DynamicPPL.AbstractContext
+    rng::R
+end
+DynamicPPL.NodeTrait(::MHContext) = DynamicPPL.IsLeaf()
+
+function DynamicPPL.tilde_assume!!(
```
mhauru, line 415: Why is this needed? Doesn't MH just use `init` to get initial values like any other sampler, and then evaluate logpdfs on proposed steps?

Also, this looks like dynamical dispatch, since the context type depends on `haskey(vi, vn)`, which could be a performance issue.

src/mcmc/particle_mcmc.jl
```diff
@@ -446,48 +450,44 @@ function set_trace_local_varinfo_maybe(vi::AbstractVarInfo)
...
-        DynamicPPL.unset_flag!(vi, vn, "del") # Reference particle parent
-        # TODO(mhauru):
-        # The below is the only line that differs from assume called on SampleFromPrior.
-        # Could we just call assume on SampleFromPrior with a specific rng?
-        r = rand(trng, dist)
-        vi[vn] = DynamicPPL.tovec(r)
+    trng = get_trace_local_rng_maybe(ctx.rng)
+    resample = get_trace_local_resampled_maybe(true)
+
+    dispatch_ctx = if ~haskey(vi, vn) || resample
```
mhauru, line 410: Like with MH, I wonder about dynamic dispatch here. Might be unavoidable and/or inconsequential in this case though.

src/mcmc/repeat_sampler.jl
```diff
@@ -81,3 +81,58 @@ function AbstractMCMC.step_warmup(
...
 end
+
+# Need some extra leg work to make RepeatSampler work seamlessly with DynamicPPL models +
+# samplers, instead of generic AbstractMCMC samplers.
+
+function DynamicPPL.init_strategy(spl::RepeatSampler{<:Sampler})
+    return DynamicPPL.init_strategy(spl.sampler)
+end
+
+function AbstractMCMC.sample(
```
mhauru, line 101: I don't see why these became necessary now. Is it something about the type hierarchy around `Sampler`?

test/mcmc/ess.jl
mhauru: While we are at it, shall we add a `≈ -3.0` on the last line?

It's a shame that these don't match exactly any more, but I guess if something about the order of sampling changes in Gibbs or some such then what can you do.

mhauru: Seems you can't chain comparisons under `@test` when using `atol`. I'm superficial enough that that makes me lean towards not checking against `-3.0`.

test/mcmc/is.jl
```diff
@@ -1,63 +1,52 @@
...
+            as = Vector{Float64}(undef, n)
+            bs = Vector{Float64}(undef, n)
 
-        for i in 1:n
-            as[i], bs[i], logps[i] = reference()
+            for i in 1:n
+                as[i] = rand(rng, Normal(4, 5))
+                bs[i] = rand(rng, Normal(as[i], 1))
+            end
+            # logevidence = logsumexp(logps) - log(n)
```
mhauru: Can this line go?

test/mcmc/is.jl
```diff
@@ -1,63 +1,52 @@
...
+        end
 
-        Random.seed!(seed)
-        chain = sample(model, alg, n; check_model=false)
-        sampled = get(chain, [:a, :b, :loglikelihood])
+        alg = IS()
+        N = 1000
+        model = normal()
+        chain = sample(StableRNG(468), model, alg, N)
+        ref = reference(N)
```
mhauru, line 36: Getting quite nit-picky, but could the RNG be an argument to `reference`, to ensure the two instance of `StableRNG(468)` remain in sync?

test/Project.toml
```diff
@@ -77,3 +76,6 @@ StatsBase = "0.33, 0.34"
 StatsFuns = "0.9.5, 1"
 TimerOutputs = "0.5"
 julia = "1.10"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: Leaving this comment just as a reminder that this needs to be removed before merging.

Project.toml
```diff
@@ -90,3 +90,6 @@ julia = "1.10.8"
 [extras]
 DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: Likewise this.

HISTORY.md
```diff
@@ -1,5 +1,30 @@
...
 
+## DynamicPPL 0.38
+
+Turing.jl v0.41 brings with it all the underlying changes in DynamicPPL 0.38.
+
+The only user-facing difference is that initial parameters for MCMC sampling must now be specified in a different form.
+You still need to use the `initial_params` keyword argument to `sample`, but the allowed values are different.
+For almost all samplers in Turing.jl (except `Emcee`) this should now be a `DynamicPPL.AbstractInitStrategy`.
+
+TODO LINK TO DPPL DOCS WHEN THIS IS LIVE
```
mhauru: Likewise a reminder comment.

HISTORY.md
```diff
@@ -1,5 +1,30 @@
...
+
+There are three kinds of initialisation strategies provided out of the box with Turing.jl (they are exported so you can use these directly with `using Turing`):
+
+  - `InitFromPrior()`: Sample from the prior distribution. This is the default for most samplers in Turing.jl (if you don't specify `initial_params`).
+  - `InitFromUniform(a, b)`: Sample uniformly from `[a, b]` in linked space. This is the default for Hamiltonian samplers. If `a` and `b` are not specified it defaults to `[-2, 2]`, which preserves the behaviour in previous versions (and mimics that of Stan).
+  - `InitFromParams(p)`: Explicitly provide a set of initial parameters. **Note: `p` must be either a `NamedTuple` or a `Dict{<:VarName}`; it can no longer be a `Vector`.** Parameters must be provided in unlinked space, even if the sampler later performs linking.
```
mhauru: Do I recall correctly that you did end up implementing the option of providing an unwrapped `NamedTuple` or `Dict` as well?

Oh, also, just came to mind: Does it need to be a `Dict`, or can it be an `AbstractDict`?

penelopeysm:
> I'm wondering about how to merge this.

Personally, I'm not too fussed. I think we discussed this last time round and IIRC you said you'd prefer to keep `breaking` 'clean' and mergeable into main at any given time (apologies if I am misremembering). If that's the case, then we should keep this as the base branch for DPPL 0.38 fixes, until 0.38 is released.

HISTORY.md
penelopeysm: Ah, yes, that's changed now. Will update accordingly. (It's AbstractDict.)

test/mcmc/is.jl
penelopeysm: Gladly deleted

src/mcmc/abstractmcmc.jl
penelopeysm: Yup, fixed

src/mcmc/mh.jl
penelopeysm: The rationale is in the comment right below this:

```julia
    # Allow MH to sample new variables from the prior if it's not already present in the
    # VarInfo.
```

This preserves the old behaviour of MH (see https://github.com/TuringLang/Turing.jl/blob/cabe73fd07b3ddb37b51f6c0c9db66891e179f3c/src/mcmc/mh.jl#L416-L422) where it would sample new variables in models where the set of variables is not fixed (thus, e.g., initialisation might result in a different set of variables).

It directly mimics the old code for SampleFromPrior: https://github.com/TuringLang/DynamicPPL.jl/blob/c1d9b61933a6251fbeac249e8ebf03631f1f25bf/src/context_implementations.jl#L131-L171 which includes the `haskey(vi, vn)` check.

> dynamic dispatch

I don't really understand what is so bad with this. (For what it's worth, I searched online and I can't find any actual explanation, so feel free to enlighten me.) I am just calling a different method, depending on the value of a Boolean which — sure — you can't know it for sure until runtime.

But if the 'not known until runtime' is the main problem, then it seems like literally any code that uses `if haskey(vi, vn)` is already problematic — would it be any better or worse if I were to inline the definitions here?

```julia
if haskey(vi, vn)
    # inline the definition of tilde_assume on DefaultContext here
else
    # inline the definition of tilde_assume on Init here
end
```

Surely that's okay, right, because otherwise it seems incredibly limiting for programming if you can't even use conditionals. So I assume I'm missing something. Like, what's the difference between 'dynamic dispatch' and 'code that uses conditionals because values aren't known until runtime'.

src/mcmc/repeat_sampler.jl
penelopeysm: Yeah, `RepeatSampler` is weird, because it doesn't subtype `InferenceAlgorithm`. So some of the code like default `sample` methods has to be copied over for it. In the past, the methods used to be written for `Union{Sampler{<:InferenceAlgorithm},RepeatSampler}`. I think that's worse because it was less flexible, you couldn't change one without affecting the behaviour of the other.

test/mcmc/is.jl
penelopeysm: The two instances of `StableRNG` are not the same, so there's no need for them to be the same thing: in this line

```julia
@test isapprox(mean(chain[:a]), mean(ref.as); atol=0.1)
```

the respective values are

```julia
julia> mean(chain[:a])
4.019881006005818

julia> mean(ref.as)
4.02356421170837
```

This is because of various things such as check_model, initialisation, ... so the intent of the test isn't to check that they generate exactly the same values, it's to check that `sample(..., IS())` (with a given seed) generates values that are qualitatively similar to hand-rolled IS (with a different given seed), and it just so happens that those given seeds are the same.

penelopeysm: I added a comment to explain the above.

test/mcmc/ess.jl
penelopeysm: Actually, I checked again, and I have no idea why but the values _are_ now the same. Weird. I'm going to add the old test back.

penelopeysm: Also added some accuracy checks for `z` and `y`, wasn't fussed enough to work out what the expected mean for `x` was.

penelopeysm:
Okay, I'm quite happy with where CI on this PR has gotten to. There are a handful of residual failures, which are mostly to do with `Sampler` methods that I removed a bit too early. #2689 properly handles the removal of `Sampler` so I don't feel inclined to chase them down here.

src/mcmc/emcee.jl
```diff
@@ -39,20 +39,18 @@ _get_n_walkers(spl::Sampler{<:Emcee}) = _get_n_walkers(spl.alg)
 function DynamicPPL.init_strategy(spl::Sampler{<:Emcee})
     return fill(DynamicPPL.InitFromPrior(), _get_n_walkers(spl))
 end
+# TODO(penelopeysm / DPPL 0.38) This is type piracy (!!!) The function
+# `_convert_initial_params` will be moved to Turing soon, and this piracy SHOULD be removed
+# in https://github.com/TuringLang/Turing.jl/pull/2689, PLEASE make sure it is!
```
mhauru: Leaving a comment to track that this gets done before merging.

HISTORY.md
```diff
@@ -13,13 +13,16 @@ TODO LINK TO DPPL DOCS WHEN THIS IS LIVE
...
 
   - `InitFromPrior()`: Sample from the prior distribution. This is the default for most samplers in Turing.jl (if you don't specify `initial_params`).
+
   - `InitFromUniform(a, b)`: Sample uniformly from `[a, b]` in linked space. This is the default for Hamiltonian samplers. If `a` and `b` are not specified it defaults to `[-2, 2]`, which preserves the behaviour in previous versions (and mimics that of Stan).
-  - `InitFromParams(p)`: Explicitly provide a set of initial parameters. **Note: `p` must be either a `NamedTuple` or a `Dict{<:VarName}`; it can no longer be a `Vector`.** Parameters must be provided in unlinked space, even if the sampler later performs linking.
+  - `InitFromParams(p)`: Explicitly provide a set of initial parameters. **Note: `p` must be either a `NamedTuple` or an `AbstractDict{<:VarName}`; it can no longer be a `Vector`.** Parameters must be provided in unlinked space, even if the sampler later performs linking.
+    
+      + For this release of Turing.jl, you can also provide a `NamedTuple` or `AbstractDict{<:VarName}` and this will be automatically wrapped in `InitFromParams` for you.
```
mhauru:
```suggestion
      + For this release of Turing.jl, you can also provide a `NamedTuple` or `AbstractDict{<:VarName}` and this will be automatically wrapped in `InitFromParams` for you. This is an intermediate measure for backwards compatibility, and will eventually be removed.
```
I assume since you said "for this release" that this is the case. If not, then I would remove the "for this release" part.

src/mcmc/mh.jl
mhauru: I'm not sure exactly how bad dynamic dispatch is either, and I wondered about the same question you are asking. Clearly some of it has to be okay. One difference to inlining is that without inlining the right method will have to be looked up from some method table at runtime. I'm not sure what exactly that causes, but maybe there are substantial costs there? Maybe those method tables are quite heavy data structures, compared to what the code itself is dealing with?

One the first point, I don't understand why MH needs to do this differently from every other sampler. I feel like whether to fall back to InitContext when evaluating logp for sampling purposes and hitting a `tilde_assume!!` for an unknown variable should be a decision made on a higher level than a sampler's implementation. I get that this is not new behaviour in this PR, though it's clunkyness is exposed in a different way because we now need to create a whole new context for MH.

HISTORY.md
penelopeysm: Technically, it was Hong who said it, but I'm also quite happy to remove it eventually.

src/mcmc/mh.jl
penelopeysm: > dynamic dispatch

I searched a bit more and came across this: https://discourse.julialang.org/t/curious-about-the-internals-of-dynamic-dispatch/102888/8. In this example, there is some object `a::A` but `A` is an abstract type, so the compiler can't figure out which method `f(a)` refers to ahead of time and this is only resolved at runtime.

But here the compiler can perfectly figure out which methods are being used, because both `InitContext` and `DefaultContext` are concrete types. My impression is that what's going on here is not dynamic dispatch — the dispatch itself is static: when I write `tilde_assume(...)` the compiler has enough information to know which method is being referred to. Once `haskey(vi, vn)` is evaluated, there is no more looking things up in method tables. I think the dynamic dispatch thing only happens if the types of the arguments to `tilde_assume!!` are themselves not concrete.

The only downside from the compiler's point of view is that it can't aggressively cut out one branch ahead of time, because it can't figure out which branch it will go down.

Also, there would be issues with type stability if both `tilde_assume!!`s return different types. But that's not a dispatch problem, that's just plain old type stability and that wouldn't be affected if the function definitions were inlined.

> I don't understand why MH needs to do this differently from every other sampler

I don't know either. I am in principle not opposed to cutting this out and just using `DefaultContext`, but I would probably prefer to not make this decision in this PR.

> performance

If it's of any reassurance, I ran the following code on main plus this PR; main comes in at around 0.89 seconds, this PR at 0.83 seconds. (of course, I have no clue why or what made it better! 😄)

```julia
using Turing

J = 8
y = [28, 8, -3, 7, -1, 1, 18, 12]
sigma = [15, 10, 16, 11, 9, 11, 10, 18]

@model function eight_schools_centered(J, y, sigma)
    mu ~ Normal(0, 5)
    tau ~ truncated(Cauchy(0, 5); lower=0)
    theta = Vector{Float64}(undef, J)
    for i in 1:J
        theta[i] ~ Normal(mu, tau)
        y[i] ~ Normal(theta[i], sigma[i])
    end
end

model_esc = eight_schools_centered(J, y, sigma)
@time sample(model_esc, MH(), 10000; progress=false);
```

mhauru: Ah, yes, what you say about dynamic dispatch vs just calling different methods in different branches makes sense. Thanks for thinking this through better than I did.

src/mcmc/particle_mcmc.jl
mhauru: Resolved in the other thread.

mhauru:
A note, mostly for myself: I'm happy with the code in this PR. We should collect other changes related to 0.38 compat into it before merging, since it still leaves things in a broken state. However, once other PRs have been merged into this so that tests pass, and the above reminder comments have been addressed, I'm happy to merge. No need to read through this PR again.

test/Project.toml
penelopeysm: done

Project.toml
penelopeysm: done

src/mcmc/emcee.jl
penelopeysm: done

penelopeysm:
@mhauru Following on from the meeting just now ... is there anything else you'd like to include in 0.41? If not, I'll fix the remaining todo link in the changelog and then we should be good to go.

Optimisation still doesn't use AbstractInitStrategy, but I'm quite happy to leave that out as part of a near-future refactoring of the optimisation API.

mhauru:
On the contrary, if I we've managed to get all tests to pass on a DPPL version with so many changes, I would advocate for releasing this to have a checkpoint and doing more incremental changes.

penelopeysm:
Ah, yes, that's definitely very wise. I'll update the changelog and do the merges.

PR #2680, opened 2025-10-02
pass initial_state through for NUTS sampling

### LINKED ISSUE #2679: `initial_state` is not used when `resume_from` is not specified
`sample` does not use `initial_state` if specified without `resume_from`. EDIT: This applies when `sampler` is of type `Sampler{<:AdaptiveHamiltonian}`.

I encountered this when I tried to resume one chain after running multiple chains, and thought I'd do this using `initial_state` instead of `resume_from`, especially in light of #2675.

MWE:
```julia
using Turing, Random

@model function gdemo(x)
    s² ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s²))
    for i in eachindex(x)
        x[i] ~ Normal(m, sqrt(s²))
    end
end

rng = Random.default_rng();
Random.seed!(rng, 1);
data = 4 .+ 0.1 .* randn(rng, 100);

chain1 = sample(rng, gdemo(data), NUTS(), 10, nadapts=1000, save_state=true)
display(Array(chain1))

chain2 = sample(rng, gdemo(data), NUTS(), 10, nadapts=0, initial_state=chain1.info.samplerstate)
display(Array(chain2)) # initial params is drawn from prior instead of using initial_state
```

penelopeysm:
Closes #2679


sunxd3 (approved):
LGTM

PR #2683, opened 2025-10-07
Use `step_warmup` for AdaptiveHamiltonian

### LINKED ISSUE #2678: Use `step_warmup` for AdaptiveHamiltonian rather than `nadapts`
NUTS currently has a parameter called nadapts which controls how many MCMC steps include an adaptation bit.

discard_adapt then controls whether those adaptation steps are discarded at the start.

For consistency with the AbstractMCMC API this should be changed to be num_warmup + discard_initial. And the adaptation steps should call `AbstractMCMC.step_warmup` rather than `AbstractMCMC.step`.

penelopeysm:
Closes #2678

Note, CI not running because this needs https://github.com/TuringLang/AbstractMCMC.jl/pull/178 to go through.

This PR:

- Removes the `nadapts` argument from HMCDA and NUTS constructors. Instead of specifying that, the user should specify the `num_warmup` keyword argument to `sample()`.
- Removes the `discard_adapt` keyword argument when sampling with HMCDA/NUTS, in favour of `discard_initial`, which has the same meaning for all samplers and is therefore more coherent.
- Since adaptive Hamiltonians must do different behaviours for adaptation / warmup steps, this PR also implements `AbstractMCMC.step_warmup` for these samplers.

This also needs:

- [x] Changelog

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,22 @@
 # 0.41.0
 
+## Breaking changes
+
+**HMCDA and NUTS constructors**
+
+HMCDA and NUTS previously used to take an optional first positional argument, which specified the number of adaptation steps to take.
+Alternatively, this could also be specified via the `nadapts` keyword argument to `sample()`.
+This has been removed: you should use the `num_warmup` keyword argument to `sample()` instead, which is consistent with all other samplers in Turing.jl.
+Warmup steps are always adaptation steps for HMCDA and NUTS, so there is no need to specify the number of adaptation steps separately.
+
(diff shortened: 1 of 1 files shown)


penelopeysm:
Oh, for goodness' sake, this is way more annoying than I had realised. `DynamicPPL.Sampler`, simply, straight up, does NOT allow for `step_warmup` at all because `DynamicPPL.initialstep` is only called via `step` and not `step_warmup`,

penelopeysm:
I think this will just have to wait for `Sampler` to be removed as a whole.

PR #2685, opened 2025-10-13
Temporarily change tests to use 1.11

mhauru:
Will open an issue to track the fact that this should be reverted ASAP once Mooncake and Libtask are back online.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+# 0.40.5
+
+Bump Optimization.jl compatibility to include v5.
+
 # 0.40.4
 
 Fixes a bug where `initial_state` was not respected for NUTS if `resume_from` was not also specified.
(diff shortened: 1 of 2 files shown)


mhauru:
Fair point, done. I also comment out `'pre'` since it's currently the same as `'1'`, i.e. 1.12.0.

mhauru:
CI is failing due to this: https://github.com/SciML/Optimization.jl/issues/1056

Since this isn't in a rush, and I expect that to be fixed within 24h, let's just wait until tomorrow. I may be on annual leave, but if someone else wants to retrigger CI once the above issue is closed and merge if it passes, feel free.

penelopeysm:
> I am also not sure why some CIs seem to stuck.

I think it's because the '1' test is set as a 'required' test, somewhere in the repository settings. Thus GitHub's UI puts a dummy entry for that because it expects that test to be run.

(It could certainly be more intuitive)

PR #2689, opened 2025-10-16
Remove `Sampler`, remove `InferenceAlgorithm`, transfer `initialstep`, `init_strategy`, and other functions from DynamicPPL to Turing

penelopeysm:
https://github.com/TuringLang/DynamicPPL.jl/pull/1037 deletes `DynamicPPL.Sampler` and its interface. This PR makes the necessary upstream adjustments.

Once this PR and the DPPL counterpart are merged, **all of the MCMC sampling behaviour will be self-contained in Turing**: this means that we can easily refactor sampler code in Turing without being held back by interface code in DynamicPPL. This will finally pave the way for things like sampling with `LogDensityFunction` #2555 and allow long-standing things to be fixed like weird parameters to NUTS #2678. It's been a few months coming so I am very pleased with this!

Literally all of the actual code changes in this PR are perfunctory boilerplate, like copy-pastes from DynamicPPL, changing `x::Sampler{<:X}` to `x::X`, etc. However, there is one conceptual bit that I think is worth explaining:

## Special behaviour when sampling from Turing models with Turing samplers

As discussed in, e.g. https://github.com/TuringLang/Turing.jl/issues/2413, there are certain special behaviours that are enabled when somebody calls `sample(model, NUTS(), 1000)`. For example:

 - `check_model` runs checks on the model,
 - `initial_params` is set to `DynamicPPL.init_strategy(spl)` (now `Turing.Inference.init_strategy(spl)`),
 - `chain_type` is set to `MCMCChains.Chains`, ...

Previously the way this was accomplished was by
1. declaring `NUTS <: InferenceAlgorithm`,
2. `sample(rng, ::AbstractModel, ::InferenceAlgorithm, N)` would then wrap NUTS in `DynamicPPL.Sampler`
3. `sample(rng, ::AbstractModel, ::DynamicPPL.Sampler, N)` would then do all the special behaviour

This PR changes it such that the special behaviour is triggered in `sample(rng, ::DynamicPPL.Model, ::AbstractSampler)`, in `src/mcmc/abstractmcmc.jl`. That is to say, NUTS is no longer 'a special class of sampler'. If you write your own sampler that knows how to handle Turing models (i.e. if you define `step(rng, ::DynamicPPL.Model, ::MySampler)`), then you will get all the special behaviour when you call `sample(model, MyNewSampler(), 1000)`.

This approach is a lot simpler than what came before, but as pointed out in #2413 there are some concerns with method ambiguities. For example, if somebody defines

```julia
sample(::AbstractModel, ::MyNewSampler, N) = ...
```

then calling `sample(dynamicppl_model, MyNewSampler(), 1000)` will lead to method ambiguities. However, it's been almost a year since then, and there are a number of reasons why I'm not really fussed about the ambiguities and why I no longer agree with the arguments in that thread (indeed I don't agree with my own comments from that thread).

I've gated the rest behind a details tag to avoid overwhelming. If you don't see a problem with the above, then feel free to ignore.

<details><summary>Method ambiguities</summary>

1. As explained in #2413 the only meaningful case where somebody might write such a `sample` method is if their sampler was a 'meta-sampler', i.e., it works with _all_ models without knowing anything about the internal structure of said models. In practice this **cannot** be done because `AbstractModel` has no interface and if you don't know anything about the model, you can't do anything meaningful with it. For example, in principle Gibbs shouldn't _need_ to know anything about the model because it just shuttles it between component samplers. In practice we have to specialise on the model type and include a ton of stuff like `setparams_varinfo` to make it even work.

2. Method ambiguities have a very straightforward solution which is to declaring an extra, more specific method. I think that is a perfectly fine price to pay if you are writing a (genuine) meta-sampler and want it to work with Turing models specifically. For example this is the case with `RepeatSampler`, it's just five extra lines of code to make it work.

3. Note that in the original proposal in #2413 the method ambiguities would be avoided _but_ if somebody wrote a model and wanted to hook into the special behaviour, they would have to write the more specific method anyway! So IMO this is no worse.

4. I haven't exactly seen an explosion of activity in the Julia ecosystem around creating meta-samplers, so I'm quite unconvinced that in practice this will cause any problems

</details>

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -90,6 +90,3 @@ julia = "1.10.8"
 [extras]
 DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
-
-[sources]
-DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -76,6 +76,3 @@ StatsBase = "0.33, 0.34"
 StatsFuns = "0.9.5, 1"
 TimerOutputs = "0.5"
 julia = "1.10"
-
-[sources]
-DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}


src/mcmc/ess.jl
```diff
@@ -103,3 +103,18 @@ struct ESSLikelihood{L<:DynamicPPL.LogDensityFunction}
...
+function AbstractMCMC.step(
+    rng::AbstractRNG,
+    model::DynamicPPL.Model,
+    sampler::EllipticalSliceSampling.ESS;
+    kwargs...,
+)
+    return error(
+        "This method is not implemented! If you want to use the ESS sampler in Turing.jl, please use `Turing.ESS()` instead. If you want the default behaviour in EllipticalSliceSampling.jl, wrap your model in a different subtype of `AbstractMCMC.AbstractModel`, and then implement the necessary EllipticalSliceSampling.jl methods on it.",
+    )
+end
```
penelopeysm: For full transparency, I will confess that I had to add two methods to remove method ambiguities caused by this PR. Both arise from the default definition of `AbstractMCMC.step`

https://github.com/TuringLang/Turing.jl/blob/b5e07603718bb290aa7d4bfc2b5af4e2bcee7000/src/mcmc/abstractmcmc.jl#L141-L163

which I would like to remove (because it doesn't allow for warmup), but I don't want to do it in this PR / version.

src/mcmc/abstractmcmc.jl
```diff
@@ -47,18 +107,66 @@ function AbstractMCMC.sample(
...
+
+    # Fill it with initial parameters. Note that, if `InitFromParams` is used, the
+    # parameters provided must be in unlinked space (when inserted into the
+    # varinfo, they will be adjusted to match the linking status of the
+    # varinfo).
+    _, vi = DynamicPPL.init!!(rng, model, vi, initial_params)
+
+    # Call the actual function that does the first step.
+    return initialstep(rng, model, spl, vi; initial_params, kwargs...)
+end
```
penelopeysm, line 172: This method of `step` is actually a little bit evil. It used to be less bad because it only applied to `Sampler{<:InferenceAlgorithm}`, but now it applies to all `AbstractSampler`, which actually _does_ cause some method ambiguities (which I've pointed out in my other comments).

On top of that, this is just generally a bit inflexible when it comes to warmup steps since it's only defined as a method for `step` and not `step_warmup`.

I think that in the next version of Turing this method should be removed. However, I've opted to preserve it for now because I don't want to make too many conceptual changes in this PR (the diff is already too large).

src/mcmc/ess.jl
penelopeysm, line 120: This is the first method ambiguity caused by `step`. This needs to exist because of https://github.com/TuringLang/EllipticalSliceSampling.jl/blob/6b8fc40e9568122331ef705a2f668d854b692538/src/abstractmcmc.jl#L20-L27

src/mcmc/repeat_sampler.jl
```diff
@@ -40,6 +41,14 @@ function AbstractMCMC.step(
...
     return AbstractMCMC.step(rng, model, sampler.sampler; kwargs...)
 end
+# The following method needed for method ambiguity resolution.
+# TODO(penelopeysm): Remove this method once the default `AbstractMCMC.step(rng,
+# ::DynamicPPL.Model, ::AbstractSampler)` method in `src/mcmc/abstractmcmc.jl` is removed.
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG, model::DynamicPPL.Model, sampler::RepeatSampler; kwargs...
+)
+    return AbstractMCMC.step(rng, model, sampler.sampler; kwargs...)
+end
```
penelopeysm, line 51: This is the second method ambiguity caused by `step`.

penelopeysm:
@mhauru pinging you also for thoughts on the method ambiguity thing, if you have any

src/mcmc/repeat_sampler.jl
mhauru: I'm confused as to why this is a method ambiguity. `RepeatSampler` is a subtype of `AbstractSampler` and should thus take precedence given the other positional arguments are the same. Kwargs don't do dispatch. I think I'm missing something.

penelopeysm: Errr, sorry, I was very unclear. It's the method above this that used AbstractModel was the problem. This one is the solution.

mhauru: My bad for not reading the PR properly. If you just need the two methods to disambiguate, and the plan is to remove the offending method soonish (a plan I like, `initialstep` is confusing), then I'm not too offended.

src/mcmc/external_sampler.jl
```diff
@@ -115,19 +140,18 @@ getparams(::DynamicPPL.Model, transition::AdvancedMH.Transition) = transition.pa
 function AbstractMCMC.step(
     rng::Random.AbstractRNG,
     model::DynamicPPL.Model,
-    sampler_wrapper::Sampler{<:ExternalSampler};
+    sampler_wrapper::ExternalSampler;
     initial_state=nothing,
-    initial_params=DynamicPPL.init_strategy(sampler_wrapper.alg.sampler),
+    initial_params=DynamicPPL.init_strategy(sampler_wrapper.sampler),
```
sunxd3:
```suggestion
    initial_params=Turing.Inference.init_strategy(sampler_wrapper.sampler),
```
perhaps?

src/mcmc/gibbs.jl
```diff
@@ -315,21 +309,20 @@ support calling both step and step_warmup as the initial step. DynamicPPL initia
...
     return vi
 end
 
 function AbstractMCMC.step(
     rng::Random.AbstractRNG,
     model::DynamicPPL.Model,
-    spl::DynamicPPL.Sampler{<:Gibbs};
-    initial_params::DynamicPPL.AbstractInitStrategy=DynamicPPL.init_strategy(spl),
+    spl::Gibbs;
+    initial_params=Turing.Inference.init_strategy(spl),
```
sunxd3, line 321: slightly tricky, this would technically override the `init_strategy` behavior for the component samplers, right? for instance HMC, when used for a component will default to sampling from prior instead of uniform?

sunxd3 (commented):
I am not too bothered with the method ambiguity.

src/mcmc/external_sampler.jl
penelopeysm: That's quite awkward, this implies that it wasn't tested lol

src/mcmc/gibbs.jl
penelopeysm: Yup indeed. I think this was already the case prior to this, because initial_params would be generated as a single vector (using `initialsampler(::Sampler{Gibbs})`, which would be `SampleFromPrior`) and then just split up between the components. I'm actually not sure how to fix this, but I can open an issue to track it.

penelopeysm: Most likely the solution would be to implement a custom `InitStrategy` for Gibbs

penelopeysm: https://github.com/TuringLang/Turing.jl/issues/2693

src/mcmc/external_sampler.jl
penelopeysm: Oh, actually, it seems that the default value just wasn't needed, because sample() always passes initial_params through to step(), and sample() already has a default value. Removed now.

sunxd3 (approved):
It looks great, thanks!

PR #2691, opened 2025-10-18
Forbid PG/SMC on 1.12

penelopeysm:
It'll error when it gets to Libtask anyway, so this is just a more useful error.

### DIFF
--- a/src/mcmc/particle_mcmc.jl
+++ b/src/mcmc/particle_mcmc.jl
@@ -2,6 +2,8 @@
 ### Particle Filtering and Particle MCMC Samplers.
 ###
 
+LIBTASK_BROKEN = v"1.12.0" <= VERSION < v"1.12.2"
+
 ### AdvancedPS models and interface
 
 """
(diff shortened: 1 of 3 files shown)


src/mcmc/particle_mcmc.jl
```diff
@@ -2,6 +2,8 @@
 ### Particle Filtering and Particle MCMC Samplers.
 ###
 
+LIBTASK_BROKEN = v"1.12.0" <= VERSION < v"1.12.2"
```
mhauru, line 5: Could we also check against `Pkg.develop`'s output, to check the Libtask version? Turing 0.40.6 might start to work if Libtask just puts out a bug fix release.

penelopeysm: Yeah, I am actually kind of unsure about this PR. I actually think that the correct solution is to fix the Julia compat bounds on Libtask itself. i.e. 0.9 should not declare compatibility with 1.12 until it is known to work with 1.12. That would require patching the General registry, which I don't mind doing.

penelopeysm: The downside of that is that the latest version of Turing won't be able to load on 1.12 at all since it requires Libtask 0.9. Which is also quite bad, argh.

mhauru: Yeah, not loading is bad. Also, you'd need to retrofix existing releases of Libtask, which could e.g. break existing environments that people already have in very surprising ways.

penelopeysm:
v glad to not have to do this

Issue #2684, opened 2025-10-13
Undocumented sampler arguments


mhauru:
For instance, the `nadapts` and `discard_adapt` arguments from [here](https://github.com/TuringLang/Turing.jl/blob/d8fbe78570a89ba3d36f07c1e62fa9f822233e1d/src/mcmc/hmc.jl#L95) seem to not be explained in any docstring or documentation page. There may be many other such arguments, just spotted these two.

PR #2692, opened 2025-10-20
added requested docstrings

AoifeHughes:
WRT: https://github.com/TuringLang/Turing.jl/issues/2684

### DIFF
--- a/src/mcmc/hmc.jl
+++ b/src/mcmc/hmc.jl
@@ -115,7 +115,9 @@ This method handles adaptation and warm-up for adaptive Hamiltonian samplers.
 
 - `initial_params`: Initial parameter values for sampling. See `DynamicPPL.initialstep` for details.
 
-Additional keyword arguments are passed to the underlying sampling implementation.
+Additional keyword arguments (e.g., `verbose`, `progress`, `chain_type`) are passed to the underlying
+sampling implementation. For more information on available options, see the
+[sampling options documentation](https://turinglang.org/docs/usage/sampling-options).
 
 # Note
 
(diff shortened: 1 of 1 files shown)


src/mcmc/hmc.jl
```diff
@@ -82,7 +82,46 @@ end
...
+  Defaults to `true`. When `true`, the adaptation samples are not included in the final results,
+  as they may not be from the target distribution whilst the sampler is still adapting.
+
+- `discard_initial::Int`: Number of initial samples to discard from the chain. Defaults to `-1`
+  (automatic). When `-1`, this becomes `nadapts` if `discard_adapt` is `true`, otherwise `0`.
+  Use this to manually specify how many initial samples to discard.
+
+- `initial_params`: Initial parameter values for sampling. See `DynamicPPL.initialstep` for details.
+
+Additional keyword arguments are passed to the underlying sampling implementation.
```
mhauru: Are all the other keyword arguments specific to HMC, or do they include generic ones joint to all samplers, like `verbose` or something?

Relatedly, may be worth linking to this in the docstring: https://turinglang.org/docs/usage/sampling-options Maybe already above?

src/mcmc/hmc.jl
```diff
@@ -175,6 +214,41 @@ function find_initial_params(
...
+    )
+
+Perform the initial step for Hamiltonian Monte Carlo sampling.
+
+This function initialises the Hamiltonian, finds a suitable step size (if not provided),
+and performs the first sampling step.
+
+# Keyword Arguments
+
+- `initial_params`: Initial parameter values to use for sampling. If `nothing` (the default),
```
mhauru: `initial_params` and `verbose` work the same way for all samplers, not just HMC, so I would not document them here, but rather refer to the docstring of the more generic `initialstep` function or method. That docstring may need writing too, but would make sense to postpone that after github.com/TuringLang/Turing.jl/pull/2676, because the way `initial_params` works is going to change quite drastically.

mhauru (changes requested):

PR #2699, opened 2025-10-22
Update variational inference interface to match `AdvancedVI@0.6`

Red-Portal:
This PR does the following:
* The interface to `AdvancedVI` is updated to match the new v0.6 version. 
* The way `vi` calls `optimize` is slightly changed so that algorithms that expect an unconstrained space can be used without being aware of `Bijectors`. In particular, the user is now expected to provide a `q <: Bijectors.TransformedDistribution`. `vi` will then unwrap the underlying unconstrained distribution, which is passed to `AdvancedVI`. Once the algorithm terminates, the result will be wrapped back and then returned. (See also [this PR](https://github.com/TuringLang/AdvancedVI.jl/pull/219) for a more detailed explanation.)
* The main file is renamed from `VariationalInference.jl` to `Variational.jl` to match the submodule name `Variational`.
* The deprecated v0.2-era interface is removed.


src/variational/Variational.jl
```diff
@@ -248,76 +251,61 @@ end
 """
     vi(
```
mhauru, line 285: Since the interface is changing, could you target this PR to the `breaking` branch and add an entry to HISTORY.md explaining to users what has changed and how they need to adapt?

src/variational/VariationalInference.jl
```diff
@@ -248,76 +251,61 @@ end
...
-        operator::AdvancedVI.AbstractOperator = AdvancedVI.ProximalLocationScaleEntropy(),
-        adtype::ADTypes.AbstractADType = Turing.DEFAULT_ADTYPE,
         kwargs...
     )
 
-Approximating the target `model` via variational inference by optimizing `objective` with the initialization `q`.
+Approximate the target `model` via the variational inference algorithm `algorithm` by starting from the initial variational approximation `q`.
 This is a thin wrapper around `AdvancedVI.optimize`.
+The default `algorithm` assumes `q` uses `AdvancedVI.MvLocationScale`, which can be constructed by invoking `q_fullrank_gaussian` or `q_meanfield_gaussian`.
+For other variational families, refer to `AdvancedVI` to determine the best algorithm and options.
```
mhauru: Could this link to the relevant docs?

src/Turing.jl
mhauru, line 120: Aqua is complaining that ADVI doesn't exist anymore.

test/variational/advi.jl
```diff
@@ -41,86 +45,62 @@ using Turing.Variational
...
-            AdvancedVI.ClipScale(),
-            AdvancedVI.DoG(),
-        ),
-        (
-            "ADVI with STL entropy",
-            AdvancedVI.RepGradELBO(10; entropy=AdvancedVI.StickingTheLandingEntropy()),
-            AdvancedVI.ClipScale(),
-            AdvancedVI.DoG(),
+            "KLMinRepGradDescent",
+            KLMinRepGradDescent(AutoReverseDiff(); operator, n_samples=10),
```
mhauru: Since the `adtype` variable is set, could these two use it too?

test/variational/vi.jl
```diff
@@ -10,12 +10,16 @@ using Distributions: Dirichlet, Normal
...
 using MCMCChains: Chains
 using Random
+using ReverseDiff
 using StableRNGs: StableRNG
 using Test: @test, @testset
 using Turing
 using Turing.Variational
 
 @testset "ADVI" begin
+    adtype = AutoReverseDiff()
```
mhauru, line 20: This is optional, but if the below tests aren't too slow, you could loop over different choices of `adtype`. You can find in `test/test_utils/ad_utils.jl` a variable called `adbackends`, that lists the ones we test against with MCMC.

src/variational/Variational.jl
Red-Portal: Yeah, forgot the drill for this; my bad. Done now!

src/variational/VariationalInference.jl
Red-Portal: Done!

src/Turing.jl
Red-Portal: Yeah that has to go.

test/variational/advi.jl
Red-Portal: Yeah let's do that

test/variational/vi.jl
Red-Portal: I think we can afford not to do that since the AD backends are already tested extensively upstream. (BTW, it seems `test/test_utils/ad_utils.jl` doesn't exist?)

mhauru: Whoops, sorry. That's where it used to live, can't recall now what we did with it. Anyway, happy with your conclusion that there's no need for this.

HISTORY.md
```diff
@@ -1,4 +1,81 @@
...
+**AdvancedVI 0.5**
+
+Turing.jl v0.42 updates `AdvancedVI.jl` compatibility to 0.5.
+Most of the changes introduced in `AdvancedVI.jl@0.5` are structural, with some changes spilling out into the interface.
+The summary of the changes below are the things that affect the end-users of Turing.
+For a more comprehensive list of changes, please refer to the [changelogs](https://github.com/TuringLang/AdvancedVI.jl/blob/main/HISTORY.md) in `AdvancedVI`.
+
+  - A new level of interface for defining different variational algorithms have been introduced in `AdvancedVI` v0.5. As a result, the method `Turing.vi` now receives a keyword argument `algorithm`. The object `algorithm <: AdvancedVI.AbstractVariationalAlgorithm` should now contain all the algorithm-specific configurations. Therefore, keyword arguments of `vi` that were algorithm-specific such as `objective`, `operator`, `averager` and so on, have been moved as fields of the relevant `<: AdvancedVI.AbstractVariationalAlgorithm` structs.
+  - The default hyperparameters of `DoG`and `DoWG` have been altered.
+  - The depricated `AdvancedVI@0.2`-era interface is now removed.
```
mhauru:
```suggestion
  - The deprecated `AdvancedVI@0.2`-era interface is now removed.
```

HISTORY.md
```diff
@@ -1,4 +1,81 @@
...
+## Initial step in MCMC sampling
+
+HMC and NUTS samplers no longer take an extra single step before starting the chain.
+This means that if you do not discard any samples at the start, the first sample will be the initial parameters (which may be user-provided).
+
+Note that if the initial sample is included, the corresponding sampler statistics will be `missing`.
+Due to a technical limitation of MCMCChains.jl, this causes all indexing into MCMCChains to return `Union{Float64, Missing}` or similar.
+If you want the old behaviour, you can discard the first sample (e.g. using `discard_initial=1`).
+
+# 0.4# 0.40.5
```
mhauru:
```suggestion
# 0.40.5
```

HISTORY.md
```diff
@@ -1,4 +1,81 @@
...
+## Breaking Changes
+
+**AdvancedVI 0.5**
+
+Turing.jl v0.42 updates `AdvancedVI.jl` compatibility to 0.5.
+Most of the changes introduced in `AdvancedVI.jl@0.5` are structural, with some changes spilling out into the interface.
+The summary of the changes below are the things that affect the end-users of Turing.
+For a more comprehensive list of changes, please refer to the [changelogs](https://github.com/TuringLang/AdvancedVI.jl/blob/main/HISTORY.md) in `AdvancedVI`.
+
+  - A new level of interface for defining different variational algorithms have been introduced in `AdvancedVI` v0.5. As a result, the method `Turing.vi` now receives a keyword argument `algorithm`. The object `algorithm <: AdvancedVI.AbstractVariationalAlgorithm` should now contain all the algorithm-specific configurations. Therefore, keyword arguments of `vi` that were algorithm-specific such as `objective`, `operator`, `averager` and so on, have been moved as fields of the relevant `<: AdvancedVI.AbstractVariationalAlgorithm` structs.
```
mhauru: Sorry to be picky, but could I ask for a code example of a typical use of `vi`? Something like

> If you previously did
> ```
> vi(mymodel, blahblah; objective=o, operator=or, average=ar)
> ```
> that would now be done instead with
> ```
> vi(mymodel, blahblah; algorithm=blah(o, or, ar)
> ```

I think your explanation already says this, but just trying to really hold the user by the hand through the breakage.

mhauru:
```suggestion
  - A new level of interface for defining different variational algorithms has been introduced in `AdvancedVI` v0.5. As a result, the function `Turing.vi` now receives a keyword argument `algorithm`. The object `algorithm <: AdvancedVI.AbstractVariationalAlgorithm` should now contain all the algorithm-specific configurations. Therefore, keyword arguments of `vi` that were algorithm-specific such as `objective`, `operator`, `averager` and so on, have been moved as fields of the relevant `<: AdvancedVI.AbstractVariationalAlgorithm` structs.
```

penelopeysm:
I'm actually sorry to butt in on this, but are there new features available in 0.5, or is it just an interface change?

Red-Portal:
@penelopeysm No new features.

penelopeysm:
Okay. The reason why I asked is because this is an interface change for Turing and that is kind of annoying for end users if they don't get any benefit. I wonder in that case if it is worth doing either:

1. Not bumping the compat in Turing, waiting for the next release of AdvancedVI that has new features; or

2. Mapping the existing Turing interface to the new AdvancedVI interface, so that it's not a breaking change. I don't know how possible this is.

Other than that, I second everything that @mhauru brought up here, so I won't also review the details of this.

HISTORY.md
Red-Portal: This is a bit tricky because the interface now does not fully expose all of the controls that were previously available. Let me think if there's something I can do.

Red-Portal:
@penelopeysm That's reasonable. I think we could go with the following for now:

> Not bumping the compat in Turing, waiting for the next release of AdvancedVI that has new features;

Red-Portal:
Hi @penelopeysm @mhauru @yebai , I've updated the code to match the new v0.6 version. Let me know if you have any comments!

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
-        adtype::ADTypes.AbstractADType = Turing.DEFAULT_ADTYPE,
         kwargs...
     )
 
-Approximating the target `model` via variational inference by optimizing `objective` with the initialization `q`.
+Approximate the target `model` via the variational inference algorithm `algorithm` by starting from the initial variational approximation `q`.
 This is a thin wrapper around `AdvancedVI.optimize`.
 
+If the chosen variational inference algorithm operates in an unconstrained space, then the provided initial variational approximation `q` must be a `Bijectors.TransformedDistribution` of an unconstrained distribution.
+For example, the initialization supplied by  `q_meanfield_gaussian`,`q_fullrank_gaussian`, `q_locationscale`.
```
mhauru, line 302: Just trying to understand the meaning of this: What would be choices of `q` that would _not_ be valid for an unconstrained-space-requiring algorithm?

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
+
+The default `algorithm`, `KLMinRepGradProxDescent` ([relevant docs](https://turinglang.org/AdvancedVI.jl/dev/klminrepgradproxdescent/)), assumes `q` uses `AdvancedVI.MvLocationScale`, which can be constructed by invoking `q_fullrank_gaussian` or `q_meanfield_gaussian`.
+For other variational families, refer the documentation of `AdvancedVI` to determine the best algorithm and other options.
```
mhauru:
```suggestion
For other variational families, refer to the documentation of `AdvancedVI` to determine the best algorithm and other options.
```

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
+
 # Arguments
 - `model`: The target `DynamicPPL.Model`.
 - `q`: The initial variational approximation.
-- `n_iterations`: Number of optimization steps.
+- `max_iter`: Maximum number of steps.
 
 # Keyword Arguments
-- `objective`: Variational objective to be optimized.
+- `adtype`: Automatic differentiation backend to be applied to the log-density. The default value for `algorithm` also uses this backend for differentiation the variational objective.
```
mhauru:
```suggestion
- `adtype`: Automatic differentiation backend to be applied to the log-density. The default value for `algorithm` also uses this backend for differenting the variational objective.
```

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
+- `algorithm`: Variational inference algorithm.
```
mhauru:
```suggestion
- `algorithm`: Variational inference algorithm. The default is `KLMinRepGradProxDescent`, please refer to [AdvancedVI docs](https://turinglang.org/AdvancedVI.jl/stable/) for all the options.
```

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
+For example, the initialization supplied by  `q_meanfield_gaussian`,`q_fullrank_gaussian`, `q_locationscale`.
+
+The default `algorithm`, `KLMinRepGradProxDescent` ([relevant docs](https://turinglang.org/AdvancedVI.jl/dev/klminrepgradproxdescent/)), assumes `q` uses `AdvancedVI.MvLocationScale`, which can be constructed by invoking `q_fullrank_gaussian` or `q_meanfield_gaussian`.
+For other variational families, refer the documentation of `AdvancedVI` to determine the best algorithm and other options.
+
 # Arguments
 - `model`: The target `DynamicPPL.Model`.
 - `q`: The initial variational approximation.
-- `n_iterations`: Number of optimization steps.
+- `max_iter`: Maximum number of steps.
```
mhauru, line 312:
```suggestion
- Any additional arguments are passed on to `AdvancedVI.optimize`.

```

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
 
 # Keyword Arguments
-- `objective`: Variational objective to be optimized.
+- `adtype`: Automatic differentiation backend to be applied to the log-density. The default value for `algorithm` also uses this backend for differentiation the variational objective.
+- `algorithm`: Variational inference algorithm.
 - `show_progress`: Whether to show the progress bar.
```
mhauru, line 316:
```suggestion
- `show_progress`: Whether to show the progress bar.
- Any additional keyword arguments are passed on to `AdvancedVI.optimize`.

```

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
-    n_iterations::Int;
-    objective=AdvancedVI.RepGradELBO(
-        10; entropy=AdvancedVI.ClosedFormEntropyZeroGradient()
+    max_iter::Int,
+    args...;
+    adtype::ADTypes.AbstractADType=DEFAULT_ADTYPE,
+    algorithm::AdvancedVI.AbstractVariationalAlgorithm=KLMinRepGradProxDescent(
+        adtype; n_samples=10
     ),
+    unconstrained::Bool=requires_unconstrained_space(algorithm),
```
mhauru, line 338: Can this one have a docstring line as well please.

src/variational/Variational.jl
```diff
@@ -248,76 +284,82 @@ end
...
-        q,
-        n_iterations;
-        show_progress=show_progress,
-        adtype,
-        optimizer,
-        averager,
-        operator,
-        kwargs...,
+    prob, q, trans = if unconstrained
+        @assert q isa Bijectors.TransformedDistribution "The algorithm $(algorithm) operates in an unconstrained space. Therefore, the initial variational approximation is expected to be a Bijectors.TransformedDistribution of an unconstrained distribution."
```
mhauru: Could this be replaced by assert be replaced by an `if-throw`? By the [assert docstring](https://docs.julialang.org/en/v1/base/base/#Base.@assert), asserts should only be used for debugging.

test/variational/advi.jl
```diff
@@ -10,12 +10,16 @@ using Distributions: Dirichlet, Normal
 using LinearAlgebra
 using MCMCChains: Chains
 using Random
+using ReverseDiff
 using StableRNGs: StableRNG
 using Test: @test, @testset
 using Turing
 using Turing.Variational
 
-@testset "ADVI" begin
```
mhauru, line 18: I often find it helpful that all the tests in a file are wrapped in a single big testset. Makes the output easier to read. Did you have a particular reason to prefer otherwise here?

HISTORY.md
```diff
@@ -18,6 +18,75 @@ As long as the above functions are defined correctly, Turing will be able to use
...
+
+```julia
+vi(model, q, n_iters; algorithm=KLMinRepGradProxDescent(adtype; n_samples=10))
+```
+
+Additionally,
+
+  - The default hyperparameters of `DoG`and `DoWG` have been altered.
+  - The deprecated `AdvancedVI@0.2`-era interface is now removed.
+  - `estimate_objective` now returns the value to be minimized by the optimization algorithm. For example, for ELBO maximization algorithms, `estimate_objective` will return the *negative ELBO*. This is breaking change from the previous behavior where the ELBO was returns.
```
mhauru:
```suggestion
  - `estimate_objective` now always returns the value to be minimized by the optimization algorithm. For example, for ELBO maximization algorithms, `estimate_objective` will return the *negative ELBO*. This is breaking change from the previous behavior where the ELBO was returned.
```

HISTORY.md
```diff
@@ -18,6 +18,75 @@ As long as the above functions are defined correctly, Turing will be able to use
...
+    operator=AdvancedVI.ProximalLocationScaleEntropy(),
+)
+```
+
+is now
+
+```julia
+vi(model, q, n_iters; algorithm=KLMinRepGradProxDescent(adtype; n_samples=10))
+```
+
```
mhauru, line 70: The return value of `vi` has also substantially changed right? Unless I missed it being mentioned somewhere else, I think it should get quite a prominent mention here.

HISTORY.md
mhauru: I think `estimate_objective` isn't exported by Turing. Is there some other way in which this change is visible to Turing users?

HISTORY.md
```diff
@@ -18,6 +18,75 @@ As long as the above functions are defined correctly, Turing will be able to use
...
+  - When using algorithms that expect to operate in unconstrained spaces, the user is now explicitly expected to provide a `Bijectors.TransformedDistribution` wrapping an unconstrained distribution. (Refer to the docstring of `vi`.)
+
+### New Features
+
+`AdvancedVI@0.6` adds numerous new features including the following new VI algorithms:
+
+  - `KLMinWassFwdBwd`: Also known as "Wasserstein variational inference," this algorithm minimizes the KL divergence under the Wasserstein-2 metric.
+  - `KLMinNaturalGradDescent`: This algorithm, also known as "online variational Newton," is the canonical "black-box" natural gradient variational inference algorithm, which minimizes the KL divergence via mirror descent under the KL divergence as the Bregman divergence.
+  - `KLMinSqrtNaturalGradDescent`: This is a recent variant of `KLMinNaturalGradDescent` that operates in the Cholesky-factor parameterization of Gaussians instead of precision matrices.
+  - `FisherMinBatchMatch`: This algorithm called "batch-and-match," minimizes the variation of the 2nd order fisher divergence via a proximal point-type algorithm.
```
mhauru:
```suggestion
  - `FisherMinBatchMatch`: This algorithm called "batch-and-match," minimizes the variation of the 2nd order Fisher divergence via a proximal point-type algorithm.
```

mhauru (changes requested):
Hi @Red-Portal, thanks for this. The code looks good, just some comments here and there on documentation, since I think it's important to get it right for such a big change.

src/variational/Variational.jl
Red-Portal: Something like `LogNormal`, `Dirichlet`, and of course all discrete support distributions.

src/variational/Variational.jl
Red-Portal: Yes, should indeed have done that. Done now!

src/variational/Variational.jl
Red-Portal: Oh! Okay, that's good to know. I have relaxed this to an exception and added a test.

test/variational/advi.jl
Red-Portal: It's already wrapped in `runtests.jl` by `variational_algorithms` and we're testing essentially all of the functionalities in the `Variational` module. So I think an additional `@testset` wrap is redundant at the moment.

HISTORY.md
Red-Portal: Not in an `export` sense, but the tutorial  [here](https://turinglang.org/docs/tutorials/variational-inference/) brought it up, so I thought it would be nice to state explicitly.

HISTORY.md
Red-Portal: Oooh that's a good point. Let me state that in the `HISTORY`.

src/variational/Variational.jl
mhauru: Oh I see, I didn't realise you even could use those, I was only thinking of the functions provided by `vi` like `q_something_gaussian`.

test/variational/vi.jl
```diff
@@ -48,6 +48,9 @@ begin
             q, _, _ = vi(gdemo_default, q0, 100; show_progress=Turing.PROGRESS[], adtype)
             c1 = rand(q, 10)
         end
+        @test_throws "unconstrained" begin
```
mhauru, line 51: Seems from CI that `@test_throws` is missing from imports.

Red-Portal: Oops my bad

Red-Portal:
@mhauru Thank you so much for the review! Very excited that this finally made it!

penelopeysm:
Merged as per private communication with @Red-Portal :-)

PR #2704, opened 2025-10-24
Move external sampler interface to AbstractMCMC

penelopeysm:
Following on from:

 - https://github.com/TuringLang/AbstractMCMC.jl/pull/182 Adding the interface function `AbstractMCMC.getstats`
 - https://github.com/TuringLang/AdvancedHMC.jl/pull/471 and https://github.com/TuringLang/AdvancedMH.jl/pull/119 Implementing them for AdvancedHMC and AdvancedMH

this PR changes Turing's external sampler interface to exclusively use AbstractMCMC functions. **With this PR, anyone who defines an external sampler will only need to depend on AbstractMCMC (which they presumably already do, because it is a sampler) and LogDensityProblems (which is already a dep of AbstractMCMC).** No need for a Turing extension.

To be precise, it makes the following changes:

 - Previously where one had to define `Turing.Inference.getparams`, now one has to define `AbstractMCMC.getparams`.
 - Previously there was no way to include sampler stats in the resulting chain, now one can define `AbstractMCMC.getstats`.
 - The default for `Turing.Inference.isgibbscomponent` is changed to `true`, so that external sampler packages don't need to override it (unless absolutely necessary).

As an example implementation, this PR contains a test mock (note how it doesn't require a Turing dep):

https://github.com/TuringLang/Turing.jl/blob/06752c41a97cd0dc37bb8700ab7a2d06f50f4f76/test/mcmc/external_sampler.jl#L20-L74

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -40,9 +40,9 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [compat]
 ADTypes = "1"
-AbstractMCMC = "5"
+AbstractMCMC = "5.9"
 AbstractPPL = "0.11, 0.12, 0.13"
-AdvancedMH = "0.6, 0.7, 0.8"
+AdvancedMH = "0.8.9"
 AdvancedPS = "0.7"
 AdvancedVI = "0.4"
 Aqua = "0.8"


src/mcmc/external_sampler.jl
```diff
@@ -1,5 +1,5 @@
 """
-    ExternalSampler{S<:AbstractSampler,AD<:ADTypes.AbstractADType,Unconstrained}
+    ExternalSampler{Unconstrained,S<:AbstractSampler,AD<:ADTypes.AbstractADType}
```
penelopeysm, line 2: Moved this type parameter earlier so that we can dispatch on it more easily.

src/mcmc/external_sampler.jl
```diff
@@ -14,45 +14,59 @@ $(TYPEDFIELDS)
...
+   call `sample(model, MySampler(), N)`.
 
-2. Implement a generic `AbstractMCMC.step` method for `AbstractMCMC.LogDensityModel`. This
-   struct wraps an object that obeys the LogDensityProblems.jl interface, so your `step`
+2. Implement a generic `AbstractMCMC.step` method for `AbstractMCMC.LogDensityModel` (the
+   same signature as above except that `model::AbstractMCMC.LogDensityModel`). This struct
+   wraps an object that obeys the LogDensityProblems.jl interface, so your `step`
    implementation does not need to know anything about Turing.jl or DynamicPPL.jl. To use
    this with Turing.jl, you will need to wrap your sampler: `sample(model,
    externalsampler(MySampler()), N)`.
```
penelopeysm, line 28: I'll write proper docs about this separately (in the docs repo).

PR #2707, opened 2025-11-01
Test on 1.12

penelopeysm:
Now that Libtask works on 1.12, I think we can and should run CI on it.

Mooncake can be disabled for the time being (it will still be tested in on 1.10).

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -35,23 +35,20 @@ jobs:
           - name: "everything else"
             args: "--skip mcmc/gibbs.jl mcmc/Inference.jl ad.jl"
         runner:
-          # TODO(mhauru) All the ones below that run on 1.11 should actually be run on 1.
-          # The current setup is a temporary arrangement to deal with issues where Mooncake
-          # and Libtask are broken on 1.12.
           # Default
-          - version: '1.11'
+          - version: '1'
             os: ubuntu-latest
             num_threads: 1
           # Multithreaded
-          - version: '1.11'
+          - version: '1'
             os: ubuntu-latest
             num_threads: 2
           # Windows
-          - version: '1.11'
+          - version: '1'
             os: windows-latest
             num_threads: 1
           # macOS
-          - version: '1.11'
+          - version: '1'
             os: macos-latest
             num_threads: 1
           # Minimum supported Julia version


test/ad.jl
```diff
@@ -11,7 +11,7 @@ using ..Models: gdemo_default
 import ForwardDiff, ReverseDiff
 
 # Skip Mooncake on 1.12 as it is not compatible yet
-const INCLUDE_MOONCAKE = VERSION >= v"1.12"
+const INCLUDE_MOONCAKE = VERSION < v"1.12"
```
penelopeysm, line 14: Silly mistake I made previously.

penelopeysm:
@mhauru Do you know if the test failure is expected?

mhauru:
Nope, haven't seen that before. I hope this isn't indeterministic to reproduce. Looks like MCMCThreads gets Libtask tasks running on multiple threads, and two happened to write to the global MistyClosures cache within Libtask at the same time. Maybe need to do some locks or `Atomic` stuff with that global cache.

mhauru:
Wanna try to make a little MWE by creating tasks on multiple threads, or shall I? My (sad) prediction is that the CI failure will go away if you rerun it and get a bit lucky.

penelopeysm:
I'm struggling to make a Libtask-only MWE, this doesn't work:

```julia
using Libtask
function task(i)
    # define a new method, ensuring that Libtask has to construct
    # a new MistyClosure and add it to the cache
    function f(::Val{j}) where {j}
        produce(j)
        return nothing
    end
    # construct a TapedTask for it
    return TapedTask(nothing, f, Val(i))
end
fetch.([Threads.@spawn task(i) for i in 1:200])
```

This is doing concurrent writes to `Libtask.mc_cache`, but it's not picked up by Julia's base library because only certain code paths trigger the detection of concurrent writes (https://github.com/JuliaLang/julia/blob/ba1e628ee49351af0b704afd2b2903d253bd3564/base/dict.jl#L182).

In fact even this doesn't trigger it:

```julia
d = Dict{Int,Int}()
empty!(d)
function seti(i)
    d[i] = i
end
fetch.([Threads.@spawn seti(i) for i in 1:1000000])
```

Will continue trying.

penelopeysm:
Patching that `Base.rehash!` method by adding a 0.1 second sleep inside allows me to replicate the error with just Libtask. Obviously, that doesn't constitute an MWE. But I think the important thing is to have some kind of function with a complicated enough cache key, such that the middle bit of `rehash!` takes long enough to cause a race condition.

penelopeysm:
https://github.com/TuringLang/Libtask.jl/issues/202

penelopeysm:
So the real underlying problem (concurrent writes to Dicts) can occur on both 1.11 and 1.12, which of course leads to the question of why does this error on 1.12 but not 1.11? I assume that there's some performance regression on `hash` in 1.12 that causes this error to manifest, and we were lucky enough that 1.11 was fast enough to not error, but 1.12 wasn't. Obviously this should still be fixed in Libtask, but just interesting to note.

mhauru:
New Libtask version should fix this, rerunning the failing CI.

penelopeysm:
Gibbs on Windows takes a full hour, meh.

PR #2708, opened 2025-11-03
Remove Optim.jl interface + minor tidying up of src/optimisation/Optimisation

### LINKED ISSUE #2635: Remove Optim.jl interface
shouldn't really be needed any more.

penelopeysm:
notes in comments.

Further refactoring will happen, but in separate PRs.

Closes #2635.

### DIFF
--- a/src/optimisation/Optimisation.jl
+++ b/src/optimisation/Optimisation.jl
@@ -503,7 +503,9 @@ function estimate_mode(
         vi = DynamicPPL.link(vi, model)
     end
 
-    ldf = DynamicPPL.LogDensityFunction(model, getlogdensity, vi; adtype=adtype)
+    # Note that we don't need adtype here, because it's specified inside the
+    # OptimizationProblem
+    ldf = DynamicPPL.LogDensityFunction(model, getlogdensity, vi)
     log_density = OptimLogDensity(ldf)
 
     prob = Optimization.OptimizationProblem(log_density, adtype, constraints)


test/ext/OptimInterface.jl
penelopeysm, line 1: I checked and the base optimisation tests are a superset of these tests, so no need to incorporate any of these deleted tests into the main suite.

src/optimisation/Optimisation.jl
```diff
@@ -19,136 +19,64 @@ using Statistics: Statistics
...
-        adtype::ADTypes.AbstractADType=Turing.DEFAULT_ADTYPE,
-    )
-        # No varinfo
-        return OptimLogDensity(
-            model,
-            getlogdensity,
-            DynamicPPL.ldf_default_varinfo(model, getlogdensity);
-            adtype=adtype,
-        )
-    end
```
penelopeysm, line 111: this is just duplicating LogDensityFunction code. I thought it simpler to just construct the LDF and then wrap it.

HISTORY.md
```diff
@@ -1,5 +1,14 @@
 # 0.42.0
 
+## Optimisation interface
+
+The Optim.jl interface has been removed (so you cannot call `Optim.optimize` directly on Turing models).
```
yebai, line 23: I suggest that we define our own `optimize` function on top of `DynamicPPL.LogDensityFunction` to avoid the breaking change. The `optimize` function provides a natural alternative to `AbstractMCMC.sample`.

penelopeysm: I don't really agree:

1. What is the purpose of avoiding the breaking change? The functionality is already present in `maximum_likelihood(model)` or `maximum_a_posteriori(model)` so this is just duplicate functionality.

2. A `Turing.optimize` function would still be a breaking change because this is not the same as `Optim.optimize`.

yebai: Thanks, Penny, for clarifying. 

My main point is not breaking change, instead `optimise` mirrors the `Turing.sample` API well, and provides a generic interface for optimisation, and VI algorithms. Sorry for the confusion. Though I agree we ought to remove the `Optim` interface and define our own `optimize` method.

penelopeysm: Sure, in that case we could start by renaming `estimate_mode` to `optimize`, and exporting it. That would give us out of the box

```julia
optimize(model, MLE(); kwargs...)
optimize(model, MAP(); kwargs...)
```

and then the VI arguments could be bundled into a struct that would be passed as the second argument.

src/optimisation/Optimisation.jl
```diff
@@ -568,7 +503,10 @@ function estimate_mode(
         vi = DynamicPPL.link(vi, model)
     end
 
-    log_density = OptimLogDensity(model, getlogdensity, vi)
+    # Note that we don't need adtype here, because it's specified inside the
+    # OptimizationProblem
+    ldf = DynamicPPL.LogDensityFunction(model, getlogdensity, vi)
+    log_density = OptimLogDensity(ldf)
 
     prob = Optimization.OptimizationProblem(log_density, adtype, constraints)
```
penelopeysm, line 511: I'm actually unsure if this is 'best practice', although I've kept it this way since that's how it was written prior to this PR. What this effectively does is to tell Optimization.jl to use adtype to differentiate through the LogDensityFunction. It seems to me that it would be more appropriate to construct an LDF with the appropriate adtype, and then tell OptimizationProblem to use the implementation of `logdensity_and_gradient` when it needs a gradient.

src/optimisation/Optimisation.jl
```diff
@@ -19,137 +19,65 @@ using StatsAPI: StatsAPI
...
-export maximum_a_posteriori, maximum_likelihood
-# The MAP and MLE exports are only needed for the Optim.jl interface.
-export MAP, MLE
+export maximum_a_posteriori, maximum_likelihood, MAP, MLE
 
 """
     ModeEstimator
 
 An abstract type to mark whether mode estimation is to be done with maximum a posteriori
-(MAP) or maximum likelihood estimation (MLE). This is only needed for the Optim.jl interface.
```
mhauru, line 30: I'm a bit confused by the comment that is being removed. Wouldn't we want these types to exist regardless? Or was the intention of the comment to say that we should move to using distinct functions for MLE and MAP and not have this type? I appreciate you didn't write the comment that is being removed, and for extra irony, I may have written it.

src/optimisation/Optimisation.jl
mhauru: That does feel like a better approach. Looks like OptimizationFunction does support both ways of doing it, though I don't see a way to make use of `logdensity_and_gradient` also computing the log density: https://docs.sciml.ai/Optimization/v4.8/API/optimization_function/#optfunction

src/optimisation/Optimisation.jl
penelopeysm: Ah yes, I was confused too. I actually started this PR by deleting them! 😄 then I realised we can't delete them. No big deal, this PR will fix the confusion...

src/optimisation/Optimisation.jl
penelopeysm: Hmmmm, but if it then wants something like Hessians (for, er, ... full Newton's method?) then we can't provide that, unless we rewrite OptimLogDensity to also cache hessian preparation and stuff. Maybe best to just let Optimization do its thing then.

penelopeysm: That's very slightly disappointing, but I suppose also it's probably quite unlikely that this would be a performance pain point (compared to the actual cost of evaluating the gradient), so happy to drop it.

src/optimisation/Optimisation.jl
penelopeysm: (I think in general these types are good to have, I'm quite happy to have these kind of enums)

Issue #2634, opened 2025-07-27
Optimisation interface reworking


penelopeysm:
There are a number of improvements that can/should be made to Turing's optimisation interface.

This issue is mainly to collect sub-issues on this topic.

penelopeysm:
Closed by #2747.

PR #2710, opened 2025-11-05
Include parameter dictionary in optimisation return value

penelopeysm:
The return value of a point optimisation currently includes an array of parameters. Arrays bad. `Dict{VarName}` good. (In particular, it currently returns a `NamedArray` which is a really weird data structure!)

This is not meant to be a permanent fix: `ModeResult` and its associated interface IMO needs a larger overhaul. (There are a few other open issues on the optimisation interface.) But getting this fix in will at least let it be more composable with the rest of Turing, while we figure out how to fix the rest. (See changelog for examples of what I mean by composable)

I reckon it's fair to not make this a breaking change because all of `ModeResult`'s existing fields still behave the same way, the only difference is that there is an extra field. So the only thing it would break is if people were doing something bizarre like `fieldnames(...)`. Also, technically, `ModeResult` isn't exported (although I think it's effectively public since it is the return value of an exported function).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
 # 0.41.1
 
-The `ModeResult` struct returned by `maximum_a_posteriori` and `maximum_likelihood` now has an additional field, `params`, which contains an `OrderedDict{VarName}` of the parameters found.
-This makes it easier to use the parameters in downstream code, e.g. when specifying initial parameters for MCMC sampling, or when using functions such as `returned`.
+The `ModeResult` struct returned by `maximum_a_posteriori` and `maximum_likelihood` can now be wrapped in `InitFromParams()`.
+This makes it easier to use the parameters in downstream code, e.g. when specifying initial parameters for MCMC sampling.
 For example:
 
 ```julia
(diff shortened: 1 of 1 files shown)


src/optimisation/Optimisation.jl
```diff
@@ -154,13 +155,22 @@ end
         V<:NamedArrays.NamedArray,
         M<:NamedArrays.NamedArray,
```
sunxd3, line 156: doesn't look consistent, probably already wrong and for a while, good to fix

sunxd3 (approved):
looks good

HISTORY.md
```diff
@@ -1,3 +1,23 @@
...
+For example:
+
+```julia
+@model function f()
+    # ...
+end
+model = f()
+opt_result = maximum_a_posteriori(model)
+
+sample(model, NUTS(), 1000; initial_params=InitFromParams(opt_result.params))
```
sunxd3: small nit is that this would somewhat makes `params` fieldname a public interface, with the warrant just following the code example, not big deal, but I thought it might be worth mentioning

penelopeysm: Ah yeah, I guess we could overload `InitFromParams(opt_result)` instead so that you don't need to access the params field. I thought we could do that in a separate release, but doing that would avoid having to document the field, so let's bundle it in

src/optimisation/Optimisation.jl
sunxd3: @penelopeysm sorry for review twice 😓  

bump this in case you missed

penelopeysm: Oh, I agree the NamedArray stuff should be removed, but that would be breaking, right? I was thinking of leaving it to another time. https://github.com/TuringLang/Turing.jl/issues/2634

penelopeysm: Also, don't ever apologise for double-reviewing!

PR #2713, opened 2025-11-11
Use FastLDF and removal of NodeTrait

penelopeysm:
I only changed HMC/NUTS here to use FastLDF for now.

I think the easier thing is if we replace LogDensityFunction in DynamicPPL, then we don't need to change it here.


mhauru:
> I think the easier thing is if we replace LogDensityFunction in DynamicPPL, then we don't need to change it here.

Agreed.

Issue #2642, opened 2025-08-01
All VarInfos are informative, but some VarInfos are more informative than others


penelopeysm:
While editing the sampler code in Turing, there's a common pattern where:

- we have a model + varinfo and construct a LogDensityFunction from it
- the sampler takes the LDF, does things, and gives us back a vector of parameters + a single logp

After that we take the returned parameters and construct a transition and a state with it. These two things are slightly different:

- **Transition:** we stick the vector of parameters back in the varinfo and then re-evaluate the model (in the `Turing.Inference.Transition` constructor)
- **State**: we stick the vector of parameters back in the varinfo and pass it on to the next step in MCMC.

In the transition case, it's clear that the varinfo does not need to have the correct logp because it will be re-evaluated anyway.

In the state case, it depends on whether varinfo's logp is accessed on the next iteration. Some samplers do (MH). Some don't (HMC).

Finally, there are some cases where we construct LDF from a varinfo. In such a case, the parameters and logp in the varinfo are completely useless; the varinfo is purely there for structure.

This creates a case where we're passing varinfos around, some have the wrong logp, and some have the wrong values, and it's quite difficult to reason about **when a varinfo really needs to have the right values and logp**.

I would love if we could make this clearer with some new types and/or wrappers. For example, consider the case where you do care about the parameters but not logp. So instead of carrying around a VarInfo with correct parameters set inside it, we should carry a struct that contains both a VarInfo (with wrong parameters) and the right parameters, and then call `get_vi_with_correct_params()` whenever we need it:

```julia
struct VIWithParams
    vi::AbstractVarInfo
    params::AbstractVector
end

get_vi_with_correct_params(vwp::VIWithParams) = DynamicPPL.unflatten(vwp.vi, vwp.params)
```

In this case `vi` only exists to provide structure, and this can be documented nicely.

But also sometimes the only thing that the next step does is to just re-extract the params with `vi[:]`. In such a case, we could simply use `vwp.params` instead of going via `unflatten` + `vi[:]` which could potentially cause issues (https://github.com/TuringLang/DynamicPPL.jl/issues/1001).

penelopeysm:
My current thinking is that we should store a LogDensityFunction (for which it is known that the parameters + logp values are bogus -- this is at least documented in DynamicPPL) plus a set of parameters if needed.

The proposed changes to make `sample()` work with LDF will probably help with this because then the varinfo structure doesn't need to be passed around as a separate argument, it can just exist in the LDF.

[...]

PR #2715, opened 2025-11-14
Update for DynamicPPL 0.39

penelopeysm:
The main change in DPPL 0.39 is OnlyAccsVarInfo and faster evaluation.

This PR uses fast evaluation in MCMC sampling where it can. MCMC sampling mostly works as can be seen from the tests.

## Of note

- MCMCChains no longer stores an overall `chain.logevidence` field

The reason is because we now use [the `bundle_samples` method in DynamicPPL](https://github.com/TuringLang/DynamicPPL.jl/blob/a6d56a2b9074d9da27eea4a6e4a2ab9a3013913f/ext/DynamicPPLMCMCChainsExt.jl#L121-L156), which has no way of reliably determining the log-evidence from the transition. If we wanted to fix this, we would have to add a `getlogevidence` function in AbstractMCMC.

I personally don't consider this a problem. The reason why log-evidence used to be stored was because chains did not provide enough information for people to calculate this themselves (specifically, chains only stored logp, not loglikelihood). Now that `chn[:likelihood]` contains the likelihood, it should be ok for people to calculate this themselves.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -39,6 +39,8 @@ You do *not* need to do this if:
   - your model has other kinds of parallelism but does not include tilde-statements inside;
   - or you are using `MCMCThreads()` or `MCMCDistributed()` to sample multiple chains in parallel, but your model itself does not use parallelism.
 
+If your model does include parallelised tilde-statements or `@addlogprob!` calls, and you evaluate it/sample from it without setting `setthreadsafe(model, true)`, then you may get statistically incorrect results without any warnings or errors.
+
 ### Faster performance
 
 Many operations in DynamicPPL have been substantially sped up.


src/mcmc/Inference.jl
```diff
@@ -115,312 +114,6 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
     return log(rand()) + logp_current ≤ logp_proposal + log_proposal_ratio
 end
 
-######################
-# Default Transition #
-######################
```
penelopeysm, line 120: All this stuff has basically been upstreamed to DynamicPPL and/or AbstractMCMC.

src/mcmc/prior.jl
```diff
@@ -12,14 +12,12 @@ function AbstractMCMC.step(
...
-    _, vi = DynamicPPL.init!!(model, vi, DynamicPPL.InitFromPrior())
-    return Transition(model, vi, nothing; reevaluate=false), nothing
+    accs = DynamicPPL.AccumulatorTuple((
+        DynamicPPL.ValuesAsInModelAccumulator(true),
+        DynamicPPL.LogPriorAccumulator(),
+        DynamicPPL.LogLikelihoodAccumulator(),
+    ))
+    vi = DynamicPPL.OnlyAccsVarInfo(accs)
+    _, vi = DynamicPPL.init!!(rng, model, vi, DynamicPPL.InitFromPrior())
+    return DynamicPPL.ParamsWithStats(vi), nothing
```
penelopeysm, line 22: Actually quite neat that the whole Prior sampler is just defined with DynamicPPL stuff now.

ext/TuringDynamicHMCExt.jl
penelopeysm, line 1: As has been the case in past PRs of this sort, this file provides a gentle introduction of the kinds of changes being made.

Generally, the current status is that: MCMC states often bundle a `varinfo`, not for the purpose of actually being an accurate varinfo, but more as a 'home' to unflatten a vector of parameters into. (See https://github.com/TuringLang/Turing.jl/issues/2642) The logp is usually not updated, because the only thing that's needed is for the next iteration to do `vi[:]`.

This PR generally attempts to remove these varinfos from states, and only ever store the parameter vector + the LDF. Often the only reason why we carried around varinfo was so that we could re-evaluate with ValuesAsInModelAcc. However, because ParamsWithStats now has [a method that takes the vector + LDF and returns the values-as-in-model](https://github.com/TuringLang/DynamicPPL.jl/blob/54ae7e30df1cfaa1f69202c8637d58afce55134d/src/chains.jl#L111-L130), we can use that without needing a varinfo now.

... That's the ideal, at least ...

The reality is that most samplers still need to carry around a varinfo, specifically so that samplers can be used inside Gibbs. (DynamicHMCExt doesn't need to, because it's not 'Gibbs-enabled'.) This suggests that a potential, and immediate, way of decoupling varinfo from the individual samplers would be to have Gibbs handle this extra varinfo overhead (i.e. make gibbs store a `(varinfo, state)` tuple, rather than just `state`).

That's probably one for the (near-ish) future. For now, at least the scope of the varinfo argument has been reduced by quite a bit: it's no longer used in the actual `AbstractMCMC.step` implementations of most samplers.

src/mcmc/emcee.jl
```diff
@@ -110,60 +105,14 @@ function AbstractMCMC.bundle_samples(
...
-    # Set up the info tuple.
-    info = (varname_to_symbol=OrderedDict(zip(varnames, varnames_symbol)),)
-    if save_state
-        info = merge(info, (model=model, sampler=spl, samplerstate=state))
+    n_walkers = _get_n_walkers(spl)
+    chains = map(1:n_walkers) do i
+        this_walker_samples = [s[i] for s in samples]
+        AbstractMCMC.bundle_samples(
+            this_walker_samples, model, spl, state, chain_type; kwargs...
+        )
```
penelopeysm, line 115: This is probably more inefficient than the old code, but I am not particularly fussed since chain construction isn't a bottleneck, and it's also way cleaner.

src/mcmc/mh.jl
```diff
@@ -223,8 +223,31 @@ function LogDensityProblems.logdensity(f::LogDensityFunction, x::NamedTuple)
...
+            DynamicPPL.setindex!(vi, vals, vns[1])
+        else
+            # Spread the values across the variables
+            length(vns) == length(vals) || error("Unequal number of variables and values")
+            for (vn, val) in zip(vns, vals)
+                DynamicPPL.setindex!(vi, val, vn)
+            end
+        end
+    end
 end
```
penelopeysm, line 251: This function is exactly the same as the one that's 'deleted' above. GitHub diffs being weird, sorry.

src/optimisation/Optimisation.jl
penelopeysm, line 1: unfortunately I had to do some surgery to the optimisation interface. I would have preferred to leave it for another time but the optim interface frequently assumed that LDF carried a varinfo with it.

penelopeysm:
I believe there is still a Mooncake failure on 1.12 with ADTypeCheckContext, but otherwise everything on CI should pass, unless I messed something up terribly. My suspicion is that it's a Mooncake issue, not Turing; however I'll only look into this later.

Edit: Confirmed locally, the test passes on 1.11 and fails on 1.12. https://github.com/chalk-lab/Mooncake.jl/issues/871

penelopeysm:
Still needs a changelog (also more bullet points for the changelog are welcome), but the code can be reviewed :)

src/mcmc/gibbs_conditional.jl
```diff
@@ -104,17 +104,17 @@ function build_variable_dict(model::DynamicPPL.Model)
...
-        get_global_varinfo(context)
-    elseif DynamicPPL.NodeTrait(context) isa DynamicPPL.IsParent
-        get_gibbs_global_varinfo(DynamicPPL.childcontext(context))
-    else
-        msg = """No GibbsContext found in context stack. Are you trying to use \
-            GibbsConditional outside of Gibbs?
-            """
-        throw(ArgumentError(msg))
-    end
+function get_gibbs_global_varinfo(context::GibbsContext)
```
mhauru, line 107: Any particular reason to this change other than code style?

test/mcmc/is.jl
```diff
@@ -64,7 +62,8 @@ using Turing
         chains = sample(test(), IS(), 1_000)
 
         @test all(isone, chains[:x])
-        @test chains.logevidence ≈ -2 * log(2)
+        logevidence = log(mean(exp.(chains[:loglikelihood])))
```
mhauru: Isn't this the same as the above `logsumexp(chain[:loglikelihood]) - log(N)`, but maybe less numerically stable or fast?

test/Project.toml
```diff
@@ -78,3 +76,6 @@ StatsBase = "0.33, 0.34"
 StatsFuns = "0.9.5, 1"
 TimerOutputs = "0.5"
 julia = "1.10"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: To remove

Project.toml
```diff
@@ -83,3 +83,9 @@ StatsAPI = "1.6"
 StatsBase = "0.32, 0.33, 0.34"
 StatsFuns = "0.8, 0.9, 1"
 julia = "1.10.8"
+
+[extras]
+DynamicHMC = "bbc10e6e-7c05-544b-b16e-64fede858acb"
+
+[sources]
+DynamicPPL = {url = "https://github.com/TuringLang/DynamicPPL.jl", rev = "breaking"}
```
mhauru: To remove

src/mcmc/gibbs_conditional.jl
penelopeysm: NodeTrait is gone, so I had to rewrite it and I think this is just what I ended up with. Separating the GibbsContext method is optional though - would you rather keep that inside the AbstractParentContext method?

test/mcmc/is.jl
penelopeysm: Yes, it's probably both less numerically stable and slower. But it's imo clearer, avoids pulling in an extra import, and doesn't cause an issue in the test.

penelopeysm:
> Do I understand correctly that the old Transition and the new ParamsWithStats will (typically?) cause the same number of evaluations, though the latter may be a bit faster due to use of OnlyAccsVarInfo? So a small positive performance change would be expected from that?

Yup, that's right. The performance difference is probably not very big - in my opinion the nice thing is that the behaviour is encapsulated in DPPL.

src/mcmc/gibbs_conditional.jl
mhauru: Ah, I was thinking of the refactor from `if-else` to method dispatch. There's something elegant about doing it with dispatch, but I sometimes find it more readable when there's just a single method with an `if-else` logic (that gets compiled away). Curious if you have a reason to prefer one. Regardless, happy to leave this as-is.

test/mcmc/is.jl
mhauru: I have a slight preference for the specialised function just because it's generally good practice to use it and the tests could lead by example. This is at disagreement level 1.5.

penelopeysm: Haha to me it's a microoptimisation rather than good practice 😅 

I'll change it but leave in a comment saying that "this is equivalent to .... but more numerically stable"

mhauru: Seems it makes very little difference for speed, but just guards against over- and underflow:
```julia
julia> function f(x)
           display(logsumexp(x))
           display(@b logsumexp(x))
           display(log(sum(exp.(x))))
           display(@b log(sum(exp.(x))))
           return nothing
       end

julia> f(randn(10_000))
9.67801864521181
37.958 μs
9.678018645211806
35.291 μs (3 allocs: 96.062 KiB)

julia> f(randn(10_000)*1000)
3682.3316144326304
43.666 μs
Inf
77.375 μs (3 allocs: 96.062 KiB)
```

penelopeysm: Yeah, but the chain isn't generating those values. Not a big deal, changed now.

src/mcmc/gibbs_conditional.jl
penelopeysm: If the intent is for the compiler to optimise it away, then I think that method dispatch is a more direct way of expressing that. Is it a documented guarantee that the compiler will optimise `if x isa T` branches away?

In terms of style, I think I gravitate towards method dispatch because it's more declarative than imperative. Same reason why `return if foo; x; else y; end` over `if foo; return x; else return y; end`.

penelopeysm:
Changelog added.

penelopeysm:
... Darn, forgot to mention the logevidence thing.

src/mcmc/gibbs_conditional.jl
mhauru: > Is it a documented guarantee that the compiler will optimise if x isa T branches away

Good point, I think not. In practice I strongly expect it to happen in simple cases like these, but there's no guarantee.

I think "more declarative" is a more precise way of expressing what I meant by "more elegant". The downside is that you may have to read the code in a weird order, and the different declarations could be scattered all over your code base.

Anyway, good chat, but no code changes needed.

HISTORY.md
```diff
@@ -2,12 +2,73 @@
...
+model = f(y)
+model = setthreadsafe(model, true)
+```
+
+You *only* need to do this if your model uses tilde-statements or `@addlogprob!` in parallel.
+You do *not* need to do this if:
+
+  - your model has other kinds of parallelism but does not include tilde-statements inside;
+  - or you are using `MCMCThreads()` or `MCMCDistributed()` to sample multiple chains in parallel, but your model itself does not use parallelism.
+
```
mhauru, line 41:
```suggestion

If your model does include parallelised tilde-statements or `@addlogprob!` calls, and you evaluate it/sample from it without setting `setthreadsafe(model, true)`, then you may get statistically incorrect results without any warnings or errors.

```

HISTORY.md
```diff
@@ -2,12 +2,73 @@
...
+### Log-density names in chains
+
+When sampling from a Turing model, the resulting `MCMCChains.Chains` object now contains the log-joint, log-prior, and log-likelihood under the names `:logjoint`, `:logprior`, and `:loglikelihood` respectively.
+Previously, `:logjoint` would be stored under the name `:lp`.
+
+### Log-evidence in chains
+
+When sampling using MCMCChains, the chain object will no longer have its `chain.logevidence` field set.
+Instead, you can calculate this yourself from the log-likelihoods stored in the chain.
+For SMC samplers, the log-evidence of the entire trajectory is stored in `chain[:logevidence]` (which is the same for every particle in the 'chain').
```
mhauru, line 73: Just to check, is this only for SMC or also for PG?

mhauru: Having thought about this for 3 more seconds, this probably makes no sense for PG. Ignore me.

penelopeysm: I actually don't really know what to do with PG. #2723 but I agree I don't think it makes sense to give an overall logevidence. it makes sense per transition but not for the chain

penelopeysm: & even if there was a well-defined logevidence, i'm like 75% sure that `mean(chain[:logevidence])` wasn't the right way to calculate it haha

penelopeysm:
It'll still be on `breaking`, so not yet released. I'll check on the AdvancedVI PR.

PR #2733, opened 2025-12-05
breaking - v0.43

penelopeysm:
This minor version does the following:

- [x] DynamicPPL v0.40
   - [x] Docs are already written up in the main docs repo and just need a release
- [x] New optimisation interface
   - [x] Docs are updated locally in the Turing repo -- we can shift over to the Quarto docs repo after releasing
- [x] New MH sampler interface

Things that NEED to be fixed before releasing:

- [x] Abstract type in SMC `TracedModel` -- ~~get around this using a tactic based on VectorParamAccumulator~~ I benchmarked and it isn't a problem.
- [x] Optimisation bug with parameter names / values (specifically for `LKJCholesky(3, x)` where it generates 6 names but 9 values) -- the Turing code for this has already been written, and generally works, but requires upstream changes
  - [x] `VectorBijectors.optic_vec` to be merged & released https://github.com/TuringLang/Bijectors.jl/pull/429
  - [x] DynamicPPL to use it for constructing LDF.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,7 +11,7 @@ However, it does place some constraints on Turing models.
 Specifically, the types of **containers that can include random variables** are now more limited:
 if `x[i] ~ dist` is a random variable, then `x` must obey the following criteria:
 
-  - They must be arrays. Dicts and other containers are currently unsupported (we have [an issue to track this](https://github.com/TuringLang/DynamicPPL.jl/issues/1263)). If you really need this functionality, please open an issue and let us know; we can try to make it a priority.
+  - They must be `AbstractArray`s. Dicts and other containers are currently unsupported (we have [an issue to track this](https://github.com/TuringLang/DynamicPPL.jl/issues/1263)). If you really need this functionality, please open an issue and let us know; we can try to make it a priority.
     
     ```julia
     @model function f()
@@ -36,7 +36,7 @@ if `x[i] ~ dist` is a random variable, then `x` must obey the following criteria
     ```
 
 However, please note that this only applies to **containers that contain random variables on the left-hand side of tilde-statements.**
-In general, there are no restrictions on containers of *observed* data, or containers that are not used in tilde-statements.
+In general, there are no restrictions on containers of *observed* data, containers that are not used in tilde-statements, or containers that are *themselves* random variables (e.g. `x ~ MvNormal(...)`).
 
   - Likewise, arrays of random variables should ideally have a constant size from iteration to iteration. That means a model like this will fail sometimes (*but* see below):
     
@@ -52,6 +52,31 @@ In general, there are no restrictions on containers of *observed* data, or conta
 
 `VarNamedTuple` and `@vnt` are now re-exported from Turing directly.
 There is a docs page explaining how to use and create `VarNamedTuple`s, which [can be found here](https://turinglang.org/docs/usage/varnamedtuple/).
+
+## Conditioning and fixing
+
+When providing conditioned or fixed variables to Turing models, we recommend that you use a `VarNamedTuple` to do so.
+The main benefit of this is that it correctly captures the structure of arrays of random variables.
+In the past, this used to depend on whether you specified variables exactly as they were seen in the model: for example, if the model had `x ~ MvNormal(zeros(2), I)`, and you conditioned separately on `x[1]` and `x[2]`, the conditioned variables would be silently ignored.
+There were also similar inconsistencies with colons in variable names.
+
+With a VarNamedTuple-based approach, both should be equivalent: you can do
+
+```julia
+vnt1 = @vnt begin
+    @template x = zeros(2)
+    x[1] := 1.0
+    x[2] := 2.0
+end
+cond_model = model() | vnt1
+
+vnt2 = @vnt begin
+    x := [1.0, 2.0]
+end
+cond_model = model() | vnt2
+```
+
+and both should work correctly.
 
 ## Optimisation interface
 
@@ -138,6 +163,11 @@ Additional changes:
   - MH now reports whether each proposal was `accepted` in the chain stats.
   - At the start of sampling, MH logs `@info` messages showing which proposal is used for each variable (disable with `verbose=false`). This helps detect misspecified proposals.
   - MH validates initial parameters against the proposal distribution; if they have zero or NaN probability, a clear error is thrown.
+
+## Gibbs sampler
+
+Both the compilation and runtime of Gibbs sampling should now be significantly faster.
+(This is largely due to the underlying changes in DynamicPPL's data structures.)
 
 ## HMC / NUTS
 


penelopeysm:
This isn't yet ready for release; DPPL isn't released yet, and the MH bit needs a review. But tests are passing.

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
+    the VarName is not in this VarNamedTuple, then it will be sampled from the prior. The
+    Bool indicates whether the proposal is in linked space (true, i.e., the strategy should
+    return a `LinkedVectorValue`); or in untransformed space (false, i.e., the strategy
+    should return an `UntransformedValue`)."
+    proposals::V
+end
+function DynamicPPL.init(
+    rng::Random.AbstractRNG, vn::VarName, prior::Distribution, strategy::InitFromProposals
+)
+    if haskey(strategy.proposals, vn)
```
sunxd3, line 187: `haskey` feels a bit dangerous, what if proposals have `x` but `vn` is `x[1]`?

also if this doesn't match, it might fail silently to the `else` branch?

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
+end
+
+# Accumulator to store priors for any variables that were not given an explicit proposal.
+# This is needed to compute the log-proposal density correctly.
+const MH_PRIOR_ACC_NAME = :MHUnspecifiedPriors
+struct StoreUnspecifiedPriors
+    vns_with_proposal::Set{VarName}
+end
+function (s::StoreUnspecifiedPriors)(val, tval, logjac, vn, dist::Distribution)
+    return if vn in s.vns_with_proposal
```
sunxd3, line 484: similar to the question above, what to do in the case of subsumption?

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
+    # already set. We can obtain this by using `RawValueAccumulator`. Furthermore,
+    # `linked_vals` is a VarNamedTuple that stores a `MHLinkedVal` for any variables that
+    # have `LinkedRW` proposals. That in turn is obtained using `MHLinkedValuesAccumulator`.
+    function init_strategy_constructor(raw_vals, linked_vals)
+        proposals = DynamicPPL.VarNamedTuple()
+        for pair in vn_proposal_pairs
+            # Convert all keys to VarNames.
+            vn, proposal = pair
+            vn = _to_varname(vn)
+            if !haskey(raw_vals, vn)
```
sunxd3, line 230: I think the behavior is correct -- if user doesn't give init info, then we skip.

But it also makes me think what if user gives the wrong key for initialization, then this is kind of silent.

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
+    new_raw_values = DynamicPPL.get_raw_values(new_vi)
+    new_linked_values = DynamicPPL.getacc(new_vi, Val(MH_ACC_NAME)).values
+    new_unspecified_priors = DynamicPPL.getacc(new_vi, Val(MH_PRIOR_ACC_NAME)).values
+
+    init_strategy_given_new = spl.init_strategy_constructor(
+        new_raw_values, new_linked_values
+    )
+
+    # Calculate the log-acceptance probability.
+    log_a = (
```
sunxd3, line 385: maybe some NaN/Inf handling? (can `lp` gets these values?)

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
+    # `log(g(x))` is already -Inf regardless of what `x'` is. We insert a check for this
+    # here.
+    initial_raw_values = DynamicPPL.get_raw_values(vi)
+    initial_linked_values = DynamicPPL.getacc(vi, Val(MH_ACC_NAME)).values
+    init_strategy = spl.init_strategy_constructor(initial_raw_values, initial_linked_values)
+    initial_unspecified_priors = DynamicPPL.getacc(vi, Val(MH_PRIOR_ACC_NAME)).values
+    initial_log_proposal_density = log_proposal_density(
+        vi, init_strategy, initial_unspecified_priors
+    )
+    if initial_log_proposal_density == -Inf
```
sunxd3: maybe also `isnan`?

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
-                # Try to convert it to a proposal anyways,
-                # throw an error if not acceptable.
-                prop = proposal(s)
-                push!(props, prop)
-            end
-        end
-
-        proposals = NamedTuple{tuple(prop_syms...)}(tuple(props...))
+**Note that when using conditional proposals, the values obtained by indexing into the
+`VarNamedTuple` are always in unlinked space.** Sometimes, you may want to define a random-walk
```
sunxd3: side question: do we have a convention on the use of `link` vs `transform` vs `constrained`?

src/mcmc/mh.jl
```diff
@@ -1,464 +1,464 @@
...
-# Use a static proposal for s² and random walk with proposal
-# standard deviation of 0.25 for m.
-chain = sample(
-    gdemo(1.5, 2.0),
-    MH(
-        :s² => AdvancedMH.StaticProposal(InverseGamma(2,3)),
-        :m => AdvancedMH.RandomWalkProposal(Normal(0, 0.25))
-    ),
-    1_000
+# Use a static proposal for s² (which happens to be the same as the prior) and a static
```
sunxd3:
```suggestion
# Use a static proposal for s (which happens to be the same as the prior) and a static
```
maybe?

src/mcmc/mh.jl
penelopeysm: It would silently do the else branch, that's true. I need to sit down and work out these edge cases. (To be fair, the old one had even weirder behaviour, but that's not an excuse.)

penelopeysm: I think it's very hard to do the correct thing though. In the else branch we could kind of do a 'stepping upwards' loop, like if `vn` is `x[1].a` we check whether `x[1]` exists or `x` exists and if so warn. But I think that would just create other edge cases, and it would also be bad for performance.

Maybe the best answer is before we start sampling (somewhere in the first `AbstractMCMC.step`) we should do one model evaluation where we iterate through the model and print out the proposal for each parameter? The user should be able to silence that with `verbose=false`, and if not then it's a good check for them.

src/mcmc/mh.jl
penelopeysm: Depends. Sometimes -Inf is fine in the sense that you just make it reject the sample (which can happen either by sampling a parameter value outside the support, or if the user just wants to signal some failure condition with `@addlogprob! -Inf`). The main problem I discovered with this was when the initial parameters already gave an lp of -Inf, which is why I inserted the special handler on line 286 onwards.

NaN on the other hand I think can only happen if the value itself is already a NaN or if there's a bug in the implementation of `logpdf`, which should *probably* not happen. But we could insert a warning here? I think the current behaviour is that it will never be accepted (because comparisons against NaN always return false), which is fairly sensible, but I could see a warning being nice.

src/mcmc/mh.jl
penelopeysm: Unfortunately not. I like transform the best out of them, so I'll try to stick to that where possible in the docstrings, but a lot of the code still uses link.

My perspective on this is that transform is generic and can be anything, but link is *the* special transform that maps to unconstrained space. That might be a bit questionable because in theory there are many transforms that will map to unconstrained space, so in practice what it really means is whatever `Bijectors.bijector` returns.

src/mcmc/mh.jl
penelopeysm: Added.

src/mcmc/mh.jl
penelopeysm: https://github.com/TuringLang/Turing.jl/pull/2774

src/mcmc/mh.jl
penelopeysm: Added a warning (also in #2774)

penelopeysm:
This is pretty much ready to go once CI passes.

Confusingly, there's a case where CI on 1.10 hangs midway through the MH tests. I have no idea why.

penelopeysm:
Oh, I thought I would finally go to bed, but actually I can reproduce it locally just by running `import Pkg; Pkg.test("Turing"; test_args=["mh"])`.

Ctrl-C'ing at the place where it hangs gives the following stack trace:

```
inference with samplers: Error During Test at /Users/pyong/ppl/lib/test/runtests.jl:54
  Got exception outside of a @test
  LoadError: InterruptException:
  Stacktrace:
    [1] lookup(pointer::Ptr{Nothing})
      @ Base.StackTraces ./stacktraces.jl:108
    [2] firstcaller
      @ ./deprecated.jl:157 [inlined]
    [3] firstcaller
      @ ./deprecated.jl:152 [inlined]
    [4] macro expansion
      @ ./deprecated.jl:132 [inlined]
    [5] macro expansion
      @ ./logging.jl:390 [inlined]
    [6] depwarn(msg::String, funcsym::Symbol; force::Bool)
      @ Base ./deprecated.jl:127
    [7] depwarn(msg::String, funcsym::Symbol)
      @ Base ./deprecated.jl:121
    [8] Distributions.MvNormal(μ::Vector{Float64}, σ::Float64)
      @ Distributions ./deprecated.jl:104
    [9] init_strategy_constructor
      @ ~/ppl/lib/src/mcmc/mh.jl:233 [inlined]
   [10] step(rng::StableRNGs.LehmerRNG, model::DynamicPPL.Model{Main.MHTests.var"#f#8", (), (), (), Tuple{}, Tuple{}, DynamicPPL.DefaultContext, false}, spl::Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}}, old_vi::DynamicPPL.VarInfo{DynamicPPL.UnlinkSome{Set{AbstractPPL.VarName{:y, AbstractPPL.Iden}}, DynamicPPL.LinkAll}, DynamicPPL.VarNamedTuples.VarNamedTuple{(:x, :y), Tuple{DynamicPPL.LinkedVectorValue{Vector{Float64}, Bijectors.VectorBijectors.OnlyWrap{Bijectors.VectorBijectors.TypedIdentity}}, DynamicPPL.VectorValue{Vector{Float64}, Bijectors.VectorBijectors.OnlyWrap{Bijectors.VectorBijectors.TypedIdentity}}}}, DynamicPPL.AccumulatorTuple{6, @NamedTuple{LogPrior::DynamicPPL.LogPriorAccumulator{Float64}, LogJacobian::DynamicPPL.LogJacobianAccumulator{Float64}, LogLikelihood::DynamicPPL.LogLikelihoodAccumulator{Float64}, RawValues::DynamicPPL.VNTAccumulator{:RawValues, DynamicPPL.GetRawValues, DynamicPPL.VarNamedTuples.VarNamedTuple{(:x, :y), Tuple{Float64, Float64}}}, MHLinkedValues::DynamicPPL.VNTAccumulator{:MHLinkedValues, Turing.Inference.StoreLinkedValues, DynamicPPL.VarNamedTuples.VarNamedTuple{(:x,), Tuple{Vector{Float64}}}}, MHUnspecifiedPriors::DynamicPPL.VNTAccumulator{:MHUnspecifiedPriors, Turing.Inference.StoreUnspecifiedPriors, DynamicPPL.VarNamedTuples.VarNamedTuple{(:y,), Tuple{Distributions.Beta{Float64}}}}}}}; discard_sample::Bool, kwargs::@Kwargs{initial_params::DynamicPPL.InitFromPrior})
      @ Turing.Inference ~/ppl/lib/src/mcmc/mh.jl:351
   [11] _step_or_step_warmup(::Int64, ::Int64, ::StableRNGs.LehmerRNG, ::Vararg{Any}; kwargs::@Kwargs{discard_sample::Bool, initial_params::DynamicPPL.InitFromPrior})
      @ AbstractMCMC ~/.julia/packages/AbstractMCMC/oqm6Y/src/sample.jl:0
   [12] _step_or_step_warmup
      @ ~/.julia/packages/AbstractMCMC/oqm6Y/src/sample.jl:124 [inlined]
   [13] macro expansion
      @ ~/.julia/packages/AbstractMCMC/oqm6Y/src/sample.jl:315 [inlined]
   [14] (::AbstractMCMC.var"#31#32"{Nothing, Int64, Int64, Int64, UnionAll, Nothing, @Kwargs{initial_params::DynamicPPL.InitFromPrior}, StableRNGs.LehmerRNG, DynamicPPL.Model{Main.MHTests.var"#f#8", (), (), (), Tuple{}, Tuple{}, DynamicPPL.DefaultContext, false}, Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}}, Int64, Float64, Int64})()
      @ AbstractMCMC ~/.julia/packages/AbstractMCMC/oqm6Y/src/logging.jl:134
   [15] with_logstate(f::Function, logstate::Any)
      @ Base.CoreLogging ./logging.jl:517
   [16] with_logger(f::Function, logger::LoggingExtras.TeeLogger{Tuple{LoggingExtras.EarlyFilteredLogger{TerminalLoggers.TerminalLogger, AbstractMCMC.var"#1#3"{Module}}, LoggingExtras.EarlyFilteredLogger{Logging.ConsoleLogger, AbstractMCMC.var"#2#4"{Module}}}})
      @ Base.CoreLogging ./logging.jl:630
   [17] with_progresslogger(f::Function, _module::Module, logger::Logging.ConsoleLogger)
      @ AbstractMCMC ~/.julia/packages/AbstractMCMC/oqm6Y/src/logging.jl:157
   [18] macro expansion
      @ ~/.julia/packages/AbstractMCMC/oqm6Y/src/logging.jl:133 [inlined]
   [19] mcmcsample(rng::StableRNGs.LehmerRNG, model::DynamicPPL.Model{Main.MHTests.var"#f#8", (), (), (), Tuple{}, Tuple{}, DynamicPPL.DefaultContext, false}, sampler::Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}}, N::Int64; progress::Bool, progressname::String, callback::Nothing, num_warmup::Int64, discard_initial::Int64, thinning::Int64, chain_type::Type, initial_state::Nothing, kwargs::@Kwargs{initial_params::DynamicPPL.InitFromPrior})
      @ AbstractMCMC ~/.julia/packages/AbstractMCMC/oqm6Y/src/sample.jl:204
   [20] sample(rng::StableRNGs.LehmerRNG, model::DynamicPPL.Model{Main.MHTests.var"#f#8", (), (), (), Tuple{}, Tuple{}, DynamicPPL.DefaultContext, false}, spl::Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}}, N::Int64; initial_params::DynamicPPL.InitFromPrior, check_model::Bool, chain_type::Type, kwargs::@Kwargs{})
      @ Turing.Inference ~/ppl/lib/src/mcmc/abstractmcmc.jl:55
   [21] sample(rng::StableRNGs.LehmerRNG, model::DynamicPPL.Model{Main.MHTests.var"#f#8", (), (), (), Tuple{}, Tuple{}, DynamicPPL.DefaultContext, false}, spl::Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}}, N::Int64)
      @ Turing.Inference ~/ppl/lib/src/mcmc/abstractmcmc.jl:44
   [22] macro expansion
      @ ~/ppl/lib/test/mcmc/mh.jl:48 [inlined]
   [23] macro expansion
      @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1527 [inlined]
   [24] (::Main.MHTests.var"#test_mean_and_std#9")(spl::Turing.Inference.MH{Turing.Inference.var"#init_strategy_constructor#44"{Tuple{Pair{AbstractPPL.VarName{:x, AbstractPPL.Iden}, Turing.Inference.LinkedRW{Float64}}}}, DynamicPPL.LinkSome{Set{AbstractPPL.VarName}, DynamicPPL.UnlinkAll}})
      @ Main.MHTests ~/ppl/lib/test/mcmc/mh.jl:47
   [25] include(fname::String)
      @ Base.MainInclude ./client.jl:494
   [26] include(fname::String)
      @ Base.MainInclude ./client.jl:494
   [27] eval
      @ ./boot.jl:385 [inlined]
   [28] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:296
   [29] _start()
      @ Base ./client.jl:557
  in expression starting at
```

HISTORY.md
```diff
@@ -1,3 +1,161 @@
...
+    for i in 1:10
+        push!(x, 0.0)
+        x[i] ~ Normal()
+    end
+    ```
+
+However, please note that this only applies to **containers that contain random variables on the left-hand side of tilde-statements.**
+In general, there are no restrictions on containers of *observed* data, or containers that are not used in tilde-statements.
+
+  - Likewise, arrays of random variables should ideally have a constant size from iteration to iteration. That means a model like this will fail sometimes (*but* see below):
```
mhauru, line 41: Would it be worth pointing out either here or above at the beginning of these bullet points that this only applies when using indexing, and doing the multivariate distribution version of the below is entirely fine?

HISTORY.md
```diff
@@ -1,3 +1,161 @@
...
+DynamicPPL v0.40 includes a major overhaul of Turing's internal data structures.
+Most notably, cases where we might once have used `Dict{VarName}` or `NamedTuple` have all been replaced with a single data structure, called `VarNamedTuple`.
+
+This provides substantial benefits in terms of robustness and performance.
+
+However, it does place some constraints on Turing models.
+Specifically, the types of **containers that can include random variables** are now more limited:
+if `x[i] ~ dist` is a random variable, then `x` must obey the following criteria:
+
+  - They must be arrays. Dicts and other containers are currently unsupported (we have [an issue to track this](https://github.com/TuringLang/DynamicPPL.jl/issues/1263)). If you really need this functionality, please open an issue and let us know; we can try to make it a priority.
```
mhauru: Worth specifying what we mean by "arrays"? Any subtype of `AbstractArray`?

HISTORY.md
```diff
@@ -1,3 +1,161 @@
...
+
+When defining a conditional posterior, instead of being provided with a Dict of values, the function must now take a `VarNamedTuple` containing the values.
+Note that indexing into a `VarNamedTuple` is very similar to indexing into a `Dict`; however, it is more flexible since you can use syntax such as `x[1:2]` even if `x[1]` and `x[2]` are separate variables in the model.
+
+## `filldist` and `arraydist`
+
+These two convenience functions are now imported and re-exported from DynamicPPL, rather than DistributionsAD.jl.
+They are now just wrappers around `Distributions.product_distribution`, instead of the specialised implementations that were in DistributionsAD.jl.
+DistributionsAD.jl is for all intents and purposes deprecated: it is no longer a dependency in the Turing stack.
+
```
mhauru, line 188: Sounds like at least Gibbs may have had a serious performance boost. Is it worth talking about performance improvements here?

mhauru, line 188: Any other advantages other than performance that would be worth raising here? Improvements to fix and conditional?

yebai:
Looks great!

One minor comment: If we drop deps on DistributionsAD, would we lose any rules for ForwardDiff (default Turing AD) in https://github.com/TuringLang/DistributionsAD.jl/issues/280?

penelopeysm:
@yebai I doubt it. At the very least, since CI is passing, we can be sure that there's nothing in Turing's CI suite that depended on that.

The CI suite does contain this list of distributions that we check against:

https://github.com/TuringLang/Turing.jl/blob/184d592cf4eb6fa30be8d1fc365f95bcde7da6ad/test/stdlib/distributions.jl#L53-L156

I think this list is probably outdated and could be expanded. Bijectors now contains very though testing for differentiability of with_logabsdet_jacobian on a ton of distributions, and the remaining thing would be to test differentiability of logpdf. (That should arguably be in Distributions.jl, but well... we might have to stick it in DynamicPPL or something.)

But I'm pretty sure that anything that used to work before should continue to work.

penelopeysm:
CI failure is due to a Julia GC segfault -- really?!....

Edit: It is indeterministic on CI and I couldn't reproduce locally, so will just ignore it going forwards

HISTORY.md
penelopeysm: Tweaked wording

HISTORY.md
penelopeysm: Yeah, changed to `AbstractArray`.

penelopeysm: Technically, it's contingent on it implementing enough of an AbstractArray interface, plus it working with BangBang. But well.

HISTORY.md
penelopeysm: Added a note. Turns out runtimes are indeed a lot better too, this is 0.11 seconds on this branch vs 0.18 seconds on main.

```julia
using Turing

@model function f()
    x ~ Normal(0, 1)
    y = zeros(10)
    for i in 1:10
        y[i] ~ Normal(x, 1)
    end
    z ~ Normal(sum(y), 1)
end

@time sample(f(), Gibbs(:x => HMC(0.1, 10), :y => HMC(0.1, 10), :z => HMC(0.1, 10)), 1000; chain_type=Any);
```

penelopeysm: I added a paragraph about this. I'm still not entirely sure I like where we are. We still don't have well-defined semantics for this. It's probably much closer to what we think should be correct, but there isn't a formal statement of what is correct, and consequently it's hard to meaningfully judge how much closer we are to that. Of course, there are lots of individual cases where we can say the behaviour is more intuitive, but lots of individual cases don't together make a formal specification.

penelopeysm: Still, that's something for another time.

Issue #2607, opened 2025-06-29
create `returned()` method for results from mode estimation


penelopeysm:
https://discourse.julialang.org/t/get-turings-model-predictions-after-optimization/130275

penelopeysm:
probably comes under https://github.com/TuringLang/Turing.jl/issues/2232

penelopeysm:
also cf. https://github.com/TuringLang/Turing.jl/issues/2363

PR #2736, opened 2025-12-09
Fix missing `fallback` argument for `InitFromParams(::ModeResult)`

penelopeysm:
Closes https://github.com/TuringLang/Turing.jl/issues/2607. ~~Note that this requires https://github.com/TuringLang/DynamicPPL.jl/pull/1171 to go through first. Hence I can't add a test. But once the DPPL one is done, I will add the following test, which works locally:~~ Test added

```julia
julia> using Turing

julia> @model function f()
           x ~ Normal()
           return x + 1
       end

julia> mapp = maximum_a_posteriori(f())
ModeResult with maximized lp of -0.92
[0.0]

julia> returned(f(), mapp)
1.0
```


penelopeysm:
@sunxd3 dppl pr went through :)

sunxd3:
thanks, taking a look shortly

Project.toml
sunxd3, line 65: bump this too?

penelopeysm: I was thinking it's not really necessary to bump, since technically using the old version is still fine, you just don't get this functionality

PR #2740, opened 2025-12-16
Add retry logic for initial parameters in ExternalSampler (#2739)

### LINKED ISSUE #2739: External sampler should try harder at generating initial params
Turing currently generates initial params and passes it to external samplers:

https://github.com/TuringLang/Turing.jl/blob/4dc7ad096f92fd571de312e7751986484ac6cb50/src/mcmc/external_sampler.jl#L159-L163

However, this might be a bit too simplistic. The HMC implementation for example currently tries a lot of times to generate initial parameters until it can find something where the logp and gradient are finite. Just giving it one attempt might make it hard to actually get started with sampling.

0AnshuAditya0:
Fixes #2739

This PR adds retry logic when generating initial parameters for external samplers, similar to what HMC already does.

**Changes:**
- Modified `src/mcmc/external_sampler.jl` to attempt finding valid initial parameters up to 10 times
- Checks that both logp and gradient are finite before accepting initial parameters
- Provides informative error message if all attempts fail

**Why:**
External samplers previously only tried once to generate initial parameters. If the first attempt produced infinite logp or gradient, sampling would fail immediately. This makes external samplers more robust.

### DIFF
--- a/src/mcmc/abstractmcmc.jl
+++ b/src/mcmc/abstractmcmc.jl
@@ -29,17 +29,15 @@ function find_initial_params(
     model::Model,
     varinfo::AbstractVarInfo,
     init_strategy::DynamicPPL.AbstractInitStrategy,
-    validator::Function;
+    validator::Any;
     max_attempts::Int=1000,
 )
     varinfo = deepcopy(varinfo)  # Don't mutate the input
     
     for attempt in 1:max_attempts
         # Validate current parameters
-        is_valid = validator(varinfo)
-        
-        if is_valid
-            return varinfo  # Success!
+        if validator(varinfo)
+            return varinfo
         end
         
         # Warn at attempt 10
--- a/test/mcmc/abstractmcmc.jl
+++ b/test/mcmc/abstractmcmc.jl
@@ -151,50 +151,37 @@ end
 end
 
 @testset "Initial parameter retry logic" begin
-        # Model that produces -Inf logp for most parameter values
-        # Only valid when x is in narrow range [-0.3, 0.3]
+        init_counter_1 = Ref(0)
         @model function bad_init_model()
             init_counter_1[] += 1
             x ~ Normal(0, 1)
             Turing.@addlogprob! (init_counter_1[] > 5) ? 0.0 : -Inf
         end
         
-        # This should succeed with retry logic (might take a few attempts)
         model = bad_init_model()
         
-        # Test with NUTS (internal HMC sampler with retry logic)
         @testset "NUTS sampler" begin
+            Random.seed!(1234)
             chain = sample(model, NUTS(), 10)
             @test size(chain, 1) == 10
-            # Check that samples are in valid range
-            x_samples = chain[:x]
-            @test all(abs.(x_samples) .< 0.3)
         end
         
-        # Test with HMC (should also work with retry logic)
-        @testset "HMC sampler" begin
-            chain = sample(model, HMC(0.1, 5), 10)
-            @test size(chain, 1) == 10
-            x_samples = chain[:x]
-            @test all(abs.(x_samples) .< 0.3)
-        end
-        
-        # Model with very narrow valid region
+        @test init_counter_1 =>  6
+
+        init_counter_2 = Ref(0)
+
         @model function very_bad_init_model()
-            x ~ Normal(0, 100)  # Very wide prior
-            # Valid only in range [-0.1, 0.1]
+            init_counter_2[] += 1
+            x ~ Normal(0, 1)
             Turing.@addlogprob! (abs(x) < 0.1) ? 0.0 : -Inf
         end
         
-        @testset "Very narrow valid region" begin
-            model = very_bad_init_model()
-            
-            # Should eventually find valid parameters
-            chain = sample(model, NUTS(), 10)
+        @testset "HMC sampler" begin
+            chain = sample(bad_init_model_hmc(), HMC(0.1, 5), 10)
             @test size(chain, 1) == 10
-            x_samples = chain[:x]
-            @test all(abs.(x_samples) .< 0.1)
         end
+
+        @test init_counter_2[] => 6
         
         # Model that's impossible to initialize (always -Inf)
         @model function impossible_model()
@@ -207,6 +194,9 @@ end
             
             # Should throw an error with informative message
             @test_throws ErrorException sample(model, NUTS(), 10)
+                sample(model, NUTS(), 10)
+                @test occursin("Failed to find initial parameters", error_msg)
+            end
         end
         
         # Model that requires many attempts
--- a/test/mcmc/hmc.jl
+++ b/test/mcmc/hmc.jl
@@ -197,14 +197,6 @@ using Turing
         end
     end
 
-    @testset "error for impossible model" begin
-        @model function demo_impossible()
-            x ~ Normal()
-            @addlogprob! -Inf
-        end
-
-        @test_throws ErrorException sample(demo_impossible(), NUTS(), 5)
-    end
 
     @testset "NUTS initial parameters" begin
         @model function f()


mhauru:
Hi @0AnshuAditya0, thanks for your interest in contributing to Turing.jl.

Some of our team is currently on leave, and I'm really busy and will soon go on leave, so it may be early January before we'll have time to give your PR a proper look. So apologies if we aren't the most responsive right now, we'll get to it in a few weeks.

Some quick first thoughts though:
* Rather than writing two parallel versions of this, for HMC specifically in `hmc.jl` and for external samplers in `external_samplers.jl`, I think we should try to unify these under a single function. Would you be happy to refactor so that there would be a single function called `find_initial_params` that handles both cases? On a quick look, I think the only part of the current `find_initial_params` function that is specific to HMC is the check of whether the `phasepoint` is finite, but that's just checking the finiteness of the value and the gradient, so that function could easily be modified to handle the general case.
* Not all samplers require gradients, and not all models have well-defined gradients, so we should have a flag for whether `find_initial_params` checks for gradient finiteness at all. I think by default this should be false.
* We should have a test for this. I'm not sure how to best do that. Maybe a model which is manually crafted to give infinite log probabilities or infinite gradients e.g. 90% of the time? And then testing on that to see that we can still eventually get a good initial value. The test model could be something very artificial, doesn't have to be a meaningful statistical model at all.

0AnshuAditya0:
Thanks for the feedback @mhauru! I appreciate the guidance on refactoring this properly.

Your approach makes a lot of sense. Instead of duplicating the retry logic, I can extract it into a shared utility that handles the general case. Here's what I'm thinking:

**Proposed changes:**
1. Create a new `find_initial_params` function in `src/mcmc/initial_params.jl` that centralizes the retry logic for generating valid initial parameters
2. Add a `check_gradient::Bool` flag (defaulting to false) so gradient-free samplers can use it without unnecessary gradient evaluations
3. Refactor both `hmc.jl` and `external_sampler.jl` to use this shared function, which should eliminate code duplication and make future maintenance easier
4. Add a test using a deliberately problematic model (perhaps one with a constrained parameter space that makes most random initializations invalid) to verify the retry mechanism works reliably

The shared function would handle:
- Multiple initialization attempts (configurable max_attempts)
- Checking finiteness of logp (always)
- Optionally checking gradient finiteness (when check_gradient=true)
- Parameter linking/unlinking for constrained spaces
- Clear error messages when all attempts are exhausted

Does this implementation plan align with what you had in mind? Happy to adjust the approach if you have other preferences.

I'll work on these changes over the next few days. No rush on the review - I understand the holiday season timing. Take your time, and I'll have the refactored version ready when you're back!

0AnshuAditya0:
Hi @mhauru! I've implemented the changes you suggested. Here's what I did:

## Changes Made

### 1. Created Shared Function
- Added `src/mcmc/initial_params.jl` with `find_initial_params()` function
- Both HMC and external_sampler now use this shared function
- Used a validator pattern where each sampler passes its own validation logic

### 2. Made Gradient Checking Configurable
- External samplers check gradients using `logdensity_and_gradient()`
- HMC uses `AHMC.phasepoint()` for validation
- Each sampler can define what "valid parameters" means

### 3. Added Tests
- Created `test/mcmc/initial_params.jl` with comprehensive tests
- Tests include models that fail initialization ~90% of the time
- Verified retry logic works for both NUTS and HMC
- Tests pass locally ✓

## Why Validator Pattern?

I chose this approach because HMC and external samplers validate parameters differently:
- HMC: validates using `AHMC.phasepoint()`
- External: validates using `LogDensityProblems` interface

The validator pattern keeps them flexible while sharing the retry logic. Let me know if a simpler approach would be better!

## Testing
Tested with constrained models locally - the retry logic works as expected.

Please let me know if you'd like me to change anything. Thanks for the guidance on this!

src/mcmc/initial_params.jl
```diff
@@ -0,0 +1,90 @@
+"""
+    find_initial_params(rng, model, varinfo, init_strategy, validator; max_attempts=10)
```
penelopeysm: The default value for max_attempts in the code is 1000.

src/mcmc/initial_params.jl
```diff
@@ -0,0 +1,90 @@
...
+# For HMC with hamiltonian
+validator = vi -> begin
+    θ = vi[:]
+    z = AHMC.phasepoint(rng, θ, hamiltonian)
+    is_valid = isfinite(z)
+    diagnostics = "phasepoint finite: \$(isfinite(z))"
+    return (is_valid, diagnostics)
+end
+varinfo = find_initial_params(rng, model, varinfo, init_strategy, validator)
+```
```
penelopeysm: I think this level of detail isn't really needed, writing too much will just add noise to the file. It's an internal function, nobody will be importing it and using it.

```suggestion
Attempt to find valid initial parameters for MCMC sampling, using the provided `init_strategy` up to `max_attempt` times.

The function `validator` should take the (updated) VarInfo with new parameters, and return a Bool
indicating whether the parameters are valid.
```

src/mcmc/initial_params.jl
```diff
@@ -0,0 +1,90 @@
...
+        # Warn at attempt 10
+        if attempt == 10
+            @warn "failed to find valid initial parameters in $(attempt) tries; consider providing a different initialisation strategy with the `initial_params` keyword"
+        end
+        
+        # If this is the last attempt, throw informative error
+        if attempt == max_attempts
+            error(
+                "Failed to find valid initial parameters after $max_attempts attempts. " *
+                "Last attempt diagnostics: $last_diagnostics. " *
```
penelopeysm: I think it's a nice idea to have this error message, but it does make the required signature for `validator` more complicated. I think it's better to just make `validator` return a single boolean and skip the diagnostics.

src/mcmc/hmc.jl
```diff
@@ -157,24 +157,25 @@ function find_initial_params(
...
+        is_valid = isfinite(z)
+        diagnostics = "phasepoint finite: $(isfinite(z))"
+        return (is_valid, diagnostics)
     end
-
-    # if we failed to find valid initial parameters, error
-    return error(
-        "failed to find valid initial parameters in $(max_attempts) tries. See https://turinglang.org/docs/uri/initial-parameters for common causes and solutions. If the issue persists, please open an issue at https://github.com/TuringLang/Turing.jl/issues",
+    
+    # Use shared function from initial_params.jl
```
```suggestion
```

This comment is not necessary

src/mcmc/external_sampler.jl
```diff
@@ -156,16 +156,32 @@ function AbstractMCMC.step(
...
+    validator = vi -> begin
+        θ = vi[:]
+        try
+            logp, grad = LogDensityProblems.logdensity_and_gradient(f, θ)
+            is_valid = isfinite(logp) && all(isfinite, grad)
+            diagnostics = "logp=$logp, grad_finite=$(all(isfinite, grad))"
+            return (is_valid, diagnostics)
+        catch e
+            return (false, "evaluation failed: $e")
+        end
```
penelopeysm: I don't think it's necessary to try/catch here. If evaluation errors, the user has a bigger problem than not being able to initialise the model. Besides, catching the original error will make it harder to diagnose and fix the actual error that happened during evaluation, since the stack trace won't be shown.

```suggestion
        logp, grad = LogDensityProblems.logdensity_and_gradient(f, θ)
        return isfinite(logp) && all(isfinite, grad)
```

src/mcmc/Inference.jl
```diff
@@ -86,6 +86,7 @@ include("abstractmcmc.jl")
 ####################
 
 include("repeat_sampler.jl")
+include("initial_params.jl")
```
penelopeysm: Could you put the function in `abstractmcmc.jl` instead please? I think it's fine to put it in there, it feels a bit overspecialised to have one file for this lone function.

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+    @info "Starting initial parameter retry logic tests"
+    
+    @testset "External sampler with difficult initialization" begin
+        # Model that produces -Inf logp for most parameter values
+        # Only valid when x is in narrow range [-0.3, 0.3]
+        @model function bad_init_model()
+            x ~ Normal(0, 10)  # Prior is wide, so most samples will be outside valid range
+            # Add log probability that's -Inf outside narrow range
+            Turing.@addlogprob! (abs(x) < 0.3) ? 0.0 : -Inf
+        end
```
penelopeysm: If the point is to have a model that errors many times and then finally succeeds, I think it's best to ensure that this is the case by using a counter that is incremented every time the model is evaluated. See here for an example:

https://github.com/TuringLang/Turing.jl/blob/f3bd120ee63f5ceacbe66ffc540f22cf53081d59/test/mcmc/hmc.jl#L200-L214

This model will always fail 30 times and then it will succeed after that, which is better than trying to find the exact StableRNG seed that will fail somewhere between 0 and 1000 times.

Also, if you test this function here, you could also remove that test ^ in `test/mcmc/hmc.jl`.

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+            @test all(abs.(x_samples) .< 0.3)
+        end
+        
+        # Test with HMC (should also work with retry logic)
+        @testset "HMC sampler" begin
+            chain = sample(StableRNG(456), model, HMC(0.1, 5), 10)
+            @test size(chain, 1) == 10
+            x_samples = chain[:x]
+            @test all(abs.(x_samples) .< 0.3)
+        end
```
penelopeysm: If you change the model to always behave the way you want, then you don't need to use StableRNG here since you don't need to control whether the parameter is between [-0.3, 0.3] or not.

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+        @test_throws ErrorException sample(StableRNG(999), model, NUTS(0.65), 10)
+        
+        # Check that error message is informative
+        try
+            sample(StableRNG(999), model, NUTS(0.65), 10)
+        catch e
+            error_msg = sprint(showerror, e)
+            @test occursin("Failed to find valid initial parameters", error_msg)
+            @test occursin("attempts", error_msg)
+        end
```
penelopeysm:
```suggestion
        @test_throws ErrorException sample(model, NUTS(), 10)
        @test_throws "Failed to find valid initial parameters" sample(model, NUTS(), 10)
```

1. You don't need to try/catch errors to get the message; on modern-ish Julia versions you can pass a string to `@test_throws`.
2. The comments aren't necessary, the code communicates the intent clearly already.
3. You don't need StableRNG as the model always fails, there is no rng involved.

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+            Turing.@addlogprob! (abs(x) < 0.05) ? 0.0 : -Inf
+        end
+        
+        model = difficult_model()
+        
+        # Should see warning at attempt 10
+        @test_logs(
+            (:warn, r"failed to find valid initial parameters in 10 tries"),
+            match_mode=:any,
+            sample(StableRNG(111), model, NUTS(0.65), 10)
```
penelopeysm:
```suggestion
            sample(model, NUTS(), 10)
```

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+            x ~ Normal(0, 1)
+            y ~ Normal(x, 1)
+        end
+        
+        model = easy_model()
+        
+        # Should work without any retries or warnings
+        chain = sample(StableRNG(222), model, NUTS(0.65), 10)
+        @test size(chain, 1) == 10
+    end
```
penelopeysm: This test isn't needed as it's just a standard NUTS sample call and there are many of them elsewhere in the test suite.

test/mcmc/initial_params.jl
```diff
@@ -0,0 +1,118 @@
...
+        rng = StableRNG(789)
+        model = very_bad_init_model()
+        
+        # Should eventually find valid parameters
+        chain = sample(rng, model, NUTS(0.65), 10)
+        @test size(chain, 1) == 10
+        x_samples = chain[:x]
+        @test all(abs.(x_samples) .< 0.1)
+    end
+
```
penelopeysm:
```suggestion
```

This testset essentially tests the same thing as the previous one, so I don't think there's any need for it.

src/mcmc/Inference.jl
penelopeysm: The tests can likewise also go in `test/mcmc/abstractmcmc.jl`.

test/mcmc/initial_params.jl
penelopeysm: As an alternative to deleting this test (and maybe even better!), I would suggest not calling `sample()`, instead you should test that `find_initial_params` on this model returns a varinfo. In fact, arguably, the rest of the tests in this section should be testing calls to `find_initial_params` rather than `sample()`. Because when you test sample, you are implicitly saying that '`sample` calls `find_initial_params`, so I can tell if `find_initial_params` has the right behaviour'. But what if the definition of `sample` is changed one day to not call it? Then this test would be out of sync.

penelopeysm:
Thank you for the PR @0AnshuAditya0, I have a number of comments above 🙂

0AnshuAditya0:
All feedback addressed! Here's what I changed:

✅ Moved `find_initial_params` to `abstractmcmc.jl` (more appropriate location)
✅ Simplified validator to return `Bool` only (removed diagnostics tuple)
✅ Updated all validator implementations in HMC and external_sampler
✅ Moved tests to `test/mcmc/abstractmcmc.jl`
✅ Improved test with counter-based model (deterministic 30 failures)
✅ Added direct `find_initial_params` unit tests
✅ Removed duplicate warning test from `test/mcmc/hmc.jl`
✅ Removed unnecessary `StableRNG` usage
✅ Fixed docstring (`max_attempts=1000`)

Ready for re-review! Thanks for the detailed feedback.

src/mcmc/abstractmcmc.jl
```diff
@@ -17,6 +17,50 @@ parameters for sampling are chosen if not specified by the user. By default, thi
...
+Attempt to find valid initial parameters for MCMC sampling, using the provided `init_strategy` up to `max_attempts` times.
+
+The function `validator` should take the (updated) VarInfo with new parameters, and return a Bool indicating whether the parameters are valid.
+"""
+function find_initial_params(
+    rng::AbstractRNG,
+    model::Model,
+    varinfo::AbstractVarInfo,
+    init_strategy::DynamicPPL.AbstractInitStrategy,
+    validator::Function;
```
penelopeysm:
```suggestion
    validator::Any;
```

Callable structs in Julia don't subtype `Function`, unfortunately there's no way to have a stricter type bound here.

src/mcmc/abstractmcmc.jl
```diff
@@ -17,6 +17,50 @@ parameters for sampling are chosen if not specified by the user. By default, thi
...
+)
+    varinfo = deepcopy(varinfo)  # Don't mutate the input
+    
+    for attempt in 1:max_attempts
+        # Validate current parameters
+        is_valid = validator(varinfo)
+        
+        if is_valid
+            return varinfo  # Success!
+        end
```
penelopeysm, line 41:
```suggestion
        if validator(varinfo)
            return varinfo
        end
```

src/mcmc/external_sampler.jl
```diff
@@ -156,16 +156,27 @@ function AbstractMCMC.step(
         varinfo = DynamicPPL.link(varinfo, model)
     end
 
-    # We need to extract the vectorised initial_params, because the later call to
-    # AbstractMCMC.step only sees a `LogDensityModel` which expects `initial_params`
-    # to be a vector.
-    initial_params_vector = varinfo[:]
 
-    # Construct LogDensityFunction
+    # Construct LogDensityFunction FIRST (we need this for validation)
```
penelopeysm, line 160:
```suggestion
    # Construct LogDensityFunction first (we need this for validation)
```

test/mcmc/abstractmcmc.jl
```diff
@@ -150,4 +150,129 @@ end
...
+            @test all(abs.(x_samples) .< 0.3)
+        end
+        
+        # Test with HMC (should also work with retry logic)
+        @testset "HMC sampler" begin
+            chain = sample(model, HMC(0.1, 5), 10)
+            @test size(chain, 1) == 10
+            x_samples = chain[:x]
+            @test all(abs.(x_samples) .< 0.3)
+        end
```
penelopeysm: Your changes actually made this test more unreliable. The model is still indeterministic. Previously, you worked around this by using StableRNG. Now you removed StableRNG but you didn't make the model deterministic.

test/mcmc/abstractmcmc.jl
```diff
@@ -150,4 +150,129 @@ end
...
+        
+        @testset "Very narrow valid region" begin
+            model = very_bad_init_model()
+            
+            # Should eventually find valid parameters
+            chain = sample(model, NUTS(), 10)
+            @test size(chain, 1) == 10
+            x_samples = chain[:x]
+            @test all(abs.(x_samples) .< 0.1)
+        end
```
penelopeysm: As I said before, this test is not doing anything that the previous one didn't do.

test/mcmc/abstractmcmc.jl
```diff
@@ -150,4 +150,129 @@ end
...
+            
+            # Check that error message is informative
+            try
+                sample(model, NUTS(), 10)
+            catch e
+                error_msg = sprint(showerror, e)
+                @test occursin("Failed to find valid initial parameters", error_msg)
+                @test occursin("attempts", error_msg)
+            end
+        end
```
penelopeysm, line 200: try/catch is not needed, see previous review comment.

test/mcmc/hmc.jl
```diff
@@ -197,22 +197,6 @@ using Turing
...
-        # verbose=false to suppress the initial step size notification, which messes with
-        # the test
-        @test_logs (:warn, r"consider providing a different initialisation strategy") sample(
-            demo_warn_initial_params(), NUTS(), 5; verbose=false
-        )
-    end
-
     @testset "error for impossible model" begin
         @model function demo_impossible()
             x ~ Normal()
```
penelopeysm: This test can also be deleted.

0AnshuAditya0:
thanks for the feedback.

The change removes a random-value assertion and fixes the RNG seed so the test result is stable and not dependent on sampler randomness. The intent was to keep the behavior the same while making the test deterministic.

I’m still learning and will keep improving how I communicate changes. Thanks for taking the time to review this.

penelopeysm:
You incorporated a few of my changes in https://github.com/TuringLang/Turing.jl/pull/2740/commits/fafdf93c8777ba7bde2dc1ead742fafc0595191d (great!), but then you reverted them in the next commit https://github.com/TuringLang/Turing.jl/pull/2740/commits/b676d6a9daf3657eb155756dd481cd3986356596 and didn't put them back in again. I think the first commit seems quite reasonable. Is there a reason why you reverted?

0AnshuAditya0:
i reverted the commit while trying to simplify the tests
are you referring to @addlogprob! (abs(x) < 0.3) ? 0.0 : -Inf 
I added probabilistic approach. Should it be deterministic?

penelopeysm:
The test in the earlier commit was fine before you reverted it. No don't use abs(x) please, the Ref counter was exactly what's needed.

penelopeysm (changes requested):
There are still other things from my previous review which were in the reverted commit.

PR #2742, opened 2025-12-21
Use OnlyAccsVarInfo in check_model

penelopeysm:
Because the model is contextualised with InitContext, the call to check_model doesn't actually need anything more than an OnlyAccsVarInfo. This allows models that don't work well with Metadata* to still be checked for validity (note that the check, as defined in DynamicPPL, only uses DebugAccumulator, it does not actually use metadata in any form -- unless the model has DefaultContext as its leaf context -- which is not the case here).

(cc @PTWaade I think this is the reason why check_model failed)

\* one might ask what models don't work well with metadata, but still otherwise work. There are a couple of things that come to mind: 1) parallel assume statements; and 2) special distribution types for which the from_[linked_]_vec_transform functions are not defined. These models should in principle work with `Prior()` sampling, which only ever uses OnlyAccsVarInfo. However, currently, they don't unless you disable check_model; this is the behaviour that this PR fixes.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+# 0.42.1
+
+Avoid passing a full VarInfo to `check_model`, which allows more models to be checked safely for validity.
+
 # 0.42.0
 
 ## DynamicPPL 0.39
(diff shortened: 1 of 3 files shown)


PTWaade:
This makes sense ! Lightning fast work Penelope - and I who thought you weren't working currently !

penelopeysm:
ahhhh, it's only 3 lines so I made an exception 😄

PR #2743, opened 2026-01-02
Make FlexiChains the default chain type

penelopeysm:
Mostly interested in seeing what breaks, I ran some tests offline then got bored.

As far as source code changes go, I only needed to change the definition of `DEFAULT_CHAIN_TYPE` and twiddle with some imports. So in principle this is pretty much done.

Most changes were really in the test suite, using `chn[@varname(x)]` instead of `chn[:x]`. In fact, that [isn't even necessary](https://pysm.dev/FlexiChains.jl/stable/turing/#Indexing-by-Symbol:-a-shortcut), but I just consider it good practice.

The only changes that I found to be mandatory were changing things like `chn[Symbol("m[1]")]` to `chn[@varname(m[1])]`, which is a Positive Change, and also other things that relied too heavily on the exact internal structure of MCMCChains, which is also a Positive Change, or a Neutral Change in the cases where I had to rely on the internal structure of FlexiChains.

This PR should definitely target breaking, but that would cause version conflicts with FlexiChains, since FlexiChains has a compat entry that pins Turing to its current minor version. That can be fixed by moving FlexiChainsTuringExt into Turing proper, so that that compat entry can be removed.


penelopeysm:
The remaining test failures are also seen on main (it helps that I accidentally pushed this PR to main and reverted it...) and is (as tradition) something to do with the Optimization.jl family 😮‍💨

sunxd3 (approved):
LG! (it is breaking indeed)

penelopeysm:
Thanks @sunxd3! Yeah, it is actually even more work than this. I need to transfer the extensions into Turing/DynamicPPL as well. That should hopefully just be a copypaste but I'm sure it will break something...

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.44.5"
+version = "0.44.999"
```
penelopeysm: See initial comment

penelopeysm:
I probably won't have the time to finish this up.

(well, it seems it depends on whether I want to work on the weekend lol.)

Sequence of events:

- [x] Release a version of FlexiChains without Turing compat https://github.com/penelopeysm/FlexiChains.jl/pull/193
- [x] Release this version of Turing
- [ ] Update Turing docs

Project.toml
penelopeysm: Fixed now, however the release needs to be synchronised with https://github.com/penelopeysm/FlexiChains.jl/pull/193

PR #2748, opened 2026-01-07
remove get_matching_type overload

penelopeysm:
Terrifyingly, Turing overloaded an internal DynamicPPL function, `get_matching_type`. The function was renamed in a patch release, which is semver compliant. However, that makes Turing error when loaded, because there is no longer such a function to overload. This PR removes the overload (it was dead code and not accomplishing anything anyway).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+# 0.43.3
+
+Removes some dead code.
+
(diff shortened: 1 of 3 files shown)


mhauru:
Gee, thanks for catching this quickly, that was nasty.

PR #2755, opened 2026-01-17
Add code for AbstractMCMC Callbacks support

shravanngoswamii:


src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+    vns = keys(vi)
+    return Iterators.flatmap(vns) do vn
+        val = DynamicPPL.getindex_internal(vi, vn)
+        if val isa AbstractArray
+            [string(vn, "[", i, "]") => v for (i, v) in enumerate(val)]
+        else
+            [string(vn) => val]
+        end
+    end
+end
```
penelopeysm: You can use `AbstractPPL.varname_and_value_leaves` to do this sort of "split VarNames up":

```julia
iters = map(
    AbstractPPL.varname_and_value_leaves, keys(vi), values(vi),
)
vns_and_vals = mapreduce(collect, vcat, iters)
```

and then convert the VarNames into Strings.

But personally I think it's a mistake to do this.

Firstly, any interface that requires breaking up VarNames into sub-VarNames (individual real elements) is inherently dodgy. If you absolutely have to do so (e.g. for plotting purposes), it should be done as late as possible, i.e. immediately before you need to use it.

Secondly, the use of `getindex_internal` means that the results are prone to misinterpretation. VarInfo can store transformed values, that may be different from what a user thinks `x` represents, and `getindex_internal` will always return the transformed value. For example:

```julia
julia> using DynamicPPL, Distributions

julia> @model function f()
           x ~ Beta(2, 2)
           return (x, DynamicPPL.getindex_internal(__varinfo__, @varname(x)))
       end

julia> model = f(); vi = VarInfo(model);

julia> first(DynamicPPL.evaluate!!(model, vi))
(0.5759135088785118, [0.5759135088785118])

julia> linked = DynamicPPL.link!!(vi, model);

julia> first(DynamicPPL.evaluate!!(model, linked))
(0.5759135088785118, [0.30602006830078243])
```

Sometimes, not only does it have a different value, it is also completely transformed. This can happen with Dirichlet (the size will be different), or things like ProductNamedTupleDistribution (x will be a NamedTuple, but getindex_internal will always return a Vector). In such cases, claiming that the variable is called `x[1]` etc. can be misleading.

Thirdly, shouldn't `getparams` be returning a vector of Reals anyway (without the names) -- see below?

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
+using AbstractMCMC: AbstractMCMC
+
+function _get_lp(vi::DynamicPPL.AbstractVarInfo)
+    lp = DynamicPPL.getlogp(vi)
+    if lp isa NamedTuple
+        return sum(values(lp))
+    end
+    return lp
+end
```
penelopeysm: `lp` is always a NamedTuple in the current version of DynamicPPL.

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+            [string(vn, "[", i, "]") => v for (i, v) in enumerate(val)]
+        else
+            [string(vn) => val]
+        end
+    end
+end
+
+###
+### getparams - Extract named parameters from sampler states
+###
```
penelopeysm: This is inconsistent with the docstring of getparams, which says

> Retrieve the values of parameters from the sampler's `state` as a `Vector{<:Real}`.

https://github.com/TuringLang/AbstractMCMC.jl/blob/31db8941ab589101314345476915514d3e5b3716/src/AbstractMCMC.jl#L149-L154

Isn't it also inconsistent with what `names_and_values` expects here anyway?

https://github.com/TuringLang/AbstractMCMC.jl/blob/31db8941ab589101314345476915514d3e5b3716/src/callbacks.jl#L160-L161

I think overall what I would suggest is to just return a vector of parameters here and let the variable names just be theta[1], theta[2], ... Sure, it's not very informative; but (1) it can be improved upon later, and (2) under-promising is better than trying to overcommit by giving variable names that aren't necessarily correct.

If you do this, please put the implementations of getparams and getstats inside the individual samplers' files (i.e. put the methods for HMCState near the definition of the HMCState struct, etc.)

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+        LogDensityProblems.logdensity(state.logdensity, state.params)
+    catch
+        NaN
+    end
+    return (lp=lp, step=state.step)
+end
+
+###
+### getparams/getstats from transitions (ParamsWithStats)
+###
```
penelopeysm: Why are these methods needed -- when do we need to call these functions on transitions?

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,232 @@
...
+    @testset "getparams from states" begin
+        @testset "HMCState getparams" begin
+            Random.seed!(42)
+            chain = sample(gdemo_default, NUTS(100, 0.65), 10; progress=false)
+
+            rng = Random.default_rng()
+            transition, state = AbstractMCMC.step(
+                rng,
+                gdemo_default,
+                NUTS(100, 0.65);
```
penelopeysm: 1. The call to `seed!` is not needed since you aren't checking the values it takes. If you don't need the RNG to be predictable, don't put it in the test suite, as otherwise it can imply that it needs to be predictable.

2. I think you can loop over the samplers and use the same model `@model f() = x ~ Normal()` for all of them, to avoid the code repetition.

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+    return [string(vn) => val for (vn, val) in transition.params]
+end
+
+function AbstractMCMC.getstats(transition::DynamicPPL.ParamsWithStats)
+    return transition.stats
+end
+
+###
+### hyperparam_metrics - Define TensorBoard hyperparam metrics
+###
```
penelopeysm: If these overloads are only needed for TensorBoard, they can be moved into a Turing tensorboard extension, no need to keep it in src.

In general code should be given 'as little power as possible' -- if there's no need for anybody to import and use this code outside of the TensorBoard ext, then it should be in the extension.

penelopeysm (changes requested):
As far as I can tell, the main reason for all of this code is to satisfy `names_and_values`, which is only used in the TensorBoard callback.

Firstly, I think the AbstractMCMC part should be changed. `names_and_values` should be moved into the TensorBoard extension. It's not used anywhere else in `AbstractMCMC/src`, and it's not meant to be overloaded. On the bright side, you did declare it internal so you can do this without calling it a breaking change in semver.

Once that's done, we "only" need to overload the following methods:

 - getparams and getstats -- okay (modulo my comment below about the return type of getparams)
 - hyperparams_metric and _hyperparams_impl -- not so okay. These functions are not exported, and are not obviously part of the AbstractMCMC API. The latter has a comment above it saying "internal". That means that if AbstractMCMC changes them in a patch or minor release, which could conceivably be semver compliant from the AbstractMCMC side, then Turing code will suddenly stop working. This is actually quite a common occurrence: just a week ago I had to fix this case where we overloaded a DynamicPPL internal function https://github.com/TuringLang/Turing.jl/pull/2748.

Before this can be merged into Turing, I would at least like to see the hyperparams_... functions in AbstractMCMC be made part of a public API.

I'm less fussed about this, but I also think that it should also be made really explicit, by naming, that their existence is solely for the TensorBoard logging. That is, in `src` you can do:

```julia
function tb_hyperparam_metric end
```

and then the default definition can live in AbstractMCMCTensorBoardExt.

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,232 @@
...
+using Random: Random
+using DynamicPPL: DynamicPPL
+
+using ..Models: gdemo_default
+
+@testset "AbstractMCMC Callbacks Interface" begin
+    @testset "getparams from states" begin
+        @testset "HMCState getparams" begin
+            Random.seed!(42)
+            chain = sample(gdemo_default, NUTS(100, 0.65), 10; progress=false)
```
penelopeysm:
```suggestion
```

This `sample` call seems unused?

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+    ]
+end
+
+function AbstractMCMC.hyperparam_metrics(model::DynamicPPL.Model, sampler::MH)
+    return ["extras/lp/stat/Mean"]
+end
+
+function AbstractMCMC.hyperparam_metrics(model::DynamicPPL.Model, sampler::PG)
+    return ["extras/lp/stat/Mean", "extras/logevidence/stat/Mean"]
+end
```
yebai: @shravanngoswamii I suggest removing `AbstractMCMC.hyperparam_metrics` entirely. For this purpose, hyperparameters can be treated in the same manner as model parameters, with `stats_options` used to specify which statistics should be computed. This avoids maintaining a separate mechanism for hyperparameter metrics and simplifies the interface.

src/mcmc/callbacks.jl
```diff
@@ -0,0 +1,227 @@
...
+
+###
+### _hyperparams_impl - Extract sampler hyperparameters
+###
+
+function AbstractMCMC._hyperparams_impl(
```
yebai: EDIT: See https://github.com/TuringLang/Turing.jl/pull/2755#discussion_r2705557805

~~I propose unifying `AbstractMCMC._hyperparams_impl` with `AbstractMCMC.getstats` (see the references linked above). At present, `getstats` is defined only for a single argument. A natural extension would be to introduce~~

```julia
AbstractMCMC.getstats(state, sampler)
```

~~and to omit the model argument from `AbstractMCMC._hyperparams_impl`, as it is not required for this functionality.~~

~~Note that some hyperparameters may be constant across MCMC steps. This is compatible with the proposed interface: such hyperparameters can simply return the same value at each step without special handling. Hyperparameters that are adapted during the warmup phase may vary across warmup iterations, but are expected to remain fixed thereafter.~~

src/mcmc/callbacks.jl
yebai: Correct naming is important for visualization purposes. @penelopeysm, do you think it would make sense to introduce a two-argument version of `getparams`, for example:

```julia
AbstractMCMC.getparams(state, Val{True}) = ... # return both names and values as Pairs 

AbstractMCMC.getparams(state, Val{False})  = AbstractMCMC.getparams(state)
```

This would allow callers to explicitly request parameter names when needed, while preserving the existing behavior by default.

yebai:
Thanks @shravanngoswamii — I’ve added a few refactoring comments above. The functionality is already in place, but it could benefit from some minor cleanup, which I agree with Penny on.

src/mcmc/callbacks.jl
penelopeysm: Sure, I agree the interface should be extended. I find val{true} and val{false} to not be very descriptive though, and generally I prefer to have functions with different names. I had half an interface worked out for this kind of thing before -- I would suggest the following

```julia
# Default implementation, can be overloaded if needed. Sometimes it doesn't really
# make sense to provide names e.g. if you are using AdvancedHMC directly.
function getnamedparams(state)
    params = getparams(state)
    default_names = ["x[$i]" for i in 1:length(params)]
    return zip(default_names, params)
end
# Can use the model as well if needed (but in most cases you shouldn't need to since
# the state should generally contain the necessary info).
function getnamedparams(model::AbstractModel, state)
    params = getparams(model, state)
    default_names = ["x[$i]" for i in 1:length(params)]
    return zip(default_names, params)
end

# Samplers must implement either getparams(model, state) or getparams(state).
# This is already the behaviour in current AbstractMCMC, changing this would be
# breaking, so please don't change it
function getparams(model::AbstractModel, state)
    return getparams(state)
end
```

Downstream functions should only use the two-argument versions (which allows the sampler to overload whichever method is most relevant for them).

src/mcmc/callbacks.jl
penelopeysm: Agree, I'm happy to bundle it into stats.

src/mcmc/callbacks.jl
yebai: In that case, it may be preferable to make a breaking release so that `AbstractMCMC.getparams(model::AbstractModel, state)` and `AbstractMCMC.getparams(state)` always return both names and values. If names are missing, we can default to `"θ[$i]"`. This would also make the interface consistent with `AbstractMCMC.getstats`.

However, this would likely be too much of a distraction for @shravanngoswamii at the moment, as it would require updating all existing `getparams` implementations across MCMC packages. As an interim solution, could we instead adopt the following convention:

```julia
AbstractMCMC.getparams(state, model::AbstractModel)  # return names and values
AbstractMCMC.getparams(state)                        # return values as a vector of floats 
```

We could then revisit this and update `AbstractMCMC.getparams`to always return both names and values in the next breaking release.

penelopeysm: Instead of switching the arguments around, isn't it just cleaner to have a separate function?

This feels like an easy way to get accidental bugs by writing the order of the arguments wrongly and having it do something weird instead of throwing a MethodError.

yebai: Okay, the most friendly option for @shravanngoswamii is probably to tweak [`_names_and_values`](https://github.com/TuringLang/AbstractMCMC.jl/blob/31db8941ab589101314345476915514d3e5b3716/docs/src/callbacks.md?plain=1#L244) and export it. For example:

```julia
names_and_values(
    model,
    sampler,
    transition,
    state;
    param = true,
    stats = false,
    hyperparams = false,
    extra = false,
)
```


At some point, we should unify `getparams` and `getstats` into `names_and_values` in some form.

shravanngoswamii: Thank you @penelopeysm @yebai. I have exported `names_and_values` in AbstractMCMC (https://github.com/TuringLang/AbstractMCMC.jl/pull/191)

src/mcmc/callbacks.jl
shravanngoswamii: Not really, removed them!

src/mcmc/callbacks.jl
shravanngoswamii: Transition metrics like `tree_depth`, `n_steps`, etc... are not logged anymore, Should we expose transition stats?

shravanngoswamii:
CI failure is not relevent to this PR.

src/mcmc/ess.jl
```diff
@@ -118,3 +118,16 @@ function AbstractMCMC.step(
...
+#####
+
+function AbstractMCMC.getparams(state::DynamicPPL.AbstractVarInfo)
+    return _get_params_vector(state)
+end
+
+function AbstractMCMC.getstats(state::DynamicPPL.AbstractVarInfo)
+    lp = _get_lp(state)
+    return (lp=lp,)
+end
```
yebai:
```suggestion
```

src/mcmc/Inference.jl
```diff
@@ -115,6 +115,18 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
...
+function _get_lp(vi::DynamicPPL.AbstractVarInfo)
+    lp = DynamicPPL.getlogp(vi)
+    return sum(values(lp))
+end
+
+# Helper to extract raw parameter values from VarInfo as Vector{<:Real}
+function _get_params_vector(vi::DynamicPPL.AbstractVarInfo)
+    return vi[:]
+end
+
```
yebai, line 137:
```suggestion
function AbstractMCMC.getparams(state::DynamicPPL.AbstractVarInfo)
    return _get_params_vector(state)
end

function AbstractMCMC.getstats(state::DynamicPPL.AbstractVarInfo)
    lp = _get_lp(state)
    return (lp=lp,)
end
```

src/mcmc/callbacks.jl
yebai: yes

src/mcmc/hmc.jl
```diff
@@ -505,3 +505,21 @@ end
...
+end
+
+function AbstractMCMC.getstats(state::HMCState)
+    lp = _get_lp(state.vi)
+    ϵ = try
+        state.kernel.τ.integrator.ϵ
+    catch
+        NaN
+    end
+    return (lp=lp, step_size=ϵ, iteration=state.i)
```
```suggestion
    # TODO(penelopeysm): For many Hamiltonian samplers you can get the
    # stepsize from the state. However for non-adaptive Hamiltonians the
    # info is only in the sampler and thus can't be accessed via getstats.
    # HMCState should be modified to include the step size as a field, which
    # would allow us to access this information here properly.
    ϵ = try
        state.kernel.τ.integrator.ϵ
    catch
        NaN
    end
    return (lp=lp, step_size=ϵ, iteration=state.i)
```

penelopeysm: I think the NaN fallback is not very nice. I'm happy to take responsibility for changing it in the future, but just wanted to add a comment here to make sure that we're aware of this.

src/mcmc/Inference.jl
```diff
@@ -115,6 +115,25 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
...
+    return sum(values(lp))
+end
+
+# Helper to extract raw parameter values from VarInfo as Vector{<:Real}
+function _get_params_vector(vi::DynamicPPL.AbstractVarInfo)
+    return vi[:]
+end
+function AbstractMCMC.getparams(state::DynamicPPL.AbstractVarInfo)
+    return _get_params_vector(state)
+end
```
penelopeysm:
```suggestion
function AbstractMCMC.getparams(state)
    return Turing.Inference.get_varinfo(state)[:]
end
```

There's actually code in Turing to extract a VarInfo from a sampler state, because the Gibbs sampler needs it. You can use this directly here to avoid duplicating the definitions of `state.vi`, `state.varinfo`, etc. etc. You can grep the codebase for `get_varinfo` to see the definitions.

It might not cover the really tiny samplers, but having this single definition will allow you to get rid of all the other getparams definitions.

Also, your helper function is actually making the code longer :-)

src/mcmc/hmc.jl
```diff
@@ -505,3 +505,21 @@ end
...
     return AHMC.Adaptation.NoAdaptation()
 end
+
+#####
+##### AbstractMCMC interface
+#####
+
+function AbstractMCMC.getparams(state::HMCState)
+    return _get_params_vector(state.vi)
+end
```
penelopeysm: e.g. this won't be needed anymore

src/mcmc/particle_mcmc.jl
```diff
@@ -519,3 +519,28 @@ Libtask.might_produce(::Type{<:Tuple{typeof(DynamicPPL.tilde_assume!!),Vararg}})
...
+# SMCState
+function AbstractMCMC.getparams(state::SMCState)
+    particle = state.particles.vals[state.particleindex]
+    vi = particle.model.f.varinfo
+    return _get_params_vector(vi)
+end
+
+function AbstractMCMC.getstats(state::SMCState)
+    return (logevidence=state.average_logevidence, particle_index=state.particleindex)
+end
```
penelopeysm: personally i would not bother with SMCState , it is not an MCMC sampling method and it behaves a bit quirkily anyway because it doesn't fit into the conventional AbstractMCMC interface. if you run SMC with N particles, all N "steps" are run on the "first" "MCMC step" and then all the remaining steps are basically no-ops apart from incrementing the particle index. so there is no point in monitoring this in an 'online' manner

src/mcmc/sghmc.jl
```diff
@@ -217,3 +217,35 @@ function AbstractMCMC.step(
...
+end
+
+function AbstractMCMC.getstats(state::SGHMCState)
+    lp = try
+        LogDensityProblems.logdensity(state.logdensity, state.params)
+    catch
+        NaN
+    end
+    return (lp=lp,)
+end
```
penelopeysm:
```suggestion
function AbstractMCMC.getstats(state::SGHMCState)
    # TODO(penelopeysm): This is inefficient as it requires an extra model evaluation
    lp = LogDensityProblems.logdensity(state.logdensity, state.params)
    return (lp=lp,)
end
```

No need to try catch. If the model errors when calculating the logp the user has bigger problems to worry about!

src/mcmc/sghmc.jl
```diff
@@ -217,3 +217,35 @@ function AbstractMCMC.step(
...
 end
+
+#####
+##### AbstractMCMC interface
+#####
+
+# SGHMCState
+function AbstractMCMC.getparams(state::SGHMCState)
+    return collect(state.params)
+end
```
penelopeysm:
```suggestion
AbstractMCMC.getparams(state::SGHMCState) = state.params
```

I think we don't need collect, right? `state.params` is already an `AbstractVector{<:Real}`.

src/mcmc/sghmc.jl
```diff
@@ -217,3 +217,35 @@ function AbstractMCMC.step(
...
+
+function AbstractMCMC.getstats(state::SGLDState)
+    lp = try
+        LogDensityProblems.logdensity(state.logdensity, state.params)
+    catch
+        NaN
+    end
+    return (lp=lp, step=state.step)
+end
```
penelopeysm:
```suggestion
function AbstractMCMC.getstats(state::SGLDState)
    # TODO(penelopeysm): Remove extra evaluation.
    lp = LogDensityProblems.logdensity(state.logdensity, state.params)
    return (lp=lp, step=state.step)
end
```

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,47 @@
+module CallbacksTests
+
+using Test, Turing, AbstractMCMC, Random
+
+if !isdefined(@__MODULE__, :Models)
+    include(joinpath(@__DIR__, "..", "test_utils", "models.jl"))
+    using .Models: gdemo_default
+end
+
```
penelopeysm: If gdemo_default is the only model you need, I think it's easier to just copy over the definition, no need to mess with this stuff (I've always found it pretty annoying that the models are defined in a test file, which makes it hard to modularise stuff).

Alternatively, you can use

```julia
GDEMO_DEFAULT = DynamicPPL.TestUtils.demo_assume_observe_literal()
```

It's actually exactly the same model.

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,47 @@
...
+    for (name, sampler, model) in samplers
+        @testset "$name Interface" begin
+            transition, state = AbstractMCMC.step(
+                rng, model, sampler; initial_params=Turing.Inference.init_strategy(sampler)
+            )
+
+            # Should return a flat vector of Reals (unconstrained)
+            params = AbstractMCMC.getparams(state)
+            @test params isa Vector{<:Real}
```
penelopeysm:
```suggestion
            @test params isa AbstractVector{<:Real}
```

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,47 @@
...
+
+    samplers = [
+        ("NUTS", NUTS(10, 0.65), gdemo_default),
+        ("HMC", HMC(0.1, 5), gdemo_default),
+        ("MH", MH(), gdemo_default),
+        ("ESS", ESS(), simple_gaussian()),
+        ("Gibbs", Gibbs(:m => HMC(0.1, 5), :s => MH()), gdemo_default),
+        ("SGHMC", SGHMC(; learning_rate=0.01, momentum_decay=1e-2), gdemo_default),
+        ("PG", PG(10), gdemo_default),
+    ]
```
penelopeysm, line 22: Is there a reason why we don't use the same model for all of these? I know ESS only takes normal distributions, but we could have a single model that looks like, say

```julia
@model function some_normals()
    x ~ Normal()
    y ~ MvNormal(zeros(3), I)
end
```

And just use this same model for all of them. The good thing about this, is you can actually check certain properties about the output. For example...

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,47 @@
...
+    for (name, sampler, model) in samplers
+        @testset "$name Interface" begin
+            transition, state = AbstractMCMC.step(
+                rng, model, sampler; initial_params=Turing.Inference.init_strategy(sampler)
+            )
+
+            # Should return a flat vector of Reals (unconstrained)
+            params = AbstractMCMC.getparams(state)
+            @test params isa Vector{<:Real}
+            @test !isempty(params)
```
penelopeysm: If you used the model above, then you could check here that the length of the params is exactly 4.

test/mcmc/callbacks.jl
```diff
@@ -0,0 +1,47 @@
...
+
+            # Should return a NamedTuple with at least log probability (:lp)
+            stats = AbstractMCMC.getstats(state)
+            @test stats isa NamedTuple
+            @test haskey(stats, :lp)
+            @test stats.lp isa Real
```
penelopeysm: You could also check here that `stats.lp` == `sum(logpdf.(Normal(), params))` -- which it should be -- otherwise getstats is returning something invalid!

src/mcmc/Inference.jl
shravanngoswamii: Ohh, I didn't explored it properly, Thanks for letting me know! Will update it

src/mcmc/hmc.jl
shravanngoswamii: Got it!

penelopeysm:
Thanks @shravanngoswamii -- this is looking nice and clean now!

There is one point that I think we should discuss, which is the meaning of the 'params' and 'stats'. In particular, for samplers which operate in linked space, what do we return? e.g. does getparams return a vector where x is always between 0 or 1? (if this is confusing, there's an explanation here https://turinglang.org/docs/developers/transforms/distributions/)

```julia
@model function f()
    x ~ Beta(2, 2)
end
```

Currently, the implementation of getparams and getstats returns whatever the sampler stores internally. That means that some samplers will have unlinked values (e.g. default MH, which samples statically from the prior in unlinked space) and some samplers will have linked values (e.g. NUTS). The same is true of stats -- currently we return the sampler's notion of `lp` which may or may not be linked.

I don't mind as long as we are consistent, for example, we could say that getstats and getparams for Turing samplers always return whatever is internal to the sampler -- but I do think that it is **much** better from a user perspective to get the unlinked values since the internal values have no external meaning. We used to return internal log probs in the MCMCChains object, but we switched it over to returning 'true' unlinked lps recently, and I think that was a good thing.

src/mcmc/hmc.jl
shravanngoswamii: Thank you! added a TODO added

shravanngoswamii:
Thank you @penelopeysm.

getparams(state) returns get_varinfo(state)[:]:
- HMC/NUTS/SGHMC/SGLD: Linked (unconstrained) -- these samplers explicitly link the VarInfo
- MH() (default/static proposals): Unlinked -- should_link returns false
- MH(covariance_matrix) or explicit RandomWalkProposal: Linked
- ESS/PG: Unlinked

getstats(state).lp is also inconsistent:
- HMC: sum(getlogp(vi)) = logprior + logjac + loglikelihood
- MH: logjoint_internal = logprior + loglikelihood - logjac

However, ParamsWithStats (the transition) already does the right thing -- DynamicPPL.ParamsWithStats re-evaluates the model via ValuesAsInModelAccumulator and returns unlinked parameter values with proper logjoint (no Jacobian artifacts).

For callback-based monitoring, users interact with ParamsWithStats, not the raw state. So the user-facing behavior is correct.

Maybe, we can document that getparams/getstats return internal sampler state (which may be in transformed space), while ParamsWithStats provides semantically correct values. This matches the intent that these are low-level accessors vs. high-level transition containers.

Happy to add this documentation if that sounds reasonable.

shravanngoswamii:
Tests will pass after https://github.com/TuringLang/AbstractMCMC.jl/pull/191

```julia
> julia --project test\mcmc\callbacks.jl          
└   ϵ = 1.6500000000000001
└   ϵ = 1.6500000000000001
Test Summary:                    | Pass  Total   Time
AbstractMCMC Callbacks Interface |   80     80  38.3s
```

penelopeysm:
> However, ParamsWithStats (the transition) already does the right thing -- DynamicPPL.ParamsWithStats re-evaluates the model via ValuesAsInModelAccumulator and returns unlinked parameter values with proper logjoint (no Jacobian artifacts). For callback-based monitoring, users interact with ParamsWithStats, not the raw state. So the user-facing behavior is correct.

Does it? Don't all the methods `getstats(...)`, etc., act on the state, not the transition? And that's what the callback framework uses?

src/mcmc/Inference.jl
```diff
@@ -115,6 +115,41 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
     return log(rand()) + logp_current ≤ logp_proposal + log_proposal_ratio
 end
 
+# Helper functions for AbstractMCMC callbacks
+# Helper to get log probability from VarInfo
+function _get_lp(vi::DynamicPPL.AbstractVarInfo)
+    lp = DynamicPPL.getlogp(vi)
+    return sum(values(lp))
+end
```
penelopeysm: I missed this, but your comment was right: this gives prior + likelihood + logjac which is statistically meaningless.

I don't think there's a good way to do this, unfortunately, this might be the easiest:

```suggestion
function _get_lp(vi::DynamicPPL.AbstractVarInfo)
    lp = DynamicPPL.getlogp(vi)
    if haskey(lp, :logjac)
        lp = merge(lp, (; logjac = -lp.logjac))
    end
    return sum(values(lp))
end
```

shravanngoswamii:
> Does it? Don't all the methods `getstats(...)`, etc., act on the state, not the transition? And that's what the callback framework uses?

I think the callback behavior is correct. The TensorBoard callback in AbstractMCMC calls `ParamsWithStats(model, sampler, transition, state)` and Turing overrides this at [Inference.jl](https://github.com/TuringLang/Turing.jl/pull/2755/changes#diff-fb21abc3e1f43bc81f6226ba9c0553a8347138069057316a5154355797e95acdR138-R152) for `transition::DynamicPPL.ParamsWithStats`.

This override extracts `transition.params` and `transition.stats` directly—it doesn't call `getparams(state)` or `getstats(state)`.

src/mcmc/Inference.jl
```diff
@@ -115,6 +115,46 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
...
+    if haskey(lp, :logjac)
+        lp = merge(lp, (; logjac=-lp.logjac))
+    end
+    return sum(values(lp))
+end
+
+# Consolidated getparams using get_varinfo (defined in each sampler file)
+# This covers HMC, MH, PG, Gibbs, and external samplers.
+# SGHMC/SGLD have their own implementations since they don't use VarInfo.
+function AbstractMCMC.getparams(state)
```
yebai: Please double-check whether we are missing some other MCMC states? 

```suggestion
function AbstractMCMC.getparams(state::Union{HMCState, MHState, PGState, GibbsState})
```

shravanngoswamii: > Please double-check whether we are missing some other MCMC states? 

I think we covered all!

penelopeysm: the state structs haven't been defined at this point, so ci is failing.

if you want to have a generic method to save code duplication you could give it a different (internal) name here, don't put a type bound on it, and then use it in the definition of getparams -- pretty much exactly like what's going on with _get_lp now

src/mcmc/Inference.jl
```diff
@@ -115,6 +115,45 @@ function mh_accept(logp_current::Real, logp_proposal::Real, log_proposal_ratio::
...
+    params::Bool=true,
+    stats::Bool=false,
+    extras::Bool=false,
+)
+    # Convert OrderedDict params to Vector{Pair} for AbstractMCMC.ParamsWithStats constructor
+    p = params ? [string(k) => v for (k, v) in transition.params] : nothing
+    s = stats ? transition.stats : NamedTuple()
+    e = extras ? NamedTuple() : NamedTuple()
+    return AbstractMCMC.ParamsWithStats(p, s, e)
+end
```
penelopeysm, line 136: > The TensorBoard callback in AbstractMCMC calls ParamsWithStats(model, sampler, transition, state) and Turing overrides this

In this case, since this method is overridden, the original method in AbstractMCMC never gets called, so why do we need to define getparams and getstats on the states?

penelopeysm: Also, I just suggest adding a comment:

```suggestion
# Directly overload the constructor of `AbstractMCMC.ParamsWithStats` so that we don't
# hit the default method, which uses `getparams(state)` and `getstats(state)`. For Turing's
# MCMC samplers, the state might contain results that are in linked space. Using the
# outputs of the transition here ensures that parameters and logprobs are provided in
# user space (similar to chains output).
function AbstractMCMC.ParamsWithStats(
    model,
    sampler,
    transition::DynamicPPL.ParamsWithStats,
    state;
    params::Bool=true,
    stats::Bool=false,
    extras::Bool=false,
)
    # Convert OrderedDict params to Vector{Pair} for AbstractMCMC.ParamsWithStats constructor
    p = params ? [string(k) => v for (k, v) in transition.params] : nothing
    s = stats ? transition.stats : NamedTuple()
    e = extras ? NamedTuple() : NamedTuple()
    return AbstractMCMC.ParamsWithStats(p, s, e)
end
```

shravanngoswamii: default AbstractMCMC implementation is never called for built-in samplers. The getparams(state) and getstats(state) methods aren't used in the callback path.
I kept them so users can call them directly and for consistency (setparams!!(state, getparams(state)) == state)
If you think they're unnecessary overhead, we could remove them. Or keep them for API completeness

penelopeysm: If they're never called, then I'd much prefer for them to be removed, since otherwise it's kind of confusing as to why they're there.

penelopeysm: (to be clear, the MCMC states are not part of Turing's public API and there's no expectation that users will use those structs directly, and also no expectation that we provide functionality for other people to use those structs in any way that we don't already do. The only functionality that we should define for those structs is whatever that is needed to make Turing's public API work, so things like sample(), mode_estimate(), etc.)

shravanngoswamii: Okay, removed them.

shravanngoswamii:
Right now, the ParamsWithStats override:
- Extracts params from transition.params
- Extracts stats from transition.stats (includes logprior, loglikelihood, logjoint, and AHMC diagnostics like tree_depth, n_steps, acceptance_rate)
- Returns empty extras (always NamedTuple())
- Ignores state entirely

I am not sure if ignoring state is acceptable? Is the current design acceptable, or should we populate extras with static sampler configuration from state?
What are your thoughts @penelopeysm @yebai ?

penelopeysm:
Since it's nonbreaking to add extra info in, I would be OK with this current implementation right now, and afterwards if there's a need for more information that can be changed later / in a separate PR.

yebai:
I’m happy with the PR!

penelopeysm:
Why the minor version bump?

shravanngoswamii:
> Why the minor version bump?

New funtion overloaded should be patch or minor?

penelopeysm:
It's backwards compatible new functionality. If the version number is 0.x, it only needs to be a patch bump. If the version number is N.x (N ≥ 1) it only needs to be a minor bump.

But semver says you do have a choice: if you prefer to release it as a minor version 0.(x+1) or a major version (N+1), you can.

If you do choose to release as a minor version, please can you rebase it against `breaking`? See description in https://turinglang.org/docs/developers/contributing/#pull-requests-versions-and-releases (although we only really follow this for Turing and DPPL right now)

penelopeysm:
I would recommend not doing it as a minor version though, as the next minor version will take a while to be done, since I have to do #2756 and that is very difficultl

shravanngoswamii:
> But semver says you do have a choice: if you prefer to release it as a minor version 0.(x+1) or a major version (N+1), you can.

Okay, I get it! Thanks for all the reviews, @penelopeysm. Let me know if you’re okay with merging this.

PR #2756, opened 2026-01-25
Run CI against new DPPL / VNT

penelopeysm:
The usual game of trying to see if DPPL breaks anything upstream.

(20 minutes later)

Okay, maybe '*if* DPPL breaks anything upstream' was a bit optimistic. More like seeing *what* DPPL broke upstream.


penelopeysm:
Ok, MH is totally borked. I refrain from using a stronger word. It basically needs to be rewritten. I'd say though, that InitContext actually gives us a really good way of doing this.

penelopeysm:
Turing officially works with VNTs 🎉

PR #2759, opened 2026-01-28
Avoid reevaluating model where unnecessary

### LINKED ISSUE #2639: Gibbs wastefully re-evaluates model once per component sampler per iteration
In these lines, component samplers (usually) return a tuple of `(Turing.Inference.Transition, SamplerState)`.

The construction of `Turing.Inference.Transition` requires a model re-evaluation (because it runs `values_as_in_model`, and after #2630, also recalculates logp).

However, in the Gibbs sampler we just throw that away:

https://github.com/TuringLang/Turing.jl/blob/d0510b1916d78f76b3fca67deceb47c22b0aab04/src/mcmc/gibbs.jl#L442-L451

https://github.com/TuringLang/Turing.jl/blob/d0510b1916d78f76b3fca67deceb47c22b0aab04/src/mcmc/gibbs.jl#L678-L679

We should either:

(1) Avoid triggering the Transition re-evaluation unless it's needed (not obvious how to do it).
(2) Investigate whether we can use the Transition return value to avoid the re-evaluation in `setparams_varinfo!!`.

Note that (1) may require storing the new varinfo itself in the `Turing.Inference.Transition` struct.

penelopeysm:
Closes #2639

Quoting from the changelog:

> Avoid reevaluating the model on MCMC iterations where the transition is not saved to the chain (e.g. in initial burn-in, or when using thinning).
> Also avoid each component sampler of Gibbs unnecessarily evaluating the model once per iteration.

This makes use of the `discard_sample` keyword argument to `step` / `step_warmup` https://github.com/TuringLang/AbstractMCMC.jl/pull/194

Note that this is still type stable as the value of the `discard_sample` kwarg can be constant propagated (I've checked this with code_warntype etc).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+# 0.42.7
+
+Avoid reevaluating the model on MCMC iterations where the transition is not saved to the chain (e.g. in initial burn-in, or when using thinning).
+Also avoid each component sampler of Gibbs unnecessarily evaluating the model once per iteration.
+
 # 0.42.6
 
 Fixed a bug in SMC and PG where results were not always stored correctly in Libtask traces (due to incorrect `objectid` checks).
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Turing"
 uuid = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
-version = "0.42.6"
+version = "0.42.7"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/Project.toml
+++ b/Project.toml
@@ -47,7 +47,7 @@ TuringDynamicHMCExt = "DynamicHMC"
 
 [compat]
 ADTypes = "1.9"
-AbstractMCMC = "5.9"
+AbstractMCMC = "5.13"
 AbstractPPL = "0.11, 0.12, 0.13"
 Accessors = "0.1"
 AdvancedHMC = "0.8.3"
--- a/src/mcmc/emcee.jl
+++ b/src/mcmc/emcee.jl
@@ -46,7 +46,12 @@ function Turing.Inference._convert_initial_params(
 end
 
 function AbstractMCMC.step(
-    rng::Random.AbstractRNG, model::Model, spl::Emcee; initial_params, kwargs...
+    rng::Random.AbstractRNG,
+    model::Model,
+    spl::Emcee;
+    initial_params,
+    discard_sample=false,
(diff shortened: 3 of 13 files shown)


penelopeysm:
Assuming this passes CI, @joelkandiah, you might find this a little bit useful. It doesn't help with compilation, but here's a demonstration:

```julia
using Turing

modelevals = Ref(0)

@model function f()
    modelevals[] += 1
    a1 ~ Normal()
    a2 ~ Normal()
    a3 ~ Normal()
    a4 ~ Normal()
    a5 ~ Normal()
    a6 ~ Normal()
    a7 ~ Normal()
    a8 ~ Normal()
end

modelevals[] = 0
@time sample(f(), Gibbs(
        @varname(a1) => MH(),
        @varname(a2) => MH(),
        @varname(a3) => MH(),
        @varname(a4) => MH(),
        @varname(a5) => MH(),
        @varname(a6) => MH(),
        @varname(a7) => MH(),
        @varname(a8) => MH()
    ),
    1000;
    chain_type=Any, progress=false
);
@show modelevals[]

# this PR: 0.274875 seconds (2.26 M allocations: 169.478 MiB, 4.95% gc time)
# modelevals[] = 17003
#
# main: 0.354257 seconds (2.88 M allocations: 208.250 MiB, 4.81% gc time)
# modelevals[] = 25003
```

If you're using discard_initial and/or thinning, the benefits should scale accordingly. For example, adding `thinning=10` to the above gives:

```julia
# With thinning=10
#
# this PR: 2.419929 seconds (21.36 M allocations: 1.582 GiB, 4.45% gc time)
# modelevals[] = 160859
#
# main: 3.395153 seconds (28.70 M allocations: 2.029 GiB, 4.05% gc time)
# modelevals[] = 249778
```

penelopeysm:
(I definitely think it should in theory be possible to cut the model evaluations down to 8000, or maybe 9000 (each component sampler goes once and Gibbs itself has to do it once), but that really requires deeper refactoring that would follow on from #2756. The extra 8000 evaluations are coming from this `setparams_varinfo!!` stuff that complicates matters.)

penelopeysm:
Turing v0.43 bumped the model evaluations back up to 25000 (it's because `match_linking!!` requires reevaluation), but happily #2803 cuts that back down to 17000.

Timings (1000 iterations, no thinning):

Main (v0.43.5) 0.062175 seconds (922.58 k allocations: 67.218 MiB, 13.04% gc time)
#2803 0.045499 seconds (666.81 k allocations: 44.957 MiB, 8.64% gc time)

PR #2770, opened 2026-02-09
Pass stats through to Gibbs

### LINKED ISSUE #2766: Stats from component samplers are lost with Gibbs
https://github.com/TuringLang/Turing.jl/blob/37af7fdb36a85fcc4e1aae770801a42ef5d5d595/src/mcmc/gibbs.jl#L323-L335

Here Gibbs returns a ParamsWithStats that only has parameters but no stats. In principle, it should aggregate stats from all component samplers, combine them into a single NamedTuple and return them.

This is incredibly annoying though because stats are provided via component samplers' transitions, not states, and #2759 makes sure that we *don't* actually use the **parameters** in component samplers' transitions, because that often requires reevaluating the model. Very, **very** annoying. The solution to this is to unify transitions and states, see https://github.com/TuringLang/AbstractMCMC.jl/issues/149, but, mehhhhhh.

penelopeysm:
This closes #2766 but at the cost of increasing model evaluations (it just undoes https://github.com/TuringLang/Turing.jl/pull/2759), so this is not really a correct fix.

Each component sampler's stats is prefixed with `splN` where `N = 1...` is the order of the sampler in the Gibbs constructor.

The problem is that

- `AbstractMCMC.step` for each component sampler returns a tuple of `(transition, state)`
- The transition contains both parameters and stats
- The parameters often (but don't always) require model reevaluation
- Gibbs doesn't need the parameters, but needs the stats
- However you can't get only the stats without the parameters.

So if you want to extract the sampler statistics from `AbstractMCMC.step`, you often need to reevaluate the model.

The correct solution is to get rid of transitions (https://github.com/TuringLang/AbstractMCMC.jl/issues/149), and put the stats into the state as well. In general, the state should be an object which doesn't reevaluate the model, but contains enough information to allow a downstream user to reevaluate the model if they actually need the parameters. In essence the state should be a lazy version of what's now `(transition, state)`.

### DIFF
--- a/src/mcmc/gibbs.jl
+++ b/src/mcmc/gibbs.jl
@@ -320,7 +320,7 @@ function AbstractMCMC.step(
     samplers = spl.samplers
     vi = initial_varinfo(rng, model, spl, initial_params)
 
-    vi, states = gibbs_initialstep_recursive(
+    vi, states, stats = gibbs_initialstep_recursive(
         rng,
         model,
         AbstractMCMC.step,
@@ -330,7 +330,7 @@ function AbstractMCMC.step(
         initial_params=initial_params,
         kwargs...,
     )
-    transition = discard_sample ? nothing : DynamicPPL.ParamsWithStats(vi, model)
+    transition = discard_sample ? nothing : DynamicPPL.ParamsWithStats(vi, model, stats)
     return transition, GibbsState(vi, states)
 end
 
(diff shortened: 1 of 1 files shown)


penelopeysm:
This won't be merged, but by putting up a PR I think I've accomplished what I set out to do, which is to demonstrate the idea.

PR #2774, opened 2026-02-23
Tell users about the proposals that are being used in MH

penelopeysm:
This (I think) is the best way to protect against unexpected behaviour when the user specifies proposals in a different format from the VarNames in the model. For example

```julia
using Turing

@model function f()
    x = zeros(2)
    x[1] ~ Normal()
    x[2] ~ Normal()
end

spl = MH(@varname(x) => MvNormal(zeros(2), I))
chain = sample(f(), spl, 100)
```

produces

```julia
julia> chain = sample(f(), spl, 100)
┌ Info: When sampling with MH, the following proposals will be used at each step.
└ This output can be disabled by passing `verbose=false` to `sample()`.
[ Info: varname x[1]: no proposal specified, drawing from prior
[ Info: varname x[2]: no proposal specified, drawing from prior
Sampling 100%|█████████████████████████████████████████████████████████████████████████| Time: 0:00:00
```

I think in general it's not possible to figure out on behalf of the user whether that's the behaviour they wanted or not. The user just has to manually check that against what they expected.

### DIFF
--- a/src/mcmc/mh.jl
+++ b/src/mcmc/mh.jl
@@ -158,7 +158,7 @@ struct LinkedRW{C}
 end
 
 """
-    InitFromProposals(proposals::VarNamedTuple)
+    InitFromProposals(proposals::VarNamedTuple, verbose::Bool)
 
 An initialisation strategy that samples variables from user-defined proposal distributions.
 If a proposal distribution is not found in `proposals`, then we defer to sampling from the


src/mcmc/mh.jl
```diff
@@ -303,6 +310,17 @@ function AbstractMCMC.step(
...
 
+    # We evaluate the model once with the sampler's init strategy and print all the
+    # proposals that were used. This helps the user detect cases where the proposals are
+    # silently ignored (e.g. because the VarName in the proposal doesn't match the VarName
+    # in the model).
+    if verbose && init_strategy isa InitFromProposals
+        @info "When sampling with MH, the following proposals will be used at each step.\nThis output can be disabled by passing `verbose=false` to `sample()`."
+        verbose_init_strategy = InitFromProposals(init_strategy.proposals, true)
+        oavi = DynamicPPL.OnlyAccsVarInfo(()) # No need to accumulate anything
+        DynamicPPL.init!!(rng, model, oavi, verbose_init_strategy, DynamicPPL.UnlinkAll())
```
sunxd3, line 321: Will consuming rng be problematic?

src/mcmc/mh.jl
sunxd3: not sure we should update the signature

penelopeysm: Ah, I hate this, I always forget to do it.

src/mcmc/mh.jl
penelopeysm: I don't think so, it just means you get a different (but still reproducible) number at the end.

I was hesitant to use `Random.default_rng()` because that means that you'll get different logging output even if you seed inside `sample(rng, ...)` which might be confusing.

We could also just use an arbitrary seed like `Xoshiro(468)` since this isn't actually returned to the user -- but then if you change the seed passed to `sample()` you still get the same output, which is also maybe counterintuitive?

I might be overthinking it haha.

sunxd3: > I might be overthinking it haha.

not at all. it's tricky, so I am just voicing.

current version works

penelopeysm:
The remaining failures are caused by https://github.com/TuringLang/DynamicPPL.jl/issues/1266, which I'll try to fix :)

Issue #2772, opened 2026-02-12
Correctness of Libtask with submodels depends on whether submodels' functions are inlined or not


penelopeysm:
Consider the following models, which are entirely equivalent:

```julia
using Turing, Random

@model function inner(y, x)
    y ~ Normal(x)
end
@model function nested(y)
    x ~ Normal()
    a ~ to_submodel(inner(y, x))
end
m1 = nested(1.0)
mean(sample(Xoshiro(468), m1, PG(10), 1000))

@model function single(y)
    x ~ Normal()
    y ~ Normal(x)
end
m2 = single(1.0)
mean(sample(Xoshiro(468), m2, PG(10), 1000))
```

It turns out that (at least on my system) both of these sample correctly with PG (giving a mean `x` of 0.46).

If you think about it carefully though, there's no reason why the submodel one _should_ work. The call to `Libtask.produce` is inside the `y ~ Normal(x)` statement, which is in turn inside the `inner` submodel. So, in order to get all the way to the produce call, we need to mark everything along the call stack with `Libtask.might_produce`. This is all basic Libtask functionality.

The problem is that the method `inner(model, varinfo, ...)` itself has never been marked as produce-able! So how does this sampling actually work correctly, then?

The answer is that if the submodel is small enough, then the method `inner(model, varinfo, ...)` will just be inlined.

We can demonstrate what happens when this fails to inline, by changing the output of the DynamicPPL compiler. If we were to apply this patch on the current release of DPPL (v0.39.13)

```diff
diff --git a/src/compiler.jl b/src/compiler.jl
index 84a9a485..604e0b32 100644
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -710,6 +710,7 @@ function build_output(modeldef, linenumbernode)
     # element in the returned value is always the most up-to-date `__varinfo__`.
     # See the docstrings of `replace_returns` for more info.
     evaluatordef[:body] = MacroTools.@q begin
+        @noinline
         $(linenumbernode)
         $(replace_returns(add_return_to_last_statment(modeldef[:body])))
     end
```

and rerun the PG sampling for the submodel, we get a mean `x` of 0.01, which is consistent with the call to `produce` never actually being triggered.

So essentially, we are pretty much at the mercy of the Julia compiler here. If it decides to inline the submodel, we are good. If it doesn't, we will silently give wrong results.

penelopeysm:
[...]

PR #2778, opened 2026-03-01
Fix correctness of submodels that aren't inlined under SMC/PG

penelopeysm:
Due to #2772, the following fails (unlike as described in the issue, you don't need to edit DynamicPPL code to trigger this: just putting `@noinline` in the body of the submodel is enough).

```julia
using Turing, StableRNGs, Test

@model function inner(y, x)
    @noinline
    y ~ Normal(x)
end
@model function nested(y)
    x ~ Normal()
    a ~ to_submodel(inner(y, x))
end
m1 = nested(1.0)
chn = sample(StableRNG(468), m1, PG(10), 500)
@test mean(chn[:x]) ≈ 0.5 atol=0.05
```

The reason is because the submodel evaluator function i.e. `inner(model, varinfo, y, x)` is not marked as produceable via `Libtask.might_produce`. That means that Libtask's IR transformation pass skips over the call to `inner(...)` completely, and any observations inside the submodel are lost.

Consequently, on the current release (v0.42.8), the chain above just samples from the prior, and the mean of `x` will be approximately `0`.

This PR patches it by declaring that ANY function of the form `f(::Model, ::AbstractVarInfo, args...)` is produceable. This causes the test above to pass.

As described in the comment in the source, this seems overkill: if we mark more methods as produceable than are necessary, this can lead to performance slowdowns since Libtask has to transform more function calls.

The thing that allows us to use this hacky patch, though, is that there is pretty much only **one** case where anybody is calling a function with the signature `f(::Model, ::AbstractVarInfo, args...)` **inside** a DynamicPPL model -- which is the submodel case. There are other functions that have similar signatures, but none of them are likely to be called inside a model.

The other thing to note about this PR is that marking this generic method as produceable is not sufficient to handle submodels with keyword arguments, because for such submodels, extra methods need to be marked as produceable. Furthermore, it is not possible to statically determine this because the exact method that we need to target will depend on the number of keyword arguments (see https://github.com/TuringLang/Libtask.jl/issues/197). In this case, it is therefore still mandatory for the user to use the `Libtask.@might_produce` macro themselves (which Turing re-exports).

All of this is mentioned in the patch notes.


penelopeysm:
Closed in favour of #2780

Issue #1603, opened 2021-05-03
Feature Request: Additional Macros


ghost:
Suggested macros that could improve code readability and help people using Turing.jl:

- [ `@noncentered` macro that converts a centered parametrization to a noncentered one ] 
- [ `@standardize` macro that tells the compiler to standardize a variable before doing inference with it, so that priors can be set by the programmer on the original scale (which might be more intuitive to work with) ] 
- [ `@tarray` macro that converts a DataFrame or normal Julia array into a TArray ]

devmotion:
> [ @tarray macro that converts a DataFrame or normal Julia array into a TArray ]

Isn't `TArray(x)` or `convert(TArray, x)` sufficiently simple?

torfjelde:
The two first points would be covered by https://github.com/TuringLang/DynamicPPL.jl/pull/220, though not automatic. That PR is on hold atm as I'll be quite occupied this week, but will return to it next week.

Closed-Limelike-Curves:
> > [ @tarray macro that converts a DataFrame or normal Julia array into a TArray ]
> 
> Isn't `TArray(x)` or `convert(TArray, x)` sufficiently simple?

In the case of regular arrays, I think so. For DataFrames, it may be worth implementing something which can preserve column labels more easily.

devmotion:
Maybe one can just use `TArray(StructArray(df))` but I think nobody has ever tried to use non-standard arrays with Libtask and so some things might not work.

Closed-Limelike-Curves:
> Maybe one can just use `TArray(StructArray(df))` but I think nobody has ever tried to use non-standard arrays with Libtask and so some things might not work.

Hmm -- maybe this can be done by working with some of the features in KeyedDistributions.jl, so that you can use something like `TKeyedArray(KeyedArray)` which is just a `TArray` that keeps track of the keys from the `KeyedArray`.

devmotion:
There should be no need for a different `TKeyedArray` - `TArray` is supposed to work with general `AbstractArray`s. I just don't think anyone has ever tested or used it with anything else than `Array`s, and I don't know if it's a good idea here since it would copy always the whole array, i.e., for all keys, and I am not sure if this is desired. I guess often it might be better to unpack the desired components before constructing the `TArray`.

storopoli:
What about a `@qr` macro that would QR-decompose a matrix of data `X`.

I would use it a lot in my regression models.

torfjelde:
[...]

PR #2779, opened 2026-03-02
Remove Random.seed! from test suite, use Xoshiro/StableRNG instead

### LINKED ISSUE #2726: Remove `Random.seed!` from test suite
0. Global state bad
1. If it doesn't matter what the result is, don't set a seed.
2. If it just matters that the seed is reproducible but it doesn't matter what it generates, use Xoshiro.
3. If the exact output matters, use StableRNG.

anurag-mds:
Closes #2726

Global state is bad, so I went through the test suite and replaced all 
`Random.seed!` calls with local alternatives.

The replacement depends on what the test actually needs:

- If the test only checks that results are roughly correct (using `atol=` 
  or helpers like `check_gdemo`), I used `Xoshiro` passed directly to `sample`.
- If the test checks exact reproducibility (using `==` or `isequal`), 
  I used `StableRNG` instead.
- If the seed was before optimization code like MLE/MAP (which is 
  deterministic anyway), I just deleted it.
- For `MCMCThreads` tests that reset the same rng between runs, I kept 
  the `Random.seed!(rng, seed)` pattern since that's already using an 
  explicit rng and not global state.

I left the single `Random.seed!` in `runtests.jl` untouched since 
that one is intentional (as discussed in the issue).

### DIFF
--- a/test/mcmc/Inference.jl
+++ b/test/mcmc/Inference.jl
@@ -10,6 +10,7 @@ import ForwardDiff
 using LinearAlgebra: I
 import MCMCChains
 import Random
+using Random: Xoshiro
 import ReverseDiff
 using StableRNGs: StableRNG
 using StatsFuns: logsumexp
@@ -35,10 +36,10 @@ using Turing
                 Gibbs(:s => HMC(0.1, 5), :m => ESS()),
             )
             for sampler in samplers
-                rng1 = StableRNG(5)
+                rng1 = Xoshiro(5)
                 chain1 = sample(rng1, model, sampler, MCMCThreads(), 10, 4)
 
-                rng2 = StableRNG(5)
+                rng2 = Xoshiro(5)
                 chain2 = sample(rng2, model, sampler, MCMCThreads(), 10, 4)
 
                 # For HMC, the first step does not have stats, so we need to use isequal to
--- a/test/mcmc/gibbs.jl
+++ b/test/mcmc/gibbs.jl
@@ -681,7 +681,7 @@ end
                 sampler = Gibbs(@varname(s) => hmc, @varname(m) => hmc)
                 rng = StableRNG(42)
                 chain = sample(
-		    rng,
+                    rng,
                     model,
                     sampler,
                     MCMCThreads(),
@@ -695,9 +695,9 @@ end
 
                 # "Ground truth" samples.
                 # TODO: Replace with closed-form sampling once that is implemented in DynamicPPL.
-                
+
                 chain_true = sample(
-		    StableRNG(42),
+                    StableRNG(42),
                     model,
                     NUTS(),
                     MCMCThreads(),
@@ -885,7 +885,7 @@ end
             sampler = Gibbs(@varname(m1) => sampler_inner, @varname(m2) => sampler_inner)
             rng = StableRNG(42)
             chain = sample(
-               rng,  model, sampler, 1000; discard_initial=1000, thinning=10, n_adapts=0
+                rng, model, sampler, 1000; discard_initial=1000, thinning=10, n_adapts=0
             )
             check_numerical(chain, [:m1, :m2], [-0.2, 0.6]; atol=0.1)
             check_logp_correct(sampler_inner)
--- a/test/mcmc/particle_mcmc.jl
+++ b/test/mcmc/particle_mcmc.jl
@@ -55,7 +55,6 @@ using Turing
     end
 
     @testset "logevidence" begin
-
         @model function test()
             a ~ Normal(0, 1)
             x ~ Bernoulli(1)


test/ext/dynamichmc.jl
```diff
@@ -6,13 +6,13 @@ using Test: @test, @testset
...
+using Random: Random, Xoshiro
 using Turing
 
 @testset "TuringDynamicHMCExt" begin
-    Random.seed!(100)
+    rng = Xoshiro(100)
     spl = externalsampler(DynamicHMC.NUTS())
-    chn = sample(gdemo_default, spl, 10_000)
+    chn = sample(rng, gdemo_default, spl, 10_000)
     check_gdemo(chn)
```
penelopeysm, line 17: This case (and anything that requires `check_gdemo` or other numerical checks) requires StableRNG.

It's not about whether it's an equality check or approximate equality check. The thing is, sampling with Xoshiro can (in principle) give different results depending on the OS, Julia version, etc. (see https://docs.julialang.org/en/v1/stdlib/Random/#Reproducibility).

That means that `Xoshiro(100)` might give a mean of 1.0 on one version of Julia, but a mean of 1.5 on a different version of Julia. If `check_gdemo` wants the mean to be 1.0 ± 0.1, then that would cause the latter to fail.

It's true that we don't care whether it's 1.0, 0.99, or 1.01 -- but we do still care about the value because not every value is okay.

Because we run tests on multiple OSes and Julia versions, we need to make sure that this test does not fail randomly -- the way to guarantee that is to use StableRNG, whose output is deterministic across OS and Julia versions.

test/mcmc/emcee.jl
```diff
@@ -25,18 +26,18 @@ using Turing
...
 
-        Random.seed!(1234)
-        chain1 = sample(gdemo_default, spl, 1)
-        Random.seed!(1234)
-        chain2 = sample(gdemo_default, spl, 1)
+        rng1 = StableRNG(1234)
+        chain1 = sample(rng1, gdemo_default, spl, 1)
+        rng2 = StableRNG(1234)
+        chain2 = sample(rng2, gdemo_default, spl, 1)
         @test Array(chain1) == Array(chain2)
```
penelopeysm, line 41: Conversely, this one does not require StableRNG, it only needs Xoshiro.

The reason is because we are running the same thing twice with the same seed, and then comparing that the output is the same. It doesn't matter *what* the seed gives, it only matters that the same seed gives the same thing when run multiple times.

In other words, I don't care if `rand(Xoshiro(5))` gives me 0.1, 0.2, or 0.3. In fact, it doesn't even matter whether different versions of Julia give me different results; hence StableRNG isn't necessary. However, I do care that `rand(Xoshiro(5)) == rand(Xoshiro(5))`.

test/mcmc/gibbs.jl
```diff
@@ -262,10 +262,10 @@ end
...
     )
-    Random.seed!(23)
-    chain1 = sample(gdemo_default, sampler1, 10)
-    Random.seed!(23)
-    chain2 = sample(gdemo_default, sampler1, 10)
+    rng1 = StableRNG(23)
+    chain1 = sample(rng1, gdemo_default, sampler1, 10)
+    rng2 = StableRNG(23)
+    chain2 = sample(rng2, gdemo_default, sampler1, 10)
     @test chain1.value == chain2.value
```
penelopeysm, line 267: Xoshiro.

test/mcmc/gibbs.jl
```diff
@@ -681,8 +681,9 @@ end
                 # Sampler to use for Gibbs components.
                 hmc = HMC(0.1, 32)
                 sampler = Gibbs(@varname(s) => hmc, @varname(m) => hmc)
-                Random.seed!(42)
+                rng = Xoshiro(42)
```
penelopeysm: StableRNG -- you get the pattern.

anurag-mds:
Thanks for clarifying and guiding, @penelopeysm. I will make the necessary changes and will ask if anything feels odd to me

test/ext/dynamichmc.jl
anurag-mds: Read the docs I got where I was wrong. 
Thanks for sharing the exact reference...

test/mcmc/emcee.jl
```diff
@@ -4,19 +4,20 @@ using ..Models: gdemo_default
...
-using Random: Random
+using Random: Random, Xoshiro
+using StableRNGs: StableRNG
 using Test: @test, @test_throws, @testset
 using Turing
 
 @testset "emcee.jl" begin
     @testset "gdemo" begin
-        Random.seed!(9876)
+        rng = Xoshiro(9876)
```
penelopeysm:
```suggestion
        rng = StableRNG(9876)
```

test/mcmc/emcee.jl
```diff
@@ -25,18 +26,18 @@ using Turing
         @info "Testing emcee with large number of iterations"
         spl = Emcee(10, 2.0)
         n_samples = 10_000
-        chain = sample(gdemo_default, spl, n_samples)
+        chain = sample(rng, gdemo_default, spl, n_samples)
```
penelopeysm:
```suggestion
        chain = sample(StableRNG(5), gdemo_default, spl, n_samples)
```

`rng` isn't defined here. You could do `rng = StableRNG(...)` and then `sample(rng, ...)` but I think it's just easier to stick it into the sample call directly since you don't ever need to access it again.

test/mcmc/gibbs.jl
```diff
@@ -262,10 +262,10 @@ end
...
     )
-    Random.seed!(23)
-    chain1 = sample(gdemo_default, sampler1, 10)
-    Random.seed!(23)
-    chain2 = sample(gdemo_default, sampler1, 10)
+    rng1 = Xoshiro(23)
+    chain1 = sample(rng1, gdemo_default, sampler1, 10)
+    rng2 = Xoshiro(23)
+    chain2 = sample(rng2, gdemo_default, sampler1, 10)
     @test chain1.value == chain2.value
```
penelopeysm:
```suggestion
    chain1 = sample(Xoshiro(23), gdemo_default, sampler1, 10)
    chain2 = sample(Xoshiro(23), gdemo_default, sampler1, 10)
    @test chain1.value == chain2.value
```

Similar to the above, I think it's just easier to inline it.

test/mcmc/gibbs.jl
```diff
@@ -696,8 +697,9 @@ end
 
                 # "Ground truth" samples.
                 # TODO: Replace with closed-form sampling once that is implemented in DynamicPPL.
-                Random.seed!(42)
+                Random.seed!(rng, 42)
                 chain_true = sample(
+		    rng,
```
penelopeysm: Instead of re-seeding the old rng object (mutating an object leads to statefulness that is hard to keep track of), it's more explicit and easier to just make a new one.

test/optimisation/Optimisation.jl
```diff
@@ -261,7 +258,6 @@ using Turing
         end
 
         @testset "MAP with box constraints" begin
-            Random.seed!(222)
```
penelopeysm, line 264: I think this one can't be removed yet. Optimisation still relies on randomness, because even though the optimisation procedure itself is completely deterministic, the generation of the initial point is still random, unless the user specifies it.

I think, though, you don't have to worry about changing this for now, because #2733 will get rid of it (that PR completely reworks the optimisation interface). So I would suggest for this PR just focusing on the MCMC tests, and reverting the changes to the optimisation tests.

test/optimisation/Optimisation.jl
penelopeysm, line 273: (For example, this call doesn't specify initial parameters, so they will be randomly generated, and the outcome of the optimisation can depend on the initial parameters.)

test/mcmc/particle_mcmc.jl
```diff
@@ -67,7 +67,7 @@ using Turing
             return x
         end
 
-        chains_smc = sample(test(), SMC(), 100)
+        chains_smc = sample(rng, test(), SMC(), 100)
```
penelopeysm: I think this is a case where writing directly `sample(StableRNG(100), test(), SMC(), 100)` is clearer than defining `rng` and then using it later -- because reading this test, by the time you read line 70 you see this `rng` variable and you have to go back 12 lines to see where it was defined. Whereas if you stick it in the sample call itself, then the definition is close to the point of use!

anurag-mds:
Done! Inlined the RNGs into the sample calls directly, replaced Random.seed!(rng, 42) with a fresh StableRNG(42), and reverted the optimisation tests as suggested.

test/mcmc/Inference.jl
```diff
@@ -35,11 +35,11 @@ using Turing
...
             for sampler in samplers
-                Random.seed!(5)
-                chain1 = sample(model, sampler, MCMCThreads(), 10, 4)
+                rng1 = StableRNG(5)
+                chain1 = sample(rng1, model, sampler, MCMCThreads(), 10, 4)
 
-                Random.seed!(5)
-                chain2 = sample(model, sampler, MCMCThreads(), 10, 4)
+                rng2 = StableRNG(5)
+                chain2 = sample(rng2, model, sampler, MCMCThreads(), 10, 4)
```
penelopeysm, line 43: I might have missed this one earlier, but this one should just be Xoshiro.

anurag-mds: Sure

anurag-mds:
Yeah, I'll do that real quick!

anurag-mds:
Done! Addressed all review comments — inlined RNGs directly into sample calls, replaced Random.seed!(rng, 42) with a fresh StableRNG(42), reverted the optimisation tests, and applied JuliaFormatter.

Anything more!
Do let me know it's exciting!

penelopeysm:
Thank you! Okay, we'll run CI and see if it all passes. It takes about an hour or so.

There are some CI workflows that always fail (due to some Julia bugs outside of our control); but if there aren't any unexpected failures, this would then be good to go.

anurag-mds:
> Thank you! Okay, we'll run CI and see if it all passes. It takes about an hour or so.
> 
> There are some CI workflows that always fail (due to some Julia bugs outside of our control); but if there aren't any unexpected failures, this would then be good to go.

Sure

penelopeysm:
> The CI failures appear to be pre-existing the same jobs (test mcmc/Inference, test mcmc/gibbs, etc.) are also failing on the breaking branch (Tests https://github.com/TuringLang/Turing.jl/issues/1603)

What do you mean? Gibbs isn't failing (yet); and 1603 doesn't seem related.

anurag-mds:
> > The CI failures appear to be pre-existing the same jobs (test mcmc/Inference, test mcmc/gibbs, etc.) are also failing on the breaking branch (Tests #1603)
> 
> What do you mean? Gibbs isn't failing (yet); and 1603 doesn't seem related.

My bad indeed! I misread the CI logs from the other branch.

penelopeysm:
I think we are good to go, the segfault on mcmc/Inference is known (#2655) and the optimisation failure in everything else is also known (and will be fixed in #2733). Thanks, @anurag-mds!

anurag-mds:
Thank you for guiding and assisting me. It was truly exciting to work on this issue!

PR #2790, opened 2026-03-14
(breaking) v0.44

penelopeysm:


src/mcmc/hmc.jl
```diff
@@ -8,28 +8,33 @@ Turing.allow_discrete_variables(sampler::Hamiltonian) = false
...
     adaptor::TAdapt
     ldf::L
+    # TODO(penelopeysm): This field is needed for Gibbs because each time we call
+    # gibbs_update_state!! on this, we need to reconstruct a LogDensityFunction.
+    # In general this would require reevaluating the model, unless we supply a
+    # VarNamedTuple which already contains vectorised parameters. This can probably
+    # be improved in DynamicPPL, but for now we will just store an extra VNT in
+    # the state.
+    # NOTE: The actual values of this field should never be used or relied on!
+    _vector_vnt::V
```
penelopeysm: https://github.com/TuringLang/DynamicPPL.jl/pull/1366 should fix this

penelopeysm: In particular, right now this PR stores the `_vector_vnt` inside the state separately and then reconstructs the LDF from that. That is rather silly, because the state already contains an LDF with all the necessary info. However, DPPL currently doesn't allow us to actually use that info. The above DPPL PR will allow us to extract the fields of the LDF and use it to make a new LDF, meaning that this VNT doesn't have to be cached.

penelopeysm: Fixed in #2811

PR #2803, opened 2026-04-09
Rework of Gibbs sampler interface

### LINKED ISSUE #2642: All VarInfos are informative, but some VarInfos are more informative than others

### LINKED ISSUE #2762: ESS can use LogDensityFunction throughout
It actually has zero need for a VarInfo at all (outside of things like Gibbs).

### LINKED ISSUE #2764: Gibbs interface
I've been generally thinking about how we can fix the Gibbs interface, specifically, what Gibbs expects from component samplers, and what Gibbs provides back to component samplers.

Right now, Gibbs *expects* from component samplers a sampler state which contains a VarInfo. The contained VarInfo must have the values that were freshly sampled by the component sampler. Gibbs then extracts it with `get_varinfo`.

Before each component sampler runs, Gibbs *provides* component samplers with a fresh VarInfo that contains values obtained from all other component samplers. It sticks this back into the sampler state, via the `setparams_varinfo!!` method.

This is all fine, but way too VarInfo-centric. It forces samplers to carry around a VarInfo in its state and also update its state solely for the purpose of Gibbs, which is a bit silly. For example, HMC really doesn't need a VarInfo at all. It can just deal with LDF throughout. Look at this:

https://github.com/TuringLang/Turing.jl/blob/37af7fdb36a85fcc4e1aae770801a42ef5d5d595/src/mcmc/hmc.jl#L274-L278

If you only look at `src/mcmc/hmc.jl`, there's no reason why we need to store this `vi`, nor any reason why we need to stick the params inside via `unflatten`. `state.vi` is never accessed again in the subsequent step. The state already contains the vectorised parameters inside `state.z`. _The only reason why we have these lines is to make Gibbs happy._

So, the point is, getting rid of VarInfo not only frees component samplers up to e.g. use OnlyAccsVarInfo (which is good for performance, but can't happen now because Gibbs), it also reduces the amount of pointless code and pointless fields in sampler states.

I think there are two ways one could go:

1. Generalise the interface so that instead of `get_varinfo` and `setparams_varinfo!!`, we replace them with `get_raw_values` (i.e. get a VNT of raw values, presumably via VAIMAcc) and `update_state_from_raw_values` (i.e. given a VNT of raw values, update whatever you need to update in the state).

   For HMC, the former requires reevaluation. (Note that this is actually SAFER: the current behaviour is potentially wrong if there are variables with dynamic support, because `vi`'s transforms are never ever updated after the sampling has started, and calling `unflatten` can lead to incorrect results!!) The latter would just use the stored LogDensityFunction: we evaluate the LogDensityFunction with the new VNT of raw values, get out the corresponding set of vectorised parameters plus the logp, and then set the appropriate fields in the state. This is doable via https://github.com/TuringLang/DynamicPPL.jl/issues/1232.

2. For each sampler, wrap it in a Gibbs wrapper that maintains a VarInfo. This is conceptually similar to how ExternalSampler wraps something that only ever uses LDF: note that the inner sampler's state need not contain a VarInfo, but the wrapped ExternalSampler does have a state that has a varinfo. (And note *again* that ExternalSampler uses unflatten, so, bugs bugs bugs, with dynamic support variables).

The first is probably better. It will take a lot of refactoring, but it will be a very good thing to do, imo.

### LINKED ISSUE #2801: dynamic transform correctness issue
```julia
using Turing, Random

@model function f()
    x ~ Uniform(-5, 5)
    y ~ truncated(Normal(); lower=x)
end

chn_gibbs1 = sample(Xoshiro(468), f(), Gibbs(:x => MH(), :y => HMC(0.1, 20)), MCMCThreads(), 100000, 4)
mean(chn_gibbs1)
# x 0.6065
# y 1.6269

chn_gibbs2 = sample(Xoshiro(468), f(), Gibbs(:x => MH(), :y => MH()), MCMCThreads(), 100000, 4)
mean(chn_gibbs2)
# x 0.1022
# y 1.5786

chn_gibbs3 = sample(Xoshiro(468), f(), Gibbs(:x => MH(), :y => MH(:y => LinkedRW(1.0))), MCMCThreads(), 100000, 4)
mean(chn_gibbs3)
# x 3.2686
# y 3.7165

chn_nuts = sample(Xoshiro(468), f(), NUTS(), MCMCThreads(), 100000, 4)
mean(chn_nuts)
# x 0.0201
# y 1.5043

chn_mh = sample(Xoshiro(468), f(), MH(), MCMCThreads(), 100000, 4)
mean(chn_mh)
# x -0.0053
# y  1.5042
```

One or more of these is wrong; I'm genuinely not sure which. I'm pretty sure that the discrepancies with `chn_gibbs1` and `chn_gibbs3` arise because `y` is being sampled in linked space. However, I'm still stuck as to why this happens.

penelopeysm:
This PR does a minor overhaul of the Gibbs interface, in line with various plans that have been made over the months.

Specifically, this PR makes the following changes:

## VNT over VarInfo

Instead of `GibbsContext` holding a global **`VarInfo`**, which may contain linked or unlinked values, it now holds a global **`VarNamedTuple`** which always holds **raw** values.

The motivation for this is twofold:

1. `VarInfo` holding linked values and cached transforms leads to correctness issues, such as #2801, which I discovered while trying to write this refactor. I can confirm that this PR fixes that issue.

2. `VarInfo` is slow because it does a lot of extra bookkeeping. The performance issues are in fact not limited to Gibbs: it leaks through to every sampler which Gibbs uses, because it forces every component sampler to carry around a VarInfo so that it can talk to Gibbs. This refactor therefore also means that some component samplers will be faster (see below).

## Sampler <=> Gibbs interface

The old interface that Gibbs expected for samplers was as follows:

- `get_varinfo(state)` Return a VarInfo that the Gibbs sampler will then merge into the global VarInfo.
- `setparams_varinfo!!(state, varinfo, ...)` Update the sampler's state with the new global VarInfo.

In addition to this interface Gibbs also had a function `match_linking!!`, which 'lined up' the transform status of the global VarInfo with each state's individual VarInfo.

This PR directly replaces them with VNT equivalents

- `gibbs_get_raw_values(state)` Return a VNT that the Gibbs sampler will then merge into the global VNT.
- `gibbs_update_state!!(state, vnt, ...)` Update the sampler's state with the new global VNT.

`match_linking!!` is no longer required as a separate function now: inside `gibbs_update_state!!` each sampler can just use its own transform strategy to reconstruct the correct state from the raw values.

## Component samplers

Component samplers have been updated to stop using VarInfo.

## Performance

Gibbs itself is slightly faster:

```julia
using Turing

@model function g()
    x ~ Normal()
    y ~ Normal()
end
@time sample(g(), Gibbs(:x => MH(), :y => MH()), 100_000; chain_type=Any, progress=false, verbose=false);
```

- main: 1.106086 seconds (23.50 M allocations: 1.398 GiB, 7.60% gc time)
- This PR: 0.807256 seconds (17.60 M allocations: 896.448 MiB, 7.04% gc time)

This speedup probably comes from just Gibbs, because performance of MH on these models is very similar (see below).

On top of that, as promised above, this gives some nice speedups for other samplers, and specifically for ESS. These are trivial models. Naturally with larger models the overhead from samplers will be smaller. I also specify `chain_type=Any` to cut out the constant overhead from MCMCChains, which is actually quite significant.

```julia
using Turing
@model function f()
    x ~ Normal()
    1.0 ~ Normal(x)
end
@time sample(f(), spl, 100_000; chain_type=Any, progress=false, verbose=false);
```

`spl = ESS()`:

- main: 0.711739 seconds (10.20 M allocations: 514.196 MiB, 5.33% gc time)
- this PR: 0.166084 seconds (6.00 M allocations: 259.379 MiB, 9.10% gc time)

I thought `spl = MH()` would also show some speedup, but it's way less drastic.

- main: 0.226019 seconds (5.53 M allocations: 271.970 MiB, 9.79% gc time)
- this PR: 0.218487 seconds (5.33 M allocations: 253.618 MiB, 8.29% gc time)

SMC and PG have also been switched over to using `OnlyAccsVarInfo`, but like MH, there's no significant difference in timings either for those (presumably Libtask overhead dominates everything else).

`spl = HMC(0.1, 20)` or `spl = NUTS()` are basically the same, they were already optimised before this.

## Closes

Closes #2801 
Closes #2764 
Closes #2762
Closes #2642


penelopeysm:
~~Mostly trying to fix CI now, have yet to add new tests for the issues in #2801.~~ Regression tests have been added.

penelopeysm:
Summary of this PR:

```julia
# ── #2801: Dynamic transform correctness ─────────────────────────────────────
# y's support depends on x, so its transform changes as x changes.
# On main, Gibbs caches stale linked values → wrong posterior.
# Analytical ground truth: E[x] = 5.05, E[y] ≈ 2.525
#   (The 1/x Jacobian from Uniform(0,x) cancels when marginalising over y,
#    so x keeps its prior; E[y|x] = x/2 gives E[y] = E[x]/2.)

using Turing, Random
# Check ground truth.
chn = sample(Xoshiro(468), dynamic_support(), NUTS(), 100_000; progress=false)
mean(chn)
# NUTS: x = 5.0614, y = 2.5303

@model function dynamic_support()
    x ~ Uniform(0.1, 10.0)
    y ~ Uniform(0, x)
end
chn = sample(Xoshiro(468), dynamic_support(), Gibbs(:x => MH(), :y => HMC(0.1, 10)), 100_000; progress=false)
mean(chn)
# main: x = 3.4242, y = 1.8012
# PR  : x = 5.0478, y = 2.5168


# ── #2762: ESS without VarInfo ────────────────────────────────────────────────
# ESS was forced to carry a VarInfo to talk to Gibbs. Now it uses
# LogDensityFunction throughout → ~4x faster on small models.

using Turing, Random
@model function ess_bench()
    x ~ Normal()
    1.0 ~ Normal(x)
end
sample(Xoshiro(468), ess_bench(), ESS(), 100; chain_type=Any, progress=false)  # warmup
@time sample(Xoshiro(468), ess_bench(), ESS(), 100_000; chain_type=Any, progress=false);
# main:  1.174546 seconds (11.70 M allocations: 585.919 MiB, 20.35% gc time)
# PR:    0.206023 seconds (5.80 M allocations: 254.802 MiB, 8.39% gc time)

# ── #2764: Gibbs overhead (match_linking!! removal) ──────────────────────────
# Old Gibbs called match_linking!! (a model evaluation) per component per step.
# That's gone now → one fewer model eval per component per step.

using Turing, Random
@model function two_normals()
    x ~ Normal()
    y ~ Normal()
end
sample(Xoshiro(468), two_normals(), Gibbs(:x => MH(), :y => MH()), 100; chain_type=Any, progress=false);
@time sample(Xoshiro(468), two_normals(), Gibbs(:x => MH(), :y => MH()), 100_000; chain_type=Any, progress=false);
# main:  2.892283 seconds (39.70 M allocations: 2.225 GiB, 19.35% gc time)
# PR:    0.977869 seconds (17.00 M allocations: 884.242 MiB, 20.99% gc time)


# ── #2642: Samplers no longer forced to carry VarInfo ────────────────────────
# Component samplers previously had to maintain a full VarInfo so Gibbs could
# merge it. Now they just return raw values via gibbs_get_raw_values().
# Best seen with ESS inside Gibbs (combines #2762 + #2764 savings):

using Turing, Random
@model function gibbs_ess()
    x ~ Normal()
    y ~ Normal(x, 1)
    1.0 ~ Normal(y, 0.5)
end
sample(Xoshiro(468), gibbs_ess(), Gibbs(:x => MH(), :y => ESS()), 100; chain_type=Any, progress=false);
@time sample(Xoshiro(468), gibbs_ess(), Gibbs(:x => MH(), :y => ESS()), 100_000; chain_type=Any, progress=false);
# main:  3.270862 seconds (41.93 M allocations: 2.348 GiB, 16.85% gc time)
# PR:    1.336143 seconds (25.03 M allocations: 1.140 GiB, 5.32% gc time)
```

penelopeysm:
@sunxd3 Do you think you have time to review (even with Claude)?

I'm aware that docs and tests are missing (I'll work on them tomorrow), but the source code should all be ready

penelopeysm:
For posterity, here is Claude's review, and my handling of them

- B1 is fixed in https://github.com/TuringLang/Turing.jl/pull/2803/commits/c62324b4bc3845d09affd21fddda387f3595cea9
- N1 is part of a larger issue, which I wrote up in https://github.com/TuringLang/Turing.jl/issues/2810. I think it's okay to leave that to a future version.
- N2 is documented in the Gibbs docstring now and I'll also write docs on Gibbs in general.
- N3 is fixed by the new release of DynamicPPL v0.41.2 which I use here https://github.com/TuringLang/Turing.jl/pull/2803/commits/3c00f771cd78b7501aa22b98f997fdc2603f432a

<details><summary>Claude's review</summary>

## Blockers

### B1 — `src/mcmc/hmc.jl:500-504`: adapted metric discarded on every Gibbs sweep

```julia
metricT = getmetricT(spl)
metric = metricT(LogDensityProblems.dimension(new_ldf))
```

For `DiagEuclideanMetric`/`DenseEuclideanMetric`, `metricT(dim)` returns identity `M⁻¹` (`AdvancedHMC/src/metric.jl:66-67`). Pre-PR `breaking:src/mcmc/hmc.jl:550-564` → `get_hamiltonian(..., state, ...)` → `gen_metric(n, spl, state)` (`breaking:src/mcmc/hmc.jl:433-435`) dispatches for `AdaptiveHamiltonian` to `AHMC.renew(state.hamiltonian.metric, AHMC.getM⁻¹(state.adaptor.pc))`, preserving the adapted mass matrix. Post-PR does not.

**Latency**: not currently observable — Gibbs does not set `nadapts = sampler.n_adapts` when calling components (the specialisation at `src/mcmc/hmc.jl:94-156` is top-level-only), HMC's component step defaults `nadapts=0` (`hmc.jl:165`), and `AHMC.adapt!`'s gate `i ≤ n_adapts` is therefore always false. `state.adaptor.pc` never accumulates, so `getM⁻¹(pc)` == identity == `metricT(n)`. Latent today; real once adaptation is plumbed into Gibbs or an adapted state is supplied via `initial_state`.

**Fix (one line)**: use the existing helper at `src/mcmc/hmc.jl:424-427`:

```julia
metric = gen_metric(LogDensityProblems.dimension(new_ldf), spl, state)
```

---

## Non-blockers to address before merge

### N1 — `src/mcmc/gibbs.jl:270-271`: RNG not threaded through `tilde_assume!!` "new variable" branch

```julia
# TODO(penelopeysm): How is the RNG controlled here?
value = rand(right)
```

Uses `Random.default_rng()` instead of the caller's RNG. Breaks reproducibility under `sample(rng, …)` for models introducing new vars mid-sweep; unsafe under `MCMCThreads()`. Fix: thread the outer rng through `GibbsContext` (or via `InitContext`, as `src/mcmc/particle_mcmc.jl:429-449` does). Author's own TODO.

Exercised by `test/mcmc/gibbs.jl:485-536` (`dynamic_bernoulli_normal`) — not broken, just not rng-respecting.

### N2 — `src/mcmc/gibbs.jl:501-503`: `initial_params` propagated unmodified to every component

```julia
# FIXME: This will cause issues if the sampler expects initial params in unconstrained space.
# This is not the case for any samplers in Turing.jl, but will be for external samplers, etc.
initial_params=initial_params,
```

Breaks `Gibbs(..., externalsampler(...; unconstrained=true), ...)` with user-supplied `InitFromParams(unconstrained_vec)`. Fix or downgrade to documented limitation. Author's own FIXME.

### N3 — private-field access across Turing↔DPPL boundary

Final commit `e8cdc08` is titled "avoid use of DPPL internal fields". Two callsites regressed:

- `src/mcmc/gibbs.jl:69` — `old_ldf._getlogdensity` inside `gibbs_recompute_ldf_and_params`.
- `src/mcmc/emcee.jl:76` — `vi.values` inside `ParamsWithStats(InitFromParams(vi.values), model)`.

Prefer public accessors (or add tiny DPPL accessors: `get_getlogdensity`, canonical raw-value getter).
```

</details>

sunxd3:
thanks, Penny

PR #2804, opened 2026-04-11
Add SafeLogDensity wrapper for graceful error handling in logdensity

niwq15:
## Summary

This PR introduces a `SafeLogDensity` wrapper that catches errors during logdensity evaluation and returns `-Inf` instead of throwing.

## Motivation

Inspired by the discussions on improving error handling during sampling, e.g. rejecting proposals instead of failing when encountering numerical issues, this PR provides a small step in that direction by introducing a safe wrapper for logdensity evaluation.

## Changes

- add `SafeLogDensity` wrapper
- catch exceptions in `logdensity` and return `-Inf`
- add a test demonstrating fallback behavior

## Scope

This does not yet integrate with samplers or gradient evaluation. It is intended as a minimal building block for future improvements.

## Notes

The wrapper is currently not exported to keep the API surface minimal until further discussion.

### DIFF
--- a/src/Turing.jl
+++ b/src/Turing.jl
@@ -50,6 +50,9 @@ using .Variational
 include("optimisation/Optimisation.jl")
 using .Optimisation
 
+
+include("logdensity_wrapper.jl")
+
 ###########
 # Exports #
 ###########
@@ -187,4 +190,5 @@ export
     might_produce,
     @might_produce
 
+    SafeLogDensity
 end
--- a/src/logdensity_wrapper.jl
+++ b/src/logdensity_wrapper.jl
@@ -1,0 +1,17 @@
+"""
+    SafeLogDensity(ldf)
+
+Return '-Inf' if evaluation throws an error.
+"""
+
+struct SafeLogDensity{L}
+    ldf::L
+end
+
+function LogDensityProblems.logdensity(w::SafeLogDensity, x)
+    try
+        return LogDensityProblems.logdensity(w.ldf, x)
+    catch
+        return -Inf
+    end
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4,6 +4,7 @@ using Pkg
 using Random: seed!
 using Test
 using TimerOutputs: TimerOutputs, @timeit
+using LogDensityProblems
 import Turing
 
 # Fix the global Random.seed for reproducibility.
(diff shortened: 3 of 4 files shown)


src/Turing.jl
```diff
@@ -187,4 +190,5 @@ export
     might_produce,
     @might_produce
 
+    SafeLogDensity
```
penelopeysm, line 193: This line doesn't do anything right now. The export list ends at `@might_produce` because there's no comma after it, and this is essentially just a floating thing on a new line. Did you intend something else?

src/logdensity_wrapper.jl
```diff
@@ -0,0 +1,17 @@
...
+    ldf::L
+end
+
+function LogDensityProblems.logdensity(w::SafeLogDensity, x)
+    try
+        return LogDensityProblems.logdensity(w.ldf, x)
+    catch
+        return -Inf
+    end
+end
```
penelopeysm, line 17: For Turing's purposes, on top of `LDP.logdensity`, we definitely need to forward `LDP.logdensity_and_gradient`, `LDP.capabilities`, and `LDP.dimension` to the wrapped object as well. Otherwise it can throw errors down the line.

I think with such a generic wrapper, there is a question of whether you should just wrap the entire LogDensityProblems API (e.g., including the Hessian functions). I'm not sure about this, but at the very least we'd need to cover the above.

penelopeysm (changes requested):
Thanks @niwq15!

> This does not yet integrate with samplers or gradient evaluation. It is intended as a minimal building block for future improvements.

Do you intend to do this in this PR? Personally, I would prefer if it is in the same PR. Otherwise, this PR on its own introduces what's essentially dead code but without the promise of it being used later. It shouldn't be super hard to do -- basically all keyword arguments to `sample()` are forwarded through to `AbstractMCMC.step`, so you can modify any definitions of `AbstractMCMC.step` which create `LogDensityFunction`s (e.g. inside `src/mcmc/hmc.jl`) to detect the keyword and wrap the `LogDensityFunction` if needed. Happy to explain more if useful.

Finally, there's a question about where this wrapper should be. I think that it should be in DynamicPPL --  however, given that it's not exported, it would be fine with me to introduce it here first, and move it later if needed.

yebai:
This would be better done in a separate package wrapping a DynamicPPL LogDensityFunction.

PR #2811, opened 2026-04-22
Remove unnecessary hacks in Gibbs interface

penelopeysm:


src/mcmc/hmc.jl
```diff
@@ -21,14 +21,6 @@ struct HMCState{
...
     adaptor::TAdapt
     ldf::L
-    # TODO(penelopeysm): This field is needed for Gibbs because each time we call
-    # gibbs_update_state!! on this, we need to reconstruct a LogDensityFunction.
-    # In general this would require reevaluating the model, unless we supply a
-    # VarNamedTuple which already contains vectorised parameters. This can probably
-    # be improved in DynamicPPL, but for now we will just store an extra VNT in
-    # the state.
-    # NOTE: The actual values of this field should never be used or relied on!
-    _vector_vnt::V
```
penelopeysm, line 31: basically this PR is just getting rid of this field which should not need to exist

penelopeysm:
Once this passes, I'll release 0.44.

PR #2819, opened 2026-04-27
adding info_sampler_output method to track divergences

PavanChaggar:
Hello! 

Following some discussion with @penelopeysm on slack, I'm making a PR that adds a method to warn if there are any divergences when using a Hamiltonian sampler. 

I've added a method to `src/mcmc/hmc.jl` which just warns if there are any divergences after sampling. I've also added some fallback methods to `src/mcmc/abstractmcmc.jl` that catch non-Hamiltonian sampler or cases where the chain is not an MCMCChains object.  I separated them to avoid errors as `Hamiltonian` is not defined until `hmc.jl` is loaded in `src/mcmc/inference.jl`. But I could add all methods to the `hmc.jl` file if thats preferred. 

I've added a few tests to `test/mcmc/abstractmcmc.jl` that detects the warning message for HMC and NUTs using serial and threading. I think this also makes sure warnings are not duplicated in the sampler calls. 

Thanks, 
Pavan

### DIFF
--- a/src/mcmc/abstractmcmc.jl
+++ b/src/mcmc/abstractmcmc.jl
@@ -69,6 +69,19 @@ function find_initial_params_ldf(
     )
 end
 
+
+"""
+    info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
+
+Returns the number of divergent transitions in the chain.
+"""
+function info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
+    return nothing
+end
+function info_sampler_output(chain, sampler::AbstractSampler)
+    return nothing
+end
+
 #########################################
 # Default definitions for the interface #
 #########################################
@@ -76,7 +89,9 @@ end
 function AbstractMCMC.sample(
     model::DynamicPPL.Model, spl::AbstractSampler, N::Integer; kwargs...
 )
-    return AbstractMCMC.sample(Random.default_rng(), model, spl, N; kwargs...)
+    chain =  AbstractMCMC.sample(Random.default_rng(), model, spl, N; kwargs...)
+    info_sampler_output(chain, spl)
+    return chain
 end
 
 function AbstractMCMC.sample(
@@ -108,10 +123,12 @@ function AbstractMCMC.sample(
     N::Integer,
     n_chains::Integer;
     kwargs...,
-)
-    return AbstractMCMC.sample(
+)   
+    chain = AbstractMCMC.sample(
         Random.default_rng(), model, alg, ensemble, N, n_chains; kwargs...
     )
+    info_sampler_output(chain, alg)
+    return chain    
 end
 
 function AbstractMCMC.sample(
--- a/src/mcmc/abstractmcmc.jl
+++ b/src/mcmc/abstractmcmc.jl
@@ -69,7 +69,6 @@ function find_initial_params_ldf(
     )
 end
 
-
 """
     info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
 
@@ -86,7 +85,7 @@ end
 function AbstractMCMC.sample(
     model::DynamicPPL.Model, spl::AbstractSampler, N::Integer; kwargs...
 )
-    chain =  AbstractMCMC.sample(Random.default_rng(), model, spl, N; kwargs...)
+    chain = AbstractMCMC.sample(Random.default_rng(), model, spl, N; kwargs...)
     info_sampler_output(chain, spl)
     return chain
 end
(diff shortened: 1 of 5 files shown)


src/mcmc/hmc.jl
```diff
@@ -3,6 +3,19 @@ abstract type StaticHamiltonian <: Hamiltonian end
...
 
+"""
+    info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
+
+Returns the number of divergent transitions in the chain.
+""" 
+function info_sampler_output(chain::MCMCChains.Chains, sampler::Hamiltonian)
+    n_divergent = sum(skipmissing(vec(chain[:numerical_error])))    
+    if n_divergent > 0
+        @warn "number of divergent transitions: $n_divergent; consider increasing `target_accept` or reparameterising your model"
```
penelopeysm: I think `target_accept` only makes sense for NUTS, right? Not a Bayesian person at all, so please tell me if I'm wrong.

test/mcmc/abstractmcmc.jl
```diff
@@ -217,4 +218,21 @@ end
...
+            return
+        end
+    end
+
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), NUTS(), 1000; progress=false)
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), NUTS(), MCMCThreads(),1000, 2; progress=false)
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), HMC(0.1, 5), 1000; progress=false)
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), HMC(0.1, 5), MCMCThreads(),1000, 2; progress=false)
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), HMCDA(200, 0.65, 0.3), 1000; progress=false)
+    @test_logs min_level=Logging.Warn (:warn, r"consider increasing `target_accept` or reparameterising your model") sample(f(), HMCDA(200, 0.65, 0.3), MCMCThreads(),1000, 2; progress=false)
```
penelopeysm: Could you loop over the samplers here + extract the regex pattern into a constant, to avoid the duplication?

penelopeysm (changes requested):
Thanks @PavanChaggar! This looks good. Just some small nitpicks.

Could I get you to please also bump the patch version in Project.toml and add a changelog entry in HISTORY.md?

src/mcmc/hmc.jl
penelopeysm: (BTW the organisation of the code, ie having this method in src/mcmc/hmc is perfect!)

penelopeysm:
(BTW, if you want to fix the formatter CI, there are instructions here: https://turinglang.org/docs/contributing/code-formatting/ but it's not super duper important, I can do it on my end before merging.)

src/mcmc/hmc.jl
PavanChaggar: Yeah, that's right. I'll add some more info to the warning message for specific methods. Thanks!

test/mcmc/abstractmcmc.jl
PavanChaggar: Yup will do!

PavanChaggar:
Hey @penelopeysm, 

Thanks for the review! I've made the changes to the code. I can check out the CI later.

src/mcmc/abstractmcmc.jl
```diff
@@ -69,14 +69,29 @@ function find_initial_params_ldf(
     )
 end
 
+
+"""
+    info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
+
+Returns the number of divergent transitions in the chain.
```
penelopeysm:
```suggestion
A post-sampling hook that can e.g. print info about the results of sampling.
```

src/mcmc/abstractmcmc.jl
```diff
@@ -69,14 +69,29 @@ function find_initial_params_ldf(
...
+"""
+function info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
+    return nothing
+end
```
penelopeysm:
```suggestion
```

The other method already takes care of this.

src/mcmc/hmc.jl
```diff
@@ -3,6 +3,19 @@ abstract type StaticHamiltonian <: Hamiltonian end
 abstract type AdaptiveHamiltonian <: Hamiltonian end
 Turing.allow_discrete_variables(sampler::Hamiltonian) = false
 
+"""
+    info_sampler_output(chain::MCMCChains.Chains, sampler::AbstractSampler)
```
```suggestion
    info_sampler_output(chain::MCMCChains.Chains, sampler::Hamiltonian)
```

penelopeysm (changes requested):
Thanks! Just some minor stylistic issues, which I'll change on my end before merging :)

penelopeysm (approved):
Thanks again @PavanChaggar!

PR #2820, opened 2026-04-29
Add Enzyme to list of AD tools using native API

wsmoses:
Looks like this was missed in https://github.com/TuringLang/Turing.jl/pull/2816


yebai:
Enzyme is promising but still a bit experimental/fragile -- it is available via DifferentiationInterface + ADTypes if one wants to use it.

PR #2825, opened 2026-05-11
feat: selectively re-export Distributions instead of full re-export

### LINKED ISSUE #2682: Stop re-exporting all of Distributions
https://github.com/TuringLang/Turing.jl/blob/d8fbe78570a89ba3d36f07c1e62fa9f822233e1d/src/Turing.jl#L7

The same as https://github.com/TuringLang/Turing.jl/issues/2681, but for Distributions.

I think the complete re-export of Distributions is **almost** justifiable. In particular, we definitely want to export all the **distributions themselves**, as well as functions like `truncated`, `product_distribution`, perhaps even `logpdf`.

But Distributions exports a ton of things that are not the above:

https://github.com/JuliaStats/Distributions.jl/blob/bbdd4f15f3230f42c8856f56b74770e346178f43/src/Distributions.jl#L32-L277

And I think we can afford to be more selective.

ArpanC6:
Resolves #2682

### What
In previous releases `@reexport` with Distributions meant exporting all 200+ symbols of Distributions.jl via Turing. With the change in Distributions usage an explicitly curated list of exported symbols is provided instead.

### Why
Distributions.jl exports lots of internal helper symbols which do not have any use to the users of Turing (for instance sqmahal, qqbuild, nsamples, EdgeworthMean). Exporting such symbols can cause namespace pollution and unintentionally break functionality in some cases.

### What is included

All distribution symbols (Normal, Beta, Gamma, and so on)
Core functions: logpdf, pdf, cdf, truncated, censored, product_distribution, insupport, fit, params, mean, var, std, and others.

### What is removed

Internal helper functions: sqmahal, qqbuild, QQPair, nsamples, EdgeworthMean/Sum/Z, pdfsquared


ArpanC6:
The remaining failure (test mcmc/gibbs on macos latest) appears to be a segmentation fault (signal 11) that is unrelated to this PR the same test passes on Ubuntu and Windows. The test at test/mcmc/Inference.jl 634 directly references Distributions as a module which currently requires `@reexport` using Distributions. I have restored this for now.

However to support a fully selective export approach would it be better to explicitly include using Distributions in the test file instead?

yebai:
@ArpanC6 https://github.com/TuringLang/Turing.jl/issues/2682 is not planned. Feel free to work on some other items if you like.

Issue #1996, opened 2023-05-25
Broadcasting , addprob and PPL not functionning with PG and SMC sampler


organic-chemistry:
Hello, 
I was trying to use the macro Turing.@addlogprob!  in a case where I wanted to use PG sampler, but it could not work.
I know that it is for advanced user, however
when trying to understand the problem I found that PG and SMC do not work either with broadcasting (function called demo_array) in the following code.
I implemented 4 times the function than you call coinflip in your quickstart (here called demo) with different 'complexity' of implementation. I tried four sampler: HMC MH SMC and PG.
For the four different way of implementing it, it works only for all with HMC and MH
With PG and SMC only the original function work: 

```julia
using Turing
using Pkg
Pkg.status("Turing")
@model function demo(y)
    # Our prior belief about the probability of heads in a coin.
    p ~ Beta(1, 1)

    # The number of observations.
    N = length(y)
    for n in 1:N
        # Heads or tails of a coin are drawn from a Bernoulli distribution.
        y[n] ~ Bernoulli(p)
    end
end

@model function demo_array(x)
    # 'array'
    p ~ Beta(1,1)
    x .~ Bernoulli(p)
    return 
end
@model function demo_addprob(x)
    # 'logprob'
    p ~ Beta(1,1)
    loglik = loglikelihood(Bernoulli(p), x)
    Turing.@addlogprob!(loglik)
    return 
end

function function_demo_PPL(model, varinfo, context, x)
    p, varinfo = DynamicPPL.tilde_assume!!(
        context,
        Beta(1, 1),
        Turing.@varname(p),
        varinfo,
    )
    DynamicPPL.dot_tilde_observe!!(context, Bernoulli(p), x, Turing.@varname(x), varinfo)

end

demo_PPL(x) = Turing.Model(function_demo_PPL, (; x))

data = [true for _ in 1:20]

p_model = Dict("demo"=>demo,
                "array"=>demo_array,
                "add_prob"=>demo_addprob,
                "PPL"=>demo_PPL)

p_sampler = Dict("HMC"=>HMC(0.05, 10), "SMC"=>SMC(),"MH"=>MH(),"PG"=>PG(20))

results=[]

for model_name in ["demo","array","add_prob","PPL"]
    for (sampler_name, sampler) in p_sampler
        c = sample(p_model[model_name](data),sampler, 1000)
        push!(results,[model_name,sampler_name,mean(c[:p])])
    end
end

[...]

PR #2848, opened 2026-07-16
Deprecate AdvancedPS

charlesknipp:
To facilitate a more uniform ecosystem, this PR is a means of deprecating AdvancedPS in favor of a Turing specific sequential Monte Carlo interface.

For context, this is a little more than a reorganization.

## Design Choices

- `Libtask.produce` no longer occurs in the accumulator to ensure that a varinfo reflects the produces log score
- it is no longer necessary to include `resample` and `varinfo` in the task global storage
- `ParticleContainer` no longer requires tracked RNG, which should be handled in the sampler
- resampling looks a lot different, but behaves as it did from AdvancedPS
- SMC sampler overloads `AbstractMCMC.sample` instead of stepping through and bundling with the API

## Minor User Facing Changes
- resamplers are now structs (`Multinomial`, `Systematic`, and `ESSResampler`)
- sampling with SMC now supports the ensemble key word argument to parallelize the re-weight step

## Criticisms
- I wonder if the model traces and Libtask utilities belong in an extension like `DynamicPPLLibtaskExt.jl`
- I am split on whether I should create a new struct for `ParticleContainer` or keep it as an alias for `Vector{<:Particle}`

Feel free to make changes. I am open to suggestions. Otherwise, unit tests pass locally so it should be ready to merge.


charlesknipp:
the unit tests currently failing are only off by a slightly larger absolute tolerance. Not totally convinced its the fault of the sampler, but I'd appreciate some feedback/sanity checks

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
-    Ws = AdvancedPS.getweights(particles)
-    index = AdvancedPS.randcat(rng, Ws)
-    newreference = particles.vals[index]
+# reseed new child particles according to the sampler RNG
+function split!(rng::AbstractRNG, particles::Vector{<:TracedModel})
+    children = deepcopy.(particles)
+    seeds = rand(rng, Random.Sampler(rng, UInt64), length(particles))
+    @. Random.seed!(get_rng(children), seeds)
+    return children
+end
```
shravanngoswamii: I think this will break Particle Gibbs. Reseeding the child's inner RNG is correct for avoiding forward seed collisions, but `AdvancedPS`'s Libtask extentions [`delete_seeds!`](https://github.com/TuringLang/AdvancedPS.jl/blob/e6e615bd0b7aa6072b2b34a6e69e29dc7d43c893/ext/AdvancedPSLibtaskExt.jl#L254-L263), `keys = keys[1:count-1]`) did a second thing this drops: it kept `TracedRNG.keys` aligned with `count` when a child is forked from the reference. The reference is a legal resample ancestor ([`resample!`](https://github.com/charlesknipp/Turing.jl/blob/83db45b209ee90635748810c2b7a184aa5cd154b/src/mcmc/particle_mcmc.jl#L416) draws over `ref.values`, whose last element is the reference), so `deepcopy` gives a child with `keys` of length `T` but `count = t`. [`Random.seed!(::TracedRNG, key)` in traced_rng.jl](https://github.com/TuringLang/Turing.jl/pull/2848/changes#diff-44c8c2909c45d0b31f8cee37e6008600f663c34b6ced6336e1fa397290a9d22dR47-R50) resets only the inner Philox counter, not `count`/`keys`, so the child's later `save_state!` appends at `T+1, T+2, ...` while `keys[t..T]` stay stale. When such a child is retained as [`new_ref`](https://github.com/TuringLang/Turing.jl/pull/2848/changes#diff-627c8a22e8dd5dfc5fcbcc0ae8628a4cc8816161cf6cbdd064a3b17b82a66ea4R659), `set_reference` rebuilds it over a fresh empty varinfo and regenerates the reference purely by replaying `keys[count]`, so it replays the previous reference's trajectory, not the retained particle's. That breaks CSMC invariance (silently biased posterior without any crash and grows `keys` by ~T per affected iteration.

Reproduced: a forked child has `count = 2` with `keys` length `= 4`, and `child.keys[2] == old_reference.keys[2]`.

Doing this in `split!` after reseeding may fix it.

```julia
for c in children
    r = get_rng(c)
    resize!(r.keys, r.count - 1)
end
```

I guess adding a small-`N`, multi-observation PG test asserting that `reference_{m+1}` equals the sample retained at iteration `m`.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
-    return SMC(AdvancedPS.ResampleWithESSThreshold(resampler, threshold))
-end
-function SMC(threshold::Real)
-    return SMC(AdvancedPS.resample_systematic, threshold)
+# this is the only difference between LogLikelihoodAccumulator
+function DynamicPPL.acclogp(acc::ProduceLogLikelihoodAccumulator, val)
+    task_local_storage(:logscore, val)
+    newacc = ProduceLogLikelihoodAccumulator(DynamicPPL.logp(acc) + val)
+    return newacc
 end
```
shravanngoswamii, line 109: Since `produce` moved out of `acclogp`, plain `@addlogprob!` (which still routes here) updates the accumulator but never produces, so it silently stops reweighting in particle models. That's basically https://github.com/TuringLang/Turing.jl/issues/1996 coming back. Reproduced: `P(x<0)` is 0.494 with `@addlogprob!` (prior) vs 0.0 with `@producelogprob!`. Since `@addlogprob!` still compiles fine inside a PG model, it might be worth making it produce under a particle context, or at least erroring/warning so people notice. Would also be nice if that let `@producelogprob!` go away entirely.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+                $(esc(:(__varinfo__))) = DynamicPPL.accloglikelihood!!(
+                    $(esc(:(__varinfo__))), val
+                )
+
+                # grab the log score and remove it from storage
+                loglike = pop!(task_local_storage(), :logscore)
+                if !isnothing(loglike)
+                    Libtask.produce(loglike)
+                end
+            end
```
shravanngoswamii: Two things here:

- `pop!(task_local_storage(), :logscore)` has no default, and the guard `hasacc(vi, Val(:LogLikelihood))` is also true for the normal `LogLikelihoodAccumulator`. So any non-particle run (NUTS, MH, `loglikelihood`, `logjoint`, or a non-PG Gibbs block) hits this and the un-defaulted `pop!` throws. I tried it: `loglikelihood(demo_p(1.0), (x=0.5,))` and `sample(..., MH())` both raise `KeyError: :logscore`. `pop!(..., :logscore, nothing)` fixes it (the existing `!isnothing` check then just skips). Same default would help at https://github.com/TuringLang/Turing.jl/pull/2848/changes#diff-627c8a22e8dd5dfc5fcbcc0ae8628a4cc8816161cf6cbdd064a3b17b82a66ea4R187.

- It also never restashes `:varinfo` (unlike `tilde_observe!!` at https://github.com/TuringLang/Turing.jl/pull/2848/changes#diff-627c8a22e8dd5dfc5fcbcc0ae8628a4cc8816161cf6cbdd064a3b17b82a66ea4R186), so a trailing `@producelogprob!` gets dropped from the chain's `loglikelihood` column. In my test, a constant `@producelogprob! -3.0` gives a loglikelihood column of all `0.0`. I think adding `task_local_storage(:varinfo, __varinfo__)` after the `accloglikelihood!!` line should sort it.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+    @producelogprob!(ex)
 
-    # Compute the transition and the next state.
-    stats = (; weight=weight, logevidence=state.average_logevidence)
-    transition = if discard_sample
-        nothing
-    else
-        DynamicPPL.ParamsWithStats(deepcopy(particle.model.f.varinfo), stats)
+Same as `@addlogprob!`, but intended for use with SMC samplers to ensure Libtask can produce
+the loglikelihood when iterating throguh the model
```
shravanngoswamii:
```suggestion
the loglikelihood when iterating through the model
```

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+        logZ += logmarginal
+    end
+    return particles, logZ
 end
 
-# Convenient constructor
-function AdvancedPS.Trace(
-    model::Model, varinfo::AbstractVarInfo, rng::AdvancedPS.TracedRNG, resample::Bool
+# We overload AbstractMCMC.sample because SMC does not yield a Markov chain. Since stepping
+# through a sampler is based on the underlying model, we posses a different structure which
```
shravanngoswamii:
```suggestion
# through a sampler is based on the underlying model, we possess a different structure which
```

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,674 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
-    fargs = (model.f, args...)
-    return TracedModel(model, varinfo, resample, fargs, kwargs)
+"""
+    TracedModel(rng, model)
+
+A container for stepping through a DynamicPPL model using Libtask to yield the log score at
+every iteration of a Sequential Monte Carlo sampler.
+
+Typically traced models internally `Libtask.consume` through a sampler, but iteration
+utilities ala Libtask extend to this structure for manually stepping through a model.
```
shravanngoswamii:
```suggestion
utilities like Libtask extend to this structure for manually stepping through a model.
```

src/mcmc/particle_mcmc.jl
charlesknipp: yeah that's essentially the approach I want to take; just haven't thought of a clean solution for it.

I'll think about this one today and see about getting around the issue. It's definitely something I plan on amending before merging this PR

src/mcmc/particle_mcmc.jl
charlesknipp: ahhhh good catch; I'll make sure to address that in the next commit. Although since I plan on eventually doing away with `producelogprob!` it wasn't something I extensively tested. I really don't intend on using it in anything other than SMC samplers.

I will add that I thought through the various cases of varinfo being populated, and it should guarantee that it is caught up. This was my issue with the previous implementation of PMCMC, since produce yields a not yet accumulated log likelihood.

src/mcmc/particle_mcmc.jl
charlesknipp: Definitely an over sight on my part, but I'm a bit surprised the unit tests never caught it. I'll make this change in the next commit.

Nice catch.

src/mcmc/particle_mcmc.jl
charlesknipp: From what I can tell, there is only one way to do this without imposing type piracy or extending DPPL: don't re-export `@addlogprob!` from DynamicPPL, and just redefine it with what I have for `@producelogprob!`.

I tested this one locally and it seems to work without raising any issues. The only real point of failure here would be if the user imports both Turing and DynamicPPL. Maybe I can fit this in an extension to get around any issues.

Going through the stack, there is no way to reference the context when using `@addlogprob!` since we have `@addlogprob!` -> `map_accumulator!!` -> `setaccs!!` which should only produce in the case of `SMCContext`, despite none of them having the opportunity to know the context at compile time. I can commit type piracy and get around that, but we might as well just go with the first solution then.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,543 +16,664 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
 
-"""
-$(TYPEDEF)
+abstract type AbstractResampler end
 
-Particle Gibbs sampler.
+should_resample(::AbstractVector, ::AbstractResampler) = true
 
-# Fields
+function sample_ancestors(
```
yebai: Is this intended for particle Gibbs with ancestor resampling (PGAS)?

It would be difficult to implement for general-purpose PPL models because they:

1. are not guaranteed to be Markovian, and
2. do not expose the dependency structure needed to compute ancestor weights.

Because of these limitations, I don't think we should add PGAS here. If this only refers to the standard particle Gibbs resampling step, it would be clearer to call it particle resampling (or simply resampling) rather than `ancestors` resampling.

yebai (commented):
I did a partial review today. Here are a few comments:

It would be helpful to include a high-level overview of the design of the new algorithm. I realize this is an unusual request for a code review, but particle Gibbs is a subtle algorithm, and it's difficult to assess correctness without understanding the overall design and implementation.

I think it's worth making a strong effort to avoid task-local storage and keep the implementation as functional as possible (i.e., by tracking all state explicitly in the TapedTask or particle).

I understand this is a substantial piece of work, so I'm happy to iterate on this over a few review cycles.

src/mcmc/particle_mcmc.jl
charlesknipp: Definitely not intentional. I borrow terminology from the particle filtering literature, but will gladly change it to avoid confusion with PGAS.

charlesknipp:
@yebai

> I think it's worth making a strong effort to avoid task-local storage and keep the implementation as functional as possible (i.e., by tracking all state explicitly in the TapedTask or particle).

The latest commit updates the interface to use taped globals.

> It would be helpful to include a high-level overview of the design of the new algorithm. I realize this is an unusual request for a code review, but particle Gibbs is a subtle algorithm, and it's difficult to assess correctness without understanding the overall design and implementation.

Under the hood, `smcsample` does the exact same thing as `AdvancedPS.sweep!`, the only differences are (1) when Libtask produces and (2) how traced RNG works.
1. Producing in the accumulator implies that the tasks varinfo does not yet apply the current log score. As a solution, I produce later in the stack after the likelihood accumulates; using the taped globals to store the log score until I need to produce.
2. When resampling, `split!` ensures that each child has a different seed per RNG trace and prunes dead branches when splitting from the reference. Otherwise, we call `update_keys!` to generate new seeds according to the samplers RNG (not counting the reference), as opposed to storing another trace in the particle container.

There are a handful of similarities between my implementation and AdvancedPS. I suggest comparing specifically this [AdvancedPS code](https://github.com/TuringLang/AdvancedPS.jl/blob/main/src/container.jl) and the highlighted section of [my rewrite](https://github.com/charlesknipp/Turing.jl/blob/1c0be672cdbd9b21b3935bf240c7d02fe1a7708e/src/mcmc/particle_mcmc.jl#L271-L540); which should provide a bit more clarity in terms of correctness of implementation.

To ensure this implementation is indeed valid, I wrote a few unit tests I feel are worth highlighting:
- [test/essential/container.jl](https://github.com/charlesknipp/Turing.jl/blob/ck/smc/test/essential/container.jl) illustrates how I use the new structures for particle sampling.
- [ensuring reference consistency](https://github.com/charlesknipp/Turing.jl/blob/1c0be672cdbd9b21b3935bf240c7d02fe1a7708e/test/mcmc/particle_mcmc.jl#L196-L228) ensures that the reference trajectory is replayed according to the intended RNG trace, therefore confirming the validity of the Particle Gibbs sampler.

yebai:
Many thanks, @charlesknipp -- can you help take a look at https://github.com/TuringLang/Turing.jl/pull/2853? https://github.com/TuringLang/Turing.jl/pull/2853 passes all existing and extra tests and also fixes a few other legacy issues that date back to the early days of Turing.jl. 

We could try to find a way to consolidate this PR and https://github.com/TuringLang/Turing.jl/pull/2853 in the end.

PR #2853, opened 2026-07-21
Reimplement particle MCMC (SMC and PG) natively, removing AdvancedPS

### LINKED ISSUE #6: Parallel Sequential Monte Carlo
I looked into the newly released `Threads.@threading` construct in Julia 0.5. It seems that adding parallelization feature to SMC is simple - could be done in a few lines of code (see the `parallelsmc` branch). 

However, the `threading` feature in Julia is still quite fragile (see e.g. [here](https://github.com/JuliaLang/julia/issues/14494)). I've also filed a [bug](https://github.com/JuliaLang/julia/issues/19450) in the Julia repo.

Maybe we should wait until the Julia team fix these threading bugs and revisit to this feature in a few months time.

### LINKED ISSUE #39: Storing statistics on the particle weights
As discussed in TuringLang/Turing.jl#426, it would be good if Turing would be able to store statistics on the particle weights for all particle-filtering based methods. These statistics could be stored in the `ParticleContainer` and exposed to the user through new fields in the `Chain` type.

### LINKED ISSUE #110: Carefully handle splitting of RNG seeds

### LINKED ISSUE #2781: StableRNG is not stable with SMC/PG across Julia versions
Sampling the same model with PG across different Julia versions can yield different results even if StableRNG is used. I think it's probably not because of StableRNG, nor is it because of PG itself -- it's probably (almost certainly) because of something inside AdvancedPS. I'm not inclined to look into it, but I did just want to open an issue so that this information is out there somewhere.

yebai:
Opened for discussion.

See https://github.com/TuringLang/Turing.jl/pull/2853#issuecomment-5358450673 for a subtle correctness issue with #2848. 

Reimplements `SMC` and `PG`/`CSMC` directly on Libtask + DynamicPPL and drops the AdvancedPS
dependency. A rewrite based on #2848 (thanks @charlesknipp; commits co-authored).

Each `observe` is one filtering step: under `ParticleMCMCContext`, every likelihood term
becomes a `Libtask.produce`, so a particle is a suspended model execution we advance, weight,
and resample. The conditional-SMC reference is reproduced by *replaying* a per-particle
`TracedRNG` (no stored trajectory); a fork is reseeded to branch off. Particle state lives in
the task's taped globals (no `task_local_storage`), and `@addlogprob!` reweights via the
likelihood accumulator.

Posteriors, log-evidence, and exact reference reproduction match the old implementation;
`gibbs.jl`, Aqua, and log-density-metadata tests pass; performance is on par.

Fix https://github.com/TuringLang/Turing.jl/issues/2781
Fix https://github.com/TuringLang/AdvancedPS.jl/issues/110
Fix https://github.com/TuringLang/AdvancedPS.jl/issues/39
Fix https://github.com/TuringLang/AdvancedPS.jl/issues/6


HISTORY.md
```diff
@@ -1,3 +1,23 @@
...
+The old function-based API (`resample_systematic`, `AdvancedPS.ResampleWithESSThreshold`, …) is gone.
+
+The default scheme is now **stratified** rather than systematic: it stays consistent as the number of particles grows and within the theoretical guarantees for particle Gibbs, which systematic does not.
+Exact draws may therefore differ from previous releases, but remain statistically consistent (the same target distribution).
+
+The rewrite also brings:
+
+  - **Reproducibility.** Internal seeds are derived through a counter-based (Philox) generator, so a fixed user seed gives the same draws on every Julia version and platform, and splitting one stream into many is better decorrelated. Previously, results could drift between Julia versions even under a `StableRNG` (https://github.com/TuringLang/Turing.jl/issues/2781).
+  - **Parallelism.** Multiple chains run under AbstractMCMC's `MCMCThreads()` and `MCMCDistributed()`, and `SMC(; threaded=true)` / `PG(n; threaded=true)` spread the particles across threads within each sweep. Neither changes the results; start Julia with multiple threads (e.g. `julia -t auto`) for the thread-based paths to take effect.
```
charlesknipp: This is a little misleading since running `sample(rng, model, SMC(), 200; ensemble=MCMCThreads())` does not reweight in parallel. I think we should adhere to the AbstractMCMC ensemble interface for consistency and clarity. We should be able to pass the kwargs through to `sweep!` and dispatch `reweight` given an `AbstractMCMCEnsemble`.

src/mcmc/particle_mcmc.jl
```diff
@@ -142,416 +537,169 @@ function AbstractMCMC.sample(
...
 function AbstractMCMC.step(
     ::AbstractRNG,
-    model::DynamicPPL.Model,
-    spl::SMC,
+    ::DynamicPPL.Model,
+    ::SMC,
     state::SMCState;
     discard_sample=false,
     kwargs...,
 )
```
charlesknipp: Why adhere to the AbstractMCMC interface if we don't actually use any of the underlying features? I feel like it is cleaner to overload `AbstractMCMC.sample` and just return a bundled version of the particle container. This allows us to pass the ensemble kwarg to `sweep!` and ignore both `discard_initial` and `thinning`.

src/mcmc/particle_mcmc.jl
```diff
@@ -142,416 +537,169 @@ function AbstractMCMC.sample(
...
-const CSMC = PG # type alias of PG as Conditional SMC
+"Conditional SMC, an alias for [`PG`](@ref)."
+const CSMC = PG
 
-struct PGState{V<:DynamicPPL.AbstractVarInfo,R<:Random.AbstractRNG}
-    vi::V
+struct PGState{V<:DynamicPPL.AbstractVarInfo,R<:TracedRNG}
+    varinfo::V
     rng::R
 end
```
charlesknipp: What is wrong with using `Particle` as the state container here? We retain the rng and varinfo without having to define another struct.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,77 +118,363 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
-    rng::R
-end
-# Because pMCMC uses OnlyAccsVarInfo, we need to overload this. It's fine to use Any (see
-# the docstring of get_param_eltype in DynamicPPL) because pMCMC doesn't involve AD or any
-# other tracer types.
+Leaf context marking a model evaluation as a particle-filter step: `tilde_assume!!` draws
+from the prior using the particle's [`TracedRNG`](@ref), and `tilde_observe!!` scores the
+observation, which [`ProduceLogLikelihoodAccumulator`](@ref) turns into a `Libtask.produce`.
+"""
+struct ParticleMCMCContext <: DynamicPPL.AbstractContext end
```
charlesknipp: This is a bit pedantic, but I think the name `SMCContext` is a better fit here since SMC is not an MCMC algorithm, and thus is also not a PMCMC algorithm.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,77 +118,363 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+    # Abstract on purpose: the VarInfo type can change during PG-inside-Gibbs. Accesses go
+    # through Libtask's (already type-unstable) taped globals, so this costs nothing extra.
+    varinfo::DynamicPPL.AbstractVarInfo
+    rng::TracedRNG
+    logweight::DynamicPPL.LogProbType
+    task::Libtask.TapedTask
+    # `task` is filled in once the particle exists, because the task must capture the
+    # particle as its taped globals (a back-reference).
+    Particle(vi, rng) = new(vi, rng, zero(DynamicPPL.LogProbType))
+end
```
charlesknipp:
```suggestion
mutable struct Particle{RT<:TracedRNG,WT<:Real}
    # Abstract on purpose: the VarInfo type can change during PG-inside-Gibbs. Accesses go
    # through Libtask's (already type-unstable) taped globals, so this costs nothing extra.
    varinfo::DynamicPPL.AbstractVarInfo
    rng::RT
    logweight::WT
    task::Libtask.TapedTask
end

Particle(vi::DynamicPPL.AbstractVarInfo, rng::TracedRNG) = new(vi, rng, zero(DynamicPPL.LogProbType))
```
This ensures that if the user decides to reset DynamicPPL.LogProbType to something else, that the particle weights will follow suit.

I'm also a little unsure whether this makes sense as an inner constructor.

src/mcmc/particle_mcmc.jl
```diff
@@ -15,77 +118,363 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+"""
+struct ProduceLogLikelihoodAccumulator{T<:Real} <: DynamicPPL.LogProbAccumulator{T}
+    logp::T
+end
+
+DynamicPPL.accumulator_name(::Type{<:ProduceLogLikelihoodAccumulator}) = :LogLikelihood
+DynamicPPL.logp(acc::ProduceLogLikelihoodAccumulator) = acc.logp
+
+function DynamicPPL.acclogp(acc::ProduceLogLikelihoodAccumulator, val)
+    Libtask.produce(val)  # the only difference from `LogLikelihoodAccumulator`
```
charlesknipp: My one concern is the internal consistency of the sampler's log evidence to the accumulator; see the following illustration:

```julia
trace = Turing.Inference.Particle(
    model, Turing.Inference.particle_varinfo(), Turing.Inference.TracedRNG(rng)
);

while true
    score = Turing.Inference.advance!(trace, false)
    score === nothing && break
    print("score: $(round(score, digits=3))")
    acc_val = trace.varinfo.accs[:LogLikelihood].logp
    println("\taccumulator: $(round(acc_val, digits=3))")
end
```

src/mcmc/particle_mcmc.jl
```diff
@@ -15,77 +118,363 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+end
+
+# Advance every particle by one observation; return `true` once all have finished. A model
+# whose number of observations varies across executions leaves particles out of step.
+#
+# Each particle advances only its own state (rng, varinfo, task), and its rng was already
+# seeded serially in `resample_propagate!`, so the threaded loop is race-free and gives
+# results identical to the serial one. Only the model evaluations parallelise; the shared
+# sampler rng is untouched here.
+function reweight!(particles, conditional::Bool, threaded::Bool)
```
charlesknipp: I think it would be best to dispatch a few versions of reweight which parallelize based on the `AbstractMCMCEnsemble` paradigm.

- `reweight!(particles, conditional, ensemble::MCMCSerial)`
- `reweight!(particles, conditional, ensemble::MCMCThreads)`
- `reweight!(particles, conditional, ensemble::MCMCDistributed)`

src/mcmc/particle_mcmc.jl
```diff
@@ -15,77 +118,363 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+    template,
+    ::DynamicPPL.AbstractVarInfo,
 )
-    # Make sure we load/reset the rng in the new replaying mechanism
-    isref ? AdvancedPS.load_state!(trace.rng) : AdvancedPS.save_state!(trace.rng)
-    score = consume(trace.model.ctask)
-    return score
+    particle = Libtask.get_taped_globals(Particle)
+    ctx = DynamicPPL.InitContext(
+        particle.rng, DynamicPPL.InitFromPrior(), DynamicPPL.UnlinkAll()
```
charlesknipp: Is it always the case that we `InitFromPrior`?

charlesknipp (commented):
Overall, I think this is a very nice revision of #2848. Although, there are a few minor choices I intentionally made when drafting that should be reconsidered here.
1. Passing `AbstractMCMCEnsemble` through to `reweight!` (only for SMC) for consistency
2. Producing the log score upon `tilde_observe` (or `accloglikelihood!!` for `@addlogprob!`) to ensure log score is up to date with log likelihood accumulation.

yebai:
Thanks, this was a valuable review. What changed, and where I took a different route:

**Score ↔ accumulator consistency.** The `produce` now fires from `tilde_observe!!` and an `accloglikelihood!!` overload *after* the accumulator is updated, so the produced score can't lag the accumulated log-likelihood and `@addlogprob!` terms reach the accumulator too. It emits the accumulator's delta rather than a stashed increment, so the accumulator stays the single source of truth.

**SMC output.** `AbstractMCMC.sample` now runs the sweep and bundles the whole population in one shot — `SMCState` and the step-loop are gone, and `discard_initial`/`thinning` no longer have anything to apply to. I took it a little further: a final resampling step returns an equal-weight sample (so `mean(chain[…])` needs no weighting), and chains carry `log_normalizing_constant` (renamed from `logevidence`, the standard SMC term) plus a per-observation `ess_per_step` diagnostic.

**Parallelism.** I diverged here. `MCMCThreads`/`MCMCDistributed` parallelise independent *chains*; within-sweep parallelism spreads *one sweep's particles*, which resample all-to-all every step and are live Libtask tasks — so a distributed reweight would be communication-bound and effectively infeasible. Rather than dispatch `reweight!` on the ensemble, I kept the two as separate knobs (chain-level = `sample`'s ensemble argument; within-sweep = a threads-only `multithreaded` sampler option) and fixed the HISTORY line that blurred them.

**PG state** Done — the retained `Particle` already carries the `varinfo` and `rng` needed to resume, so `PGState` is gone.

**Naming / parametrization** `SMCContext`; and `Particle{RT<:TracedRNG,WT<:Real}` so `logweight` follows `LogProbType`.

**"Is it always `InitFromPrior`?".** Yes — verified a value is never already present at `tilde_assume!!` (particle varinfos start empty, each variable is assumed once, and the CSMC reference replays RNG seeds rather than reusing stored values), so an `InitFromParams` branch would be dead code. I documented the reasoning in place instead of adding it.

yebai:
Will merge this in 72 hours.

src/mcmc/particle_mcmc.jl
```diff
@@ -182,36 +180,62 @@ back-reference; [`reseed!`](@ref) then gives it its own random stream.
...
+    strategy = if reference === nothing
+        DynamicPPL.InitFromPrior()
+    else
+        vn in reference.varnames || error(
+            "the reference execution trace changed while replaying retained values " *
+            "(new address: $vn)",
+        )
+        DynamicPPL.InitFromParams(reference.values, nothing)
+    end
+    ctx = DynamicPPL.InitContext(particle.rng, strategy, DynamicPPL.UnlinkAll())
```
charlesknipp, line 228: I left an earlier comment which questioned the exclusion of `InitFromParams` which is necessary for PG, and hence why you mentioned task based RNG wouldn't handle this alone.

src/mcmc/particle_mcmc.jl
```diff
@@ -119,56 +69,104 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+    #
+    # `varnames` is the set of addresses the retained trajectory assumed, and cannot be recovered
+    # from `values`: a slice assume such as `x[1:2] ~ MvNormal(...)` is stored under the keys `x[1]`,
+    # `x[2]` but assumed under the single address `x[1:2]`, so comparing against those keys would
+    # report a spurious trace change. Without it an address the retained trajectory never had would
+    # silently draw from the prior, corrupting the reference.
+    reference::Union{
+        Nothing,
+        @NamedTuple{values::DynamicPPL.VarNamedTuple, varnames::Set{DynamicPPL.VarName}}
+    }
```
charlesknipp, line 124: Is this necessary on a per particle basis? I think this should be handled in the sampler code, like in my version.

src/mcmc/particle_mcmc.jl
```diff
@@ -119,56 +69,104 @@ function error_if_threadsafe_eval(model::DynamicPPL.Model)
...
+    # Addresses assumed by this execution, in the same form as `reference.varnames`. Survives
+    # forking, so a particle that becomes the retained state hands the complete set to the next
+    # reference; a reference must finish having assumed exactly that set.
+    assumed_varnames::Set{DynamicPPL.VarName}
```
charlesknipp, line 128: If I understand correctly, this is necessary for the reference varinfo type to match the rest of the particles. My original draft of the PR (which is now closed) needed this in order to do subsequent CSMC steps, but ultimately led me to reimplement the traced RNG approach.

src/mcmc/particle_mcmc.jl
```diff
@@ -489,110 +604,85 @@ struct SMC{R<:AbstractResampler} <: ParticleInference
...
-    index::Int
-    logevidence::DynamicPPL.LogProbType
+    return SMC(ESSThresholdResampler(threshold, scheme); kwargs...)
 end
 
+# SMC is a single weighted sweep, not a Markov chain: rather than fake an iteration through
+# AbstractMCMC's step loop (returning the population one particle at a time), we run the sweep
+# and bundle the whole population into the chain in one shot. `discard_initial`/`thinning`
+# therefore have nothing to apply to.
 function AbstractMCMC.sample(
```
charlesknipp, line 636: I like how you took the approach I suggested, dropping the `AbstractMCMC.step` paradigm. I think you should take this a step further.

I would argue that each particle trajectory is indeed Markov, which implies that a set of particle trajectories is the same structure as a set of parallel Markov chains. Therefore, I think including `AbstractMCMCEnsemble` here and dispatching it to `sweep!` would be the correct way to parallelize this algorithm that is still consistent with the AbstractMCMC interface.

This further strengthens my earlier suggestion to replace `SMC(; multithreaded=true)` by passing `MCMCThreads()` to a dispatched `reweight`. See my branch to understand how it is dispatched, since it works for both multiprocessing and multithreading.

src/mcmc/particle_mcmc.jl
```diff
@@ -614,67 +704,80 @@ struct PG{R<:AbstractResampler} <: ParticleInference
...
+    # Passing `state` makes this the reference, pinned to the retained trajectory by value (see
+    # the `reference` field). Its own generator is never read, since every draw is supplied by value --
+    # only its forks' are, and `reseed!` gives those fresh seeds. So it carries the retained
+    # generator forward rather than taking a fresh one, which keeps the sweep from drawing anything
+    # from `rng` on its behalf; the copy just avoids aliasing `state`.
+    reference = Particle(model, deepcopy(state.rng), state)
+    # `n - 1` fresh particles, with the reference last -- the slot `resample_propagate!` retains.
+    particles = [Particle(model, particle_rng(rng)) for _ in 1:(n - 1)]
+    push!(particles, reference)
+    logZ, _ = sweep!(rng, particles, sampler.resampler, sampler.multithreaded)
```
charlesknipp, line 780: It seems like `@code_warntype` is going to have a field day here.

While I understand it is not entirely possible to have a fully parametric and type stable SMC sampler with DPPL, my version at least guarantees that particles are parametrically typed and that each element shares the SAME type. The only abstract type I had was the varinfo stored in the task globals. Everything else is parametric and consistent between elements.

When the rest of the cloud (besides the reference trajectory) is the same type, it seems wasteful to make the compiler work for it.

My suggestion is to run a profiler and check to see how much bloat is from abstract types. Maybe I'm off base here, but my prior is that it will have a whole lot of type inference.

test/mcmc/particle_mcmc.jl
```diff
@@ -252,32 +362,152 @@ end
...
+                z ~ Normal()
+                μ = z
+            end
+            return y ~ Normal(μ, 1)
+        end
+        rng = StableRNG(91)
+        retained = Particle(branch_changes(true, 0.0), particle_rng(rng))
+        run_to_end!(retained)
+        reference = Particle(branch_changes(false, 0.0), particle_rng(rng), retained)
+        @test_throws "reference execution trace changed" advance!(reference)
```
charlesknipp, line 433: I'm a bit surprised that this passes given that we drop the traced RNG component; as long as forking a reference changes the per particle RNG I think this is golden.

src/mcmc/particle_mcmc.jl
charlesknipp: Instead of the named tuple, why not use a raw value accumulator?

charlesknipp (commented):
This largely resembles my initial redesign, so I don't hate it. As I point out, there are also some design changes which better motivate some of my earlier suggestions.

## Reference Trajectories

My main criticism here is with respect to handling reference trajectories. This approach induces a ton of type unions and abstractions which are only necessary to accommodate a single reference trajectory.

In my PR, I introduce `ReferencedContainer` which assumes reference trajectories are the same type as other particles, so forking them during resampling ensures my particles are all type stable. To be fair, this is only possible with `TracedRNG`.

The PG inconsistency you pointed out was due to the missing `DynamicPPL.InitFromParams` in the `tilde_assume` pipeline, not the RNG tracing. Otherwise, the particle samplers in `main` are also incorrect.

## AbstractMCMC Interface

My comment on `AbstractMCMC.sample` says it all. I think this is absolutely the right call, but it should be pushed further to adhere to the interface.

Passing `PG(; multithreading=true)` in conjunction with `MCMCThreads()` is not only nonsensical, but potentially dangerous. Playing devils advocate, I wonder if we could see some nice gains when pairing multithreaded sweep and parallel chains with `MCMCDistributed()`.

## Hooking into Libtask

While not mentioned in my comments, I have some concerns about the libtask usage here.

My original concern from the review was initially addressed, then subsequently replaced with the desynchronized accumulated varinfo. I understand this choice, but it still doesn't sit right with me.

I also don't like the self referential aspect of storing the entire particle in the taped globals. I think a much cleaner approach is outlined in my PR where we only need two things: the RNG and the varinfo; everything else is unnecessary storage. At the very least, I wonder if you could abuse Ref to avoid redundant copies.

## Final Remarks

Compared to my implementation, this rewrite drops much of the elegance and modularity without provable gains to speed or efficiency. While I understand much of the design choices, the bloated containers and heavy indentation in functions like `reweight` and `resample_propagate` just leave a lot to be desired.

With all of that being said, if this is sufficient to pass the unit tests and guarantee correctness, it would be a waste not to merge.

src/mcmc/particle_mcmc.jl
yebai: It sounds like that would require combining  `conditioning` with `InitFromParams` for reference particles.  I haven't explored that, but would be happy if there is a clean mechanism.

yebai:
We will do another iteration of the design, likely as a new PR.

yebai:
Following up on this, I've added a targeted regression test to #2855.

The test turned out to be sensitive to more than the defect it was written for. #2848 has a subtler variant of the same problem: the reference particle is rebuilt by replaying the stored random numbers rather than by reusing the retained values, which is what `main` does, so it lands on a different trajectory as soon as another Gibbs component re-conditions the latents. It is easy to miss because it only shows up when a Gibbs-updated parameter sits inside the latent dynamics and the latents are dependent — but the effect is clearly measurable: mean absolute error on the state marginals is `0.017–0.022`, against `0.002–0.005` for a correct sweep.

On the same test, #2853 comes out clean at `0.002–0.005` across seeds, as does `main` once #2855 is applied.

The mechanism that works is the one `main` uses — reuse the retained values via `InitFromParams` rather than replaying the stream. That does conflict with `ReferencedContainer` as it stands: a reference carrying values no longer shares the type of the free particles, which is the property the parametric container relies on.

yebai:
Thanks, @charlesknipp, for #2848 and the reviews. 

It has taken a while, as I was away for some time. You asked for evidence on speed and efficiency rather than taking an
assertion on trust, and that was the right thing to ask. Here it is, together with a benchmark you
can run yourself. It is small, takes no test utilities, and works unmodified against all three
implementations, since each accepts a bare ESS threshold: the resampling policy is matched rather
than left at whatever each one defaults to. 


| | linear Gaussian, T=200 | | two-state HMM, T=100 | |
|---|---|---|---|---|
| | SMC, N=64 | PG(16) × 100 | SMC, N=64 | PG(16) × 100 |
| `main` (v0.46.1) | 0.77 s / 832 MiB | 17.90 s / 23.1 GiB | 0.50 s / 542 MiB | 12.76 s / 16.8 GiB |
| #2848  | 0.54 s / 851 MiB | 14.35 s / 19.6 GiB | 0.34 s / 535 MiB | 6.49 s / 12.1 GiB |
| this PR | 0.37 s / 345 MiB | 10.89 s / 11.6 GiB | 0.29 s / 304 MiB | 7.68 s / 9.4 GiB |

Against `main` this branch is 1.6–1.7× quicker and allocates two to two-and-a-half times less, on
every cell. Against yours it leads three cells by 1.2–1.5×, and trails on the fourth: discrete-latent
PG is 7.68 s here against your 6.49 s, an 18% lead for #2848. I have changes in hand that close most
of that cell, but as this PR stands today the figure is yours.

Two caveats belong with the table. #2848 sits on v0.46.0, so its non-particle code differs a little
from v0.46.1. More importantly, `main`'s PG figures are what they are only because of #2855: before
that fix, descendants of the reference were clones of it, the population stayed near-uniformly
weighted, the ESS gate rarely fired, and `main` looked about twice as fast as it does here while
sampling the wrong distribution. A benchmark against v0.46.0 gives `main` a flattering number for
that reason alone.

On the type-stability question: your instinct that the hot path deserved measuring was right, though
the mechanism differs from the one predicted. Dynamic dispatch is 0.7% of profile frames; `deepcopy`
in the fork path is 15%. That is where the remaining work points.

**Your other points.** Several were well made and are addressed in a follow-up PR that builds on
this one and on #2848:

- Storing the whole particle in the taped globals. Our tilde handlers took a varinfo argument and
  ignored it, so `@addlogprob!` terms were missing from a chain's `loglikelihood` and `:=` values
  from chains entirely. Yours threads the argument and reports both. We now do the same, and it is
  worth 9–14% of runtime.
- Producing at `accloglikelihood!!`, from your first review. Now fixed as you
  proposed.
- The per-particle reference bookkeeping, and "bloated containers". Both fields can be collapsed into the
  retained values alone, which removes a `Set` that was deep-copied on every fork — about 8% of
  allocations. 
- The shapes of `reweight!` and `resample_propagate!`, flattened.
- `InitFromParams`: your diagnosis was correct and has been fixed since 23 July.
- The ensemble interface cannot be built as proposed. A suspended particle will not cross
  a task/process boundary: the in-sweep `MCMCDistributed` path on your branch fails with the same
  error as ours does. 

Your wider argument about elegance and modularity is not something a benchmark settles, and it is
worth continuing to make on its own terms. The script is below: if other models or particle counts
tell a different story on your machine, I would like to see it.

<details>
<summary><code>pmcmc_bench.jl</code></summary>

Julia 1.12.7, single thread, one machine, sequential runs. 

```julia
# Particle MCMC benchmark. Runs unmodified on Turing 0.46 (AdvancedPS-based) and on the
# native rewrite: both accept a bare ESS threshold, so the resampling policy matches.
#
#   julia --project=/path/to/some/Turing.jl pmcmc_bench.jl

using Turing, Random, Printf, Pkg

@model function lgssm(y)                    # continuous latents, T observations
    x = Vector{Float64}(undef, length(y))
    x[1] ~ Normal(0, 1)
    y[1] ~ Normal(x[1], 1)
    for t in 2:length(y)
        x[t] ~ Normal(0.9 * x[t - 1], 0.5)
        y[t] ~ Normal(x[t], 1)
    end
end

@model function hmm(y)                      # discrete latents, T observations
    z = Vector{Int}(undef, length(y))
    z[1] ~ Categorical([0.5, 0.5])
    y[1] ~ Normal(z[1] == 1 ? -1.0 : 1.0, 0.8)
    for t in 2:length(y)
        p = z[t - 1] == 1 ? 0.7 : 0.3
        z[t] ~ Categorical([p, 1 - p])
        y[t] ~ Normal(z[t] == 1 ? -1.0 : 1.0, 0.8)
    end
end

const THRESHOLD = 0.5                       # resample when ESS <= 0.5 * nparticles
const NPARTICLES = 64                       # SMC population
const PG_PARTICLES = 16
const PG_ITERS = 100

function measure(f; reps=3)
    f()                                     # compile
    t = minimum(@elapsed(f()) for _ in 1:reps)
    return t, (@allocated f()) / 2^20
end

function run_case(name, model)
    ts, as = measure(() -> sample(Xoshiro(3), model, SMC(THRESHOLD), NPARTICLES; progress=false))
    tp, ap = measure(
        () -> sample(Xoshiro(3), model, PG(PG_PARTICLES, THRESHOLD), PG_ITERS; progress=false)
    )
    @printf("%-22s  SMC N=%d: %6.2f s %8.0f MiB   PG(%d) x%d: %6.2f s %8.0f MiB\n",
            name, NPARTICLES, ts, as, PG_PARTICLES, PG_ITERS, tp, ap)
end

function turing_version()
    proj = Pkg.project()                    # the repo itself, when run with --project there
    proj.name == "Turing" && return string(proj.version)
    return string(only(p.version for (_, p) in Pkg.dependencies() if p.name == "Turing"))
end
@printf("Turing %s, julia %s, %d thread(s)\n", turing_version(), VERSION, Threads.nthreads())
run_case("linear Gaussian T=200", lgssm(randn(Xoshiro(1), 200)))
run_case("two-state HMM T=100", hmm(randn(Xoshiro(1), 100)))
```

PR #2859, opened 2026-08-26
Native particle MCMC (SMC/PG), revised: dropping AdvancedPS

### LINKED ISSUE #6: Parallel Sequential Monte Carlo

### LINKED ISSUE #39: Storing statistics on the particle weights

### LINKED ISSUE #110: Carefully handle splitting of RNG seeds

### LINKED ISSUE #2781: StableRNG is not stable with SMC/PG across Julia versions

yebai:
`SMC` and `PG`/`CSMC` now run directly on Libtask and DynamicPPL. Resamplers are
types, stratified by default; conditional sweeps draw ancestors multinomially.
Philox seeding makes draws reproducible across Julia versions. Chains carry
`log_normalizing_constant`. Roughly twice as quick as `main`, allocating half.

Builds on #2848. Supersedes #2853.

Fix #2781
Fix TuringLang/AdvancedPS.jl#110
Fix TuringLang/AdvancedPS.jl#39
Fix TuringLang/AdvancedPS.jl#6
Close #2848. Close #2853.


sunxd3 (approved):
The core logic looks good to me.

Agent found two minor issues I think worth a second look. They are unlikely to impact user too much though.
Here are the MWEs.

```julia
using Turing, Distributions, StableRNGs, Test

@model function constrained_posterior()
    x ~ Normal()
    @addlogprob! (; logprior = x >= 0 ? 0.0 : -Inf)
    0.0 ~ Normal(x, 1)
end

smc = sample(
    StableRNG(1), constrained_posterior(), SMC(), 1_000; progress=false
)
pg = sample(
    StableRNG(1), constrained_posterior(), PG(10), 200; progress=false
)

for chain in (smc, pg)
    xs = vec(collect(chain[:x]))
    logjoints = vec(collect(chain[:logjoint]))

    @show count(x -> x < 0, xs)
    @test all(x -> x >= 0, xs)
    @test all(i -> xs[i] >= 0 || logjoints[i] != -Inf, eachindex(xs))
end
```

```julia
SMC: 507 / 1000 negative draws
PG:   95 / 200 negative draws
```

`logprior` didn't affect particle weight (`loglikelihood` is fine otherwise)

---

```julia
using Turing, Distributions, MCMCChains, StableRNGs, Test

@model function one_observation()
    x ~ Normal()
    0.0 ~ Normal(x, 1)
end

chain = sample(
    StableRNG(1),
    one_observation(),
    SMC(),
    32;
    chain_type=MCMCChains.Chains,
    progress=false,
)

ess = vec(collect(chain[:ess_per_step]))

@show names(chain, :internals)
@show ess
@test any(x -> !ismissing(x), ess)
```

```julia
names(chain, :internals) =
[:log_normalizing_constant, :ess_per_step,
 :logprior, :loglikelihood, :logjoint]

ess =
Union{Missing, Float64}[missing, missing, ..., missing]
```
agent explained that this is because [`ess_per_step` ](https://github.com/TuringLang/Turing.jl/blob/90053807b0e7af8ef047d260a8df102387f6588e/src/mcmc/particle_mcmc.jl#L718-L724)being vector [will be replaced by `missing` ](https://github.com/TuringLang/DynamicPPL.jl/blob/66a5b73b01658d8bc17396ea211417b1824ee09b/ext/DynamicPPLMCMCChainsExt.jl#L22-L31)

yebai:
Thanks, @sunxd3 -- both fixed.

PR #2862, opened 2026-08-27
Use internal_values_as_vector in emcee

shravanngoswamii:
Removes the last `vi[:]` call in Turing, which unblocks deprecating it in DynamicPPL. See TuringLang/DynamicPPL.jl#1454 and the review discussion on TuringLang/DynamicPPL.jl#1451.


shravanngoswamii:
Not needed anymore, vi[:] is staying. See TuringLang/DynamicPPL.jl#1454.

PR #2863, opened 2026-08-28
Gibbs: condition components instead of `GibbsContext`

### LINKED ISSUE #2693: Gibbs does not use initialisation strategies for component samplers
### Minimal working example

```julia
using Turing, Bijectors

dist = Beta(1, 1)
binv = Bijectors.inverse(bijector(dist))

@model function f()
    h ~ dist
    m ~ dist
end

spl = Gibbs(@varname(h) => HMC(0.1, 5), @varname(m) => MH())

chn = sample(f(), spl, MCMCThreads(), 1, 100)
all(h -> binv(-2) < h < binv(2), chn[:h]) # false

chn2 = sample(f(), spl, MCMCThreads(), 1, 100; initial_params=fill(InitFromUniform(), 100))
all(h -> binv(-2) < h < binv(2), chn2[:h]) # true
```

### Description

The Gibbs sampler does not make use of initialisation strategies for component samplers. That's because it uses `Turing.Inference.init_strategy(::Gibbs)` which resolves to `InitFromPrior`, and `init_strategy(::HMC)` is never called.

In the above example the HMC variable should always be initialised to `(binv(-2), binv(2))` but it isn't as can be seen from the check.

The solution to this is to write a custom `AbstractInitStrategy` for Gibbs which, I think, maps varnames to init strategies of the component samplers: so this would wrap `@varname(h) => InitFromUniform(), @varname(m) => InitFromPrior()`. Then inside `DynamicPPL.init` it would check the varname and dispatch to the appropriate strategy.

### Julia version info

1.11.7

### Manifest

0.41.0 (currently using https://github.com/TuringLang/Turing.jl/pull/2689)

Note the same is true on previous versions as the first call to `all(...)` also returns false on 0.40.

### LINKED ISSUE #2766: Stats from component samplers are lost with Gibbs

### LINKED ISSUE #2810: Gibbs third branch
Running on #2803, I struggled to hit the third branch in the Gibbs `tilde_assume!!` call (these lines)

https://github.com/TuringLang/Turing.jl/blob/e8cdc08a24291da8818e7a7b3c11ebe9746e5e9a/src/mcmc/gibbs.jl#L265-L274

but managed to do so with this:

```julia
using Turing, Random
@model function f()
    x ~ Normal()
    y ~ Normal()
    if x > 0
        z ~ Normal()
    end
end
spl = Gibbs(:x => MH(), :y => MH(), :z => MH())

sample(Xoshiro(470), f(), spl, 100; check_model=false) |> mean
```

Now, the bug I was hoping to find was that the rng is not controlled and therefore the value of `z` chosen will not be deterministic. However, the _real_ bug here is that `set_global_vnt!` stores a VarNamedTuple of a different type from the original one (because the VNT has a new variable inside it). This causes Gibbs to crash because the type parameter of the ref is fixed at construction time.

I think the solution to the crash is fairly simple — we just need to make the Ref an abstract type in the struct. This was done for pMCMC in Turing v0.43 and there were no real performance differences at all.

https://github.com/TuringLang/Turing.jl/blob/a979e75e0538bac971cece43e89af4d8e370c8d6/src/mcmc/particle_mcmc.jl#L27-L38

However, I'm not sure what the performance impacts are here, so that needs to be studied. Note that this change will affect type stability of `get_global_vnt` (but it is indeed possible that there are enough function barriers in the Gibbs codebase to make this not a meaningful problem). Once the crash is fixed, we could then fix the original reproducibility issue.

I don't think any of this is breaking, and it's a code path that is only hit in a vanishing number of use cases, so I'm ok with leaving this until after v0.44 is released.

### LINKED ISSUE #2860: Rework Gibbs: replace `GibbsContext` with `condition`, complete the component interface
## Current implementation and its problems

Gibbs conditions each component via `GibbsContext`, a parent context holding an implicit mutable `Ref{VarNamedTuple}`. `GibbsContext` and `condition` on the complement are the same operation, differing in exactly one case: a latent that first appears mid-trace, which the `Ref` smuggles out to sibling components.

Problems

- `GibbsContext` duplicates `condition`: a bespoke context and its mutable `Ref` where the core primitive would do;
- component init strategies ignored — Gibbs forces its own `InitFromPrior` on every component (#2693);
- component stats lost — they live on the discarded transition, so HMC/MH diagnostics never reach the chain (#2766).

## Principles for a new implementation

1. No shared mutable state. The global view is an immutable snapshot — read-only across components and particles, replaced functionally between sweeps — the value-replay the CSMC reference already uses on `main`. The mid-trace-variable case is eliminable per sampler kind (log-density: fixed-dimensional trace; simulation: absorbed into its own trace).
2. Condition with the core primitive: `make_conditional = condition(model, subset(global_vnt, non_target))`. Deletes the `GibbsContext` struct, the `Ref`, and the `extract_prefixes` hack.
3. A four-channel component interface: init-in (per-variable, from each component's `init_strategy`), condition-in, values-out (`gibbs_get_raw_values`), stats-out (`gibbs_get_stats`).

Related: https://github.com/TuringLang/AbstractMCMC.jl/issues/205, https://github.com/TuringLang/DynamicPPL.jl/issues/895

### LINKED ISSUE #2866: Require samplers to opt in to varying-dimensional targets
Most samplers assume that the set of variables in a model remains fixed after initialisation. If a later evaluation adds a variable such as `z` or a new array element such as `x[5]`, the sampler may silently use an invalid state or fail downstream with an unrelated error.

For example:

```julia
using Random
using Turing

@model function varying_target()
    x ~ Normal()
    y ~ Normal()

    if x > 0
        z ~ Normal()
    end
end

sample(Xoshiro(470), varying_target(), MH(), 100; progress=false)
```

Here `z` may be absent during initialisation and appear after `x` changes sign. `MH` does not implement a varying-dimensional transition, so it should reject the new variable immediately. Samplers like `Gibbs(:x => MH(), :y => MH(), :z => MH())` should error loudly if initialisation does not contain `z`. In the case of `z` happening to be in the initialisation, it will be retained and updated by its assigned sampler `MH()`, but its log density contribution will be ignored for MCMC steps where `x<0`. 

Non-Gibbs samplers should compare the variables in the new `VarNamedTuple` with those in the previous state. Encountering a new variable or leaf should raise an informative error unless the sampler explicitly opts in to varying-dimensional targets. `SMC`, `CSMC`, and `PG` should opt in; the default should be to reject structural changes.

The ongoing Gibbs-conditioning PR, [#2863](https://github.com/TuringLang/Turing.jl/pull/2863), adds the corresponding check for Gibbs. This should become a general sampler invariant rather than remain specific to Gibbs.

This also makes cases such as [#2810](https://github.com/TuringLang/Turing.jl/issues/2810) fail at the cause instead of silently producing invalid sampling behaviour or crashing later. Also related to https://github.com/TuringLang/Turing.jl/issues/2869.

yebai:
Gibbs samples one group of variables at a time, holding the rest fixed. That suits a model whose parts want different samplers: a discrete variable with `MH`, a smooth block with `NUTS`. It is also easy to get subtly wrong, because a badly chosen grouping does not fail loudly; it converges confidently to the wrong answer.

This rewrite conditions each component with `DynamicPPL.condition` and works out which variables are live by asking the model rather than by bookkeeping. The upshot: a bad partition is refused, naming the variable at fault, instead of being sampled.

Gibbs takes a partition of the model's variables into blocks $B_1,\dots,B_K$, one component sampler each, and sweeps them in turn. To step block $B_k$ it conditions the model on the current values of everything outside the block, so the component targets

$$p(\theta_{B_k} \mid \theta_{-B_k}, y)$$

Conditioned variables reach `tilde_observe!!`, which is what makes particle components reweight on them correctly.

### The snapshot comes from the model, not from the components

Values are threaded through the sweep as an immutable `VarNamedTuple`. After each step, the snapshot is rebuilt as the set of variables a model evaluation reaches at the proposed values — it is not the old snapshot merged with what the component reported.

Which variables exist is a property of the model at the current values, so only an evaluation settles it. A component may legitimately keep a value for a variable it is not currently sampling, to reuse if that branch comes back; were such a value taken into the snapshot, it would be conditioned into another component's step indefinitely, and the variable's disappearance would never be noticed.

```mermaid
flowchart LR
  S["snapshot: variables<br/>the model reaches"] --> C["condition on<br/>everything outside B_k"]
  C --> P["component samples B_k"]
  P --> E["evaluate model at<br/>the proposed values"]
  E --> G{"variable set<br/>changed?"}
  G -->|"no"| S
  G -->|"yes"| K{"sampled by B_k, and<br/>component declares it?"}
  K -->|"both"| S
  K -->|"otherwise"| X["reject, naming<br/>the variable"]
```

### A block has to be expressible

A component may not own part of a value the model stores as a unit. Under `x ~ MvNormal(zeros(2), I)` the values hold `x` as one key, so freeing it for a component that samples `x[1]` would free `x[2]` too; that is refused. Element-wise `x[1] ~` and `x[2] ~` give a key each and split cleanly. This depends only on the model's tilde statements, not on which sampler holds the containing block.

### A support change belongs to the component that made it

When the reached set is random, a block's support depends on the rest of the state:

$$\theta_{B_k} \in \Theta_{B_k}(\theta_{-B_k})$$

Let $b \sim \mathrm{Bernoulli}(0.3)$ decide whether $\theta_2$ exists. Then $p(b \mid \theta, y)$ compares densities over spaces of different dimensions, so it is not a conditional of one joint distribution. `Gibbs(:b => MH(), :θ => PG(20))` used to sample it and return $P(b=1) \approx 0.1$ against an exact $0.394$. Two conditions must now hold together: the stepping component declares `allow_varying_dimension`, and it samples the variable that came or went. Otherwise, put that variable and whatever decides its existence in one block, sampled by `PG` or `CSMC`.

Within a step: may the component's own proposal move between supports? That is `allow_varying_dimension` — `PG` and `CSMC` yes, `MH` no. Some partitions are refused before sampling starts; others only when a step reveals the change, because a block's shape is discovered by running it.

Note: DynamicPPL is pinned to an unreleased branch, so the `min` CI job fails until that
releases.

### [Gibbs Geweke Tests](https://github.com/user-attachments/files/31801729/gibbs-geweke-report.pdf)

Fixes #2860; fixes #2693; fixes #2766; fix #2866; fix #2810


src/mcmc/gibbs.jl
sunxd3: this is not true after this PR

src/mcmc/gibbs.jl
```diff
@@ -1,22 +1,44 @@
...
+
+"""
+    isgibbscomponent(spl::AbstractSampler)
+
+Deprecated name for [`is_gibbs_component`](@ref).
+
+The default `is_gibbs_component` delegates here, so a sampler that opted out by defining
+`isgibbscomponent` is still honoured; define `is_gibbs_component` instead.
+"""
 isgibbscomponent(::AbstractSampler) = true
```
sunxd3: understandable to keep, but feels like a footgun. we can add a TODO to signal future removal?

sunxd3 (commented):
It has come a long way to this point.

The code looks right. As far as I can tell, this PR and `main` are at parity on the usual cases.

Tricky cases like overlapping varnames and dynamic models still have corner cases. I found two particular examples.

## MWE 1: components whose varnames overlap by subsumption

```julia
using Turing, DynamicPPL, AbstractMCMC, Random
using Turing.Inference: gibbs_get_raw_values

@model function pair()
    x = Vector{Float64}(undef, 2)
    x[1] ~ Normal()
    x[2] ~ Normal()
    return 0.0 ~ Normal(x[1] + x[2], 0.1)
end
```

A: HMC on `x[1]`, HMC on `x`.

```julia
sample(Xoshiro(1), pair(), Gibbs(@varname(x[1]) => HMC(0.05, 3), @varname(x) => HMC(0.05, 3)), 20)
```

PR:
```
BoundsError: attempt to access DynamicPPL.VarNamedTuples.PartialArray{RangeAndTr
```
main:
```
No value was provided for the variable `x[2]`.
```

B: MH on `x[1]`, HMC on `x`. Component 1 is declared on `x[1]` only; print what it returns each sweep.

```julia
spl = Gibbs(@varname(x[1]) => MH(), @varname(x) => HMC(0.05, 3))
rng = Xoshiro(1); m = pair()
_, st = AbstractMCMC.step(rng, m, spl)
for i in 1:3
    _, st = AbstractMCMC.step(rng, m, spl, st)
    println(keys(gibbs_get_raw_values(st.states[1])))
end
```

PR:
```
VarName[x[1], x[2]]
VarName[x[1], x[2]]
VarName[x[1], x[2]]
```
main:
```
No value was provided for the variable `x[2]`.
```

I am fine with not supporting this particular shape. But `main` throws an (uninformative) error, while this PR silently samples a different block.

## MWE 2: a variable whose existence is controlled from another component (related to #2810)

```julia
using Turing, DynamicPPL, AbstractMCMC, Random, Statistics
using Turing.Inference: gibbs_get_raw_values

@model function f()
    x ~ Normal()
    y ~ Normal()
    if x > 0
        z ~ Normal()
    end
end
```

1: `z` is declared to belong to component 3. The seed gives `x < 0` initially, so `z` does not exist yet. Each sweep: `x`, keys in the global state, keys returned by component 1 (the `x` block).

```julia
spl = Gibbs(:x => MH(), :y => MH(), :z => MH())
rng = Xoshiro(470); m = f()
_, st = AbstractMCMC.step(rng, m, spl)
for i in 1:6
    _, st = AbstractMCMC.step(rng, m, spl, st)
    println(round(st.vnt[@varname(x)]; digits=2), "  ", keys(st.vnt), "  ", keys(gibbs_get_raw_values(st.states[1])))
end
```

PR:
```
1.76  VarName[x, y, z]  VarName[x, z]
-0.7  VarName[x, y, z]  VarName[x]
1.77  VarName[x, y, z]  VarName[x]
0.94  VarName[x, y, z]  VarName[x]
-1.0  VarName[x, y, z]  VarName[x]
-0.52  VarName[x, y, z]  VarName[x]
```
main:
```
MethodError: Cannot `convert` an object of type 
  VarNamedTuple{(:x, :y, :z),Tu
```

The first time `x` crosses 0, the `x` block draws `z` from the prior and returns it; `z` then stays in the global state permanently, including in sweeps where `x < 0` and the model never visits `z`.

2: no observations, so `P(b = 1) = 0.5` exactly.

```julia
@model function g()
    b ~ Bernoulli(0.5)
    u ~ Bernoulli(0.5)
    if b == 1
        z ~ Bernoulli(0.9)
    end
end
chn = sample(Xoshiro(1), g(), Gibbs(:b => MH(), (:u, :z) => MH()), 40_000; check_model=false)
println(mean(chn[@varname(b)]))
```

PR:
```
0.33755
```
main:
```
0.33755
```

This is the stationary distribution on the augmented state `(b, z)` in which the inactive `z` is kept and conditioned on with no density: `P(b = 1) = 1/3`.

3: same model as 1, HMC owns `z`.

```julia
sample(Xoshiro(470), f(), Gibbs(:x => MH(), (:y, :z) => HMC(0.1, 5)), 100; check_model=false)
```

PR:
```
KeyError: key z not found
```
main:
```
MethodError: Cannot `convert` an object of type 
  VarNamedTuple{(:x, :y, :z),Tu
```

On the PR, HMC's `LogDensityFunction` fixes its variable layout when the component is first stepped (`z` absent) and `gibbs_recompute_ldf_and_params` reuses that layout on every later update, so `z` has no slot.

yebai:
@sunxd3, this should be ready for another look. 

Gibbs is now more robust wrt varying dimensional models and gives much (useful) louder error messages when the sampler is misspecified.

sunxd3:
This looks good, thanks a lot for the work!

I found a few remaining correctness issues and edge cases, which I’ve put in a details block below. The Gibbs sample is a really complicated problem: it interfaces with many samplers, and dynamic models are just hard.

<details><summary>Details</summary>
<p>

## 1. Default prior-proposal `MH()` is incorrectly rejected on dynamic traces

Default `MH()` is a special case. In the following example, a block-local MH update should be permitted, even if its practical use is limited.

```julia
@model function dynamic_model(y=2.0)
    b ~ Bernoulli(0.3)
    θ = zeros(2)

    if b == 0
        θ[1] ~ Normal()
        y ~ Normal(θ[1], 0.5)
    else
        θ[1] ~ Normal()
        θ[2] ~ Normal()
        y ~ Normal(θ[1] + θ[2], 0.5)
    end
end

model = dynamic_model()

standalone = sample(
    Xoshiro(1), model, MH(), 20;
    check_model=false, progress=false, chain_type=Any,
)

sample(
    Xoshiro(1),
    model,
    Gibbs((@varname(b), @varname(θ)) => MH()),
    20;
    check_model=false,
    progress=false,
    chain_type=Any,
)
```

Observed:

```text
standalone MH = SUCCESS, 20 draws, b states [0, 1]
one-block Gibbs MH = FAIL
ArgumentError: The variable θ[2] stopped existing during a step of MH ...
```

## 2. `GibbsConditional` can see values different from those used by the model

This is a corner case involving `merge` and missing values. A missing conditioned or fixed value can overwrite a non-missing model argument, even though DynamicPPL treats the missing context value as absent during model execution.

```julia
@model function arg_model(y)
    x ~ Normal()
    y ~ Normal(x, 1)
end

seen = Ref{Any}(nothing)
cond_x(vnt) = (seen[] = vnt; Normal())

model = condition(arg_model(1.0); y=missing)

sample(
    Xoshiro(1),
    model,
    Gibbs(@varname(x) => GibbsConditional(cond_x)),
    2;
    progress=false,
)

@show model.args.y
@show seen[][@varname(y)]
```

Observed:

```text
model.args.y = 1.0
seen[][@varname(y)] = missing
```

## 3. Dynamic changes are not checked against overlapping fixed-layout owners

A dynamic component can change an address that is also owned by HMC. This exposes an unfortunate interaction: the dynamic component must manage the variables that control address existence, while the overlapping HMC component retains a fixed-layout state.

```julia
@model function overlap_dynamic()
    b ~ Bernoulli(0.5)
    if b == 1
        z ~ Normal()
    end
end

sample(
    Xoshiro(1),
    overlap_dynamic(),
    Gibbs(
        (@varname(b), @varname(z)) => PG(10),
        @varname(z) => HMC(0.1, 3),
    ),
    10;
    check_model=false,
    progress=false,
    chain_type=Any,
)
```

Observed after two completed draws:

```text
ArgumentError: Some indices in the output vector were not set. This likely means
that the vector values provided are not consistent with the LogDensityFunction.
```

## 4. Equal leaf sets can conceal an incompatible fixed layout

```julia
using LinearAlgebra

@model function regroup_dynamic()
    b ~ Bernoulli(0.5)
    x = Vector{Float64}(undef, 2)

    if b == 1
        x[1] ~ Normal()
        x[2] ~ Normal()
    else
        x ~ MvNormal(zeros(2), I)
    end
end

sample(
    Xoshiro(1),
    regroup_dynamic(),
    Gibbs(
        @varname(b) => MH(),
        @varname(x) => HMC(0.1, 3),
    ),
    10;
    check_model=false,
    progress=false,
    chain_type=Any,
)
```

Both branches have the leaves `x[1]` and `x[2]`, but the layouts differ:

```text
x[1] → 1:1, x[2] → 2:2

versus

x → 1:2
```

Observed after two completed draws:

```text
MethodError: no method matching
getindex(::DynamicPPL.RangeAndTransform{DynamicPPL.DynamicLink}, ::Int64)
```

## 5. A valid block schedule depends on another sampler's reporting format

HMC, ESS, and external samplers reconstruct values through `ParamsWithStats`, which densifies a complete element-wise `PartialArray`. MH and PG preserve the element-level representation.

```julia
@model function pair()
    x = Vector{Float64}(undef, 2)
    x[1] ~ Normal()
    x[2] ~ Normal()
    0.0 ~ Normal(x[1] + x[2], 0.1)
end

mh_blocks = Gibbs(
    @varname(x[1]) => MH(),
    (@varname(x[1]), @varname(x[2])) => MH(),
)

hmc_blocks = Gibbs(
    @varname(x[1]) => MH(),
    (@varname(x[1]), @varname(x[2])) => HMC(0.05, 3),
)

sample(Xoshiro(1), pair(), mh_blocks, 5; progress=false)
sample(Xoshiro(1), pair(), hmc_blocks, 5; progress=false)
```

Observed:

```text
MH blocks  = SUCCESS
HMC blocks = FAIL
ArgumentError: Gibbs cannot condition on part of x: the values store it as a unit,
but this component does not sample x[2].
```

## 6. Pre-existing: ESS retains a stale condition-dependent prior

This defect predates the PR. When the Gaussian prior of an ESS block depends on another Gibbs block, `gibbs_update_state!!(::ESS, ...)` recomputes the LDF and parameter vector but retains `state.priors`. The ellipse therefore remains centred on the prior from the initial Gibbs state.

```julia
using StableRNGs, Statistics

@model function chain2()
    m ~ Normal(0, 1)
    x ~ Normal(m, 1)
    1.0 ~ Normal(x, 1)
end

for seed in (1, 2)
    chain = sample(
        StableRNG(seed),
        chain2(),
        Gibbs(@varname(m) => MH(), @varname(x) => ESS()),
        40_000;
        discard_initial=2_000,
        progress=false,
    )

    @show seed mean(chain[@varname(x)]) mean(chain[@varname(m)])
end
```

The exact posterior means are `E[x | y=1] = 2/3` and `E[m | y=1] = 1/3`.

Observed:

```text
seed 1: mean(x) = 0.914496, mean(m) = 0.367646
seed 2: mean(x) = 0.470938, mean(m) = 0.231775
```

The result's dependence on the initial draw is consistent with a stale ellipse centre. `PriorDistributionAccumulator` must be recomputed during ESS synchronization.


</p>
</details>

PR #2877, opened 2026-09-02
Fix biased MH samples from ignored aggregate proposals

### LINKED ISSUE #2876: Misspecified MH proposals can bias sampling
An aggregate proposal key can be present in the parameter `VarNamedTuple` without matching any executed `~` statement. `init` then draws each site from its prior, while `log_proposal_density` still includes the unused aggregate proposal. The recorded density therefore differs from the proposal kernel that generated the sample.

```julia
using Distributions
using DynamicPPL
using Random
using Statistics
using Turing

@model function model()
    x = zeros(2)
    x[1] ~ Exponential(0.5)
    x[2] ~ Exponential(0.5)
end

proposal = filldist(Exponential(1.0), 2)
chain = sample(
    MersenneTwister(0x5eed),
    model(),
    MH(@varname(x) => proposal),
    100_000;
    discard_initial=10_000,
    progress=false,
    verbose=false,
)

mean(chain[@varname(x[1])]), mean(chain[@varname(x[2])])
# Approximately (1.03, 0.94); expected (0.5, 0.5)
```

The proposal for `x` is ignored because the model executes separate statements for `x[1]` and `x[2]`. Both sites are therefore sampled from their priors. The density calculation nevertheless includes both prior densities and the declared density `q(x)`. In this static example, the chain consequently targets a density proportional to $p(x)L(x)/q(x)$.

This was traced to the MH rewrite introduced in [`481be6b3`](https://github.com/TuringLang/Turing.jl/commit/481be6b349934a448b293e0cb41e99e67d1bbdfb), merged through [PR #2756](https://github.com/TuringLang/Turing.jl/pull/2756), and released in v0.43.0. Using the equivalent `:x => proposal` syntax, the example produces means near `(0.5, 0.5)` in v0.42.9 but approximately `(1.03, 0.94)` in v0.43.0. Before the rewrite, MH grouped the indexed sites under `x` and applied the aggregate proposal to the whole vector. The new implementation selects proposals by exact `VarName` during model execution, but still includes the unused aggregate proposal when computing its density.

Proposal-density accounting should include only proposals actually selected during model evaluation. The initial support check and both directions of the MH ratio should use the same information.

yebai:
### The bug

An aggregate proposal can be addressable through a `VarNamedTuple` without being one of its exact keys. For a model containing `x[1] ~ ...` and `x[2] ~ ...`, `haskey(raw_vals, @varname(x))` is therefore true even though the model never executes a statement for `x`.

MH retains the proposal for `x`, but `InitFromProposals` does not use it when evaluating the indexed statements. Both sites are drawn from their priors. Density accounting nevertheless includes the unused aggregate density:

$$
g_{\mathrm{recorded}}(x) = p(x_1)p(x_2)q(x)
$$

The extra density ratio changes the invariant density from $p(x)L(x)$ to one proportional to $p(x)L(x)/q(x)$.

### The proof

```julia
using Distributions, DynamicPPL, Random, Statistics, Turing

@model function model()
    x = zeros(2)
    x[1] ~ Exponential(0.5)
    x[2] ~ Exponential(0.5)
end

chain = sample(
    MersenneTwister(0x5eed), model(),
    MH(@varname(x) => filldist(Exponential(1.0), 2)), 100_000;
    discard_initial=10_000, progress=false, verbose=false,
)

mean(chain[@varname(x[1])]), mean(chain[@varname(x[2])])
```

Before the fix, this gives approximately `(1.03, 0.94)` instead of `(0.5, 0.5)`. After the fix, it gives `(0.5020, 0.4999)`.

The regression test compares this chain against `MH()` under the same `StableRNG`. The chains differ before the fix and agree exactly afterward.

### The fix

Use membership in `keys(raw_vals)`, which contains the exact executed `VarName`s, instead of `haskey`, which also accepts structurally reconstructable aggregates. This excludes unused proposals before constructing `InitFromProposals`, keeping initialization and both directions of the MH ratio consistent.

Fixes #2876.

### DIFF
--- a/src/mcmc/mh.jl
+++ b/src/mcmc/mh.jl
@@ -164,6 +164,15 @@ struct LinkedRW{C}
 end
 LinkedRW(var::Real) = LinkedRW(var * I)
 
+# For example, `LinkExact(Set([@varname(x)]))` links `x` but not `x[1]`,
+# whereas `DynamicPPL.LinkSome` links both.
+struct LinkExact{V<:Set{<:VarName}} <: DynamicPPL.AbstractTransformStrategy
+    vns::V
+end
+function DynamicPPL.target_transform(strategy::LinkExact, vn::VarName)
+    return vn in strategy.vns ? DynamicPPL.DynamicLink() : DynamicPPL.Unlink()
+end
+
 """
     InitFromProposals(proposals::VarNamedTuple, verbose::Bool)
 
@@ -222,11 +231,12 @@ function MH(pair1::SymOrVNPair, pairs::Vararg{SymOrVNPair})
     # have `LinkedRW` proposals. That in turn is obtained using `MHLinkedValuesAccumulator`.
     function init_strategy_constructor(raw_vals, linked_vals)
         proposals = DynamicPPL.VarNamedTuple()
+        raw_vns = Set(keys(raw_vals))
         for pair in vn_proposal_pairs
             # Convert all keys to VarNames.
             vn, proposal = pair
             vn = _to_varname(vn)
-            if vn ∉ keys(raw_vals)
+            if vn ∉ raw_vns
                 continue
             end
             proposal_dist = if proposal isa Distribution
@@ -254,7 +264,7 @@ function MH(pair1::SymOrVNPair, pairs::Vararg{SymOrVNPair})
     link_strategy = if isempty(linkedrw_vns)
         DynamicPPL.UnlinkAll()
     else
-        DynamicPPL.LinkSome(linkedrw_vns, DynamicPPL.UnlinkAll())
+        LinkExact(linkedrw_vns)
     end
     return MH(init_strategy_constructor, link_strategy, all_vns)
 end
--- a/src/mcmc/mh.jl
+++ b/src/mcmc/mh.jl
@@ -226,7 +226,7 @@ function MH(pair1::SymOrVNPair, pairs::Vararg{SymOrVNPair})
             # Convert all keys to VarNames.
             vn, proposal = pair
             vn = _to_varname(vn)
-            if !haskey(raw_vals, vn)
+            if vn ∉ keys(raw_vals)
                 continue
             end
             proposal_dist = if proposal isa Distribution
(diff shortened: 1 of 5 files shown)


src/mcmc/mh.jl
```diff
@@ -222,11 +231,12 @@ function MH(pair1::SymOrVNPair, pairs::Vararg{SymOrVNPair})
     # have `LinkedRW` proposals. That in turn is obtained using `MHLinkedValuesAccumulator`.
     function init_strategy_constructor(raw_vals, linked_vals)
         proposals = DynamicPPL.VarNamedTuple()
+        raw_vns = Set(keys(raw_vals))
         for pair in vn_proposal_pairs
             # Convert all keys to VarNames.
             vn, proposal = pair
             vn = _to_varname(vn)
-            if !haskey(raw_vals, vn)
+            if vn ∉ raw_vns
```
yebai, line 239: This is the issue, as it includes ignored aggregate proposals.

sunxd3:
This is subtle, thanks for looking into this.

I think the current PR might still have some issues. For instance, 

```julia
@model function ranged_site()
    x = zeros(2)
    x[1:2] ~ MvNormal(zeros(2), I)
end

MH(@varname(x[1:2]) => MvNormal(zeros(2), 4I))
```

The proposal exactly matches the tilde LHS, but `keys(raw_vals)` contains `x[1]` and `x[2]`, so the PR drops it. It draws from the prior but does not include that prior in the proposal-density correction. Consequently, the chain targets $N(0, \frac{1}{2} I)$ rather than $N(0, I)$.

Actual output from 30,000 samples on the PR head:

```text
E[x1²]=0.9963  E[x2²]=1.0044  acceptance=0.4015
```

So the variance is wrong (should be both ~0.5).

---

A pre-existing bug involving `:=` discovered during review

```julia
@model function deterministic_site()
    x ~ Exponential(0.5)
    y := x
end

MH(@varname(y) => Exponential(1.0))
```

`RawValueAccumulator(true)` includes `y`, so this proposal is retained. However, `y` is not a tilde-site, meaning the proposal is never used for sampling while its density is still included in the acceptance ratio.

```text
proposal for y: E[x]=0.9839  acceptance=0.6689
MH() control:   E[x]=0.5007  acceptance=1.0000
```

yebai:
Thanks @sunxd3. Closing this in favour of https://github.com/TuringLang/Turing.jl/pull/2883. See the comments there for reasons.

Issue #1583, opened 2021-04-10
MH currently does not support link/invlink properly on a per-variable basis


torfjelde:
Ref: https://github.com/TuringLang/Turing.jl/pull/1582

Currently, something like
```julia
 MH(
       :m => AdvancedMH.StaticProposal(Normal()),
       :s => AdvancedMH.RandomWalkProposal(Normal())
)
```

won't lead to `s` being linked as one might except. The "fix" in https://github.com/TuringLang/Turing.jl/pull/1582 makes it so that we at least link if *all* are `RandomWalkProposal`. 

This could be easily done after https://github.com/TuringLang/DynamicPPL.jl/pull/231 if we decide to take that approach.

Making this issue as a reminder.

penelopeysm:
I've basically rewritten MH with init strategies + accumulators in #2756, and this should be handled correctly, albeit without forwarding to AdvancedMH (whose NamedTuple-based interface is too inflexible for Turing models, and was too tightly tied to the NamedTuple-of-metadata structure). The current interface is something like this:

```julia
spl = MH(
    # Static proposal
    @varname(x) => Normal(),

    # Conditional proposal, depends on previous step's value in unlinked space.
    # vnt contains all variables' values so you could have some wacky stuff here
    @varname(y) => (vnt -> Normal(vnt[@varname(y)], 1.0),

    # RW proposal, but operates on linked values.
    @varname(z) => LinkedRW([1.0;;])
)
```

penelopeysm:
Fixed in v0.43

yebai:
The per-`VarName` generalisation introduced to resolve this issue created a correctness risk, documented in #2876. An aggregate proposal can be included in the Hastings ratio without being used to generate the candidate:

  ```julia
  using Distributions, DynamicPPL, Random, Statistics, Turing

  @model function aggregate_sites()
      x = zeros(2)
      x[1] ~ Exponential(0.5)
      x[2] ~ Exponential(0.5)
  end

  chain = sample(
      MersenneTwister(0x5eed),
      aggregate_sites(),
      MH(@varname(x) => filldist(Exponential(1.0), 2)),
      100_000;
      discard_initial=10_000,
      progress=false,
      verbose=false,
  )

  mean(chain[@varname(x[1])]), mean(chain[@varname(x[2])])
  # (1.034, 0.935); expected approximately (0.5, 0.5)
```

[...]

PR #2882, opened 2026-09-03
mh: validate exact proposal sites

yebai:
Draft alternative to #2877 for #2876. This records `VarName`s seen at executed `~` sites and requires keyed MH proposals to match exactly, preventing unused aggregate proposals from entering the acceptance ratio. Tests cover indexed and ranged sites, `:=` variables, and #1583-related linking behavior.

Open questions remain around preserving linear-index identity, distinguishing inactive proposals under conditioning and Gibbs, and avoiding repeated site-key scans. These should be resolved before this approach is merged.


yebai:
Closing this in favour of https://github.com/TuringLang/Turing.jl/pull/2883. See the comments there for reasons.

